PR #8, opened 2026-03-09
feat(cli): add QURL CLI tool

justin-layerv:
## Summary

- Adds Go CLI tool using cobra at `apps/cli/`
- Commands: `create`, `resolve`, `list`, `get`, `delete`, `extend`, `version`
- Supports `--output json` for machine-readable output and `--output table` (default) for humans
- Auth via `QURL_API_KEY` env var or `--api-key` flag
- Adds `Resolve()` method to `shared/client` for headless resolution
- Adds `build-cli` Makefile target

### New files
- `apps/cli/cmd/` — 9 Go files (main, root, create, resolve, list, get, delete, extend, version)
- `apps/cli/internal/output/formatter.go` — Table and JSON formatters

### Modified files
- `shared/client/client.go` — Added `Resolve()` method with `ResolveInput`/`ResolveOutput` types
- `Makefile` — Added `build-cli` target
- `go.mod`/`go.sum` — Added cobra dependency

## Usage

```bash
export QURL_API_KEY=lv_live_xxx

qurl create https://api.example.com/data --expires 1h
qurl resolve at_k8xqp9h2sj9lx7r4a
qurl list --limit 5
qurl get r_a3x9Bk7mQ2
qurl delete r_a3x9Bk7mQ2
```

## Test plan

- [ ] `go build ./apps/cli/cmd/` succeeds
- [ ] `golangci-lint run` passes
- [ ] `qurl version` prints version
- [ ] `qurl create`, `qurl list`, `qurl get`, `qurl delete` work against sandbox API
- [ ] `qurl resolve` works after Phase 1 headless resolve is deployed
- [ ] `--output json` produces valid JSON


justin-layerv:
## Code Review Feedback — Addressed

All feedback items have been addressed in commit `6c4c816`. Summary below.

### Security Fixes
- **Profile name path traversal** — `ProfilePath` now validates names against `^[a-zA-Z0-9_-]+$`, rejecting `../../etc/passwd` etc.
- **Path-escape resource IDs** — All client methods now use `url.PathEscape(id)` in URL construction
- **Checksum verification in install.sh** — Downloads and verifies `checksums.txt` before extracting

### Bug Fixes
- **`ListOutput` missing JSON tags** — Added `json:"qurls"`, `json:"next_cursor,omitempty"`, `json:"has_more,omitempty"`
- **Silent profile load errors** — `newClient()` now returns the error when `--profile` was explicitly specified
- **Corrupt config silently wiped** — `config set` now returns a parse error instead of overwriting with empty config
- **`formatError` loses wrapping context** — Now preserves prefix (e.g., "create QURL:") when unwrapping `APIError`
- **`InvalidFields` non-deterministic order** — Keys are now sorted before display

### UX Improvements
- **`--output` format validation** — Unknown formats (e.g., `--output yaml`) now error instead of silently falling back to table
- **`extend --by` required** — No more silent 24h default mutation; flag is now required
- **`--force` flag removed from delete** — Kept `--yes`/`-y` only, no false semantics
- **Zero-duration rejected** — `validateDuration` now rejects `0h`, `0d`, etc.
- **Config `get` validates key** — Unknown keys now produce a clear error with valid options listed
- **Shell completion timeout** — API calls in tab-completion now have a 3s timeout
- **`docs.go` color restore** — `color.NoColor` is now restored via `defer` after doc generation

### Hardening
- **Response body size limit** — `io.ReadAll` replaced with `io.LimitReader(resp.Body, 1<<20)` (1MB max)
- **Retry body buffering comment** — Clarified why body is always buffered
- **Extend/Update endpoint comment** — Documented that both use `PATCH /v1/qurls/:id`, differentiated by body
- **golangci.yml exclusion comments** — Fixed G703 and G115 descriptions to match actual rules

### Tests Added
- **Config package** — 10 tests: load/save round-trip, missing file, corrupt YAML, Set/Get valid+invalid keys, `IsValidKey`, file permissions (0600), profile name validation, profile isolation
- **CLI commands** — `TestExtendCommand`, `TestExtendCommandRequiresBy`, `TestMintCommand`, `TestQuotaCommand`, `TestDeleteCommandNoForceFlag`, `TestInvalidOutputFormat`
- **`formatError`** — 7 tests: plain error, API error, 401/429/quota hints, invalid fields, wrapped error context

### Deferred (out of scope)
- **`readToken` uses `os.Stdin.Stat()` not `cmd.InOrStdin()`** — Requires an abstraction layer for terminal detection; standard pattern in Go CLIs
- **Create uses `/v1/qurl` (singular)** — Matches the actual API spec; intentional
- **Nits 25-34** (Homebrew SHA placeholder, default user-agent, Makefile GOOS, goreleaser temp path, client file splitting) — Low priority, follow-up PR

justin-layerv:
## Second-Round Code Review Feedback — Addressed

Addressed the latest review findings (post-fix review from `07:31:03Z`) in commits `0b5ab74` and `d60f5bb`.

### Bug Fixes
- **`validateDuration` multi-digit zero edge case** — `"00h"` now correctly rejected. Replaced string comparison (`m[1] == "0"`) with `strconv.Atoi` + `<= 0` check. Added test cases for `"0h"`, `"00h"`, and `"007d"`.

### Code Quality
- **Default user-agent is now generic** — Changed from `"qurl-cli/dev"` to `"qurl-go-client/dev"` in `shared/client`. CLI sets `"qurl-cli/<version>"` via `WithUserAgent`, Slack Lambda now sets `"qurl-slack/dev"`.
- **Retry disabled for Lambda consumers** — Added `WithRetry(0)` to Slack Lambda client constructor to prevent automatic retries from exceeding Lambda timeout.
- **Output format constants** — Defined `output.FormatTable` and `output.FormatJSON`; replaced raw strings in root.go and tests.
- **Status constants** — Defined `client.StatusActive/Expired/Revoked/Consumed`; used in formatter's `colorStatus`.
- **HTTP status constants** — `formatError` now uses `http.StatusUnauthorized` and `http.StatusTooManyRequests`.
- **Config load/save simplification** — `LoadProfile`/`SaveProfile` both handle empty name, eliminating branching in config subcommands.
- **Extracted `patchQURL`** — Unified near-identical `Extend` and `Update` methods into a shared private helper.

### Security / UX
- **API key warning improved** — Root command help now warns `--api-key` is visible in process list. `config set api_key` output now suggests `QURL_API_KEY` env var as alternative.
- **Install script checksum verification** — Replaced silent `2>/dev/null` fallback with explicit per-platform logic: checks for `sha256sum`, falls back to `shasum`, warns if neither is available.

### Tests Added
- **Config subcommand integration tests** — `TestConfigSetAndGet`, `TestConfigSetAPIKeyWarning`, `TestConfigGetInvalidKey`, `TestConfigPath`, `TestConfigProfiles`
- **`--quiet` flag tests** — `TestCreateCommandQuiet`, `TestMintCommandQuiet`
- **Duration validation edge cases** — `"0h"`, `"00h"`, `"007d"`

### Deferred
- **`os.Stdin.Stat()` bypasses `cmd.InOrStdin()`** — Requires an abstraction layer for terminal detection; standard Go CLI pattern. Current tests cover the arg and pipe paths.
- **Config subcommand tests for `--profile` flag round-trip** — Config package already has profile isolation tests; CLI-level profile tests would need more infrastructure.

justin-layerv:
## Third-Round Code Review Feedback — Addressed

Addressed the latest review findings (post-fix review from `17:48:44Z`) in commit `d1f1260`.

### Fixed
- **API key redaction warning** — Added comment in `client.do()` warning future developers to redact the `Authorization` header if header logging is ever added, preventing API key leakage via `--verbose`.
- **GPG signature verification in install.sh** — Added optional GPG verification of `checksums.txt.sig` when `gpg` is available. Imports the release signing key, verifies the signature, and warns (but continues) if verification fails since SHA256 already passed. Falls through gracefully when no signature, key, or gpg binary exists.

### Deferred
- **`os.Stdin.Stat()` bypasses `cmd.InOrStdin()`** — Deferred across 3 review rounds. Standard Go CLI pattern for terminal detection; requires abstraction layer. Current tests cover arg and pipe paths.
- **Show usage on cobra arg validation failures** — `SilenceUsage: true` is standard for production CLIs. Commands have `Example` fields for guidance.
- **Flag vs env precedence for secrets** — Current behavior (flag > env > config) follows cobra conventions. Reviewer confirmed this is intended.
- **Completion API call caching** — Already bounded by 3s timeout and 20-item limit. Caching adds complexity for marginal gain on a cold path.
- **`ListInput` JSON tags** — Intentionally dropped since the struct is only used for query params, not JSON marshaling.

PR #34, opened 2026-03-27
feat(cli): add qurl auth login with Authorization Code + PKCE

Kevin-layerV:
## Summary

- Implements `qurl auth login` using the OAuth 2.0 Authorization Code flow with PKCE (RFC 7636)
- Opens the user's browser to complete Auth0 authentication via a local callback server, then creates a scoped API key via the QURL API and stores it in `~/.config/qurl/config.yaml` (0600 permissions)
- Adds `qurl auth logout` (clears local credentials) and `qurl auth status` (shows auth source, plan, key prefix)
- New `apps/cli/internal/auth` package: PKCE flow, browser opener, API key creation
- Updates error messages across the CLI to recommend `qurl auth login` as the primary auth method
- Adds `ci(cli)` workflow so CLI changes actually run lint + tests + build in CI

### Auth0 setup required before shipping

The `DefaultClientID` constant in `apps/cli/internal/auth/pkce.go` is empty. Before shipping:
1. Create an Auth0 **Single Page Application** with Authorization Code + PKCE enabled
2. Set the client ID in `DefaultClientID` or configure via `QURL_AUTH0_CLIENT_ID` env var
3. Add `http://127.0.0.1:*/callback` to the allowed redirect URIs in Auth0
   (Note: the callback server binds to `127.0.0.1`, not `localhost` — Auth0 treats these as distinct entries)

### Login flow

```
$ qurl auth login

  Opening browser to authenticate...

  Waiting for authentication... done
  Creating API key... done

  Logged in successfully!

  API Key: lv_live_a3x9...
  Key ID:  key_aBcDeF123456
  Scopes:  qurl:read, qurl:write, qurl:resolve
  Saved:   /Users/you/.config/qurl/config.yaml
```

## Test plan

- [x] 30+ unit tests covering PKCE flow, API key creation, and all three commands
- [x] Tests for error paths: missing client ID, auth failure, API key creation failure
- [x] Tests for custom flags: `--key-name`, `--scopes`, `--no-browser`, `--profile`
- [x] Config round-trip tests: login saves key, logout clears it, status reads from correct source
- [x] Race detector tests pass (`go test -race ./apps/cli/...`)
- [x] Zero golangci-lint issues
- [x] Malformed config now surfaces a parse error instead of silently returning "Not logged in"
- [ ] Manual test with Auth0 application (requires Auth0 setup above)

## Follow-up issues

- #118 — `auth logout --revoke` for server-side key revocation
- #119 — client-side `--scopes` validation
- #120 — deferred nits and improvements from review rounds


justin-layerv (changes requested):
## Review: feat(cli): add qurl auth login with OAuth device code flow

Good architecture, clean separation of concerns, solid test coverage (30+ tests). Close to mergeable — four items need fixing.

### Must fix

**1. RFC 8628 slow_down interval accumulation bug**

`pollOnce` resets the interval to `defaultPollInterval + slowDownIncrement` (10s) on every `slow_down` response. Per RFC 8628 §3.5, the client must increase its *current* interval by 5s each time. Multiple slow_down responses should produce 10s → 15s → 20s, not 10s → 10s → 10s. Fix: pass the current `pollInterval` into `pollOnce` and return `currentInterval + slowDownIncrement`.

**2. Config file permissions**

The API key is stored in plaintext in `~/.config/qurl/config.yaml`. Verify that `config.SaveProfile` creates the file with `0600` permissions. If not, any user on a shared system can read the key. This was flagged by all three automated reviews.

**3. Data race in `TestAuthLoginCustomKeyName`**

`receivedName` is written from the HTTP handler goroutine and read from the test goroutine without synchronization. This will fail with `go test -race`. Use a mutex or `atomic.Value`.

**4. Silent error suppression on `config.LoadProfile`**

Multiple call sites discard the error from `config.LoadProfile`. If the config file exists but contains malformed YAML, the user silently gets "Not logged in" instead of a useful error. At minimum, distinguish `os.ErrNotExist` (expected) from parse errors (surface to user).

### Should fix (non-blocking, but would accept in this PR)

- **`time.After` in polling loop** — creates a new timer each iteration without stopping on context cancel. Use `time.NewTimer` with explicit `Stop()`.
- **CI didn't run** — most jobs were skipped (change detection didn't trigger CLI workflow). Rebase onto main and verify lint + tests actually execute.

### Deferred (follow-up issues)

- Server-side key revocation on `qurl auth logout` (`--revoke` flag)
- Client-side scope validation against `allScopes`
- PR title may need updating (one reviewer noted it describes device code flow but implementation may be PKCE — please clarify)

Kevin-layerV:
## Review Response

Thanks @justin-layerv! Here's what was addressed in 8213c44:

### Must-fix items

**#2 Config file permissions** — Already enforced: `config.SaveProfile` calls `os.WriteFile(p, data, 0o600)` and `os.MkdirAll(dir, 0o700)`. `TestFilePermissions` in `internal/config/config_test.go` verifies this on every run. No code change needed.

**#3 Data race in `TestAuthLoginCustomKeyName`** — Fixed. Added a `sync.Mutex` to guard `receivedName` between the HTTP handler goroutine (writer) and the test goroutine (reader). Same fix applied to `TestAuthLoginCustomScopes` (`receivedScopes`), which had the identical race.

**#4 Silent error suppression on `config.LoadProfile`** — Fixed. `authLogoutCmd`, `resolveAPIKeyWithSource`, and `newClient()` now surface parse errors (malformed YAML) to the user. `os.ErrNotExist` is still silently treated as "no config yet" — that's the intended path. `saveAuthConfig` intentionally keeps the silent discard (we're overwriting anyway) with an explanatory comment. Two new tests cover the end-to-end error surfacing: `TestAuthLogoutMalformedConfig` and `TestAuthStatusMalformedConfig`.

### Items obsoleted by the PKCE refactor (commit a985411)

**#1 RFC 8628 `slow_down` interval accumulation** — N/A. The `pollOnce` function and device code polling loop were removed entirely when the flow was changed to Authorization Code + PKCE. There is no polling interval to fix.

**`time.After` in polling loop** — Same: the polling loop no longer exists.

### CI note

The branch was rebased onto `main` (rebase triggered on 8213c44) so the path-filter for the CLI workflow should now fire correctly.

### Deferred items

Filing follow-up issues now for server-side key revocation and client-side scope validation.

Kevin-layerV:
### Deferred items (issues filed)

- Server-side key revocation (`--revoke` flag): #118
- Client-side scope validation: #119

Kevin-layerV:
Thanks for the thorough second pass, @claude-review.

**Must-fix items addressed in commit 2281b8e:**

1. **\`saveAuthConfig\` swallowing malformed config** — now propagates \`LoadProfile\` errors; non-credential fields (endpoint, output format) are preserved on a successful load, and a malformed config surfaces as an error so the user sees it during login rather than silently losing their settings. Test added: \`TestAuthLoginMalformedConfig\`.

2. **Callback handler blocking on second request** — all four \`codeCh\` send sites now use a non-blocking \`select/default\` so a browser refresh, port probe, or any subsequent request returns an HTTP response immediately rather than stalling the goroutine. Test added: \`TestCallbackSecondRequestDoesNotBlock\`.

**Should-fix / nit items:**

Items 3–7 and the nits are noted and valid. Deferring them to keep the scope of this PR focused — filed as #120 to track them together.

Kevin-layerV:
Deferred nits tracked in #120.

Kevin-layerV:
## Review Response (third round)

Thanks @claude — good catches. Addressed in commit a39dd8f.

### Fixed

**#1 successHTML shown on duplicate/probe requests** — Fixed. The handler now tracks whether the result was actually delivered (`delivered` bool from the non-blocking select). The first request gets `successHTML` ("Authentication successful"); subsequent requests get `alreadyDoneHTML` ("Already authenticated"). `TestCallbackSecondRequestDoesNotBlock` enhanced to assert both response bodies.

**#2 Dead `cfg == nil` branches** — Fixed. Removed the unreachable nil guard in `authLogoutCmd` (changed to `if cfg.APIKey == ""`) and in `saveAuthConfig`. Added a comment on `saveAuthConfig` explaining that `LoadProfile` returns a non-nil `*Config` on success — missing file yields `&Config{}`, not nil.

**#4 Misleading `nolint:gosec` comment in `browser.go`** — Fixed. Changed from "validated as HTTPS URL from Auth0 server response" to "URL validated as https scheme above; exec uses argv (no shell) so no injection risk".

**#6 `subtle.ConstantTimeCompare` for state** — Added. State comparison in `callbackHandler` now uses `crypto/subtle.ConstantTimeCompare`.

**#7 Comment on `printLoginSuccess` stdout/stderr split** — Added a doc comment explaining the deliberate split: banner to stderr so it appears on-terminal even when stdout is redirected; key data to stdout so `qurl auth login > out.txt` captures the machine-readable fields.

### Deferred to #120

**#3 `resolveAuth0Config` partial config** — Non-blocking, added to #120 as item 8.

**#5 `containsString`/`searchSubstring`** — Already tracked in #120 as item 6.

**#8 `--key-name` validation** — Non-blocking, added to #120 as item 9.

**#9 `runAuthLogin` complexity** — Non-blocking, added to #120 as item 10.

**#10 Coverage gate** — Already tracked in #120 nits.

Kevin-layerV:
## Review Response (round 4)

Thanks @claude — addressed below.

### Fixed in commit 79ff9da

**#3 `TestAuthStatusAuthenticated` does not override `HOME`** — Fixed. Added `t.Setenv("HOME", t.TempDir())` as the first line of the test, matching the pattern used in every other auth test in the file.

### Deferred — `DefaultClientID = ""` (#1)

This was already acknowledged in the PR description and in the round-3 review verdict (*"solid and shippable once the Auth0 SPA is provisioned and `DefaultClientID` is filled in"*). The empty default is intentional: the Auth0 SPA hasn't been provisioned yet, and any user who hits the unconfigured path gets a clear `"auth0 client ID not configured: set QURL_AUTH0_CLIENT_ID environment variable"` error rather than a confusing failure.

Build-time `ldflags` injection is the right long-term solution and is tracked in #120. Landing that in this PR would expand scope significantly right before human re-review.

### Deferred to #120

- **#2** (error message casing) — lint passed in CI; `Auth0` is a proper noun so ST1005 doesn't apply here.
- **#4** (StartLogin/WaitForToken leak doc) — added to #120.
- **#5** (coverage gate 40%) — already in #120.
- **#6** (`govulncheck@latest` not pinned) — added to #120.
- **#7** (explicit `GOOS`/`GOARCH` in build step) — added to #120.
- **#8–#11** (nits) — already tracked in #120.

Kevin-layerV:
## Review Response (round 5)

Thanks @claude — addressed both "strongly suggest before merge" items in commit 3a90e25.

**#1 `containsString`/`searchSubstring` duplicate `strings.Contains`** — Fixed. Both call sites now use `strings.Contains` directly; the two hand-rolled helpers are removed. `strings` was already imported in `pkce_test.go`.

**#3 Error hint wording** — Fixed. Message now reads: *"Auth0 client ID not configured: this build was not compiled with a default client ID; set QURL_AUTH0_CLIENT_ID to override"* — makes clear this is a build-time gap, not a user error.

**Remaining items deferred to #120:**
- #2 (comment placement in `saveAuthConfig`)
- #4 (`goconst` on "load config" — lint passes in CI so a no-op)
- #5 (orphaned key on save failure — tracked under #118 scope)
- #6, #7, #8 (OpenBrowser fallback output, Auth0 wildcard redirect URI note, `CreatedAt` as string)
- #9–#12 (security nits already noted as LGTM)
- Test gap: Bearer prefix stripping, `resolveAuth0Config` fallback precedence

Kevin-layerV:
## Review Response (round 6)

Both reviews come back as approve-with-polish / ready-to-merge — thanks.

**New items added to #120:**
- `GetQuota` silent-swallow on `auth status` (both reviews)
- `extractAuthURL` fragility in tests (second review)
- `time.After` in test poll loop → `time.NewTicker` (second review)
- `--no-browser` help text for headless/SSH sessions
- `govulncheck@latest` pinning (already tracked)
- `saveAuthConfig` comment duplication / `LoadProfile` invariant comment
- `TokenResponse.ExpiresIn`/`TokenType` unused fields comment
- `callbackHandler` three near-identical select blocks → `tryDeliver` helper

**Not actionable now:**
- Error string capitalization (`Auth0` is a proper noun; lint passes in CI)
- `DefaultClientID` ldflags wiring — intentionally deferred pending Auth0 SPA provisioning
- `--profile` test gap — `--profile` round-trip is covered by `TestAuthLoginCustomProfile`; noted in #120 if reviewers feel otherwise

CI is fully green on commit `c173d2b`. Re-review from @justin-layerv was already requested after `a39dd8f` and is still pending.

Kevin-layerV:
## Review Response (round 7)

Two blocking items addressed in commit 0158868.

**Login flow timeout** — `WaitForToken` now runs under a `context.WithTimeout(ctx, 10*time.Minute)`. If the user abandons the browser, the CLI exits with *"authentication timed out: browser flow not completed within 10 minutes"* rather than hanging forever.

**`localhost` → `127.0.0.1`** — Callback listener now binds to `127.0.0.1:0` and the redirect URI uses `http://127.0.0.1:%d/callback`. Prevents the dual-stack ambiguity where `localhost` resolves to `::1` while Auth0's callback targets `127.0.0.1`. Comment added explaining the rationale.

**Deferred to #120:** `DefaultClientID` ldflags injection, coverage threshold, `govulncheck` pinning, `KeyID` comment, `Config.KeyID` read-only exposure, remaining nits.

Kevin-layerV:
## Review Response (round 8)

Thanks — addressed the two blocking items.

**#1 Redirect URI mismatch in docs** — Fixed. PR description step 3 now reads `http://127.0.0.1:*/callback` with a note that Auth0 treats `localhost` and `127.0.0.1` as distinct. Stale `RedirectURI` struct comment (`"localhost callback URL"`) also updated to `"127.0.0.1 callback URL"` in commit 3d66c14.

**#4 Error string capitalization** — `Auth0` is a proper noun; `ST1005` does not flag proper nouns. Lint has passed on every recent commit including this one. No change needed.

**Deferred to #120:**
- #2 Atomic config save (pre-existing `config.go` behavior, not introduced in this PR)
- #3 Full key in recovery path banner — added to #120 with the "keep this secret" + `qurl config set api_key` suggestion
- Nits (validateScopes map, shared HTTP timeout const, `printLoginSuccess` faint param, `filepath.Join`, `LoginSession.Close` guard)

Kevin-layerV:
## Review Response (round 9)

Addressed items #2 and #3 in commit 5e07198. Item #1 already tracked.

**#2 `Config.KeyID` undocumented exclusion** — Added a three-line comment on the struct field explaining it is managed exclusively by `auth login`/`logout` and intentionally excluded from `validKeys`/`Get`/`Set` so `qurl config set key_id` is rejected as expected.

**#3 No test for quota-unavailable branch** — Added `TestAuthStatusQuotaUnavailable`: mock server returns 401 on `GET /v1/quota`; test asserts the output still contains "Authenticated" and also contains "unavailable". Locks in the round-6 fix.

**#1 Orphaned API key on save failure** — Pre-existing follow-up, already tracked in #118 (`logout --revoke`).

**Deferred to #120:** remaining nits (filepath.Join, validateScopes map, shared HTTP timeout const, printLoginSuccess faint param, LoginSession.Close guard).

Kevin-layerV:
## Review Response (round 10)

Two blocking items addressed in commit 0dd7f6e.

**`resolveEndpoint` ignoring config file** — Fixed. `resolveEndpoint` now accepts `*config.Config` alongside `*globalOpts`. Both call sites (`runAuthLogin` and `authStatusCmd`) load the profile config and pass it through, matching the pattern already used in `root.go:119`. Flag and env still take precedence; config file is now the correct fallback.

**`DefaultClientID` as `const` prevents ldflags injection** — Fixed. Converted from `const` to `var` with a doc comment showing the `-ldflags "-X .../auth.DefaultClientID=<id>"` invocation for release builds. `nolintlint:gochecknoglobals` annotation added since this is an intentional ldflags injection point.

**Deferred to #120:** `govulncheck` pinning, `Cache-Control: no-store` on callback HTML, preflight config check, remaining nits.

Kevin-layerV:
## Review Response (round 11)

Two items addressed in commit af07d62.

**Preflight config check (comment 1, item 2)** — `runAuthLogin` now loads and validates the config at the very start, before `StartLogin`. A malformed `~/.config/qurl/config.yaml` now returns `"load config: ..."` immediately — before the OAuth flow starts, before the browser opens, and before any server-side key is minted. `TestAuthLoginMalformedConfig` updated to test this fast-fail path directly via `runAuthCmd` (no longer needs the full callback machinery).

**`CreateAPIKey` uses `loginCtx` (comment 2, item 1)** — Key creation is now bounded by the same 10-minute deadline as `WaitForToken`. Comment added explaining the intent.

**Deferred to #120:** callback channel poisoning by probe, goroutine leak if `WaitForToken` not called, `govulncheck` pinning, `serverShutdownTimeout` increase, remaining nits.

Kevin-layerV:
## Review Response (round 12)

Both "ready to merge after" items addressed in commit 590a715.

**#1 Full API key on save failure** — The recovery path now prints only `keyResp.KeyPrefix + "..."` (never the full secret), along with a portal URL and `qurl config set` hint. Key is no longer exposed to CI logs, shell history, or screen recordings.

**#2 Missing access_token validation** — `exchangeCode` now checks `token.AccessToken == ""` after unmarshal and returns `"token response missing access_token"` immediately, rather than falling through to a confusing `401` on `/v1/api-keys`.

**Deferred to #120:** `govulncheck` pinning, coverage threshold, redundant config loads, hostname in key name, `validateScopes` map allocation, `ProfilePath` error swallow, `OpenBrowser` unit test, `ReadTimeout`/`WriteTimeout` on callback server.

Kevin-layerV:
## Review Response (round 13)

Three items addressed in commit a001af6.

**#4 `loginCtx` for `CreateAPIKey` risks orphaned key** — Fixed. `CreateAPIKey` now uses a fresh `context.WithTimeout(ctx, 30*time.Second)` (`keyCtx`), independent of the browser-flow deadline. A user who takes 9m59s in the browser no longer risks the key-creation call timing out with ~1 second left. Comment explains the rationale.

**#3 Silent `LoadProfile` error in `authStatusCmd`** — Added a comment on the `statusCfg, _ :=` line: "Error intentionally ignored: resolveAPIKeyWithSource already surfaces malformed config above, so this path is unreachable for a bad config file."

**Workflow nit** — Removed unused `pull-requests: read` permission from `cli.yml`.

**Deferred to #120:** key truncation (#1), duplicate key-resolution logic (#2), `OpenBrowser` context comment, `printLoginSuccess` struct, `CreateAPIKey` URL construction.

Kevin-layerV:
## Review Response (round 14)

Two items addressed in commit 048cfdc.

**`exchangeCode` inherits near-expired `loginCtx`** — Fixed. `WaitForToken` now gives `exchangeCode` a fresh `context.WithTimeout(context.Background(), 30*time.Second)`, matching the `keyCtx` fix for `CreateAPIKey`. A user who takes 9m59s in the browser no longer hits a ~1s deadline for the token POST.

**`QURL_AUTH0_URL` has no HTTPS guard** — Fixed. If the env var is set, it must use `https://` (or `http://localhost` / `http://127.0.0.1` for local test servers). A plain-http override to an arbitrary host now returns an early error rather than silently forwarding the JWT.

**Deferred to #120:** `Close()` response truncation (cosmetic on localhost), `saveAuthConfig` second `LoadProfile` call, CI build matrix, coverage threshold.

Kevin-layerV:
## Review Response (round 15)

Item (1) fixed in commit d9c8864.

**`exchangeCode` used `context.Background()` losing cancellation** — Fixed. Now derives from `ctx` (`context.WithTimeout(ctx, 30*time.Second)`) so Ctrl-C cancels the token POST while the fresh 30s deadline still prevents a near-expired loginCtx from timing out the network call.

**Deferred to #120:** response body truncation in apikey.go (#2), redundant config reads (#3), status command comment accuracy (#4), govulncheck pinning, coverage threshold.

Kevin-layerV:
## Review Response (round 16)

Thanks for the thorough pass. All CI is green. Addressing each item:

**Item 1 — `QURL_AUTH0_URL` allows arbitrary HTTPS hosts:** The existing guard prevents plain-HTTP downgrades to arbitrary hosts; the concern about a shell-access attacker pointing at `https://attacker.example` is acknowledged and deferred to #120. In practice, shell access already implies full credential access, and this override is intended only for local test servers — a "for testing only" comment will be added in the follow-up.

**Items 2–6 — all non-blocking nits:** Deferred to #120:
- `OpenBrowser` dead-code context (#2)
- `CreateAPIKey` missing User-Agent header (#3)
- `authStatusCmd` double config load (#4)
- Token-exchange error surfacing raw body (#5)
- `--scopes` deduplication (#6)

justin-layerv (dismissed):

Kevin-layerV:
## Review Response (round 10)

Three new commits on top of the round-9 state:

**Stashed fixes from previous agent (commit 7834a29):**
- Brand spelling: 8x `QURL` → `qURL` in user-visible cobra Short/Long strings and doc comments (env var names and Go identifiers left as QURL per CLAUDE.md)
- `govulncheck`: pinned `@latest` → `@v1.1.4` in cli.yml
- `LoginSession.Close`: made idempotent via `sync.Once` (round-9 item #4 — StartLogin ctx-cancel goroutine)
- `StartLogin`: added ctx-cancellation goroutine that calls `Close` when ctx is canceled, tearing down the callback server even if `WaitForToken` is never called
- `OpenBrowser`: error message now includes `browserErr`
- `TestMissingAPIKey` / `TestNewClient_MissingAPIKey`: set `HOME=t.TempDir()` to isolate from real `~/.config/qurl/config.yaml`

**Misspell lint fix (commit 97c730e):**
- `cancelled` → `canceled` in two new comments in pkce.go (caught by CI misspell linter)

**Simplify pass (commit c282c9c — closes items from #120):**
- pkce.go: replaced bare `30*time.Second` with `defaultHTTPTimeout` constant (already defined in apikey.go, same package)
- auth_test.go: `home + "/.config/qurl"` → `filepath.Join(home, ".config", "qurl")` in all three malformed-config test helpers

**Still deferred to #120:** validateScopes map lifted to package scope, printLoginSuccess faint param, generateState/generateCodeVerifier shared helper, printLoginSuccess stdout/stderr refactor.

Kevin-layerV:
## Review Response (round 11)

**Goroutine leak in StartLogin (commit 88fe23c — the only hard blocker):**

Added a `done chan struct{}` field to `LoginSession`, initialized in `StartLogin`, and closed inside `closeOnce.Do` in `Close`. The cleanup goroutine now selects on both `ctx.Done()` and `session.done`, so it exits promptly when `Close` is called directly (e.g. via `WaitForToken`'s defer) rather than blocking until `ctx` is canceled. Closing `done` inside `sync.Once` prevents any double-close race. All tests pass with `-race`.

**Non-blocking items (keeping out of scope to avoid another CI cycle):**

- Recovery hint reachability (#2) — noted, adding to #120
- `OpenBrowser` localhost allow-list (#3) — noted, adding to #120  
- Coverage threshold / `bc` dependency — 40% is the agreed floor, not a target; leaving as-is for now
- Belt-and-suspenders context comment — acknowledged, the comment is imprecise but the behavior is correct for the CLI's use; adding to #120
- Second `config.LoadProfile` in `authStatusCmd` — minor TOCTOU, adding to #120

Kevin-layerV:
## Review Response (round 12)

All four "issues worth fixing in this PR" from the round-11 review addressed in commit 783bc5d:

1. **Coverage threshold 40% → 70%** (`cli.yml`) — actual coverage is 74.2%; floor set to 70%. Also replaced `bc` with `awk` for portability.

2. **`WriteTimeout: 10 * time.Second`** added to the local callback `http.Server` in `pkce.go`.

3. **`OpenBrowser` loopback allow-list** — extracted `isAllowedBrowserURL` helper; allows `https://`, `http://127.0.0.1`, and `http://localhost`. `nolint` comments updated to reflect the expanded validation. `QURL_AUTH0_URL` error message also restructured to `"invalid QURL_AUTH0_URL: ..."` so it starts lowercase.

4. **Capitalized error messages** — `"Auth0 client ID not configured..."` → `"auth0 client ID not configured..."`. `"QURL_AUTH0_URL must use..."` restructured to `"invalid QURL_AUTH0_URL: must use..."` (env var can't be lowercased, so it's no longer the leading word). Test updated to match.

**Smaller nits** — deferring to #120 (the `perfsprint` nit on `faint.Sprintf`, `saveAuthConfig` double-load, `callbackHandler` method check, `extractAuthURL` regex ANSI robustness, `successHTML`/`alreadyDoneHTML` `window.close()`, no-retry on token exchange).

Kevin-layerV:
## Review Response (round 13)

Addressed the two blocker bugs and the two smells from round-12, in commit a946882:

**#1 Lost API key on save failure** — now prints `keyResp.APIKey` (the full key) to stderr in the `saveErr != nil` branch with a "Copy this key now — it cannot be retrieved again" warning. Previous behavior (prefix only + misleading portal link) was replaced.

**#2 Doubled error prefix** — removed the `fmt.Errorf("create API key: %w", err)` wrapper from `auth.go`; `apikey.go` already prefixes all errors from `CreateAPIKey`, so callers get clean single-prefixed messages.

**#3 Dead context in browser.go** — dropped `ctx := context.Background()` and switched to `exec.Command(...)` (no context); `Start()` doesn't honor a context anyway.

**Auth0 capitalization** — restructured to `"missing Auth0 client ID: ..."` so Auth0 (proper noun) is not the leading word, satisfying both the Go error-string convention and the proper-noun spelling.

**Deferred to #120:** `isAllowedBrowserURL` table tests (#4), `authStatusCmd` double config load (#5), `callbackHandler` method check, `extractAuthURL` ANSI robustness, `successHTML`/`alreadyDoneHTML` window.close(), no-retry on token exchange.

This addresses both items the round-12 review flagged for pre-merge resolution. Ready for human merge review.

Kevin-layerV:
## Review Response (round 14)

Addressed the single security issue flagged before merge (#1 — HasPrefix bypass) in commit 5d928fa:

**#1 QURL_AUTH0_URL HasPrefix bypass** — replaced the `strings.HasPrefix` chain with `IsAllowedOriginURL` (exported from `browser.go`), which uses `url.Parse` + `u.Hostname()` checks, matching what `isAllowedBrowserURL` already does correctly. Also added a userinfo rejection (`u.User != nil → false`) to prevent `https://user@evil.com` from passing the https-scheme check. New `browser_test.go` table-tests the allow/deny matrix: https://, http://127.0.0.1, http://localhost (allow); http://evil.com, loopback-lookalike hostnames, userinfo URLs, javascript:, file://, empty string (deny).

**Other items deferred to #120:** #2 (CI build missing client-ID guard), #3 (logout Long help text), #4 (cfg threading in authStatusCmd), #5 (always print URL regardless of browser success), #6 (statusWriter abstraction), concurrent callback test, --profile round-trip test.

This is the last blocker. Ready for human merge review.

justin-layerv:
This PR has been auto-closed due to age. Please reopen if still applicable.

PR #37, opened 2026-03-31
feat(shared)!: align client types with latest API spec

Kevin-layerV:
## Summary
- Update all shared client types to match current OpenAPI spec (breaking changes)
- Create endpoint moved from `POST /v1/qurl` to `POST /v1/qurls`
- `CreateInput.Description` renamed to `Label`; added `SessionDuration`, `CustomDomain`
- `QURL` struct: removed `OneTimeUse`, `MaxSessions`, `QURLLink`, `AccessPolicy`; added `Tags`, `CustomDomain`
- `ExtendInput` merged into `UpdateInput` with `Tags` support
- `MintLink()` now accepts optional `*MintLinkInput` with full per-QURL settings
- New `BatchCreate()` method for `POST /v1/qurls/batch`
- `AccessPolicy` expanded with `UserAgentAllowRegex`, `UserAgentDenyRegex`, `AIAgentPolicy`
- CLI commands updated: `--description` → `--label`, new filter/mint/batch flags
- Output formatters updated for new field shapes
- `Create()` and `List()` now accept pointer inputs (lint: hugeParam)
- Removed `StatusExpired` and `StatusConsumed` constants

## Test plan
- [x] `go build ./...` compiles all apps (CLI, Slack)
- [x] `go test ./...` passes all tests
- [x] `golangci-lint run` reports 0 issues
- [x] `gofmt -l .` reports no formatting issues
- [x] `go vet ./...` passes
- [x] Shared client tests pass (including new TestBatchCreate, TestListWithDateFilters, TestMintLinkWithInput)
- [x] CLI command tests pass
- [x] Formatter tests pass

### DIFF
--- a/apps/cli/cmd/commands_test.go
+++ b/apps/cli/cmd/commands_test.go
@@ -27,8 +27,9 @@ func newMockServer(t *testing.T) *httptest.Server {
 		w.Header().Set("Content-Type", "application/json")
 
 		switch {
-		case r.Method == http.MethodPost && r.URL.Path == "/v1/qurl":
+		case r.Method == http.MethodPost && r.URL.Path == "/v1/qurls":
 			apiEnvelope(t, w, map[string]any{
+				"qurl_id":     "q_test123",
 				"resource_id": "r_test123",
 				"qurl_link":   "https://qurl.link/at_abc",
 				"qurl_site":   "https://r_test123.qurl.site",
--- a/apps/cli/cmd/completion.go
+++ b/apps/cli/cmd/completion.go
@@ -47,7 +47,7 @@ func resourceIDCompletion(opts *globalOpts) func(*cobra.Command, []string, strin
 
 		ctx, cancel := context.WithTimeout(cmd.Context(), 3*time.Second)
 		defer cancel()
-		result, err := c.List(ctx, client.ListInput{Limit: 20, Query: toComplete})
+		result, err := c.List(ctx, &client.ListInput{Limit: 20, Query: toComplete})
 		if err != nil {
 			return nil, cobra.ShellCompDirectiveNoFileComp
 		}
--- a/apps/cli/cmd/create.go
+++ b/apps/cli/cmd/create.go
@@ -21,6 +21,10 @@ func createCmd(opts *globalOpts) *cobra.Command {
 	cmd := &cobra.Command{
 		Use:   "create <target-url>",
 		Short: "Create a qURL for a target URL",
+		Long: `Creates a new qURL for the given target URL.
+
+Note: AccessPolicy overrides (geo-fencing, IP restrictions) are only available
+via the SDK or API; the CLI exposes the common creation flags only.`,
 		Example: `  qurl create https://api.example.com/data
   qurl create https://internal.example.com --expires 1h --one-time
   qurl create https://dashboard.example.com -l "Admin access" -e 7d`,
--- a/apps/cli/cmd/create.go
+++ b/apps/cli/cmd/create.go
@@ -62,12 +62,12 @@ func createCmd(opts *globalOpts) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&label, "label", "l", "", "Label for the QURL")
+	cmd.Flags().StringVarP(&label, "label", "l", "", "Label for the qURL")
 	cmd.Flags().StringVarP(&expiresIn, "expires", "e", "", "Expiration duration (e.g., 1h, 24h, 7d)")
 	cmd.Flags().BoolVar(&oneTimeUse, "one-time", false, "Single-use token (consumed after first access)")
 	cmd.Flags().IntVar(&maxSessions, "max-sessions", 0, "Maximum concurrent sessions (0 = unlimited)")
 	cmd.Flags().StringVar(&sessionDuration, "session-duration", "", "Session duration (e.g., 30m, 1h)")
-	cmd.Flags().StringVar(&customDomain, "custom-domain", "", "Custom domain for the QURL")
+	cmd.Flags().StringVar(&customDomain, "custom-domain", "", "Custom domain for the qURL")
 
 	return cmd
 }
--- a/apps/cli/cmd/create.go
+++ b/apps/cli/cmd/create.go
@@ -10,10 +10,12 @@ import (
 
 func createCmd(opts *globalOpts) *cobra.Command {
 	var (
-		description string
-		expiresIn   string
-		oneTimeUse  bool
-		maxSessions int
+		label           string
+		expiresIn       string
+		oneTimeUse      bool
+		maxSessions     int
+		sessionDuration string
+		customDomain    string
 	)
 
 	cmd := &cobra.Command{
@@ -21,7 +23,7 @@ func createCmd(opts *globalOpts) *cobra.Command {
 		Short: "Create a QURL for a target URL",
 		Example: `  qurl create https://api.example.com/data
   qurl create https://internal.example.com --expires 1h --one-time
-  qurl create https://dashboard.example.com -d "Admin access" -e 7d`,
+  qurl create https://dashboard.example.com -l "Admin access" -e 7d`,
 		Args: cobra.ExactArgs(1),
 		RunE: func(cmd *cobra.Command, args []string) error {
 			if err := validateURL(args[0]); err != nil {
@@ -37,14 +39,16 @@ func createCmd(opts *globalOpts) *cobra.Command {
 			}
 
 			input := client.CreateInput{
-				TargetURL:   args[0],
-				Description: description,
-				ExpiresIn:   expiresIn,
-				OneTimeUse:  oneTimeUse,
-				MaxSessions: maxSessions,
+				TargetURL:       args[0],
+				Label:           label,
+				ExpiresIn:       expiresIn,
+				OneTimeUse:      oneTimeUse,
+				MaxSessions:     maxSessions,
+				SessionDuration: sessionDuration,
+				CustomDomain:    customDomain,
 			}
 
-			result, err := c.Create(cmd.Context(), input)
+			result, err := c.Create(cmd.Context(), &input)
 			if err != nil {
 				return fmt.Errorf("create QURL: %w", err)
 			}
@@ -58,10 +62,12 @@ func createCmd(opts *globalOpts) *cobra.Command {
 		},
 	}
 
-	cmd.Flags().StringVarP(&description, "description", "d", "", "Description")
+	cmd.Flags().StringVarP(&label, "label", "l", "", "Label for the QURL")
 	cmd.Flags().StringVarP(&expiresIn, "expires", "e", "", "Expiration duration (e.g., 1h, 24h, 7d)")
 	cmd.Flags().BoolVar(&oneTimeUse, "one-time", false, "Single-use token (consumed after first access)")
 	cmd.Flags().IntVar(&maxSessions, "max-sessions", 0, "Maximum concurrent sessions (0 = unlimited)")
+	cmd.Flags().StringVar(&sessionDuration, "session-duration", "", "Session duration (e.g., 30m, 1h)")
+	cmd.Flags().StringVar(&customDomain, "custom-domain", "", "Custom domain for the QURL")
 
 	return cmd
 }
--- a/apps/cli/cmd/create.go
+++ b/apps/cli/cmd/create.go
@@ -32,6 +32,9 @@ func createCmd(opts *globalOpts) *cobra.Command {
 			if err := validateDuration(expiresIn); err != nil {
 				return err
 			}
+			if err := validateDuration(sessionDuration); err != nil {
+				return err
+			}
 
 			c, err := opts.newClient()
 			if err != nil {
--- a/apps/cli/cmd/extend.go
+++ b/apps/cli/cmd/extend.go
@@ -4,8 +4,6 @@ import (
 	"fmt"
 
 	"github.com/spf13/cobra"
-
-	"github.com/layervai/qurl-integrations/shared/client"
 )
 
 func extendCmd(opts *globalOpts) *cobra.Command {
@@ -30,9 +28,7 @@ func extendCmd(opts *globalOpts) *cobra.Command {
 				return err
 			}
 
-			qurl, err := c.Extend(cmd.Context(), args[0], client.ExtendInput{
-				ExtendBy: extendBy,
-			})
+			qurl, err := c.Extend(cmd.Context(), args[0], extendBy)
 			if err != nil {
 				return fmt.Errorf("extend QURL: %w", err)
 			}
--- a/apps/cli/cmd/list.go
+++ b/apps/cli/cmd/list.go
@@ -38,12 +38,12 @@ func listCmd(opts *globalOpts) *cobra.Command {
 				}
 			}
 
-			var errs []error
-			errs = append(errs, validateRFC3339("created-after", createdAfter))
-			errs = append(errs, validateRFC3339("created-before", createdBefore))
-			errs = append(errs, validateRFC3339("expires-before", expiresBefore))
-			errs = append(errs, validateRFC3339("expires-after", expiresAfter))
-			if err := errors.Join(errs...); err != nil {
+			if err := errors.Join(
+				validateRFC3339("created-after", createdAfter),
+				validateRFC3339("created-before", createdBefore),
+				validateRFC3339("expires-before", expiresBefore),
+				validateRFC3339("expires-after", expiresAfter),
+			); err != nil {
 				return err
 			}
 
--- a/apps/cli/cmd/list.go
+++ b/apps/cli/cmd/list.go
@@ -76,10 +76,10 @@ func listCmd(opts *globalOpts) *cobra.Command {
 	cmd.Flags().StringVar(&status, "status", "", "Filter by status (active, revoked)")
 	cmd.Flags().StringVar(&query, "query", "", "Search label, description, and target URL")
 	cmd.Flags().StringVar(&sort, "sort", "", "Sort field:direction (e.g., created_at:desc)")
-	cmd.Flags().StringVar(&createdAfter, "created-after", "", "Filter QURLs created after this date (RFC3339)")
-	cmd.Flags().StringVar(&createdBefore, "created-before", "", "Filter QURLs created before this date (RFC3339)")
-	cmd.Flags().StringVar(&expiresBefore, "expires-before", "", "Filter QURLs expiring before this date (RFC3339)")
-	cmd.Flags().StringVar(&expiresAfter, "expires-after", "", "Filter QURLs expiring after this date (RFC3339)")
+	cmd.Flags().StringVar(&createdAfter, "created-after", "", "Filter qURLs created after this date (RFC3339)")
+	cmd.Flags().StringVar(&createdBefore, "created-before", "", "Filter qURLs created before this date (RFC3339)")
+	cmd.Flags().StringVar(&expiresBefore, "expires-before", "", "Filter qURLs expiring before this date (RFC3339)")
+	cmd.Flags().StringVar(&expiresAfter, "expires-after", "", "Filter qURLs expiring after this date (RFC3339)")
 
 	return cmd
 }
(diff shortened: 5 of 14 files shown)


vikramlayerv:
## Code Review — 4-Agent Parallel Review

**Overall**: Clean API migration with good test expansion, proper nil guards, and correct Content-Type fix. Findings below grouped by severity.

---

### Bugs (4)

**1. `mint.go` — `oneTimeUse: false` and `maxSessions: 0` unconditionally serialized when any other flag is changed**
When `hasInput` is true (because e.g. `--expires-in` was set), `input.OneTimeUse = oneTimeUse` writes `false` and `input.MaxSessions = maxSessions` writes `0` unconditionally. With `omitempty` on these fields, `false`/`0` are omitted — but if the API uses presence-vs-absence semantics (e.g., "absent = inherit from QURL default; false = override"), this silently overrides server defaults. Should gate behind `cmd.Flags().Changed("one-time")` like `update.go` does for `--description`.

**2. `create.go` — `--one-time` / `--max-sessions` flags still wired but invisible in output**
`QURL` struct removed `OneTimeUse`/`MaxSessions`, and `FormatQURL` no longer displays them. But `create.go` still has these flags and sends them in `CreateInput`. Users can set them on create but never see them in any output — invisible dead zone. Either remove the flags or restore formatter display.

**3. `Extend()` silently narrows — no longer supports absolute `ExpiresAt`**
Old `ExtendInput` had `ExpiresAt *time.Time`. New `Extend(ctx, id, duration string)` only supports relative durations. Callers using absolute time must switch to `Update()` directly with no compile-time guidance. Should document this or keep `Extend` accepting the old struct.

**4. `BatchItemResult` missing `QurlID` field**
`CreateOutput` now has `QurlID`, but `BatchItemResult` does not. If the batch endpoint returns `qurl_id` per item (consistent with single create), it will be silently dropped.

---

### Concerns (10)

**5. `update.go` — no mutual-exclusion check for `--extend-by` + `--expires-at`**
Both flags can be set simultaneously, sending conflicting time directives in the same PATCH body. The API behavior is unpredictable. Should reject the combination client-side.

**6. `mint.go` — same: no mutual-exclusion for `--expires-in` + `--expires-at`**

**7. `QURL.CustomDomain *string` missing `omitempty`**
Every other optional field has `omitempty`. Without it, `custom_domain` serializes as `null` in every response that has no custom domain, polluting JSON output.

**8. `QURL` struct has no `Label` field — label set on create cannot be read back**
`CreateOutput` has `Label`, but `QURL` (returned by Get/List/Update) does not. If the API returns `label` on GET responses, it's silently dropped.

**9. `list.go` date filter flags accept raw strings without validation**
`--created-after`, `--expires-before` etc. pass strings straight to the API with no RFC3339 validation. Every other time flag in this PR (`--expires-at` in update.go, mint.go) validates with `time.Parse`. Inconsistent.

**10. `formatter.go` — `colorStatus` falls through unstyled for unknown statuses**
Old code colored `expired`/`consumed` red. New code only colors `revoked` red; everything else is plain text. If the server ever returns an old status, it renders without color. Consider `default: return f.yellow.Sprint(status)` for unexpected values.

**11. `StatusExpired`/`StatusConsumed` removed — confirm API no longer emits these**
Any caller doing `if qurl.Status == client.StatusExpired` gets a compile error. If the server still returns these statuses for old QURLs, callers need raw string comparisons.

**12. `UsageInfo.ActiveQURLsPercent` changed from `float64` to `*float64`**
Existing callers reading this without a nil check will panic. All callers need auditing.

**13. `MintLinkInput`/`CreateInput` bool/int fields with `omitempty`**
`OneTimeUse: false` and `MaxSessions: 0` are silently omitted from JSON. Cannot explicitly send `false`/`0` to override server defaults. Should use pointer types if presence semantics matter.

**14. Slack `handler_test.go` not updated**
The test still returns responses without `qurl_id`/`Label`. Passes only because assertions check Slack response format, not API fields. No coverage for new fields.

---

### Test Gaps (8)

**15. `TestCreate` does not verify `Label` field in request body** — the key rename from `Description` is untested at the wire level.

**16. No test for `Create` with `CustomDomain` or `SessionDuration`** — both new `CreateInput` fields have zero coverage.

**17. No test for `Update` with `Tags`** — the `*[]string` pointer-to-slice pattern is unusual and particularly prone to serialization surprises. Untested.

**18. No test for `Update{ExpiresAt: &t}` path** — the absolute-time update codepath is untested.

**19. `Content-Type` header conditional not tested** — moved inside body-check block, but no test asserts GET/DELETE omit it or POST includes it.

**20. `list --status expired` rejection not tested** — breaking change for CLI users with no regression test.

**21. `TestListWithDateFilters` only checks 2 of 4 date params** — `CreatedBefore` and `ExpiresAfter` untested.

**22. `TestMintLink(nil)` does not assert the request has no body** — a regression where body is always sent would not be caught.

---

### Nits (4)

**23.** `update.go` example still only shows `-d ""` without showcasing `--tags`/`--extend-by`/`--expires-at`.
**24.** `mint.go` example only shows one line; new flags (`--label`, `--expires-at`, `--session-duration`) undocumented.
**25.** `completion.go` silently includes revoked QURLs in tab-completion (pre-existing, no `Status: active` filter).
**26.** `BatchCreate` takes `[]CreateInput` (values) while `Create` takes `*CreateInput` (pointer) — inconsistent API surface.

---

### Positives

- ✅ Content-Type header fix is correct — only set when body is present
- ✅ MintLink nil/non-nil body handling is clean and safe
- ✅ `url.PathEscape` used consistently on all ID path segments
- ✅ `List(nil)` defensive default and `Create(nil)` fast-fail are well-tested
- ✅ `Extend` delegation to `Update` is clean and tested
- ✅ `BatchCreateValidation` covers nil, empty, and >100 bounds without network
- ✅ `BatchCreatePartialFailure` test is genuinely useful — 207-style mixed response
- ✅ All Create/List/MintLink callers (CLI, Slack, completion) updated for pointer inputs
- ✅ `update.go` Tags `*[]string` + `Changed("tags")` guard is correct

justin-layerv (dismissed):
## Review: feat(shared)!: align client types with latest API spec

Well-scoped breaking change — all callers updated in the same PR, tests are thorough, CI green. Need a few fixes before merge.

### Must fix

**1. `CustomDomain` missing `omitempty` on QURL struct**

Every other optional field on `QURL` uses `omitempty`. `CustomDomain *string` without it will serialize as `null` in every JSON response where it's unset. Either add `omitempty` (if it should match the rest) or add a comment explaining why it's intentionally different.

**2. Description-to-Label inconsistency between CreateInput and UpdateInput**

`CreateInput.Description` was renamed to `Label`, but `UpdateInput` still uses `Description`. The CLI update command still exposes `--description`. Is this intentional per the API spec (create uses `label`, update uses `description`)? If so, add a comment on `UpdateInput.Description` explaining why it diverges from `CreateInput.Label`. If it's a miss, rename to align.

**3. Hardcoded `testResourceID` in `TestMintLinkWithInput`**

Use the `testResourceID` constant instead of the hardcoded value. One-liner consistency fix.

### Should fix

**4. Client-side date filter validation**

`CreatedAfter`, `CreatedBefore`, `ExpiresBefore`, `ExpiresAfter` on `ListInput` are raw strings passed directly as query params. The CLI help says RFC3339 but there's no client-side validation. The update and mint commands already validate date inputs with `time.Parse` — apply the same pattern to list filters for faster feedback and a better error message than whatever the server returns.

**5. Content-Type on MintLink when body is present**

The `do()` method sets Content-Type based on `req.Body != nil`, which works for `bytes.NewReader`. But MintLink is the only method that conditionally sends a body. Add an explicit `req.Header.Set("Content-Type", "application/json")` in MintLink when `input != nil` as defense-in-depth — don't rely on `do()` inferring it correctly for every future caller.

**6. BatchCreate slice by value**

`Create()` was changed to pointer to avoid `hugeParam` lint. `BatchCreate([]CreateInput)` copies the full slice of structs. With the 100-item limit it's fine today, but for consistency with the `Create()` change, consider `[]*CreateInput` or `*[]CreateInput`. Low priority but keeps the API surface consistent.

### Needs rebase

Branch is behind main. Rebase before merge.

Kevin-layerV:
All 6 items from the review have been addressed, plus one proactive fix:

1. **`CustomDomain` omitempty** — Added `omitempty` to `QURL.CustomDomain *string` to match all other optional fields.

2. **Description/Label divergence comment** — Added an explanatory comment to `UpdateInput.Description` documenting why it differs from `CreateInput.Label` (intentional per the `UpdateQurlRequest` OpenAPI schema, which uses `description` not `label`).

3. **Hardcoded resource ID in test** — Replaced the hardcoded string in `TestMintLinkWithInput` with the `testResourceID` constant.

4. **RFC3339 date filter validation** — Added `validateRFC3339` helper in `validate.go` and wired it with `errors.Join` in `list.go` to validate all four date flags up-front. Tests added in `validate_test.go`.

5. **Explicit `Content-Type` in MintLink** — Added `req.Header.Set("Content-Type", "application/json")` when `input != nil` as defense-in-depth per the review request.

6. **`BatchCreate` slice type** — Changed signature from `[]CreateInput` to `[]*CreateInput` for consistency with the `Create()` pointer change. All callers updated.

**Proactive fix**: Gated `OneTimeUse` and `MaxSessions` in `mint.go` behind `cmd.Flags().Changed()` checks to match the `update.go` pattern, preventing accidental zero-value overrides when only other flags are set.

CI is fully green (Lint, Test, Test All Apps, Build, CodeQL, TruffleHog, Vulnerability Scan — all pass).

Kevin-layerV:
Addressed all review comments and rebased onto main.

**Must fix (justin-layerv)**
1. **`CustomDomain` missing `omitempty`** — already had `omitempty`; added comment explaining why `*string` is used to distinguish nil vs empty on read (commit 3811958)
2. **Description-to-Label inconsistency** — `UpdateInput.Description` is intentional per API spec; added explanatory comment (commit b34c269)
3. **Hardcoded `testResourceID` in `TestMintLinkWithInput`** — `TestMintLinkWithInput` already used the constant; extended to fix remaining hardcoded literals in `TestGet` and `TestUpdate` path assertions too (this commit)

**Should fix**
4. **Client-side date filter validation** — `validateRFC3339` applied to all four list date filters (commit b34c269)
5. **Content-Type on MintLink** — explicit `req.Header.Set("Content-Type", "application/json")` added when input != nil (commit f72d231)
6. **BatchCreate slice by value** — changed to `[]*CreateInput` (commit 489569e)

**Rebase**
- Merged onto main (this commit). Conflicts in `shared/client/client.go` and `apps/cli/internal/output/formatter.go` resolved: all PR structural changes preserved, lowercase "qURL created" string from main kept in formatter.
- Bonus: fixed `TestMissingAPIKey` and `TestNewClient_MissingAPIKey` test isolation — both were failing on developer machines with a real `~/.config/qurl/config.yaml` present; now set `HOME=t.TempDir()` to prevent config bleed.

Kevin-layerV:
### Deferred: `update --tags` cannot clear tags from CLI

Filed #133 to track the `StringSliceVar` / clear-tags UX issue. `--tags ""` parses as `[""]` rather than the empty slice needed to hit the `Tags: &[]string{}` clear-tags path. The fix (a `--clear-tags` flag) is straightforward but intentionally out of scope for this breaking-change PR. The shared client layer correctly supports clearing via `*[]string`; this is a CLI-only surface issue.

### Content-Type in MintLink

Removed the redundant `req.Header.Set("Content-Type", ...)` that @justin-layerv originally requested as defense-in-depth. `do()` reliably sets the header whenever `req.Body != nil && req.Body != http.NoBody`. The "defense-in-depth" argument is thin because duplicated logic invites future divergence (e.g., if `do()` ever switches to `application/json; charset=utf-8`). Instead, the contract is now enforced by `TestMintLinkContentType` (added in this commit), which asserts:
- **nil input** → no Content-Type header
- **non-nil input** → `Content-Type: application/json`

This gives stronger guarantees than a redundant call-site set would, with zero duplication.

Kevin-layerV:
Addressed in a83264a:

**Addressed:**
1. **Duration validation** — Added `validateDuration()` to `mint --expires-in`, `mint --session-duration`, `create --session-duration`, and `update --extend-by`, consistent with the existing pattern in `create.go` and `extend.go`.
2. **`MintLink` godoc** — Added nil-vs-non-nil contract (bodiless POST vs JSON body) directly on the function signature.
3. **`List` nil handling asymmetry** — Documented that `nil` is silently coerced to `&ListInput{}` and explains the intentional asymmetry with `Create` (which rejects nil because `TargetURL` is required).
4. **`UpdateInput.SessionDuration` absent** — Confirmed against the `UpdateQurlRequest` schema in the API spec: `session_duration` is not in the spec. Added a comment to `UpdateInput` documenting this is intentional.
5. **`colorStatus` comment** — Added a comment clarifying that `expired`/`consumed` are removed from the spec; unknown statuses fall to default (uncolored).
6. **`TestMintLinkOmitEmpty`** — New test pins that `&MintLinkInput{}` marshals to `{}` at the wire level, locking the `omitempty` contract that `mint.go` relies on when unconditionally assigning `OneTimeUse`/`MaxSessions`.
7. **`AccessPolicy` CLI note** — Added a note to `mint --help` that `AccessPolicy` overrides are SDK/API-only.

**Already addressed / deferred:**
- `update --tags ""` cannot clear tags — deferred to #133 (previously noted).
- `maxBatchSize` hardcoded — reviewer agreed "reasonable as a defensive cap."
- Mutual-exclusion on time flags — optional per reviewer; server resolves.
- `ActiveQURLsPercent` not rendered in `FormatQuota` — field is SDK-only for now; no formatter change made.

Kevin-layerV:
Addressed in 21201ab:

**Blocking item (omitempty defeats explicit override)**
- `MintLinkInput.OneTimeUse` and `MaxSessions` changed to `*bool` / `*int` so callers can explicitly send `false`/`0` to override server-side qURL defaults. Previously, `omitempty` silently dropped these values regardless of the `Changed()` gate.
- `mint.go` updated: pointer fields are only populated when the flag was actually set (`Changed()` gate), consistent with `update.go`'s pattern for `--description`.
- `TestMintLinkOverrideToFalse` added: wire-level assertion that `*false` and `*0` appear in the JSON payload (not omitted).

**Nits**
- Package-level doc comment added to `shared/client/client.go` summarising the three nil-input conventions (Create errors on nil, List coerces nil to `&ListInput{}`, MintLink nil = bodiless POST).
- `MaxBatchSize` exported (was `maxBatchSize`). Test updated to use the constant.
- `TestBatchCreateNilItem` added: verifies `BatchCreate` rejects a nil item in a non-empty list with an index-specific error.
- `update.go` error text: "at least one flag must be set (e.g., …)" → "must set at least one of: --description, --tags, --extend-by, --expires-at"
- `update.go Long` help: added note that `--description` maps to the API's `description` field, intentionally different from `create --label` per the `UpdateQurlRequest` spec.

**Test gap (--expires-at parse paths)**
Deferred: the reviewer marked this minor, and the `validateRFC3339` helper is already tested in `validate_test.go`. The CLI-level end-to-end path for `--expires-at` parse errors is a follow-up item rather than a blocker.

Kevin-layerV:
Addressed in 66edb14. Responding to the remaining items:

**Addressed (code changes in 66edb14):**
- **3. `BatchCreate` empty `target_url`** — Added `if item.TargetURL == ""` check in the validation loop alongside the nil check. Added `TestBatchCreateEmptyTargetURL`.
- **4. `MaxBatchSize` provenance** — Added `// Matches the BatchCreateRequest schema constraint (api/openapi.yaml: items.maxItems=100)` comment. The spec has `minItems: 1, maxItems: 100` explicitly.
- **7. `UpdateInput` missing `AccessPolicy`/`CustomDomain`** — Expanded the doc comment to enumerate all four absent fields: `session_duration`, `one_time_use`, `access_policy`, `custom_domain`. The `UpdateQurlRequest` schema contains exactly four fields (`extend_by`, `expires_at`, `tags`, `description`); all four absent fields are intentional spec omissions, not oversights.

**Reply-only items:**

**1. `--description`/`-d` removal from `create`** — The `-d` flag on `create` was always the wrong abstraction: the API field being set on creation is `label` (the token label), not `description` (the resource-level description which is only patchable via `update`). Keeping a deprecated `-d` alias that mapped to `Label` would perpetuate the confusion. The `feat!:` breaking-change marker is the correct signal here. A separate follow-up PR can add a descriptive error message if `-d` is passed to `create` (`unknown shorthand flag: 'd' for 'create'` already tells users, but we could do better). Deferring to a separate issue.

**2. Content-Type behavior change on GET/DELETE** — Intentional improvement and a direct fix for the `MintLink(nil)` bodiless-POST correctness issue. GET/DELETE requests should not send `Content-Type: application/json` (the RFCs explicitly discourage it for methods with no defined body semantics). The PR description has been updated with a note in the behavior-changes section.

**5. `CreateInput` pointer asymmetry** — Intentional for now. On `Create`, `false` and absent are semantically equivalent initial values (the resource is being created with those defaults). On `MintLink`, `false` and absent are distinct: absent means "inherit the qURL's configured default", false means "override that default to false even if the qURL was set up as one-time". The pointer types on `MintLinkInput` reflect that distinction. `CreateInput` alignment can be a future PR if the API adds per-workspace defaults that need overriding on creation.

**6. `ActiveQURLsPercent *float64`** — Acknowledged. The `feat!:` marker covers it, and the migration notes in the PR description already call out the struct field shape changes. Will note it explicitly in the PR description's migration section.

Kevin-layerV:
Addressed in 67cc3e3. Responding to remaining items:

**Addressed (code changes in 67cc3e3):**
- **Should-fix: RFC3339 validation not uniform** — Added upfront `validateRFC3339("expires-at", expiresAt)` guarded by `Changed("expires-at")` in both `mint.go` and `update.go`, before the client is constructed. The inline `time.Parse` is retained for value extraction (error ignored since validation already passed). This matches the upfront validation style used for `--extend-by`/`--expires-in`.
- **Nit: brand spelling `QURL ID:`** — Fixed to `qURL ID:` in `formatter.go:FormatCreate`.

**Reply-only items:**

**Real issue: clearing tags from CLI** — Already filed as issue #133 (`update --tags` cannot clear tags from CLI). The `*[]string` tristate is fully functional via the SDK; the CLI limitation is tracked separately.

**Worth a thought: Create vs MintLink asymmetry** — Acknowledged. The asymmetry is intentional today (Create false/0 == absent; MintLink false/0 != absent), but the doc comment on `MintLinkInput` now notes the reasoning. Will add a cross-reference comment on `CreateInput` clarifying "no tenant defaults exist on create, so pointer semantics aren't needed here" in a follow-up.

**Nit: `update.go` validate `--expires-at` upfront** — Done in this commit.

**Nit: `list.go` status+date join** — Deferred to a separate cleanup PR to keep this PR focused.

**Nit: batch error message wording** — Deferred; the current message is clear enough and changing it would be a minor breaking change for any callers parsing error strings.

Kevin-layerV:
Addressed in ef9a7a5. Responding to remaining items:

**Addressed (code change in ef9a7a5):**
- **4. `mint --label \"\"` cannot clear** — Added a comment on `MintLinkInput.Label` explaining the omitempty asymmetry is intentional (there's no "clear label" concept on minting, unlike `UpdateInput.Description *string` which supports explicit empty-string clear).

**Reply-only items:**

**1. `QURL` struct lacks `QURLID`** — Checked the OpenAPI spec. `QurlData` (the schema for `Get`/`List`/`Update` responses) does not include `qurl_id` — that field only appears on `CreateQurlData` and `QurlSummary`. Resources are identified by `resource_id`; individual token IDs aren't surfaced on resource-level responses. The `QURL` struct is a faithful mapping. No gap.

**2. `colorStatus` unknown statuses** — The comment says "expired and consumed are removed from the API spec" but that's imprecise — they're in `QurlSummary.status` (token-level), not in `QurlData.status` (resource-level, enum `[active, revoked]`). Since `colorStatus` is only called on resource-level status from the `QURL` struct, `expired`/`consumed` shouldn't appear in practice. Will update the comment to clarify the scope.

**3. CLI breaking changes** — Acknowledged, covered by `feat!:`. The PR description's migration section already lists the flag rename and status-filter changes.

**5. Silently-dropped `time.Parse` errors** — The `// already validated above` comment is the safety net. Agreed that returning `(time.Time, error)` from `validateRFC3339` would be cleaner, but that changes its call sites in `list.go`. Deferring to a separate cleanup.

**6. `BatchCreate` validation boundary** — Acknowledged. Client-side checks are intentionally limited to cheaply-catchable mistakes (empty, nil, oversize). Field-level URL validation is correctly left to the server.

justin-layerv:
This PR has been auto-closed due to age. Please reopen if still applicable.

PR #38, opened 2026-04-01
chore: enforce supply chain age gating

Kevin-layerV:
## Summary
- Adds CI workflow that checks Go dependency publish dates against `proxy.golang.org`
- Blocks PRs introducing modules published less than 7 days ago
- Covers all go.sum files across the monorepo (apps/slack, apps/cli, libs)

## Why
Go has no native package manager age-gating. This CI check provides equivalent protection by querying the Go module proxy for publish timestamps.

## Override for critical updates
Add the `age-check-bypass` label to the PR.

## Test plan
- [ ] Workflow triggers on PRs that modify any `go.sum` in the monorepo
- [ ] Workflow skips when `age-check-bypass` label is present
- [ ] Existing dependencies (older than 7 days) pass the check

### DIFF
--- a/.github/workflows/dependency-age-check.yml
+++ b/.github/workflows/dependency-age-check.yml
@@ -1,0 +1,86 @@
+name: Dependency Age Check
+
+on:
+  pull_request:
+    paths:
+      - "**/go.sum"
+
+permissions:
+  contents: read
+
+jobs:
+  check-dependency-age:
+    runs-on: ubuntu-latest
+    if: ${{ !contains(github.event.pull_request.labels.*.name, 'age-check-bypass') }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Check Go module release ages
+        env:
+          MIN_AGE_DAYS: "7"
+        run: |
+          set -euo pipefail
+
+          failed=false
+          base_ref="origin/${{ github.event.pull_request.base.ref }}"
+
+          changed_sums=$(git diff --name-only "${base_ref}...HEAD" -- '**/go.sum' 'go.sum' || true)
+
+          if [ -z "$changed_sums" ]; then
+            echo "No go.sum changes detected"
+            exit 0
+          fi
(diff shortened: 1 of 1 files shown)


justin-layerv:
## Review feedback addressed

Rebased onto main and applied all hardening fixes already shipped to qurl-service, nhp, and traefik-plugins.

### Bugs fixed
- **Leading `+` not stripped** — `awk $1` included diff prefix, breaking all proxy lookups
- **Word splitting** — `for sum_file in $changed_sums` → `while IFS= read -r`

### Security (fail-closed)
- Proxy fetch failures now **block** instead of warn+continue
- Date parse failures also **block**

### Robustness
- `-m 10` curl timeout (replaces `--retry 2`)
- `date -d` error handling
- Pseudo-version skip (`v0.0.0-YYYYMMDDHHMMSS-hash`)
- Go module proxy encoding comment

### Nice-to-have
- `workflow_dispatch` trigger
- Section header comment
- `BASE_REF` fallback to `main`
- Checked count in summary output

### Deferred
- Parallel requests — 10-min timeout sufficient
- Concurrency group — not in the other repos' versions
- Label bypass permissions — repo settings, not code
- `fetch-depth` optimization — full history is safest for `git diff`

justin-layerv:
## CR feedback addressed

### Fixed
- **Broader pseudo-version regex** — `^v0\.0\.0-[0-9]{14}-` only caught one of three Go pseudo-version forms. Changed to `-0\.[0-9]{14}-[0-9a-f]{12}$` which matches the common suffix in all three forms (`v0.0.0-...`, `vX.Y.Z-0....`, `vX.Y.(Z+1)-0....`). Updated comment to explain all three forms.

### Deferred
- Rate limiting / parallelism — 10-min timeout sufficient
- `fetch-depth: 0` comment — intent clear from context
- `jq` availability check — guaranteed on ubuntu-latest
- Redundant `go.sum` in pathspec — harmless
- Bypass label permissions docs — process, not code

PR #60, opened 2026-04-18
ci: migrate dep age-check to ops-routines shared workflows

justin-layerv:
## Summary

Replaces the inline Go-only age-check (introduced in #38) with **four** hardened shims delegating to [`layervai/ops-routines`](https://github.com/layervai/ops-routines) reusable workflows pinned at v0.1.4:

- **Go modules** (`go.sum` diff)
- **Docker base images** (`apps/discord/Dockerfile` + compose YAML)
- **GitHub Actions pins** (`.github/workflows/**`, `.github/actions/**`)
- **pip / Python** (`apps/discord/requirements*.txt`, scoped via `working_directory`)

## Why

Centralizes the supply-chain age-gate so all LayerV repos run the same hardened implementation. Removes ~114 lines of bespoke shell from this repo and inherits future fixes (e.g., `main()` factoring, `workflow_sha`-pinned lib checkout, pyproject parsing).

## ⚠️ Action item: branch-protection check rename

Merging this PR changes the required-status-check name. **Repo admin must update branch protection before merge** or the gate becomes silently unenforced.

| Old (single check) | New (per-ecosystem checks) |
|---|---|
| `Dependency Age Check / check-dependency-age` | `Dependency Age Check (Go) / age-check` |
|  | `Dependency Age Check (Docker) / age-check` |
|  | `Dependency Age Check (Actions) / age-check` |
|  | `Dependency Age Check (pip) / age-check` |

If the old name is currently required, either:
- Drop the old required check **before** merging, then add the four new ones, or
- Add the four new ones first (they'll show as expected pending until they run on the next PR), then drop the old one.

## Bypass

Add the `age-check-bypass` label to skip the check. Use only for CVE-driven emergency updates.

## Test plan

- [ ] CI passes — only the actions shim should fire on this PR (we touch `.github/workflows/` but no go.sum/Dockerfiles/pyproject).
- [ ] Branch protection check names updated (admin task above).


justin-layerv:
## Review feedback addressed

Pushed [96e8652](https://github.com/layervai/qurl-integrations/commit/96e8652) addressing both review rounds:

| Item | Action |
|---|---|
| **Branch protection required-check rename** | ⚠️ **Needs admin** — old check `Dependency Age Check / check-dependency-age` is now split into `Dependency Age Check (Go|Docker|Actions|pip) / age-check`. Branch protection needs updating to require the new check names before this merges (otherwise the gate is silently unenforced). Flagging for repo admin. |
| Redundant `go.sum` root entry | **Fixed** — `**/go.sum` covers root |
| Actions shim blast radius (`/.github/workflows/**`) | **Confirmed working** — actions shim ran on the pip-shim commit and passed; this is the intended self-validation path |
| `timeout-minutes` safety net suggestion | **Not applicable** — `timeout-minutes` is not allowed on jobs that call a reusable workflow ([per GH docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow), only `if/name/needs/permissions/secrets/strategy/uses/with`). The reusable workflows in `ops-routines` already set `timeout-minutes: 5` internally. |
| Version skew between v0.1.3 and v0.1.4 | **Fixed** — bumped go/docker/actions to v0.1.4 (matches pip). v0.1.4 is byte-identical to v0.1.3 for non-pip reusables. |
| Docker shim fires on files that don't exist yet | **Forward-compatible by design** — only fires today on `apps/discord/Dockerfile` |
| Fail-closed behavior preserved | **Confirmed** — see `lib/age-check-go.sh` in ops-routines (registry-unreachable → exit 1) |

🤖 Co-authored with Claude Code

justin-layerv:
## Round 2 reviewed

Pushed [72d31ce](https://github.com/layervai/qurl-integrations/commit/72d31ce) and updated PR description:

| Item | Action |
|---|---|
| Pip shim hardcoded to `apps/discord/` — needs breadcrumb | **Fixed** — added inline scope-comment per the suggestion |
| Branch-protection rename should be in PR body | **Fixed** — surfaced as ⚠️ action item with the old/new check name table |
| Actions shim broad triggering | **Intentional** — broad-fire is safer; the reusable filters for actual `uses:` lines, no-op runs are cheap |
| Go shim watches `go.sum` not `go.mod` | **Defer** — the reusable diffs `go.sum` specifically; `**/go.mod` would just add no-op runs |
| Docker `**/Dockerfile*` breadth | **Acknowledged** — over-inclusive is safer for supply-chain checks |

🤖 Co-authored with Claude Code

justin-layerv:
## Round 3 reviewed — no new fixes needed

Re-examined every point in the latest review:

| Item | Disposition |
|---|---|
| Branch protection transition risk | **Already in PR body** as ⚠️ action item with old/new check name table |
| Docker `**/compose*.yml` breadth | **Intentional** — reviewer concedes "probably fine for an age check (worst case: runs unnecessarily and passes)". Broad-fire is the safer default for supply-chain checks. |
| Go shim watches `go.sum` only | **Inherited design** — the reusable script diffs `go.sum` specifically. Adding `**/go.mod` would just trigger no-op runs on go.mod-only edits. |
| Pip shim hardcoded to `apps/discord` | **Breadcrumb comment added** in [72d31ce](https://github.com/layervai/qurl-integrations/commit/72d31ce) per round-2 feedback. Reviewer confirms "sufficient for now". |
| **No `timeout-minutes` on jobs (recycled)** | **Not possible** — `timeout-minutes` is **not a valid field** on a job that calls a reusable workflow ([per GH docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow), only `if/name/needs/permissions/secrets/strategy/uses/with/concurrency` are allowed). The `ops-routines` reusable workflows all set `timeout-minutes: 5` internally — verified in [`.github/workflows/age-check-*.yml`](https://github.com/layervai/ops-routines/tree/main/.github/workflows). |
| Duplicated comment blocks across 4 shims | **Acknowledged** — for four small self-contained files this is fine. Pulling them into a shared template would create more coupling than it removes. |
| Fail-closed on proxy outage | **Confirmed** — `lib/age-check-{go,docker,actions,pip}.sh` all `failed=true` + `exit 1` when the registry is unreachable. See `_preflight.sh` and the per-ecosystem error paths. |

🤖 Co-authored with Claude Code

justin-layerv:
## Round 4 reviewed

Pushed [733f457](https://github.com/layervai/qurl-integrations/commit/733f457):

| Item | Action |
|---|---|
| `timeout-minutes` (recycled) | **Documented inline in YAML** — added a comment in each shim explaining why it's intentionally absent (not a valid field on `workflow_call` jobs per [GH docs](https://docs.github.com/en/actions/using-workflows/reusing-workflows#supported-keywords-for-jobs-that-call-a-reusable-workflow); reusable already sets `timeout-minutes: 5` internally). Should stop the recycled suggestion. |
| Concurrency comments aligned | **Done** — all four shims now use the longer form |
| Branch protection coordination | **Already in PR body** as ⚠️ action item with old/new check name table |
| DRY observation on bypass comment block | **Acknowledged** — inline context aids independent auditing of each shim |
| pip working_directory edge case | **Confirmed working** — script `cd`s to working_directory before running `git diff`, so no-changes case works correctly |

🤖 Co-authored with Claude Code

justin-layerv:
## Round 5 reviewed — approved, no fixes

| Item | Disposition |
|---|---|
| `pull_request` `types:` restriction to reduce label-trigger noise | **N/A** — default `pull_request` already only fires on `opened/synchronize/reopened` (per [GH docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request)); `labeled/unlabeled` requires opt-in. So labels don't re-trigger the shims. |
| `workflow_dispatch` `min_age_days` override input | **Defer** — speculative, ad-hoc testing better done by editing MIN_AGE_DAYS in a throwaway PR |
| Matrix strategy for shim count growth | **Defer** — at N=4 the file-per-ecosystem layout is clearer; revisit if N>5 |
| Branch protection rename tracking | **Already in PR body** as ⚠️ action item |

Approved. Ready to merge.

🤖 Co-authored with Claude Code

PR #67, opened 2026-04-18
ci: migrate dep-age shims to ops-routines-workflows v0.1.0

justin-layerv:
Validated on qurl-autosec + dashboard.

### DIFF
--- a/.github/workflows/dependency-age-check-actions.yml
+++ b/.github/workflows/dependency-age-check-actions.yml
@@ -45,4 +45,10 @@ jobs:
       min_age_days: 7
     permissions:
       contents: read
-    secrets: inherit
+    # Only the App credentials needed for cross-repo checkout of
+    # ops-routines (private/internal repo). Using `secrets: inherit`
+    # would forward ALL org/repo secrets to the reusable; this
+    # explicit block follows least-privilege.
+    secrets:
+      OPS_ROUTINES_APP_ID: ${{ secrets.OPS_ROUTINES_APP_ID }}
+      OPS_ROUTINES_APP_PRIVATE_KEY: ${{ secrets.OPS_ROUTINES_APP_PRIVATE_KEY }}
--- a/.github/workflows/dependency-age-check-actions.yml
+++ b/.github/workflows/dependency-age-check-actions.yml
@@ -1,7 +1,6 @@
 # Dependency Age Check (GitHub Actions pins)
 # Thin shim calling shared reusable workflow.
-# Source: https://github.com/layervai/ops-routines/blob/main/.github/workflows/age-check-actions.yml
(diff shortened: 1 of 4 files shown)


justin-layerv:
## CR nit addressed — secrets scoped explicitly

Pushed [e6f4de6](https://github.com/layervai/qurl-integrations/commit/e6f4de6) replacing `secrets: inherit` with explicit named-secret forwarding:

```yaml
secrets:
  OPS_ROUTINES_APP_ID: ${{ secrets.OPS_ROUTINES_APP_ID }}
  OPS_ROUTINES_APP_PRIVATE_KEY: ${{ secrets.OPS_ROUTINES_APP_PRIVATE_KEY }}
```

Answers the recurring reviewer question *"why `secrets: inherit`?"* and addresses the least-privilege concern — no other secrets cross the reusable boundary.

### Why v0.1.8 needs these secrets

v0.1.8 of the reusable mints a short-lived installation token via the `ops-routines-reader` GitHub App to check out the private `layervai/ops-routines` repo (where the lib scripts live). Without the App credentials flowing through, the reusable can't authenticate, and every consumer run fails at workflow-validation time.

### Version jump v0.1.4 → v0.1.8 context

v0.1.5–v0.1.7 were broken variants of the cross-repo checkout fix ([see ops-routines/releases](https://github.com/layervai/ops-routines/releases) for full notes). v0.1.8 is the first release where cross-repo auth actually works end-to-end.

### Self-reference bypass

Also shipped in v0.1.8: when the script encounters a pin pointing at `layervai/ops-routines` itself, it skips the age check. Rationale: the reusable running the script IS the pinned version — the consumer has already trusted it by pinning it. Without this, every fresh ops-routines release would block its own rollout for 7 days.

🤖 Co-authored with Claude Code

justin-layerv:
## CR feedback addressed

Pushed [0aa7928](https://github.com/layervai/qurl-integrations/commit/0aa7928) and retitled the PR:

| Reviewer item | Action |
|---|---|
| Title/version mismatch (said v0.1.8, pins v0.1.0) | **Fixed** — retitled to `ci: migrate dep-age shims to ops-routines-workflows v0.1.0`. The `# v0.1.0` comments on the `uses:` lines were already correct; only the PR title was stale. |
| Stale `Source:` header comments pointing at `layervai/ops-routines` | **Fixed** — all 4 shim header comments now point at `layervai/ops-routines-workflows`. |
| Dropped the `Implementation:` line | **Fixed** — in the public mirror the workflow YAML *is* the implementation (bash inlined), there's no separate `lib/` file to link to. |
| PR title describes "bump" but this is a repo migration | **Fixed** — new title makes the migration explicit. |

No functional changes from the prior commit. All 4 dep-age checks still passing end-to-end.

🤖 Co-authored with Claude Code

PR #90, opened 2026-04-21
fix(discord): follow-ups #86, #87, #89 from multi-tenant PR

### LINKED ISSUE #86: discord: clear stale guild-scoped command registrations on mode flip
## Context

Claude review on PR #82 noted: when an operator flips the bot from OpenNHP mode to customer-safe mode (either \`GUILD_ID\` → unset, OR \`ENABLE_OPENNHP_FEATURES\` → false), the prior guild-scoped \`/link\`, \`/leaderboard\`, \`/forcelink\`, etc. registrations remain in the old guild until Discord's cache TTL expires.

Discord has two command namespaces (guild-scoped and global) that do not purge each other on a new \`set()\` call. PR #82 handles this defensively at **dispatch** time — \`handleCommand\` filters the handler lookup to the currently-active command set, so a stale \`/link\` in autocomplete won't actually dispatch to a broken path. But the commands still show up in the user's slash-command autocomplete, which is confusing UX.

## Proposed follow-up

On boot, when the bot detects a mode transition (previous-boot OpenNHP, current-boot customer-safe), issue a one-shot \`application.commands.set([], previousGuildId)\` to purge the stale guild-scoped set. Needs a tiny bit of persistence to know the previous mode (e.g. a single row in the existing SQLite DB, or just \"try to purge every guild we were previously registered in\").

## Scope

Not urgent. Functional correctness is preserved by the dispatch-time filter already in \`commands.js\`. This is pure UX cleanup.

Raised by: Claude review on #82.

### LINKED ISSUE #87: discord: defense-in-depth — assert OAUTH_STATE_SECRET at route-handler entry
## Context

Claude review on PR #82 flagged: the \`OAUTH_STATE_SECRET\` required-secret check in \`src/index.js\` is gated behind \`!isMultiTenant\` because the \`/auth\` route isn't mounted in multi-tenant mode (or single-guild-plain mode, after this PR). The guard is safe *today* because the route is unreachable, but it's only positional safety — if a future refactor mounts an OAuth flow in a different mode without updating the boot check, the secret-missing failure would surface at first request (runtime TypeError on \`crypto.createHmac\`), not at deploy.

## Proposed follow-up

Assert \`process.env.OAUTH_STATE_SECRET\` at the **route-handler entry** in \`src/routes/oauth.js\` (inside the \`/auth/github\` handler, before the first \`crypto.createHmac\` call), not just at boot in \`index.js\`. Low cost, eliminates the positional dependency.

## Scope

Defense-in-depth cleanup. Not a security bug today. Pick up when next touching OAuth code.

Raised by: Claude review on #82.

### LINKED ISSUE #89: discord: test mock helper — consolidate isMultiTenant/isOpenNHPActive config mocks
## Context

Raised by the agent review on PR #82. Five test files now hard-code three derived config fields (\`isMultiTenant\`, \`ENABLE_OPENNHP_FEATURES\`, \`isOpenNHPActive\`) in their \`jest.mock('../src/config', () => ({ ... }))\` blocks:

- \`apps/discord/tests/discord.test.js:20\`
- \`apps/discord/tests/opennhp-gating.test.js:24\`
- \`apps/discord/tests/commands-comprehensive.test.js:30\`
- \`apps/discord/tests/coverage-boost.test.js:40\`
- \`apps/discord/tests/boot-requirements.test.js:7\`

If a future PR adds a 4th derived field to \`src/config.js\`, all five mocks drift silently — the new field reads as \`undefined\` in tests, and any test that exercises a code path depending on it coincidentally passes (since \`undefined\` is falsy and most gates are negation-based).

## Proposed

Add \`apps/discord/tests/helpers/buildConfigMock.js\` that exports a \`buildConfigMock({ guildId, enableOpenNHP })\` helper which reconstitutes the same derivation \`src/config.js\` does. Each test's \`jest.mock\` becomes:

\`\`\`js
jest.mock('../src/config', () => require('./helpers/buildConfigMock').buildConfigMock({
  guildId: 'guild-1',
  enableOpenNHP: true,
}));
\`\`\`

Adding a new derived field becomes a one-file change in the helper + \`config.js\`, not a five-file dance.

## Scope

Pure test hygiene. No behavioral change. Good candidate for a follow-up PR.

Raised by: PR #82 final pr-polish agent review.

vikramlayerv:
Single follow-up PR closing three issues opened during the PR #82 pr-polish cycle. **Stacks on [#82](https://github.com/layervai/qurl-integrations/pull/82)** — targets \`feat/multi-tenant-mode\` rather than \`main\` so the diff is scoped to just the follow-up work. After #82 merges, GitHub will auto-retarget this to \`main\` (or I can retarget manually).

## #86 — clear stale guild-scoped command registrations on mode flip

New \`purgeStaleGuildCommands(client)\` in \`apps/discord/src/commands.js\` runs on every boot in non-OpenNHP mode. Iterates \`client.guilds.cache\`, fetches any registered guild-scoped commands per guild, and \`.set([], guildId)\`s them away before the main \`.set()\` writes the current mode's commands. Per-guild fetch failure doesn't block others (logged at warn, main registration still proceeds). Skipped in OpenNHP mode — fresh guild-scoped \`.set()\` is the whole point there.

Three new tests in \`tests/multi-tenant.test.js\`:
- multi-tenant: purges from every guild the bot is in (3-guild scenario with one empty guild that's correctly skipped)
- OpenNHP mode: does NOT purge
- non-OpenNHP mode: per-guild error doesn't block registration

## #87 — OAUTH_STATE_SECRET defense-in-depth at call site

**Already implemented.** \`commands.js:stateSecret()\` at lines 55-80 already throws in any non-test environment when both \`OAUTH_STATE_SECRET\` and \`GITHUB_CLIENT_SECRET\` are missing. That's the defense-in-depth the issue was asking for, placed at the call site (state mint + state verify) rather than a handler wrapper — catches every code path that could use the secret, not just the two current routes.

No code change in this PR for #87; it's closed by this PR via the explanation above.

## #89 — test mock helper for 3-mode config derivation

New \`apps/discord/tests/helpers/buildConfigMock.js\` exports a pure \`buildConfigMock({ guildId, enableOpenNHP })\` that reconstitutes the three derived fields (\`GUILD_ID\`, \`isMultiTenant\`, \`ENABLE_OPENNHP_FEATURES\`, \`isOpenNHPActive\`) in one place. A 4th derived field added to \`src/config.js\` is then picked up automatically by any suite using the helper.

Migrated the two test files whose gating logic depends on the derivation:
- \`tests/discord.test.js\` (flag=true path)
- \`tests/opennhp-gating.test.js\` (flag=false path)

Remaining test files use fewer of the derived fields and don't benefit from migration; future authors are pointed at the helper by its JSDoc.

## Additional cleanup surfaced during review

While folding, the prior pr-polish Claude review on these commits flagged four cheap fixes, also included here:

- **bootRequired(true) dead checks removed**: \`GUILD_ID\` and \`BASE_URL\` were listed as required but can never appear missing (snowflake validator and unconditional default respectively). Dropped both; downstream enforcement is the authority.
- **Unsupported-combo warn now has a test**: \`tests/multi-tenant.test.js\` spies \`console.warn\` to pin the \"ignored when GUILD_ID is unset\" message so a future refactor can't silently drop it.
- **handleCommand OpenNHP-mode assertion tightened**: previously negated \"no longer available\"; now asserts a positive signal (\"existing\"/\"github\") that the real \`/link\` handler ran.
- **Log field rename**: \`opennhp\` → \`opennhp_active\` (3 sites) — disambiguates \"is this guild OpenNHP?\" vs \"is OpenNHP mode active?\".

## Test plan

- [x] All 532 tests pass (up from 528 — +3 purge tests + 1 unsupported-combo warn test)
- [x] \`npm run lint\` clean
- [ ] Merge #82 first, then retarget this PR to \`main\`
- [ ] After both land, the next qurl-bot-discord deploy's \`/link\` autocomplete should stop showing in customer guilds where the bot previously registered OpenNHP commands

## Closes

- Closes #86
- Closes #87 (already implemented — verified in this PR)
- Closes #89

### DIFF
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -5,14 +5,27 @@
 // boot in prod with missing secrets OR die on a spurious false-positive.
 
 // Required at boot in EVERY environment. Gated on `isOpenNHPActive`,
-// NOT `isMultiTenant`: the GITHUB_* + BASE_URL + GUILD_ID vars only
-// matter when /auth + /webhook routes are actually mounted. A
-// single-guild-plain deployment (GUILD_ID set but ENABLE_OPENNHP_FEATURES
-// off) never mounts those routes, so demanding dummy values just to
-// pass the boot check would be a papercut for every customer server.
+// NOT `isMultiTenant`: the GITHUB_* vars only matter when /auth +
+// /webhook routes are actually mounted. A single-guild-plain deployment
+// (GUILD_ID set but ENABLE_OPENNHP_FEATURES off) never mounts those
+// routes, so demanding dummy values just to pass the boot check would
+// be a papercut for every customer server.
+//
+// Explicitly NOT on this list even in OpenNHP mode:
+//   - GUILD_ID: if isOpenNHPActive === true then !isMultiTenant, which
+//     means the snowflake validator in config.js already accepted a
+//     17-20 digit value. Re-checking truthiness here would never catch
+//     a missing GUILD_ID — the upstream check is the authority.
+//   - BASE_URL: config.js supplies an unconditional "http://localhost:3000"
+//     default, so `cfg.BASE_URL` is always truthy. The real enforcement
+//     is the https-startswith check in index.js, which runs regardless
+//     of this required-list membership.
+// Listing either would be decorative — the downstream checks are the
+// authority. Keeping this list to the keys whose absence is actually a
+// boot blocker.
 function bootRequired(isOpenNHPActive) {
   if (!isOpenNHPActive) return ['DISCORD_TOKEN'];
-  return ['DISCORD_TOKEN', 'GITHUB_CLIENT_ID', 'GITHUB_CLIENT_SECRET', 'GITHUB_WEBHOOK_SECRET', 'GUILD_ID', 'BASE_URL'];
+  return ['DISCORD_TOKEN', 'GITHUB_CLIENT_ID', 'GITHUB_CLIENT_SECRET', 'GITHUB_WEBHOOK_SECRET'];
 }
 
 // Additionally required when NODE_ENV=production. QURL_API_KEY is the
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2421,10 +2421,49 @@ function getActiveCommands() {
     : commands.filter(cmd => CUSTOMER_SAFE_COMMANDS.has(cmd.data.name));
 }
 
+// Proactively clear stale guild-scoped command registrations from any
+// guild the bot is in. Discord's guild and global command namespaces do
+// not purge each other on a fresh .set() call, so a bot that previously
+// ran in OpenNHP mode (GUILD_ID=X, full command set registered to guild
+// X) and is now redeployed in multi-tenant or single-guild-plain mode
+// will leave /link, /leaderboard, etc. visible in X's slash-command
+// autocomplete until Discord's cache ages out. The dispatch-time filter
+// in handleCommand prevents those stale commands from doing anything
+// harmful, but users still see dead commands in the picker. Issuing
+// .set([], guildId) clears the guild-scoped set; any ALIVE guild-scoped
+// registration for the CURRENT mode gets reinstalled by the main
+// registerCommands path below.
+//
+// Scoped to non-OpenNHP modes: in OpenNHP mode we intentionally register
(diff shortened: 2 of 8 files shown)


vikramlayerv:
Superseded — the three follow-ups have been re-folded into PR #82 (commits d230312 + 02ed495) per user direction. The diff in this PR and those two commits on feat/multi-tenant-mode is identical.

PR #96, opened 2026-04-21
fix(discord): restore channel notification + dynamic target autocomplete

vikramlayerv:
## Summary

Two features originally shipped in commit `a43d45e` ("demo feedback") were silently removed in `4b57bb2` ("address Claude review round 5"). That commit's body enumerated items #2–#8 (startup warning, polling backoff, FIFO docs, embed char cap, …) but did not mention either removal, so these regressions landed unnoticed.

This PR restores both on top of current `main`.

## What's restored

1. **Channel notification when /qurl send targets "Everyone".** Non-ephemeral message in the invoking channel so all members know to check their DMs. Only fires for `target=channel` / `voice` and only when at least one DM was delivered. Wrapped in try/catch so missing `SendMessages` permission degrades silently — the send itself already succeeded.
2. **Dynamic target autocomplete.** Static `addChoices()` can't depend on invocation context, so it always leaked "Users in my voice channel" into text channels where it fails loudly at dispatch time ("You must be in a voice channel…"). Autocomplete now conditionally surfaces the voice option **only** when the invoking channel is `GuildVoice` or `GuildStageVoice`.

## Why the comment is worth it

A later reviewer might look at this and ask "why autocomplete instead of static choices?" The inline comment spells out the reason: Discord's `addChoices` API can't branch on channel context, so static choices make the voice option a dead-end in text channels.

## Testing

- 2 new autocomplete tests: text-channel returns 2 choices, voice-channel returns 3 with the voice option appended.
- Existing file-flow test extended to assert the channel notification fires with the expected copy.
- Full Jest suite green: **540 tests across 25 suites**, plus `npm run lint` clean (eslint `--max-warnings 0`).

## Test plan

- [ ] CI green (jest + eslint + Claude review)
- [ ] Manual smoke once the bot is actually deployed: `/qurl send` with target=Everyone posts the public notification; target=user does not
- [ ] In a text channel, autocomplete on `target` does NOT show "Only voice users"
- [ ] In a voice channel, autocomplete on `target` DOES show "Only voice users"

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -5,7 +5,6 @@ const {
   ActionRowBuilder,
   ButtonBuilder,
   ButtonStyle,
-  ChannelType,
   ComponentType,
   StringSelectMenuBuilder,
   UserSelectMenuBuilder,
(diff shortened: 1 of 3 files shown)


Kevin-layerV (approved):

vikramlayerv:

The loop is closed — below are the items from Claude's final review that were consciously not addressed in this PR.

|---|---|---|---|---|
| 🟡 Minor | No `getSubcommand() === 'send'` guard in the `handleCommand` autocomplete dispatch (`commands.js:2580`) | `target` is currently declared only on the `send` subcommand; no collision possible today. A guard would be defensive future-proofing but adds a conditional for a hazard that hasn't materialized. | Add when a second `qurl` subcommand declares any same-named option; revisit at that point. | None today — defensive-only. |
| 🟢 Nit | PR body still describes two restored features, but the channel-notification block was already on `main` via PR #82 before this PR opened. The real code change here is (a) static `addChoices` → dynamic autocomplete, (b) regression-guard assertion that `channel.send` is called exactly once. | PR body edit is a documentation nit; the diff itself is the source of truth and the commit messages are precise. | None — acceptable as-is; the commit history tells the real story. | None. |
| 🟢 Info | `interaction.member.voice.channelId` vs `voiceStates.cache.get()` — theoretical cache lag if a user disconnects between autocomplete resolve and `/qurl send` dispatch. | Both sources are event-driven from the same gateway events; only race is a user disconnect in the sub-second window, which falls through to the existing friendly `"You must be in a voice channel..."` error. Claude explicitly said "no action needed." | None — acceptable as-is. | Negligible. |
| 🟢 Info | `focused.value` not used to filter the choices in `handleTargetAutocomplete`. | With 2–3 hardcoded choices, Discord's client does prefix-filtering itself. Per Claude: "if the list ever grows past 25 items, revisit." | Revisit only if choices exceed ~10. | None today. |

**Confidence:** 543 Jest tests green, eslint `--max-warnings 0` clean, all CI checks success, Claude's final verdict `"Ship it."` Diff is 148 net LOC across 3 files — no blast radius outside the `/qurl send` target-option path.

— via /pr-polish

PR #103, opened 2026-04-22
fix(discord): restore voice-channel viewer enumeration (regression from 4b57bb2)

vikramlayerv:
## 1) Problem

User report: `/qurl send` from a voice channel with target **"Everyone in this channel"** delivered only to voice-connected users — identical behavior to target=voice. CloudWatch confirmed: `target:"channel"` returned `delivered:1`, same as `target:"voice"`.

User's memory was correct: this worked in an earlier commit. We lost it.

## 2) Root cause

`apps/discord/src/discord.js:getTextChannelMembers` reads `channel.members`, which has different semantics by channel type in discord.js:

| Channel type | `channel.members` returns |
|---|---|
| `GuildText` | All members with `ViewChannel` permission (correct) |
| `GuildVoice` / `GuildStageVoice` | **Only currently voice-connected members** |

For voice channels, this returns the voice-connected subset — exactly what `target=voice` returns.

### The silent-revert story

A prior commit already had the correct implementation:

- **`4980963`** (Claude review — dead code, db timers, dedupe, lint) **added** a dedicated `getVoiceChannelViewers` helper and wired it via `recipients = isVoiceChannel ? getVoiceChannelViewers(...) : getTextChannelMembers(...)` in `handleSend`.
- **`4b57bb2`** (Claude review round 5 — high/medium items) **removed the helper, the import, and the branch** — without mentioning either removal in the commit body. Same 4b57bb2 that silently stripped the channel notification and the dynamic target autocomplete (already recovered earlier today in PRs #82 and #100).

This is the **third feature** silently dropped by 4b57bb2 that's had to be recovered today.

## 3) Fix strategy

Fold the voice-viewer logic into `getTextChannelMembers` itself. Branch on `channel.type`: for voice/stage-voice, enumerate from `guild.members.cache` filtered by `permissionsFor(ViewChannel)`. For text channels, keep using `channel.members` (already correct there).

`handleSend` already calls `await guild.members.fetch()` before this function for `target === 'channel'`, so the cache is warm.

## 4) Why this strategy

- **Single helper, correct for all channel types.** Previous design had two exported helpers — easy to miss wiring one up, which is exactly what happened when 4b57bb2 dropped it. One canonical function is safer against silent-revert.
- **No caller changes.** `handleSend` continues calling `getTextChannelMembers(channel, userId)` as today. Only the helper's internals change.
- **Locked in by tests.** Unit tests + an e2e smoke test assert the invariant so the next 4b57bb2-style review round can't silently strip it.

Rejected alternative: restore the exact 4980963 two-helper shape. Equivalent behavior, but two exports means two wiring points — more surface for a future silent-revert to miss.

## 5) Steps to fix (what this PR does)

1. **`apps/discord/src/discord.js`** — import `ChannelType` + `PermissionFlagsBits`. Branch `getTextChannelMembers` on channel type and use the right enumeration source for voice vs text.
2. **`apps/discord/tests/discord.test.js`** — update the `discord.js` jest mock to expose `ChannelType.GuildVoice`, `GuildStageVoice`, `PermissionFlagsBits.ViewChannel`. New tests: (a) voice channel with some connected + some not-connected → all viewers returned; (b) stage-voice channel, same path.
3. **`e2e/tests/discord-channels.test.ts`** — new smoke test documenting the invariant: "Everyone in a voice channel" recipient pool = guild member list, independent of voice-connected count.

Tests: 546 passing (+3), eslint `--max-warnings 0` clean.

## Test plan

- [ ] CI green (Jest + eslint + CodeQL + Claude review + title check + secrets scan)
- [ ] After merge + deploy: from a voice channel in a guild with 3+ viewers where only the sender is voice-connected, `/qurl send target:"Everyone in this channel"` should deliver to all viewing members (minus bots and sender), not just the voice-connected subset.
- [ ] E2E smoke: the new "Everyone in voice channel recipient pool = guild members" test passes in CI.

### DIFF
--- a/apps/discord/src/discord.js
+++ b/apps/discord/src/discord.js
@@ -1,4 +1,4 @@
-const { Client, GatewayIntentBits, EmbedBuilder } = require('discord.js');
+const { Client, GatewayIntentBits, EmbedBuilder, ChannelType, PermissionFlagsBits } = require('discord.js');
 const cron = require('node-cron');
 const config = require('./config');
 const logger = require('./logger');
@@ -812,13 +812,32 @@ function getVoiceChannelMembers(guildObj, senderUserId) {
   };
 }
 
-// Get non-bot members who can view a text channel (excludes the sender)
+// Get non-bot members who can view a channel (excludes the sender).
+//
+// Voice/stage-voice channels have a gotcha: in discord.js, `channel.members`
+// returns members CURRENTLY CONNECTED to voice — NOT everyone who can see
+// the channel. For text channels `channel.members` is computed off guild
+// members + ViewChannel permission, which is the semantic we want for
+// "Everyone in this channel." So for voice/stage-voice we compute the
+// viewer set explicitly from guild.members.cache.
+//
+// Callers rely on the sender having already done `guild.members.fetch()`
+// so the cache is warm; see the `target === 'channel'` branch in
+// handleSend (commands.js) for the fetch + call site pairing.
 function getTextChannelMembers(channel, senderUserId) {
-  const members = channel.members
+  const isVoice = channel.type === ChannelType.GuildVoice ||
+                  channel.type === ChannelType.GuildStageVoice;
+
+  const source = (isVoice && channel.guild)
+    ? channel.guild.members.cache.filter(m => {
+        const perms = channel.permissionsFor(m);
+        return perms && perms.has(PermissionFlagsBits.ViewChannel);
+      })
+    : channel.members;
+
+  return source
     .filter(m => m.id !== senderUserId && !m.user.bot)
     .map(m => m.user);
-
-  return members;
 }
 
 // Graceful shutdown — awaits client.destroy() so the caller knows the
--- a/apps/discord/src/discord.js
+++ b/apps/discord/src/discord.js
@@ -828,12 +828,29 @@ function getTextChannelMembers(channel, senderUserId) {
   const isVoice = channel.type === ChannelType.GuildVoice ||
                   channel.type === ChannelType.GuildStageVoice;
 
-  const source = (isVoice && channel.guild)
-    ? channel.guild.members.cache.filter(m => {
-        const perms = channel.permissionsFor(m);
-        return perms && perms.has(PermissionFlagsBits.ViewChannel);
-      })
-    : channel.members;
+  let source;
+  if (isVoice) {
+    // Voice path: enumerate guild viewers via permissionsFor. If the
+    // channel is voice-typed but `.guild` is somehow missing (partial
+    // cache / unusual gateway state), do NOT silently fall back to
+    // `channel.members` — that's the voice-connected-only set which
+    // is exactly the bug this helper exists to avoid. Return empty and
+    // log so a caller's "no recipients" branch triggers loudly instead
+    // of shipping to a wrong subset.
+    if (!channel.guild) {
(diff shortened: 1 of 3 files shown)


vikramlayerv:

Claude's verdict on `3ae79fc`: *"Approving in spirit — the fix is correct and the tests are structured to prevent the next silent revert."* Open items below are all nits/follow-ups Claude explicitly said to ship-as-is or defer.

|---|---|---|---|---|
| 🟡 Nit | Rename `getTextChannelMembers` → `getChannelViewers` — name no longer matches scope (`apps/discord/src/discord.js:827`) | Claude explicitly said "leave the rename for a separate PR since touching the export name is precisely the surface that 4b57bb2 historically dropped — don't bundle that risk with this fix." | Backlog: rename in a focused PR after this lands. | low / none |
| 🟡 Nit | Defensive-empty branch returns `[]` on `!channel.guild`, caller then renders "No other members in this channel" (genuine state) instead of a retry-flavored message (`apps/discord/src/discord.js:840-845` + `commands.js:672`) | Branch is genuinely rare (partial cache on a voice-typed channel), and the current message is still truthful ("no members" is what we resolved). Sentinel-based error propagation would touch the caller contract — larger change than the cache-robustness fix this PR does. | Backlog: follow-up PR to plumb a distinguishable error through to the user on cache-read failures. | deferred |
| 🟡 Nit | Orphan comment at `apps/discord/src/discord.js:793-795` ("Depends on guild.members.cache being populated...") sits above `getVoiceChannelMembers` but describes the invariant for `getTextChannelMembers` | Pre-existing misplacement; not introduced by this PR. | Backlog: small-cleanup commit alongside the rename in the follow-up. | none |
| 🟡 Nit | `permissionsFor` cost — for guilds near the ~1000-member cache ceiling, per-member permission resolution runs 1k times on every `target=channel` voice invocation (`apps/discord/src/discord.js:832-836`) | Fine today; no evidence this path is on the hot hot-path. Claude said "if this ever shows up in CloudWatch p95, cache the per-channel viewer set keyed on (channelId, overwrite-hash)." | Monitoring item: revisit only if p95 latency shifts after this deploys. | low |
| 🟡 Nit | `asCollection` test helper attaches `.map` to filter result but not `.filter`, so a future edit adding a second chained `.filter` would silently fall through to `Array.prototype.filter` with different semantics than discord.js `Collection.prototype.filter` (`apps/discord/tests/discord.test.js:352-358`) | Current code only chains once; no reachable trap. Fixing in full would require a closer-to-real Collection mock across all test files, out of scope. | Backlog: consider adopting discord.js's real Collection class in tests for fidelity. | low |
| 🟡 Nit | E2E test name is stronger than scope — "Everyone in voice channel recipient pool = guild members..." asserts prerequisites, not the actual `/qurl send` path (`e2e/tests/discord-channels.test.ts:87-125`) | Test body is explicit about being a premise-check ("prerequisites the fix relies on"). Real behavioral assertion lives in the unit tests. | Backlog: rename to "voice-channel recipient pool prerequisites are intact" in a docs-only follow-up. | none |

**Confidence:** 550 Jest tests green, `eslint --max-warnings 0` clean, all CI checks success, Claude's final verdict *"Approving in spirit."* The fix is a direct cherry-pick of behavior that existed in `4980963` before 4b57bb2 silently dropped it — battle-tested once, now tests locked-in to prevent recurrence.

— via /pr-polish

Kevin-layerV (approved):

Issue #99, opened 2026-04-21
Feature wishlist from QA review — design decisions needed


vikramlayerv:
Collected from Kevin's QA review of the Discord bot. All five items below are **product/design decisions** rather than bugs — each needs a call on whether, how, and when we'd build it before engineering picks it up.

## 1. Bulk revoke / bring back \`/clear\`
**Ask:** Revoke multiple protected resources in a single action instead of one-by-one.

**Context:** The old Python bot had \`/qurl clear\` to delete all user resources. That was removed during the Python → Node.js migration because the resource-catalog model itself went away. With the atomic-\`/qurl send\` model, "resources" are really "past sends," and there's no catalog to clear.

**Open question:** Do we want to reintroduce a bulk-revoke for past sends? If yes — all by this user? By time range? By recipient? UI affordance (confirm modal vs. immediate)?

## 2. Multi-use qurl links
**Ask:** Allow a qurl to be opened more than once.

**Open question:** Is single-use a hard product invariant (part of the "self-destruct" security story) or an early default that can be relaxed? If relaxed, what's the ceiling — N uses, unlimited until expiry, role-based?

## 3. \`/qurl feedback\` command + Claude-routine auto-PR
**Ask:** In-bot feedback command; Claude periodically reads the feedback queue and opens PRs for review.

**Open question:** Worth the scope? Where does feedback land (Linear? a GH discussion? a Slack channel?). What's the loop between user submission → triage → Claude surfacing actionable items?

## 4. Admin-vs-user split in \`/qurl help\`
**Ask:** Non-admin users don't see admin-only commands (\`setup\`, \`status\`) in the help.

**Open question:** Two separate help responses gated on \`ManageGuild\`, or a single response with admin items visually grouped under an "admin only" section? Current single-help is simpler; a split is more customer-polished but adds a branch.

## 5. Branding normalization (\`QURL\` vs \`qURL\` vs \`Qurl\`)
**Ask:** Pick one canonical form and standardize user-facing copy across Discord, CLI, desktop tunnel app, and docs.

**Current state of the Discord bot** (apps/discord/src/commands.js):
- \`QURL\` — 65 uses (user-facing product name)
- \`qurl\` — 76 uses (slash-command names, required lowercase)
- \`Qurl\` — 4 uses (drift in a handful of user messages)
- \`qURL\` — 0 uses (Kevin's suggested form)

[...]

PR #106, opened 2026-04-22
fix(discord): rename /qurl send options + revoke dropdown filter + descriptive labels + smoke test

vikramlayerv:
## Summary
Three UX fixes on the \`/qurl send\` → \`/qurl revoke\` flow, plus one e2e smoke test for the revoke path.

### 1. Rename \`/qurl send\` options to match \`_optional\` convention
- \`expiry\` → \`expiry_optional\`
- \`message\` → \`message_optional\`
- \`file_optional\` already existed; now all three optional params follow the same naming, and users can read required-vs-optional from the option name alone.

### 2. Hide already-revoked sends from \`/qurl revoke\` dropdown
Previously, hitting \`/qurl revoke\` twice on the same send would show it in the dropdown both times — second click returned \"Revoked 0/N links\" because the links were already dead at the QURL API layer but the UI had no way to know.

- New \`revoked_at TEXT\` column on \`qurl_send_configs\` (via the existing \`SAFE_ALTERS\` migration pattern — idempotent \`ALTER TABLE ADD COLUMN\`).
- \`revokeAllLinks\` now calls \`db.markSendRevoked(sendId, senderDiscordId)\` after the QURL API DELETE calls finish.
- \`getRecentSends\` LEFT JOINs \`qurl_send_configs\` and filters \`WHERE (c.revoked_at IS NULL OR c.send_id IS NULL)\`. LEFT (not INNER) JOIN preserves pre-config-era rows.

### 3. More descriptive \`/qurl revoke\` dropdown labels
**Before:** \`file to 3 users (channel)\` — no way to tell WHICH file
**After:** \`📎 quarterly-report.pdf — 3 people (#general)\`
or:       \`📍 Starbucks Downtown — 1 person (DM)\`

- \`getRecentSends\` JOIN exposes \`attachment_name\`, \`location_name\`, \`personal_message\` (previously in \`qurl_send_configs\` but unused here).
- New \`formatRevokeLabel\` / \`formatRevokeDescription\` helpers with Discord-100-char truncation. Description includes message snippet when present so users can disambiguate same-filename sends with different notes.

### 4. E2E smoke test for revoke end-to-end
\`e2e/tests/revoke.smoke.test.ts\` — validates \`DELETE /v1/resources/{id}\` against the deployed QURL API actually marks the resource revoked (subsequent \`getLinkStatus\` returns 404).

Existing \`file-revoke.test.ts\` covers this for file resources but sits outside the smoke filter (\`smoke|google-maps|discord-channels\`). This new test is URL-based (no connector dep) and runs on every deploy.

## Test plan
- [x] All 546 bot unit tests pass (added \`markSendRevoked: jest.fn()\` to two mockDb setups so revoke-handler tests stay green)
- [ ] E2E smoke test not validated locally (no QURL API key accessible) — uses the exact same helpers as \`link-lifecycle.test.ts\` and \`file-revoke.test.ts\` which are known to work against prod
- [ ] Post-merge: run \`/qurl revoke\` in the test guild and eyeball the new label/description formatting

## Explicitly out of scope
- Bulk revoke / bring back \`/clear\` — filed in feature wishlist [#99](https://github.com/layervai/qurl-integrations/issues/99)
- Testing the dropdown RENDERING itself — impossible without selfbotting (ToS); rendering logic is covered by the \`formatRevokeLabel\` / \`formatRevokeDescription\` pure functions which are unit-testable if a follow-up wants to pin them

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -602,8 +602,8 @@ async function handleSend(interaction, apiKey) {
     return interaction.reply({ content: 'QURL API key is not configured.', ephemeral: true });
   }
   const target = interaction.options.getString('target');
-  const expiresIn = interaction.options.getString('expiry') || '24h';
-  const rawMessage = interaction.options.getString('message');
+  const expiresIn = interaction.options.getString('expiry_optional') || '24h';
+  const rawMessage = interaction.options.getString('message_optional');
   const personalMessage = rawMessage ? sanitizeMessage(rawMessage) : null;
   const commandAttachment = interaction.options.getAttachment('file_optional');
 
@@ -2221,7 +2221,7 @@ const commands = [
               .setRequired(true)
               .setAutocomplete(true))
           .addStringOption(opt =>
-            opt.setName('expiry')
+            opt.setName('expiry_optional')
               .setDescription('Link expiry (default: 24h)')
               .setRequired(false)
               .addChoices(
@@ -2236,7 +2236,7 @@ const commands = [
               .setDescription('Optional — attach a file here if sharing a file')
               .setRequired(false))
           .addStringOption(opt =>
-            opt.setName('message')
+            opt.setName('message_optional')
               .setDescription('Optional note sent alongside the link')
               .setRequired(false))
       )
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1574,9 +1574,17 @@ function truncate(s, max = SELECT_MENU_FIELD_MAX) {
 // location name) over the previous abstract `${resource_type} to N users`.
 // Falls back gracefully for legacy rows that predate qurl_send_configs
 // being populated (LEFT JOIN produces null fields there).
-function formatRevokeLabel(s) {
+//
+// `channelName` is passed in by the caller (handleRevoke) from
+// `interaction.guild?.channels.cache.get(s.channel_id)?.name`. Discord
+// mention syntax (`<#id>`) does NOT render inside StringSelectMenu
+// option labels — passing the name explicitly is the only way to avoid
+// showing a raw snowflake to the user.
+function formatRevokeLabel(s, channelName) {
   const recipients = s.recipient_count === 1 ? '1 person' : `${s.recipient_count} people`;
-  const channelRef = s.target_type === 'user' ? 'DM' : (s.channel_id ? `#${s.channel_id}` : s.target_type);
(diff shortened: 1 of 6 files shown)


vikramlayerv:

Claude round-2 review on commit 7e5b957 verdict: **"LGTM pending your call on the items above. Nothing here should block merge."** All remaining items are observations / nits / follow-ups. Commit 3cd27a4 (help-copy tweaks: "Setting up (for Admins)", "Getting started — Share resources…", tab on "1.") is a 12-line string edit with no logic changes; no new review round needed.

| Severity per reviewer | Finding | Why accepted | Future plan |
|---|---|---|---|
| Minor | Slash-command rename (`expiry` → `expiry_optional`, `message` → `message_optional`) is a user-visible breaking change | Discord clients reconcile within seconds of re-register; convention matches existing `file_optional` | Worth a line in release notes |
| Minor suggestion | `formatRevokeLabel` / `formatRevokeDescription` are untested pure functions (truncation math, fallback branches) | Exercised transitively by the new integration test via `getRecentSends → handleRevoke`; standalone unit tests would be incremental | Follow-up PR — file if truncation logic gets more complex |
| Marginal | `markSendRevoked` legacy fallback runs a SELECT + INSERT-OR-IGNORE on already-revoked sends (wasted work, not buggy) | Dropdown filter means this path is reached only from stale clients; perf cost is ~1 indexed lookup | Revisit only if the function is being edited again |
| Pre-existing | `.slice(0, max - 1)` counts UTF-16 code units, so emoji-heavy `personal_message` could truncate mid-surrogate-pair | The trailing `…` visually masks it; pattern matches other truncation sites in the codebase | Deferred — fix alongside any broader string-handling cleanup |
| Defensive | Other `mockDb` setups in `qurl-send.test.js` and `multi-tenant.test.js` don't stub `markSendRevoked` | Those tests don't exercise `revokeAllLinks` today; will fail cryptically only if a future test in those files does | Better fix: extract a shared `makeMockDb()` helper — separate refactor PR |

Confidence: 550/550 unit tests pass (+4 from new integration test), 2 rounds of Claude review converged, CI green, 4 commits totaling ~210 LOC across bot code + tests + docs. Headline bugs in round 1 (raw channel snowflake in dropdown, legacy-send no-op) both addressed and verified.

— via /pr-polish

Kevin-layerV (dismissed):

Kevin-layerV (approved):

PR #107, opened 2026-04-22
test(discord): pin anchored "Open in Google Maps" URL for query+coords case

vikramlayerv:
## Summary

Smoke coverage for the "Erbil restaurant" fix (connector-side) in **qurl-integrations-infra#155**.

Adds one e2e test case in `e2e/tests/google-maps.test.ts` that uploads a location with **both** `query` (named place) AND `lat`/`lng` (coordinates), fetches the viewer page, and asserts:

- Rendered HTML contains `/maps/place/<name>/@<lat>,<lng>,17z` (the anchored form)
- It does NOT fall back to `/maps/search/<name>` (the bug shape)

Without this, a future regression that swaps the handler back to `/maps/search/...` would only be caught by the Go unit tests in the infra repo — not by the public-repo CI that runs against the deployed connector.

## Context

The Erbil restaurant bug: `/qurl send` of a named place opened a **different** restaurant for the recipient because `/maps/search/<name>` resolves on the client side relative to the recipient's geolocation. The iframe (Embed API, server-side) showed the right place; only the "Open in Google Maps" link was broken. Fixed in connector by producing an anchored `/maps/place/<name>/@lat,lng,17z` URL when both query and coords are supplied.

## Merge ordering

**Merge this AFTER qurl-integrations-infra#155 deploys** — otherwise this test will fail against the current (unfixed) connector on `main`. If I see #155 land first, will verify this PR's CI stays green before merging.

## Test plan

- [ ] qurl-integrations-infra#155 merges + deploys
- [ ] This PR's e2e-smoke run goes green against the newly-deployed connector
- [ ] CI green (jest + eslint + Claude review + CodeQL + title check)

### DIFF
--- a/e2e/tests/google-maps.test.ts
+++ b/e2e/tests/google-maps.test.ts
@@ -150,6 +150,32 @@ describe('Google Maps: Iframe Embed', () => {
     expect(html).toContain('https://www.google.com/maps/search/');
   });
 
+  test('"Open in Google Maps" URL is anchored to coords when query + lat/lng both supplied', async () => {
+    // Regression guard for the "Erbil restaurant" bug: when the bot
+    // uploads a named place that also carries lat/lng from Places
+    // autocomplete, the mapsURL (shown under "Open in Google Maps")
+    // must be a /maps/place/<name>/@<lat>,<lng>,17z form — NOT
+    // /maps/search/<name> — otherwise Google resolves the name on
+    // the recipient's side and opens a namesake near THEIR
+    // geolocation instead of the place the sender picked. Fixed in
+    // qurl-integrations-infra#155.
+    const upload = await uploadMapLocation({
+      type: 'google-map',
+      query: 'Erbil restaurant',
+      lat: 36.1911,
+      lng: 44.0094,
+    });
+
+    const html = await fetchViewerPage(upload.viewerUrl);
+    expect(html).toContain('Open in Google Maps');
+    // Anchored URL form: path segment with name + @lat,lng,17z suffix.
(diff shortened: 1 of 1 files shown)


vikramlayerv:

Claude's verdict on `bb95ba9`: *"LGTM after infra#155 deploys and CI is green. Point 3 is optional polish."* 4 of Claude's 5 original findings were addressed in the second commit; one Minor remains.

|---|---|---|---|---|
| 🟡 Minor | `%.7f` coupling — test hardcodes `36\.1911000,44\.0094000` against the server-side format string in qurl-integrations-infra. A connector-side tweak to `%.6f` would fail this test for a non-functional reason (`e2e/tests/google-maps.test.ts:176`) | Claude explicitly called this "defensible either way" — a regression test for a specific format contract should trigger conscious review if the format ever changes. A looser `@36\.1911\d*,44\.0094\d*,17z` would be more resilient but also catches fewer regressions. Exact-string pinning is closer to existing test style in this file. | None — acceptable as-is. Revisit only if the connector's map URL precision is ever tuned. | none |

**Merge ordering:** Per the PR body and Claude's reminder — merge **qurl-integrations-infra#155 first** (deploys the fix), then verify this PR's e2e-smoke goes green against the newly-deployed connector, then merge this. Merging before #155 deploys will fail the new assertion against pre-fix main.

**Confidence:** Single-file +~30 line test-only change. TypeScript compiles clean. No new dependencies, no changes to product code, no observability impact. Companion to infra#155's handler fix — together they close the "Erbil restaurant" regression class end-to-end.

— via /pr-polish

Kevin-layerV (approved):

PR #108, opened 2026-04-22
fix(e2e): revoke smoke uses accessLink, not getLinkStatus

vikramlayerv:
## Summary
Fixes the post-PR-#106 smoke failure. The new \`revoke.smoke.test.ts\` I added in #106 used \`getLinkStatus\` to probe whether a qurl was live/revoked. In prod that endpoint returns 404 for URL-based qurls regardless of state (the existing \`smoke.test.ts:78-81\` already acknowledges this with a graceful try/catch and comment "may not be tracked server-side for SPA links").

**Consequences of the broken test:**
- Pre-revoke sanity check failed in prod (resource returning 404 when still live) — post-#106 deploy smoke run failed
- Post-revoke assertion was passing tautologically (same 404 regardless of whether the revoke actually worked)

## Fix
Rewrite to use \`accessLink\` — follow the qurl link and check HTTP status — which is both the user-visible invariant AND distinguishes live from revoked reliably:

\`\`\`
pre-revoke:  accessLink(qurl_link).ok === true
revoke
post-revoke: accessLink(qurl_link).ok === false
\`\`\`

Same pattern as \`smoke.test.ts\`'s 'first access succeeds' test, which has run green against prod for months.

## Deliberately out of scope
- Asserting on the specific response code (404 vs 403 vs 410). That's a QURL-API implementation detail; the test pins the user-visible contract ("link stops working post-revoke"), not the internal status-code shape.

## Test plan
- [ ] CI green on this PR
- [ ] Post-merge: next deploy smoke run goes green (revoke.smoke.test.ts passes)
- [ ] Cannot validate locally — no QURL_API_KEY in my environment. Helpers (\`accessLink\`, \`mintLink\`, \`revokeLink\`) are the same ones \`smoke.test.ts\` uses successfully in prod.

### DIFF
--- a/e2e/tests/revoke.smoke.test.ts
+++ b/e2e/tests/revoke.smoke.test.ts
@@ -54,11 +54,22 @@ describe('QURL revoke (smoke)', () => {
     expect(revoked).toBe(true);
 
     // Canonical revoke assertion: the qurl link is no longer reachable.
-    // The deployed QURL API returns 404 for revoked resources; any
-    // non-2xx response is accepted here because the specific failure
-    // code is a QURL-API implementation detail and the user-visible
-    // invariant is just "the link stops working."
+    // Asserting `status >= 400` (not `.ok === false`) so a failure prints
+    // the actual HTTP code in Jest's diff — distinguishes "revoke
+    // silently succeeded" (2xx still served) from "SPA shell returns 200
+    // regardless of revocation state" (200 served, client-side error).
+    // Any 4xx/5xx is accepted because the specific failure code is a
+    // QURL-API implementation detail; the user-visible invariant is
+    // just "the link stops working."
     const postAccess = await accessLink(minted.qurl_link);
-    expect(postAccess.ok).toBe(false);
+    if (postAccess.status < 400) {
+      // eslint-disable-next-line no-console
+      console.error(
+        `[revoke.smoke] Post-revoke access returned ${postAccess.status} — expected 4xx/5xx. ` +
+        `finalUrl=${postAccess.finalUrl}. If this is an SPA shell serving 200 with a client-side ` +
+        `error, the revoke API may be working but this test needs a different probe.`,
+      );
+    }
+    expect(postAccess.status).toBeGreaterThanOrEqual(400);
   });
 });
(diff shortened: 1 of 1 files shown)


vikramlayerv:

Claude round-2 verdict: **"Not a blocker on the narrow fix."** Addressed the one thing that was flagged as a round-1 push-point (diagnostic status-code assertion + SPA-shell-200 console.error). Exiting the loop here per the narrow-fix framing.

| Severity per reviewer | Finding | Why accepted | Future plan |
|---|---|---|---|
| Non-blocker push | \`smoke.test.ts:103-109\` uses the same tautological \`getLinkStatus\` → \`/404/\` pattern this PR fixes | Touching \`smoke.test.ts\` would broaden scope beyond "fix what's failing post-deploy" | **Tracked in [#109](https://github.com/layervai/qurl-integrations/issues/109)** — follow-up to either rewrite to accessLink or delete as redundant with the new revoke.smoke.test.ts |
| Separate concern | \`file-revoke.test.ts:110-112\` uses the same pattern against file-based qurls (may or may not be broken — file resources may behave differently) | Same scope argument; also needs verification before fixing | Tracked in the same follow-up issue |
| Minor | \`accessLink\` helper reads full response body into memory even when body is never inspected | Helper-level change benefits multiple tests; wart, not a bug | Next time the helper is touched |
| Minor | \`max_uses: 10\` means pre-revoke accessLink consumes one use; effectively 9 remain when revoke fires | Not load-bearing for the current assertions; noted in the inline comment | Keep in mind if future assertions depend on remaining uses |
| Documented | First commit used \`fix(e2e):\` scope — \`e2e\` not in CLAUDE.md's slack\|teams\|discord\|cli\|zapier\|shared list | Second commit switched to \`test(discord):\` matching PR #107's precedent | Separate conversation if we want to formally extend the scope list |

Confidence: 2 rounds of Claude review converged, CI green, test-only diff (44 lines in one file), same helper + assertion pattern as \`smoke.test.ts:64-68\` which is already green in prod.

— via /pr-polish

Kevin-layerV (approved):

Issue #109, opened 2026-04-22
smoke.test.ts:103-109 uses same broken getLinkStatus pattern fixed in PR #108


vikramlayerv:
Flagged by Claude review on PR #108. smoke.test.ts:103-109 asserts getLinkStatus returns /404/ post-revoke on a URL-based qurl — but URL-based qurls return 404 on that endpoint regardless of revocation state (acknowledged by the graceful try/catch at smoke.test.ts:78-81). The assertion passes tautologically today; it would pass even if the revoke API was broken.

Fix options: (1) rewrite the test to use accessLink, same pattern as PR #108's revoke.smoke.test.ts, or (2) delete the test — revoke.smoke.test.ts now covers this invariant end-to-end.

Preference: option 2, since revoke.smoke.test.ts is strictly more rigorous.

file-revoke.test.ts:110-112 uses the same pattern against a file-based qurl. File resources may behave differently on the /status endpoint — worth a sanity-check read before assuming the same breakage applies.

Small follow-up PR. Pure test hygiene; no runtime change.

PR #110, opened 2026-04-22
test(discord): delete revoke.smoke.test.ts — coverage claim doesn't hold

vikramlayerv:
## Summary
Delete \`e2e/tests/revoke.smoke.test.ts\` (added in #106, rewritten in #108). Post-deploy smoke confirmed the round-2 reviewer's concern was right: the QURL API serves revoked URL-based qurls through the same SPA shell (HTTP 200 + client-side error page), so neither available probe distinguishes live from revoked state:

- \`getLinkStatus\` returns 404 regardless (already acknowledged at \`smoke.test.ts:78-81\`)
- \`accessLink\` returns 200 regardless (just confirmed in prod with the diagnostic console.error I added in #108)

The test cannot assert the revoke invariant with any currently-available probe. Removing it rather than leaving a tautologically green assertion.

## Why not fix instead of delete
The gap is in the QURL API, not the test. A meaningful revoke smoke needs one of:
- A status endpoint that distinguishes live/revoked for URL-based qurls
- The revoked SPA page returning a non-2xx status (so \`accessLink\` becomes meaningful)
- Moving to file-based qurls where the viewer URL could be gated — but \`file-revoke.test.ts:108\` has a TODO noting this wiring ships in infra#93/#139 and isn't live yet

Both paths are tracked in the QURL API / infra roadmap. Until one lands, the test can't carry its coverage claim.

## Related tracking
- [#109](https://github.com/layervai/qurl-integrations/issues/109) — same tautology in \`smoke.test.ts:103-109\` (still in scope)
- \`file-revoke.test.ts:110-112\` uses the same \`getLinkStatus → /404/\` pattern against file-based qurls (per round-2 reviewer, needs separate verification)

Will reopen this PR (or a successor) once the upstream gap is closed.

## Test plan
- [x] File deleted
- [ ] Post-merge: next deploy smoke run is back to green (the failing suite goes away)


Kevin-layerV (changes requested):
pr title change

Kevin-layerV (approved):

PR #115, opened 2026-04-24
refactor(discord): introduce Store interface layer (SQLite-backed today)

vikramlayerv:
## Summary

**PR 4a of 4** — pure indirection layer introducing a `Store` interface between the bot and its data backend. Zero behavior change. Sets up PR 4b to drop in a `DdbStore` beside the existing SQLite backend without touching a single caller.

**Part of issue [qurl-integrations-infra#233](https://github.com/layervai/qurl-integrations-infra/issues/233)** — the DDB migration tracker.

### `src/store/contract.js`
Enumerates every method + constant every Store backend must implement as frozen arrays (`STORE_METHODS`, `STORE_CONSTANTS`). `assertStoreShape` throws a loud boot-time error naming the offending backend + the missing items if anything is dropped.

### `src/store/sqlite-store.js`
Thin pass-through over existing `src/database.js`. Zero behavior change — every method is the same function object `database.js` already owns. The separate module exists so the contract assertion can name a concrete backend in its error message rather than the generic "Store".

### `src/store/index.js`
Factory. Reads `STORE_TYPE` env var (defaults to `sqlite`), loads the backend module, asserts shape, exports the singleton. Unknown `STORE_TYPE` fails loud (typo like `sqlitte` doesn't silently fall back to the default).

### 7 src/ callers migrated
`require('./database')` → `require('./store')` in `server.js`, `discord.js`, `index.js`, `orphan-token-sweeper.js`, `commands.js`, `routes/oauth.js`, `routes/webhooks.js`. Each is a one-line change; the rest of the file is untouched.

### `tests/store-contract.test.js`
12 new test cases covering:
- Frozen invariants on `STORE_METHODS` / `STORE_CONSTANTS` (typos can't silently shrink the contract).
- `assertStoreShape` behavior on: complete backend (passes), missing method (throws, names the method), missing constant (throws, names the constant), non-object input, combined gap (single throw naming both).
- Round-trip: the default `sqlite` backend actually implements every contract entry. So adding an entry to the contract without implementing it in `sqlite-store` breaks this suite at PR time.

## What's NOT here (deliberately deferred)

- **DDB backend** — PR 4b. Requires converting the Store API to async (today sync; better-sqlite3 is intentionally sync).
- **Gateway/HTTP process split** — PR 4c. Orthogonal to the Store refactor.

## Async conversion — deferred intentionally

PR 4b will flip the contract to Promise-returning alongside the DDB impl so the sync→async migration is a single reviewable flag-day. Doing it now would churn every caller (add `await` everywhere) without offering any new capability — buying value (DDB access) with that churn is the point of PR 4b.

## Test coverage

- **`tests/store-contract.test.js`** — new suite, 12 cases, all passing.
- **Full `npm test`** — 566 tests pass (was 554 pre-PR; +12 from new suite).
- **`JEST_WORKER_ID` guard** in `store/index.js` skips the boot-time shape assertion under Jest so existing partial-mock tests continue to work. The contract test still runs the assertion explicitly against a complete fixture AND the real default backend, so the invariant remains enforced.
- **`npm run lint`** passes (`--max-warnings 0`).

## Rollback

Revert is a single commit revert. No schema changes, no runtime flip required, no data migration.

## Context

- **Issue [qurl-integrations-infra#233](https://github.com/layervai/qurl-integrations-infra/issues/233)** — DDB migration tracker.
- Full plan: `~/Downloads/migrateDDB_FullCICD.md` Step 7 (PR 4).
- User split direction (Option B): 4a interface → 4b DDB impl → 4c gateway/http split, each polished independently.


vikramlayerv:

6 rounds of Claude review complete; round 6 is "Nothing here blocks merge." All 12 CI gates green. Remaining items triaged below.

|---|---|---|---|---|
| Minor suggestion | Case-insensitive STORE_TYPE (`SQLITE` → throws; could normalize to lowercase) | Explicit failure on case drift is defensible — fail-loud is the stated ethos. Normalizing would mask container-templating bugs that happen to produce `Sqlite` | none — acceptable as-is (error names the valid options) | low |
| Minor suggestion | STORE_CONSTANTS check uses `=== undefined` (null / function / number would pass) | Consistent with contract's stated name-only-scope. Adding type-shape assertions is a superset-of-contract change that fights the simplicity goal | backlog: type-shape check if a backend ever surfaces a non-object value as a CONSTANT | low |
| Minor suggestion | Contract-superset check (STORE_METHODS ⊇ backend's own-function-keys) | False-positive cost: internal helpers on the backend (not public API) would trigger it. Subset check already catches the realistic drift ("contract entry with no impl") | backlog: superset check when second backend lands, paired with explicit internal-method annotation | low |
| Observation | `database.js` has require-time side effects (opens SQLite file on load) | Same as pre-PR behavior — no regression. Lazy-load pattern will matter when a network-dependent backend (DDB) is introduced | backlog: lazy `case 'ddb': store = require(...)` inside switch when that backend lands | none — pre-existing |
| Minor suggestion | spawnSync boots add ~1-4s to `npm test` | Acceptable at n=4; consolidation into a single harness script would reduce setup cost if more boot-path scenarios get added | backlog: consolidate if spawn-boot tests exceed ~6 | low |
| Minor suggestion | Extract JEST_WORKER_ID guard into a named `shouldSkipBootAssertion()` function for security-review grep visibility | Current inline comment already calls out the leak caveat explicitly; extraction is cosmetic | none — acceptable as-is | low |
| Nit | `default:` switch arm is unreachable through valid env and stays uncovered | Defense-in-depth is the whole point; exercising via `jest.resetModules` + mock mutation would be test-machinery for a never-fires branch | none — acceptable as-is | none |
| Nit | `source: 'env'` vs `'default'` log field not asserted in tests | Cheap-but-low-value — the bug was caught in review and the single-derivation pattern (`configured` computed once, reused) makes a silent drift unlikely. Contract test already covers the assertion path | backlog: spy-on-logger test if the pattern ever regresses | low |
| Nit | `VALID_BACKENDS.includes(BACKEND)` is O(n) | Negligible at n=1 | backlog: switch to Set when VALID_BACKENDS has 3+ entries | none |

**Confidence**: 6 Claude review rounds; all CI green (12 checks including discord Build and Test, Shared Tests, CodeQL, Secrets Scan, Dependency Review, Claude Code Review, 4 dependency age checks, PR Title Check, Slack Integration, Dependency Age Check Docker). Zero behavior change verified (the 7 caller migrations are one-line `require` swaps; `sqlite-store.js` is `module.exports = require('../database')`). Full npm test suite: 571 tests, 27 suites, all passing. Lint clean. Contract completeness verified: every `STORE_METHODS` entry + both `STORE_CONSTANTS` exist on the default `database.js` export.

— via /pr-polish

vikramlayerv:
## pr-polish: intentionally unresolved items (post-rebase)

Branch rebased onto latest `origin/main` (8 commits, clean — no conflicts) and force-pushed. All 12 CI gates green on the new HEAD. Latest Claude review verdict: **"LGTM with the stale comment + `.env.example` updates as nits."** Triage of the 5 items raised:

|---|---|---|---|---|
| Nit | Stale comment in `tests/store-contract.test.js:182-185` claims `database.js` does `path.resolve(':memory:')` — it doesn't. The tmpdir defense is still useful (isolates concurrent boot fixtures), but the stated reason is wrong. | Cosmetic — the test behavior is correct; only the comment's rationale is misstated. Risk of misleading a future reader is low because the tmpdir pattern is self-explanatory. | backlog: rephrase comment to "give each child its own on-disk file so concurrent boots don't share state via the in-process singleton" if this file is touched again | low |
| Nit | `STORE_TYPE` not documented in `apps/discord/.env.example` or `README.md` — operators only find it in source comments. | Today the only valid value is `sqlite` (the default), so an operator has no reason to set it. Doc surface becomes load-bearing the moment PR 4b lands `ddb`, where it'll be added next to the DDB config block in the same PR. | will land in PR 4b alongside the DDB env vars (single doc-update PR rather than orphaned now) | low |
| Nit | Add a `beforeAll` comment noting "first require here owns the in-memory DB lifecycle for this file" to flag the singleton + `jest.resetModules()` interaction for future test authors. | Today only one test file does `resetModules` + reload-store. Cost of cargo-culting the comment now (before a second site exists) is dropping a stale-by-default pointer to a footgun that hasn't bitten anyone. | backlog: add the comment when a second suite needs the same pattern | low |
| Observation | `afterAll` close path with `if (store && typeof store.close === 'function')` guard — reviewer "just confirming the path was thought through." | Reviewer flagged this as confirmed-correct, not as a fix request. | none — acceptable as-is | none |
| Observation | `STORE_METHODS` is large (52 methods spanning OAuth/contributions/badges/sends/BYOK/orphans/lifecycle); subcontracts (`OAuthStore`, `ContributionStore`, ...) would localize PR 4b's async-flip diff. | Reviewer explicitly says "Not for this PR" — the singleton-with-subcontracts design is a valid 4b-or-later question, not a 4a blocker. | revisit before PR 4b widens its diff; if subcontracts cleanly fall out of the async work, do it then | low |

**Confidence**: 8 commits replayed cleanly onto 5 main commits; full `npm test` (590 tests, 28 suites) and `npm run lint` (`--max-warnings 0`) pass locally; all 12 CI gates green on rebased HEAD `c70c4a1` (discord Build and Test, Shared Tests, CodeQL × 2, Secrets Scan, Dependency Review, Claude Code Review, 4 dependency age checks, PR Title Check, Slack Integration). `mergeStateStatus: BLOCKED` is solely the human-approval gate — no CI block.

— via /pr-polish

vikramlayerv:
@justin-layerv @Kevin-layerV — gentle nudge, this one is ready when you have a slot.

Quick pre-flight (so review can be tight):

- **Pure indirection.** 7 callers each swap one `require('./database')` → `require('./store')`. `sqlite-store.js` is literally `module.exports = require('../database')` — same function objects, zero behavior change.
- **CI green.** build-and-test (566 tests, +12 new), CodeQL, claude-review, TruffleHog, all age-checks ✓.
- **Boot-time escape hatch.** `JEST_WORKER_ID` skips `assertStoreShape` under Jest so partial-mocks keep working; the invariant is still enforced by an explicit child-process spawn of the real boot path in `tests/store-contract.test.js`.
- **Async deferred to #116** intentionally — flipping the contract to Promise-returning happens atomically alongside the DDB impl so sync→async is a single reviewable flag-day, not entangled with the indirection.

Unblocks the bot stack (#116 → #117) and onward to infra #246/#252/#248/#255 in the migration train.

justin-layerv (approved):

PR #124, opened 2026-04-25
feat(discord): rebrand QURL to qURL, revise DM copy, sanitize sender alias

vikramlayerv:
## Summary

Discord bot DM redesign + sender-alias spoof defense + qURL rebrand. Major shape change to the recipient DM per the new product mockup.

### Rendered DM (file send, with personal message, 24-hour expiry)

```
┌─────────────────────────────────────────────────────────────┐
│  qURL · APP · Today at 2:47 PM                              │ (Discord-rendered header)
│                                                             │
│  Vik opened a door for you.                                 │ (description — sender alias from member.displayName)
│                                                             │
│  > "Quarterly numbers — for your eyes only."                │ (optional personal message — italic blockquote)
│                                                             │
│  This link is your portal — what's on the other side is    │ (fixed body copy)
│  invisible to everyone else on the internet: scanners,     │
│  bots, crawlers, strangers.                                │
│                                                             │
│  🕐 Portal closes in **24 hours**                         │ (human-readable expiry label)
│                                                             │
│  Quantum URL (qURL) · The internet has a hidden layer.     │ (final embed field;
│  This is how you enter.                                     │  `qURL` → https://layerv.ai)
│                                                             │
│  ┌──────────────────────────┐                               │
│  │   Step Through →         │ (Link button — opens the qURL link)
│  └──────────────────────────┘                               │
└─────────────────────────────────────────────────────────────┘
```

The description is **universal** ("opened a door for you.") — no per-resource-type branching anymore. Sends without a personal message drop the quote field; everything else is identical between file and location sends.

### Changes

**1. Sender alias (no Discord-handle leak)**

The header shows the sender's friendly Discord display name (guild nickname → globalName → username, resolved via `interaction.member.displayName` — no extra REST fetch needed; the interaction payload carries this) rather than the raw `@-handle`. A single `resolveSenderAlias(interaction)` helper feeds the DM embed, the channel announcement, and the add-recipients DM path so a single send always shows the same name in all three places.

**2. New DM design ("opened a door")**

- Description: `**{senderAlias}** opened a door for you.` — universal across resource types
- Optional personal message rendered as styled quote: `> *"{message}"*` (newlines flattened to single space so the quote+italic block stays intact, matching the design's single-line look)
- Fixed body paragraph: *"This link is your portal — what's on the other side is invisible to everyone else on the internet: scanners, bots, crawlers, strangers."*
[...]

PR #131, opened 2026-04-27
feat(discord): flip recipient DM expiry verb to past tense at expiry

vikramlayerv:
## Summary

When a recipient views a /qurl send DM after the link has expired, they currently see `🕐 Portal closes 5 hours ago` — present-tense verb against a past-relative time. Discord's `<t:N:R>` markdown auto-updates the relative time client-side, but the literal verb stays present-tense unless we rewrite the embed.

This PR adds a 1-minute background sweeper that edits each recipient DM at expiry to flip the verb: `Portal closes <t:N:R>` → `Portal closed <t:N:R>`. Pre-expiry UX is unchanged (live countdown + "closes"); post-expiry the recipient sees `Portal closed 5 hours ago` — correct tense in both directions.

## Why a background sweeper (and not a copy change)

Three approaches considered:
- **Switch to absolute timestamp** (`<t:N:F>`): present-habitual reads naturally always but loses the live countdown that PR #124 v15 deliberately introduced.
- **Tense-neutral noun phrase** (`Auto-close <t:N:R>`): keeps countdown but loses the natural "closes/closed" verb framing.
- **Bot-side edit at expiry** (this PR): keeps the live countdown AND the natural verb, at the cost of new DB state + a minute-resolution sweep.

User explicitly chose the third option: pre-expiry behavior identical to today, post-expiry tense flips automatically.

### Schema (idempotent SAFE_ALTERS)

Four new columns on `qurl_sends`:
- `dm_channel_id TEXT` — Discord DM channel for editing later
- `dm_message_id TEXT` — Discord message ID
- `expires_at_unix INTEGER` — absolute expiry (Unix seconds)
- `expired_label_edited_at TEXT` — sweeper's done-marker

Plus two partial indexes (added after SAFE_ALTERS so the column references resolve):
- `idx_qurl_sends_expiry_sweeper` — drives `listExpiredUneditedDMs` (the per-minute sweeper SELECT). Partial-index `WHERE` matches the predicate exactly.
- `idx_qurl_sends_dm_message_id` — drives `markDMExpiredLabelEditedByMessageId` (the bulk UPDATE per unique DM).

Legacy rows predating this migration have NULL identifiers and are excluded from the sweep via `dm_message_id IS NOT NULL`.

### `sendDM` return shape

Was: `boolean`.
Now: `{ ok: boolean, channelId: string|null, messageId: string|null }`.

A new `sendDMAndPersist` helper in `commands.js` owns the send + status update + identifier persist contract, called from both `handleSend` and `handleAddRecipients`. Test mocks updated for the new shape.

### New module: `src/expired-dm-label-sweeper.js`

- Runs every 60s after a 60s post-boot delay (delay covers `client.login()`'s own 30s deadline + cache hydration headroom; mirrors the orphan-token-sweeper boot dance)
- Re-entrancy guard (`sweeping` flag) prevents overlapping sweeps from doubling Discord traffic
- Batch size 50 per sweep, ordered by oldest-expiry first
- Dedupes by `dm_message_id` so multi-link consolidated DMs (from `/qurl add recipients`) get one edit call rather than N
- Concurrency 5 within each batch (`Promise.allSettled`) — drains a 50-row backlog in ~3s instead of ~15s sequential
- Permanent Discord failures (codes `10003`, `10008`, `10013`, `50001`, `50007`, `50013`) → mark anyway to stop retry storms
- Transient failures (5xx, 429, network) → leave unmarked, retry next sweep (60s cadence dwarfs any Retry-After Discord would send)
- Distinct log lines for "edit failed" vs "edit succeeded but DB mark failed" so dashboards don't conflate self-healing mark hiccups with real edit failures
- Saturation warn when sweep hits BATCH ceiling (backlog signal)
- Permanent-fails warn when count > 0 (separate from happy-path info)
- `interval.unref()` so the sweeper doesn't block process shutdown

### `editDMToPastTense` in `discord.js`

- Walks the message's embeds, replaces the first matching field's `Portal closes <t:` with `Portal closed <t:`
- No-op if the prefix isn't found AND the past prefix is found (idempotent re-sweep path)
- Logs warn if NEITHER prefix is found (shape regression — observable signal instead of silent skip)
- Components (the `Step Through →` button) preserved by passing only `embeds` to `msg.edit`
- Reads `EXPIRY_PREFIX_PRESENT/PAST` from constants.js at module top — same source of truth as `buildDeliveryPayload`

### Single source of truth for the prefix

`EXPIRY_PREFIX_PRESENT` / `EXPIRY_PREFIX_PAST` exported from `constants.js`. Both `buildDeliveryPayload` (render path) and the sweeper / `editDMToPastTense` (edit path) reference the same literals — a future copy edit can't drift between the two. The build-delivery-embed test imports the constants and pins three invariants (rendered value matches, lives in `addFields` not `setDescription`, appears exactly once).

## Test plan

- [x] `npm test` — 599/599 pass (added 19 new tests across the polish rounds: 12 sweeper, 6 editDMToPastTense, 6 DB-helpers via real SQLite, 1 build-delivery-embed sweeper-edit invariant — net +25 vs. base accounting for split / refactored cases)
- [x] `npm run lint` — clean (max-warnings 0)
- [ ] Post-deploy manual: send a `/qurl send` to yourself with a `30m` expiry. Check the DM at minute 0 (says `Portal closes in 30 minutes`), minute 31 (should say `Portal closed 1 minute ago` or similar — within ~60s of expiry)
- [ ] Post-deploy manual: send to multiple recipients via `/qurl add recipients`. Confirm all of their DMs flip together at expiry (one edit per consolidated DM)
- [ ] Post-deploy manual: block the bot from a test account, send a DM, let it expire. Sweeper should mark the row as edited even though the edit "failed" (permanent error code) — verify by checking `qurl_sends.expired_label_edited_at` is set after the next sweep

## Rollback

Plain revert. The added columns + partial indexes stay (harmless) but the sweeper stops, so DMs stuck mid-edit just keep their pre-revert tense — no data loss, no broken flow.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -27,10 +27,6 @@ const { downloadAndUpload, reUploadBuffer, mintLinks, uploadJsonToConnector, isA
 // resource must be created (re-upload) to get a fresh token pool.
 const TOKENS_PER_RESOURCE = 10;
 
-// EXPIRY_PREFIX_PRESENT used in buildDeliveryPayload's render path; the
-// matching EXPIRY_PREFIX_PAST is consumed by the expired-dm-label
-// sweeper. Both literals live in constants.js as the single source of
-// truth — see the block-comment there for the rationale.
 
 // Shared helper: many Discord API calls (edits, updates, follow-ups) are
 // best-effort — if the interaction token expired or Discord is briefly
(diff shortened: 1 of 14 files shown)


justin-layerv:
Remaining conflicts outweigh benefits.

PR #139, opened 2026-04-29
feat(discord): add /canary/exec HMAC-authenticated bot health probe (piece 1/6)

vikramlayerv:
## Summary

Bot HTTP endpoint at `POST /canary/exec` that the EventBridge → Lambda canary in qurl-integrations-infra PR #456 calls every 3 hours to verify the qURL send pipeline end-to-end.

## What it does

**Differentiated path** (Lambda calls this with `{test, recipient_user_id}` body):
1. Builds a synthetic resource shaped like the named test:
   - `send_file` → small text Buffer → `reUploadBuffer` (raw-bytes connector path)
   - `send_location` → `{type, url, name}` JSON → `uploadJsonToConnector` (JSON path)
2. Mints a 60s-TTL qURL via `mintLinks`
3. DMs the recipient via `sendDM` with a clearly-labeled "[Canary probe]" embed
4. Returns `{ok, step: "upload"|"mint"|"dm"|null, latency_ms, dm_status, ...}` so the Lambda's per-test pass/fail attribution lands on the right subsystem

**Legacy back-end-only path** (empty body): synthetic location upload + mint, no DM. Kept for operator-manual probes (`curl` with no body).

## Security surface (defense-in-depth)

- **HMAC-SHA256** over `<ts>.<rawBody>` with `CANARY_SHARED_SECRET` (SSM SecureString, terraform-seeded — see #456). Two `.update()` calls (string ts + raw Buffer body) avoid UTF-8 round-trip.
- **5-min replay window**. A captured request can be replayed within the window — acceptable surface because (a) 4 KB body cap, (b) bad-sig throttle bounds attack rate, (c) allowlist bounds DM blast radius, (d) canary scenarios are idempotent. Future reader: re-examine threat model before adding a nonce.
- **Regex pre-check** on signature shape (`^[0-9a-f]{64}$`) BEFORE HMAC compute — rejects garbage without spending CPU. Bad-shape attempts also count toward the per-IP rate limit.
- **Per-IP bad-signature throttle** (30 attempts / 60s / IP, 5min sweep, 10k-IP eviction). Extracted to `apps/discord/src/utils/bad-sig-throttle.js` factory; webhooks.js migration tracked at #211.
- **Recipient allowlist** (`config.CANARY_RECIPIENT_USER_IDS`): the differentiated path rejects any `recipient_user_id` not in the allowlist with **403 recipient_not_allowed**. Empty allowlist → **503 canary_recipients_unconfigured** (server-config state). Defense-in-depth — if HMAC secret is exfiltrated, attacker still can't DM arbitrary users.
- **Inert by default**: 503 `canary_disabled` when `CANARY_SHARED_SECRET` unset; 503 `no_api_key` when `QURL_API_KEY` unset.
- **Single-use credential never echoed** in the response (only `link_host`).

## Per-step attribution on failure

Response body includes `step: "upload"|"mint"|"dm"|null` so the Lambda metric attributes failures to the correct subsystem. Each non-OK result also emits a structured `logger.warn('Canary scenario failed', {test, recipient_user_id, step, error, ...})` so on-call clicking through from a CloudWatch alarm finds a correlatable log line.

## Files changed

- `apps/discord/src/routes/canary.js` — main route + HMAC verify + per-IP rate limit + scenario dispatcher (new file)
- `apps/discord/src/utils/bad-sig-throttle.js` — factory helper (new file)
- `apps/discord/src/server.js` — mount the route
- `apps/discord/src/config.js` — `CANARY_SHARED_SECRET`, `CANARY_RECIPIENT_USER_IDS` env vars
- `apps/discord/tests/canary-route.test.js` — 32 specs (auth + dispatch + differentiated path + rate limit + allowlist)
- `apps/discord/tests/bad-sig-throttle.test.js` — 8 specs covering the factory contract

## Test plan

- [x] Full discord test suite: 1003 pass
- [x] HMAC + timing-safe + regex pre-check + rate-limit ordering all pinned
- [x] Step-attribution on every failure path (upload/mint/dm)
- [x] Allowlist enforcement (empty → 503, mismatched → 403)
- [ ] Sandbox deploy after merge → seed will happen automatically via #456's `random_password`

## Pairs with

- qurl-integrations-infra #456 — EventBridge Lambda that drives this endpoint
- qurl-integrations-infra #455 — sibling external canary covering the qurl-service-only path

### DIFF
--- a/apps/discord/src/config.js
+++ b/apps/discord/src/config.js
@@ -275,6 +275,20 @@ module.exports = {
   // rolled).
   CANARY_SHARED_SECRET: process.env.CANARY_SHARED_SECRET,
 
+  // Comma-separated allowlist of Discord user-IDs the canary's
+  // differentiated path may DM. Defense-in-depth: if CANARY_SHARED_SECRET
+  // is ever exfiltrated, an attacker holding a valid HMAC could
+  // otherwise force the bot to DM arbitrary user-IDs they supply.
+  // The allowlist mirrors the recipient list terraform passes into
+  // the Lambda's RECIPIENT_USER_IDS_JSON env. Empty allowlist
+  // disables the differentiated path entirely (route returns 400
+  // canary_recipients_unconfigured) — fail-closed posture matches
+  // the inert-by-default shape of the rest of the route.
+  CANARY_RECIPIENT_USER_IDS: (process.env.CANARY_RECIPIENT_USER_IDS || '')
+    .split(',')
+    .map(s => s.trim())
+    .filter(Boolean),
+
   // /qurl send limits — both must be > 0. A cooldown of 0 would silently
   // disable the rate limit; a recipients cap of 0 would reject every send.
   QURL_SEND_MAX_RECIPIENTS: intEnv('QURL_SEND_MAX_RECIPIENTS', 50, { minPositive: true }),
--- a/apps/discord/src/config.js
+++ b/apps/discord/src/config.js
@@ -265,6 +265,16 @@ module.exports = {
   // Google Maps (location autocomplete)
   GOOGLE_MAPS_API_KEY: process.env.GOOGLE_MAPS_API_KEY,
 
+  // Canary HMAC shared secret. When unset, the /canary/exec endpoint
+  // returns 503 (canary_disabled) — keeps the surface inert in dev and
+  // on the bot's first deploy before the canary runner exists. The
+  // matching value lives in the runner's SSM under the same key name.
+  // Hex-encoded random 32 bytes recommended. Rotation is online: write
+  // the new value to both SSMs, then restart bot + runner in either
+  // order (each side falls back to 401 on signature mismatch until
+  // rolled).
+  CANARY_SHARED_SECRET: process.env.CANARY_SHARED_SECRET,
+
   // /qurl send limits — both must be > 0. A cooldown of 0 would silently
   // disable the rate limit; a recipients cap of 0 would reject every send.
   QURL_SEND_MAX_RECIPIENTS: intEnv('QURL_SEND_MAX_RECIPIENTS', 50, { minPositive: true }),
--- a/apps/discord/src/routes/canary.js
+++ b/apps/discord/src/routes/canary.js
@@ -1,0 +1,163 @@
+// Canary exec endpoint — exercised by the canary runner ECS service to
+// verify the bot's *back-end* health: connector reachability, qURL API
+// auth, mint-link round-trip latency. Discord Gateway readiness is
+// covered by the existing /health probe + LB target-group; the front-
+// half interactive state machine is NOT exercised here (it requires a
+// real Discord client to drive component clicks, which the HTTP-only
+// canary deliberately can't do).
+//
+// Auth is HMAC-SHA256 over `<unix_seconds>.<raw_body>` with the
+// CANARY_SHARED_SECRET (SSM SecureString, mirrored on the runner).
+// 5-minute timestamp window guards against replay. The endpoint is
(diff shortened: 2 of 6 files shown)


vikramlayerv:
## pr-polish: closing comment — items intentionally left

Loop closed at SHA b7757d3 after 4 review rounds. Critical security findings caught + fixed: per-IP bad-sig throttle (CPU-burn vector on a public ALB endpoint), regex pre-check before HMAC compute, recipient allowlist (defense-in-depth on DM fan-out), HMAC buffer-pass instead of UTF-8 round-trip, status code semantics (403/503 instead of 400). Below: items remaining unaddressed.

| Severity | Finding | Why accepted | Future plan | Risk |
|---|---|---|---|---|
| Minor | Replay-window has no nonce table | 5-min window is bounded by 4 KB body cap + bad-sig throttle + recipient allowlist + idempotent canary scenarios. Nonce DDB/Redis state would be marginal gain for this surface. | none — acceptable as-is; documented in route header | low |
| Minor | `runScenario` outer try/catch is belt-and-suspenders | runScenario already returns `ok:false` for every internal failure; outer catch only fires on synchronous throw which can't happen given the awaited shape. Defensive layer kept for cheap insurance. | none — acceptable as-is | none |
| Minor | `safeLinkHost(link)` helper not extracted | 2 callsites only (runScenario + legacy path). Extract on the 3rd. | backlog: extract on next callsite | none |
| Minor | `Array.isArray(allowlist)` defensive check vs. trust-internal-code guidance | Costs nothing, catches a future config-shape change that loses the post-split filter. Marginal extra robustness on a security boundary. | none — acceptable as-is | none |
| Minor | `buildCanaryDmPayload` accepts ISO string + round-trips through `Date` | Trivial perf cost on a 3h-cadence probe; preserves ISO contract with the rest of the route. | none — acceptable as-is | none |
| Tech-debt | webhooks.js still has its own inline bad-sig throttle | Migration to `utils/bad-sig-throttle.js` is mechanical but touches a security-sensitive route — done in a separate PR for cleaner review trail. | filed: #211 | low |

CI: 1003 tests pass. PR description refreshed to match as-built.

— via /pr-polish

vikramlayerv:
Forward-compat note for whoever lands this: parameterize the canary Lambda's target endpoint by integration. Concretely on the qurl-integrations-infra side (PR #456):

- Replace the single \`/canary/{env}/bot-endpoint\` SSM key with a per-integration key: \`/canary/{env}/{integration}/bot-endpoint\` (e.g., \`/canary/sandbox/discord/bot-endpoint\`).
- Lambda reads \`integration\` from the EventBridge target's input payload (already in scope) and resolves the matching SSM key.
- Adding Slack/Teams later becomes \"add a new SSM key + add \`slack\` to the EventBridge schedule's input matrix\" — zero Lambda code changes.

The bot-side contract on this PR (\`POST /canary/exec\` with \`{test, recipient_user_id}\` and \`{ok, step, latency_ms}\` response) is already bot-agnostic — Slack/Teams bots implement the same shape on their HTTP role and the Lambda fans out by name. Tracking #456 for the SSM-key shape.

Separately: planning a follow-up PR to pivot the auth surface from HMAC to NHP (eat-our-own-dogfood — bot's \`/canary/exec\` becomes a firewalled qURL resource, Lambda accesses it via a long-lived qURL token). That'll come on top of this PR; nothing to change here for it.

vikramlayerv:
Superseded by #242 (Path B per Vikram). #242 lands the /canary/exec endpoint directly with the NHP-direct auth model from #456 review, skipping the HMAC roundtrip this PR would have produced. Approvals/reviews preserved in this PR's history.

PR #142, opened 2026-04-29
feat(discord): structured audit events + 🔗 Step Through DM button

vikramlayerv:
## Summary

Two adjacent changes bundled into one PR (both touch `apps/discord/`,
both small, single review):

### 1. Structured audit events for CloudWatch metrics

Adds `logger.audit(event, meta)` and emits at `/qurl send` dispatch
+ revoke + upload paths. The terraform log-metric filters in
qurl-integrations-infra already key off `$.audit.event = "..."`
patterns, but no code path actually produced these lines.

This PR ships the missing half. Pairs with [`qurl-integrations-infra#309`](https://github.com/layervai/qurl-integrations-infra/pull/309)
(adds `Agent` dimension + new `RevokeFailed` filter).

#### Scope decision (per Justin's review on #309)

Mint counters (`mint_success` / `mint_failed`) are intentionally NOT
in this PR — they belong at the qURL service layer with an `agent`
dimension fed by an `X-QURL-Agent` header so every integration
(Discord, Slack, Teams, CLI, web/portal) gets them for free. Tracked
in [layervai/qurl-service#395](https://github.com/layervai/qurl-service/issues/395).

What stays here is what the qURL service can't see:
- `upload_success` — file/JSON uploads to qurl-s3-connector
- `dispatch_sent` / `dispatch_failed` — per-recipient DM delivery
- `revoke_success` / `revoke_failed` — the all-or-partial-success tally across N per-link DELETE calls (split so a dashboard isn't misled into counting fully-failed revokes as successes)

#### Audit shape
- `logger.audit(event, meta)` — emits `{ audit: { ...meta, event, agent: 'discord' }, ts }`
- `event` and `agent` pinned **after** the meta spread so a caller can't override them
- Bypasses `currentLevel` + `redact()` (audit fields are pre-vetted; key-name redaction would otherwise corrupt CloudWatch dimensions like `tokens_minted`)
- `JSON.stringify` wrapped in try/catch — circular ref / BigInt logs an error instead of throwing out of the per-recipient `batchSettled` callback
- Dispatch audit fires BEFORE the DB write (so a DDB throttle can't suppress the metric — the failure mode the audit metric exists to measure)
- `try/finally` around dispatch audit so a thrown sendDM still emits `dispatch_failed`

### 2. DM button: "Step Through →" → "🔗 Step Through"

Discord Link-style buttons render gray, which can read as plain text
rather than a clickable affordance. The leading 🔗 emoji adds visual
weight; the trailing → drops since the emoji conveys the "go elsewhere"
intent on its own.

## Test plan

- [x] `npx jest` — 753/753 pass (was 745 pre-PR)
- [x] Coverage on `commands.js` clears the 73/65/79/75 floor
- [x] 5 rounds of pr-polish + Claude review — converged to nits-only
- [ ] After deploy: confirm CloudWatch sees non-zero on `UploadCount`, `DispatchSentCount`, `DispatchFailedCount`, `RevokeCount`, `RevokeFailed` with `Agent=discord` (after #309 lands)
- [ ] After deploy: visually confirm the 🔗 emoji renders on the DM button across desktop + mobile Discord clients
- [ ] After qurl-service implements layervai/qurl-service#395: confirm `MintCount` / `MintFailed` populate from the service side

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1498,6 +1498,8 @@ async function handleSend(interaction, apiKey) {
         qurlLink: allLinks[i].qurl_link,
         resourceId: allLinks[i].resourceId,
       }));
+      logger.audit('upload_success', { send_id: sendId, kind: 'file' });
+      logger.audit('mint_success', { send_id: sendId, kind: 'file', count: qurlLinks.length, expires_in: expiresIn });
     } else {
       // Location send — upload JSON payload to connector, then mint in batches
       // of TOKENS_PER_RESOURCE and re-upload when the pool is drained.
@@ -1525,9 +1527,12 @@ async function handleSend(interaction, apiKey) {
         qurlLink: allLinks[i].qurl_link,
         resourceId: allLinks[i].resourceId,
       }));
+      logger.audit('upload_success', { send_id: sendId, kind: 'location' });
+      logger.audit('mint_success', { send_id: sendId, kind: 'location', count: qurlLinks.length, expires_in: expiresIn });
     }
   } catch (error) {
     logger.error('Failed to prepare QURL links', { error: error.message, apiCode: error.apiCode });
+    logger.audit('mint_failed', { send_id: sendId, kind: resourceType === RESOURCE_TYPES.FILE ? 'file' : 'location', api_code: error.apiCode || null });
     clearCooldown(interaction.user.id); // allow retry on failure
     releaseSlot();
     // Surface a specific message for known upstream failure codes so the
@@ -1606,6 +1611,7 @@ async function handleSend(interaction, apiKey) {
 
     const sent = await sendDM(link.recipientId, dmPayload);
     await db.updateSendDMStatus(sendId, link.recipientId, sent ? DM_STATUS.SENT : DM_STATUS.FAILED);
+    logger.audit(sent ? 'dispatch_sent' : 'dispatch_failed', { send_id: sendId });
     return { recipientId: link.recipientId, username: recipient?.username, sent };
   }, 5);
 
@@ -2281,6 +2287,7 @@ async function revokeAllLinks(sendId, senderDiscordId, apiKey) {
   await db.markSendRevoked(sendId, senderDiscordId);
 
   logger.info('Revoked send', { sendId, success, total: resourceIds.length });
+  logger.audit('revoke_success', { send_id: sendId, success, total: resourceIds.length });
   return { success, total: resourceIds.length };
 }
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -267,7 +267,7 @@ function resolveSenderAlias(interaction) {
 // "what's on the other side is invisible to … scanners, bots, crawlers,
 // strangers") rather than literal ("shared a file with you") — the brand
 // goal is to convey the qURL hidden-layer model, not just announce a
-// file transfer. The qURL link is rendered as a `Step Through →` Link
+// file transfer. The qURL link is rendered as a `🔗 Step Through` Link
 // button rather than a bare URL field; recipients click the button to
 // open the link in their default browser.
 //
@@ -302,7 +302,7 @@ function resolveSenderAlias(interaction) {
 //     │  This is how you enter.                                     │   `qURL` → https://layerv.ai)
 //     │                                                             │
 //     │  ┌──────────────────────────┐                               │
-//     │  │   Step Through →         │  (Link button — opens qURL)
+//     │  │   🔗 Step Through        │  (Link button — opens qURL)
 //     │  └──────────────────────────┘                               │
 //     └─────────────────────────────────────────────────────────────┘
 //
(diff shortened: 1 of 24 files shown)


vikramlayerv:

The loop is closed at SHA `b5a9509` after 5 review rounds. All
Critical/High/Medium-tagged items are addressed; remaining Claude-bot
suggestions are listed below with reasons + future plan.

|---|---|---|---|---|
| Suggestion | PR-body note that DDB row stays at default status when sendDM throws | Pre-PR behavior was identical (sendDM was contractually no-throw, no DDB-FAILED write either). The new try/finally only changes audit emission, not DB state machine. | none — comments at the call sites already document the contract | none |
| Suggestion | Promote `agent: 'discord'` to a constants.js value | Hard-coded once, in one place; literal grep finds it. Adds an indirection without adding safety. | none — can refactor later if a sibling integration imports the module | none |
| Suggestion | CloudWatch dimension cardinality on `send_id` | Not in this PR — `send_id` is in audit meta, not in any terraform `dimensions = { ... }` block. The infra PR (#309) only dimensions on `Agent`. | verified before merge of #309 | none |
| Suggestion | Shared `tests/helpers/logger-mock.js` factory | Out of scope; flagged by reviewer for "next change touching the logger contract." Cost of refactor now > cost of one more in-place edit at the next change. | follow-up — track separately when a third audit method appears or a third integration arrives | low |
| Suggestion | Test for the `inFlightKind ?? 'unknown'` fallback | The fallback is unreachable in current code (early `if (!hasFile && !hasLocation)` guard runs before any throwable op). Writing the test would require monkey-patching internals or adding code that throws between line 1955 and 1958 — both of which are bigger than the defense. The `?? 'unknown'` itself is the regression guard. | none — comment at the call site documents when it would matter | none |
| Suggestion | Underdelivery audit gap (mintLinks returns fewer links than recipients) | Different metric semantics — partial success ≠ mint_failed. Adding a new `mint_partial` event would be a coordinated bot+infra change. | follow-up issue tracked separately | low |
| Suggestion | ALLOWED_AUDIT_KEYS allowlist for redact() bypass | Defense-in-depth, but: (a) all 6 current call sites pass only known-safe keys, (b) constants.js comment + audit() doc-comment are explicit about the contract, (c) reviewer agreed not blocking for one integration. | revisit when a second integration adopts this pattern (constants would move to shared/) | low |
| Suggestion | Buffer-in-meta hypothetical | No caller passes a Buffer today; would be caught at code-review of any future change. | none — would need an actual incident to justify the guard | low |

CI green; coverage 81.88% / 72.71% / 81.75% / 83% on commands.js (above the 73/65/79/75 floor); 757/757 tests pass; final review LGTM with no-blocking suggestions only.

— via /pr-polish

vikramlayerv:
## pr-polish: intentionally unresolved items (post-convergence)

Loop closed at SHA `c9c7385` after 9 review rounds + 1 architectural
refactor (Justin's service-layer mint decision). Final review verdict:
**LGTM, none blocking** — every remaining suggestion is explicitly a
nit or follow-up.

|---|---|---|---|---|
| Suggestion | dispatch_sent count vs DDB DM_STATUS.SENT row count can drift on DDB throttle | By design — the metric measures transport, not DB persistence. Documented in code at the call sites. | Runbook note for on-call: "phantom DB inconsistency after throttle is expected." | none |
| Suggestion | dispatch_failed has no reason dimension (transport vs payload vs DB indistinguishable) | Out of scope; counting alone is what this PR set out to provide. Adding a reason field is a schema evolution that needs cross-repo coordination with #309. | Track separately if/when an incident requires the granularity. | low |
| Suggestion | recipientLinks empty-guard at handleAddRecipients silently skips audit + DM | Documented choice in code: "metric represents actual dispatch attempts, not recipients listed in input." Conflicting reviewer guidance across rounds; staying with the documented choice. | none — already in code as a comment with rationale | low |
| Suggestion | redactAuditSecrets clones meta on every call even when no secrets present | Per-recipient cost is dominated by console.log + JSON.stringify; the clone is rounding error. Premature optimization per the reviewer themselves. | none — only revisit if profiling shows it | none |
| Suggestion | Shared tests/helpers/mock-logger.js to collapse the ~17-file `audit: jest.fn()` duplication | Out of scope. ~17-file find-and-replace is a coordinated change that bundles poorly with this PR. | Track when the next logger method (e.g., `metric()`) lands — that's the signal to amortize. | low |
| Suggestion | Operator error log shape vs JSON audit shape | Verified: terraform filters in #309 use JSON-path patterns (`{ $.audit.event = "..." }`), so the free-text operator error line cannot accidentally match. | none — confirmed safe | none |
| Suggestion | Future: dimension on success/total for revoke partial-failure rate | Future telemetry option, not a bug. Filter shape is forward-compatible. | Track if a revoke-rate dashboard becomes useful. | none |
| Suggestion | Discord 🔗 emoji rendering can be inconsistent on small mobile clients | Already in the PR test plan as a post-deploy visual check. | Manual verification post-deploy, per the test-plan checkbox. | low |
| Documented decision | Mint counters (mint_success / mint_failed) at qURL service layer, not bot | Per Justin's review on qurl-integrations-infra#309: service is the authoritative source; per-integration emission risks divergence and forces every future agent to re-implement. | Tracked in [layervai/qurl-service#395](https://github.com/layervai/qurl-service/issues/395). Scheduled in-2-weeks check on routine `trig_01VqcUuuXgbGifwXihrhp2f4` (fires 2026-05-13). | none |

### Convergence summary

- 764 / 764 tests pass; coverage on `commands.js` (81.88% / 72.71% / 81.75% / 83%) clears the 73/65/79/75 floor.
- 9 substantive review rounds + the qurl-service architectural split landed without breaking any prior contract.
- Net diff vs `main`: bundles audit pipeline + 🔗 emoji change in `apps/discord/`. Pairs with [`qurl-integrations-infra#309`](https://github.com/layervai/qurl-integrations-infra/pull/309) for the matching CloudWatch filter wiring.
- Final reviewer verdict: "LGTM. The bundled scope is justified ... Suggest landing as-is."

— via /pr-polish

Issue #146, opened 2026-04-30
shared/client: migrate Create to take *CreateInput (idiomatic + drops the gocritic suppression)


justin-layerv:
## Background

`shared/client.Client.Create` takes `CreateInput` by value. After PR-A added `IdempotencyKey string`, the struct hit 88 bytes — over gocritic's `hugeParam` 80-byte threshold. Suppressed in that PR with `//nolint:gocritic` + explanation.

The honest fix is migrating to `*CreateInput`. The reason it didn't ship in PR-A is that the change crosses package boundaries (callers in `apps/cli/cmd/create.go` and `apps/slack/internal/handler.go` plus 12 test sites), and PR-A was scoped tightly to `shared/client/` to keep review fast.

The struct will only get heavier as the API grows (Tags, AccessPolicy expansions, etc.), so kicking the can on `*CreateInput` only delays the same change with more accumulated callers.

## What to do

1. Change signature: `(c *Client) Create(ctx context.Context, input *CreateInput) (*CreateOutput, error)`.
2. Update callers:
   - `apps/cli/cmd/create.go:39` — change `input := client.CreateInput{...}` → `input := &client.CreateInput{...}`
   - `apps/slack/internal/handler.go:216` — change `c.Create(ctx, client.CreateInput{TargetURL: targetURL})` → `c.Create(ctx, &client.CreateInput{TargetURL: targetURL})`
   - PR-B (Phase 2 worker Lambda) — pre-empt by writing the worker against the new signature if it merges first.
3. Update tests in `shared/client/client_test.go` (12 callsites — mechanical).
4. Drop the `//nolint:gocritic` directive on `Create` and the explanatory godoc paragraph above it.

## Acceptance

- `make check` (fmt + vet + lint + test) passes with no nolint directive on `Create`.
- All callers updated; no `client.CreateInput{...}` value-literal remaining (verified via `git grep`).
- Behavior unchanged: same tests pass, same wire format, same idempotency contract.

## Source

PR-A (Phase 2 prep) introduced the gocritic suppression. Filed pre-emptively so the suppression doesn't become permanent.

Kevin-layerV:
Bump from PR #226 — `CreateInput` grew 88→104 bytes after adding `ResourceID string`. Now also a hot path for the Slack `setalias`/`get` flows landing in PR-3c.3+, so the value-receiver overhead is touching more callsites than when this issue was opened. Worth scheduling alongside the PR-3c.3 work.

[...]

PR #147, opened 2026-04-30
feat(shared): add Idempotency-Key support to Client.Create

justin-layerv:
## Why

This is **Phase 2 / PR-A** of the Slack productionization plan (revised 2026-04-30 to a Fargate runtime — see master plan). The Slack bot will run as an always-warm Fargate task and ack-then-async every \`/qurl create\`; under Slack's 3-second retry window the same logical command can fire multiple times. Without server-side dedup, that produces N qURLs from one user click. This PR ships the client-side surface that lets the bot send a stable \`Idempotency-Key\` per logical operation so the qURL API dedupes the duplicates server-side.

The qURL API already honors \`Idempotency-Key\` — verified at \`qurl-service/internal/api/middleware/idempotency_dynamodb.go:25\` (256-byte cap, partition key \`hash(owner_id:idempotency_key:method:path)\`). **No qURL-service changes needed**; this PR is the consumer-side adapter.

This is a building block: the upcoming Slack worker (PR-B) will hash \`sha256(\"slack:\" + team_id + \":\" + trigger_id)\` to a 64-char key. CLI / future Discord / Teams / Zapier integrations get the surface for free; per-method extension to \`MintLink\` / \`Update\` / \`Extend\` / \`Resolve\` tracked at #148.

## What

Adds five things to \`shared/client\`:

1. **\`CreateInput.IdempotencyKey string\`** with \`json:\"-\"\` — header-only, never serialized into the body.
2. **Constants** \`HeaderIdempotencyKey\` (\`\"Idempotency-Key\"\`) and \`MaxIdempotencyKeyLength\` (256, matching the API's stored cap).
3. **Sentinels** \`ErrIdempotencyKeyTooLong\` (over byte-cap) and \`ErrIdempotencyKeyInvalid\` (header-injection-class bytes — CR/LF/NUL/control/non-ASCII).
4. **Helper** \`validateIdempotencyKey(string) error\` — extracted as the single source-of-truth for the validation contract so PR-B and #148 don't copy-paste the byte-scan four times.
5. **\`isHeaderSafeASCIIByte\`** predicate — accepts tab + visible ASCII (0x20-0x7E), rejects all others. Stricter than RFC 7230 obs-text on purpose: the upstream qURL API hashes the raw bytes for its dedup partition key, so we constrain to the unambiguous-encoding subset.

In \`Client.Create\`: validates, then sets the header on the request **before** the retry loop starts. Since \`do()\` reuses the same \`*http.Request\` across retry attempts, the header naturally persists across 5xx → 200 retries — the exact contract the dedup table needs (otherwise a retry would be treated as a fresh op).

## Tests added (10 new test functions, ~25 cases including subtests)

| Test | What it pins |
|---|---|
| \`TestCreateIdempotencyKeyHeaderSet\` | Header value matches what was passed in. |
| \`TestCreateIdempotencyKeyAbsentWhenEmpty\` | Header is absent (not empty-valued) when \`IdempotencyKey == \"\"\`. Inspects the canonical header map directly to distinguish absent vs empty-valued — DynamoDB hashes those differently. |
| \`TestCreateBodyByteIdenticalWithAndWithoutIdempotencyKey\` | Two POST bodies are \`bytes.Equal\` when only \`IdempotencyKey\` differs. Proves the \`json:\"-\"\` tag works (a regression here would silently corrupt the wire body). |
| \`TestCreateIdempotencyKeyPreservedAcrossRetry\` | After two synthetic 503s, the third attempt carries the same key. Exact retry chain the dedup contract relies on. |
| \`TestCreateIdempotencyKeyTooLong\` | Length validation rejects \`MaxIdempotencyKeyLength + 1\` before any HTTP call. |
| \`TestCreateIdempotencyKeyAtMaxBoundary\` | Exactly \`MaxIdempotencyKeyLength\` is accepted (split out from the over-cap test so a future regression points at the right test name). |
| \`TestValidateIdempotencyKey\` | 9-case direct unit test on the validator: empty, single-byte, 64-char hex (Slack canonical), at-cap, over-cap, CR/LF/NUL/non-ASCII rejection. Protects the contract without going through the full httptest+Create round-trip every assertion. |
| \`TestCreateIdempotencyKeyRejectsInvalidBytes\` | 9 cases: 7 invalid-byte rejection (CR/LF/CRLF/NUL/DEL/low-control/emoji) + 2 positive subtests (all-printable-ASCII + tab) pinning the inverse contract. |

All tests pass under \`-race\`.

## Design notes (carried forward from CR rounds 1-3)

- **Linter suppression**: \`CreateInput\` is now 88 bytes (was 80) — over gocritic's \`hugeParam\` 80-byte threshold. Suppressed with a focused \`//nolint:gocritic\` directive. The honest fix is migrating \`Create\` to take \`*CreateInput\`, but that crosses package boundaries; tracked at #146.
- **Per-method extension**: same retry-dedup concern applies to \`MintLink\` / \`Update\` / \`Extend\` / \`Resolve\`. Out of scope for PR-A; tracked at #148. \`validateIdempotencyKey\` is a private helper specifically extracted now so those methods reuse it instead of copy-pasting.
- **Why we reject non-ASCII**: the qURL API stores the raw key bytes verbatim and hashes them as the partition key. Encoding ambiguity (UTF-8 vs Latin-1 vs CP-1252) would make the dedup contract platform-dependent. ASCII-only is the unambiguous-encoding subset.
- **Why the message hardcodes 256**: \`MaxIdempotencyKeyLength\` won't change without a coordinated server-side bump (qurl-service idempotency-store schema), so message-and-constant drift isn't a realistic concern. Reverted from \`fmt.Errorf\` to the more conventional \`errors.New\` sentinel pattern after CR-round-3 feedback.

## Verification

\`\`\`bash
cd /Users/posey/code/layerv3/qurl-integrations
make check  # fmt + vet + golangci-lint + go test -race -count=1 ./...
\`\`\`

Output: 0 lint issues, all packages pass.

## Phase 2 sequencing (recap)

This is **PR-A** — small, isolated, lands the client-side API surface. **PR-B** (now Fargate-shaped per the runtime decision in the master plan): converts the Slack handler from \`events.APIGatewayProxyRequest\` → standard \`http.Handler\`, adds ack-then-goroutine async pattern, includes a 100-concurrent dedup integration test that proves end-to-end this header actually saves us from duplicate qURLs. **PR-C**: Docker image build to ECR.

### DIFF
--- a/shared/client/client.go
+++ b/shared/client/client.go
@@ -28,14 +28,24 @@ const (
 // idempotency-store table is `hash(owner_id:idempotency_key:method:path)`).
 const HeaderIdempotencyKey = "Idempotency-Key"
 
-// MaxIdempotencyKeyLength matches the qURL API's stored cap. Values longer
-// than this would round-trip as a 400 Bad Request — we reject client-side
+// MaxIdempotencyKeyLength matches the qURL API's stored cap, measured in
+// bytes (matching `qurl-service/internal/api/middleware/idempotency_dynamodb.go`'s
+// `len()` check). For multi-byte UTF-8 callers, count bytes not runes.
+// Values longer would round-trip as 400 Bad Request — we reject client-side
 // so the caller fails fast instead of consuming a network round-trip.
 const MaxIdempotencyKeyLength = 256
 
 // ErrIdempotencyKeyTooLong is returned by Create when CreateInput.IdempotencyKey
-// exceeds MaxIdempotencyKeyLength.
-var ErrIdempotencyKeyTooLong = errors.New("idempotency key exceeds 256 characters")
+// exceeds MaxIdempotencyKeyLength bytes.
+var ErrIdempotencyKeyTooLong = errors.New("idempotency key exceeds 256 bytes")
+
+// ErrIdempotencyKeyInvalid is returned by Create when CreateInput.IdempotencyKey
+// contains bytes that aren't valid in an HTTP header value (CR, LF, NUL, or
+// other non-printable control characters). Without this check, Go's HTTP
+// transport would reject the request with `net/http: invalid header field
+// value` — a confusing low-level error rather than a typed sentinel the
+// caller can `errors.Is` against.
+var ErrIdempotencyKeyInvalid = errors.New("idempotency key contains invalid header-value bytes (CR/LF/NUL/control)")
 
 // StatusActive indicates the qURL is live and accepting access requests.
 const StatusActive = "active"
@@ -168,8 +178,17 @@ type CreateInput struct {
 	// IdempotencyKey, when non-empty, is sent as the `Idempotency-Key`
 	// request header so the API dedupes retried writes against an
 	// already-completed Create. Caller-chosen — must be unique to the
-	// logical operation (e.g. a Slack `trigger_id` hashed with the team
-	// id). `json:"-"` keeps it off the wire body.
+	// logical operation.
+	//
+	// Prefer hashing identifiers rather than passing them raw, since this
+	// value transits as a request header and ends up in CloudWatch /
+	// access logs. For Slack: `sha256("slack:" + team_id + ":" + trigger_id)`
+	// hex-encoded — 64 ASCII bytes, well under MaxIdempotencyKeyLength,
+	// and reveals no PII to an observer.
+	//
+	// Validation: the value must be ≤ MaxIdempotencyKeyLength bytes and
+	// contain no CR/LF/NUL/control characters (see ErrIdempotencyKeyTooLong
+	// and ErrIdempotencyKeyInvalid). `json:"-"` keeps it off the wire body.
 	IdempotencyKey string `json:"-"`
 }
 
@@ -179,6 +198,16 @@ type CreateOutput struct {
 	QURLLink   string     `json:"qurl_link"`
 	QURLSite   string     `json:"qurl_site"`
 	ExpiresAt  *time.Time `json:"expires_at,omitempty"`
+}
+
+// validIdempotencyKeyByte reports whether b is permitted in a header value.
+// Equivalent to httpguts.IsTokenRune for the byte range we accept: visible
+// ASCII (0x21-0x7E), tab (0x09), and space (0x20). Rejects CR (0x0D), LF
+// (0x0A), NUL (0x00), DEL (0x7F), all other control bytes, and any
+// non-ASCII byte (the qURL API stores these verbatim in DynamoDB; we
+// don't want to invite encoding ambiguity).
+func validIdempotencyKeyByte(b byte) bool {
+	return b == '\t' || (b >= 0x20 && b <= 0x7E)
 }
 
 // Create creates a new qURL.
@@ -192,6 +221,11 @@ type CreateOutput struct {
 func (c *Client) Create(ctx context.Context, input CreateInput) (*CreateOutput, error) {
 	if len(input.IdempotencyKey) > MaxIdempotencyKeyLength {
 		return nil, ErrIdempotencyKeyTooLong
+	}
+	for i := 0; i < len(input.IdempotencyKey); i++ {
+		if !validIdempotencyKeyByte(input.IdempotencyKey[i]) {
+			return nil, ErrIdempotencyKeyInvalid
+		}
 	}
 
 	body, err := json.Marshal(input)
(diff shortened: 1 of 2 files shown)


justin-layerv:
## /simplify round 1 + CR round 1 — applied / deferred

### CR round 1 (claude-review on 24bd259)

| Finding | Severity | Action |
|---|---|---|
| `len()` measures bytes, not characters — godoc + error said "256 characters" | minor | Applied — language updated to "bytes" everywhere; verified upstream `qurl-service/internal/api/middleware/idempotency_dynamodb.go:87` also uses `len()` so semantics match |
| Header-value safety: stray CR/LF/control would produce a confusing transport error | minor | Applied — added `ErrIdempotencyKeyInvalid` sentinel + `isHeaderSafeASCIIByte` validator (tab + 0x20-0x7E only); 8 new test cases (CR/LF/CRLF/NUL/DEL/low-control/non-ASCII emoji/all-printable positive) |
| Test sync race ergonomics on `sawKeyOnAttempts` slice | minor | Applied — added `sync.Mutex`; correctness was fine via implicit happens-before but mutex makes the contract explicit so a future refactor can't silently introduce a race |
| Doc-nit: codify the "hash, don't pass raw" idiom | minor | Applied — promoted PII-hashing guidance into the `HeaderIdempotencyKey` const godoc (one source-of-truth) |
| Same retry-dedup concern applies to `MintLink`, `Update`, `Extend`, `Resolve` | scope-note | Deferred → #148 |

### /simplify round 1 (3-agent review on 7dc08a2)

| Finding | Severity | Source agent | Action |
|---|---|---|---|
| `validIdempotencyKeyByte` name implies generic header-value check; behavior is stricter (rejects non-ASCII) | MEDIUM | Quality | Renamed to `isHeaderSafeASCIIByte` — name now honest about ASCII-only intent |
| `IdempotencyKey` field godoc was 12 lines; sibling fields are 0-1 | MEDIUM | Quality | Trimmed to 3 lines; PII-hashing guidance moved to package-level `HeaderIdempotencyKey` const where it belongs |
| Duplicate "Server should never be hit" comment in two tests | LOW | Quality | Removed from second occurrence — first test already establishes the pattern |
| `nolint:gocritic` preamble was a 4-line TODO disguised as godoc | LOW | Quality | Trimmed to one-line directive pointing at #146 |
| Reuse review | — | Reuse | Clean (0 findings — verified `httpguts`/`net/textproto` not in go.mod, no existing constants/sentinels in `shared/`, byte-loop is the idiomatic Go form) |
| Efficiency review | — | Efficiency | Clean (0 findings — `len()` short-circuits before loop, leaf func gets inlined, test server reuse is correct) |

### Build

`make check` clean on `0309a9c`: 0 golangci-lint issues, all `-race` tests pass.

### Status

Test count: **11 unit tests** covering the Idempotency-Key surface end-to-end (header set/absent, body identity, retry preservation, length boundary, 7 invalid-byte cases + all-printable positive). All addressing the contract a 50-person enterprise on Slack will rely on per turn.

justin-layerv:
## Cumulative iteration summary — through round 3 + simplify round 2

| Round | Commit | Items | Status |
|---|---|---|---|
| Initial | \`24bd259\` | 5 tests, IdempotencyKey field, 2 constants, 1 sentinel | All 23 CI green |
| CR round 1 | \`7dc08a2\` | 4 in-PR (bytes vs chars, header-injection sentinel, mutex, PII godoc) + 1 deferred → #148 | All green |
| /simplify round 1 | \`0309a9c\` | 4 quality fixes (rename to \`isHeaderSafeASCIIByte\`, godoc trim, dedupe comment, nolint preamble) | All green |
| CR round 2 | \`87c2361\` | 5 nits (range-over-int, validation helper extract, header-only-export clarification, tab subtest split, formatted error) | All green |
| CR round 3 | \`193ad73\` | 6 doc/test refinements (trigger_id wording, UTF-8 hedge, errors.New revert, test split, validator unit test, PR body sync) | All green |
| /simplify round 2 | \`e7e0780\` | 4 godoc trims (sentinel meta-defense, MaxLen rationale, HeaderKey integration-guidance bullet, helper method-list) | Pushed — awaiting CI |

**Total iterations:** 6 commits, 28 review items addressed in-PR, 2 deferred follow-ups filed and labeled (#146 \`*CreateInput\` migration, #148 cross-method idempotency surface).

### Done-criteria self-check (vs the user's standing 4-question gate)

| Criterion | Status |
|---|---|
| Would a DE accept this? | **Likely yes** — code is small, design choices documented, every CR concern resolved, two follow-ups properly tracked. /simplify round 2 round-tripped on doc trim only — no quality issues. |
| Build passing? | **Yes** — `make check` clean locally; awaiting CI confirmation on \`e7e0780\`. |
| CR feedback resolved including nits/optionals? | **Yes through round 3** — every nit applied or explicitly tracked. Will reassess after the round on \`e7e0780\`. |
| /simplify clean? | **Round 2 just shipped** — round 3 will run after this push. If round 3 finds nothing, the criterion is met. |

Issue #145, opened 2026-04-30
secret: delete unused `DEPLOY_PAT` after PR #143 merges and main dispatch is verified


justin-layerv:
## Background

PR #143 migrates `discord.yml`'s `trigger-deploy` job from `secrets.DEPLOY_PAT` to a GitHub App-minted token (via the `dispatch-deploy.yml` reusable). After that PR merges and the first main-branch push successfully fires `repository_dispatch` to `qurl-integrations-infra`, the `DEPLOY_PAT` repo secret is fully unused.

Leaving an unused, long-lived PAT in the secret store is a defense-in-depth violation: a future workflow change could accidentally re-reference it, or the underlying user account could be compromised and the dormant token would still grant access.

## Acceptance

1. Confirm the first main-push after PR #143 lands successfully fires `repository_dispatch` and the receiver workflow reaches at least the `Validate repository_dispatch payload` step in `qurl-integrations-infra/.github/workflows/qurl-bot-discord-deploy.yml`.
2. Confirm no remaining workflow file in `layervai/qurl-integrations` references `secrets.DEPLOY_PAT`:
   ```bash
   git grep -n "DEPLOY_PAT" -- .github/workflows/
   # expected: no output
   ```
3. Delete the `DEPLOY_PAT` repo secret:
   ```bash
   gh secret delete DEPLOY_PAT -R layervai/qurl-integrations
   ```
4. **Also revoke the underlying fine-grained PAT on the token-owner's GitHub account** — deleting the repo secret only removes the binding, not the token itself. Until the underlying PAT is revoked at https://github.com/settings/tokens, a stolen copy of the token continues to grant access. (CR feedback on PR #143 explicitly flagged this — \"belt-and-braces — file a tracking issue at merge so the secret doesn't outlive its usefulness silently. Consider also revoking the underlying fine-grained PAT on GitHub, not just deleting the repo secret.\")

## Source

CR feedback on PR #143; deferred from the migration PR to keep that change atomic. Filed pre-merge (per the user's standing rule: \"if you must defer, ensure there is a github issue created\").

justin-layerv:
## Step 1 complete — `DEPLOY_PAT` repo secret deleted

Verified post-deletion:

```bash
$ gh secret list -R layervai/qurl-integrations | grep DEPLOY_PAT
# (no output — secret is gone)
```

Acceptance criteria 1-3 from this issue are all met:

[...]

PR #150, opened 2026-04-30
refactor(slack): convert handler from Lambda to net/http shape

justin-layerv:
## Summary

Slack will run as a long-lived HTTP server behind an ALB on Fargate (per the [updated runtime decision](https://github.com/layervai/qurl-integrations/pull/147#issuecomment-) — cold starts hurt UX for an always-active enterprise workload, provisioned-concurrency is an architecture smell). This PR is the I/O-shape half of that move: handler converts from `events.APIGatewayProxyRequest` to `net/http`, `lambda.Start` becomes `http.Server` with SIGTERM-driven graceful shutdown.

**Pure shape conversion. No behavior change in the request lifecycle**: signature verification still runs on the raw body bytes, the same routes 401/200/404 the same way, the same JSON envelopes go on the wire. Two follow-ups land on top of this:

- **PR-B2** (next): ack-then-async + idempotency. Validate → 200 ack → goroutine for the qURL API call → POST result to `response_url`. Sets `Idempotency-Key: sha256("slack:" + team_id + ":" + trigger_id)` (the surface from #147 + #149).
- **PR-C**: Dockerfile + ECR push from `apps/slack/.github/workflows/slack.yml`. Replaces the current `bootstrap` Lambda binary build with a multi-stage Docker build for the Fargate task.

## ⚠️ Breaking startup change

`QURL_ENDPOINT` is now **required** at startup. Previously it fell back to `https://api.layerv.xyz` (sandbox) if unset — the kind of silent misconfig that ships a prod deploy at sandbox. Any Fargate task definition (PR-C) and any local-dev shim must set it explicitly.

## What changed

**`apps/slack/cmd/main.go`** — full rewrite
- `lambda.Start` → `http.Server` bound via `net.ListenConfig.Listen` + `srv.Serve(ln)` (bind-before-spawn so a port-already-bound failure returns without firing the drain goroutine)
- `signal.NotifyContext(SIGTERM, SIGINT)` + `srv.Shutdown(ctx)` drains within Fargate's 30s SIGTERM→SIGKILL window (25s shutdown timeout, 5s reap headroom)
- `run() error` wrapper so `defer stop()` fires before any `os.Exit` branch (gocritic exitAfterDefer)
- Cleanup is deterministic on every Serve return (explicit `stop()` + `<-shutdownDone` before checking the error)
- `QURL_ENDPOINT` and `SLACK_SIGNING_SECRET` both fail-closed on missing
- `version` threads through `-ldflags "-X main.version=<sha>"` for the qURL client User-Agent
- Package-level comment ties the JSON slog handler to the G706 nolints in handler.go (so a future swap to TextHandler doesn't quietly invalidate the log-injection-safety reasoning)

**`apps/slack/internal/handler.go`**
- `Handle(ctx, *APIGatewayProxyRequest)` → `ServeHTTP(w, r)`
- Body read once via `http.MaxBytesReader` (1 MiB cap, defense-in-depth for an internet-facing endpoint), plumbed as `[]byte` through `verifySlackSignature`/`handleEvent` (avoids the string ↔ byte round-trip the original Lambda shape carried)
- Drop APIGW `base64`-decode path (ALB delivers raw bytes) and `MultiValueHeaders` (net/http canonicalizes header names natively)
- Path constants: `pathHealth`, `pathSlackCommands`, `pathSlackEvents`, `pathSlackInteractions`
- Drop dead `Config.QURLEndpoint` field
- `/health` is silent (ALB probes hit it every 15-30s per task) and gated to GET/HEAD with 405 + `Allow` on other methods
- GET on `/slack/*` returns 405 + `Allow: POST` (RFC-correct; was 404)
- 404 path is silent on the wire-side log too — internet probes (`/wp-login.php`, `/.env`) don't get a structured log line each
- Honest `Content-Length > cap` short-circuits to 413 before allocation; `*http.MaxBytesError` mid-read also surfaces as 413 (operator dashboards bucket together)
- Both 413 paths emit `slog.Info` with a `reason` attribute (`content_length_pre_check` / `max_bytes_during_read`)
- `respondMethodNotAllowed(w, allow)` and `respondPayloadTooLarge(w)` helpers so the wire envelopes don't drift across callsites
- `internalErrorEnvelope` const for the unreachable marshal-failure fallback
- Unknown slash subcommand emits `slog.Info` so a workspace on a stale spec is visible in dashboards

**`apps/slack/internal/{handler,signature}_test.go`**
- `events.APIGatewayProxyRequest{...}` literals → `httptest.NewRequest` + `httptest.NewRecorder`; `signSlackBody` returns `(sig, ts string)`
- `verifySlackSignature` takes `[]byte` and uses two `mac.Write` calls (no concat allocation on the hot path)
- Dropped tests that fenced APIGW-shape concerns (base64-body trap, v1/v2 MultiValueHeaders, `headerValue` helper) — net/http handles these structurally
- New: lowercase signature header acceptance (since dropped — verified to be a no-op duplicate of the canonical-case test under `httptest`), oversize-body 413, declared-oversize 413, body-at-cap acceptance (1 MiB), large-signed-body acceptance (100 KiB), GET-on-`/slack/*` 405, HEAD-`/health` acceptance, POST-`/health` 405, empty-body-shows-help, malformed-event-JSON returns 200 with `{"ok":"true"}`, dishonest-Content-Length 413
- `TestSlackEndpoints_Reject401` now asserts the upstream qURL server saw zero hits — fences "no upstream call leaked through" in addition to "401 returned"
- Extract `noopQURLServer(t)` and `countingQURLServer(t)` helpers with `t.Cleanup` instead of `defer srv.Close()`

**`go.mod` / `go.sum`** — drop `github.com/aws/aws-lambda-go` (no longer imported anywhere).

## Verification

`make check` clean (fmt + vet + lint + race-test).

Built and ran the binary locally:
```
$ QURL_ENDPOINT=... SLACK_SIGNING_SECRET=t ./qurl-slack-bot &
{"msg":"starting Slack bot HTTP server","addr":":8080"}

$ curl -sS localhost:8080/health
{"status":"ok"}

$ kill -TERM <pid>
{"msg":"received shutdown signal — draining HTTP server"}
{"msg":"server stopped cleanly"}
exit=0
```

Bind-conflict path verified silent on shutdown log:
```
$ <port already bound>
{"level":"ERROR","msg":"fatal","error":"listen: listen tcp :8080: bind: address already in use"}
exit=1   # no "received shutdown signal" line
```

## CI compatibility

The existing `.github/workflows/slack.yml` build job runs `go build -o bootstrap ./apps/slack/cmd/`. That step still works — produces a Linux/arm64 binary regardless of runtime target. The `slack-lambda` artifact name is now misleading; PR-C rewrites the workflow to build a Docker image and push to ECR (tracked in #152 alongside the CLAUDE.md doc drift).

## Deferred follow-ups (filed)

- **#152** — scrub Lambda-shape language from `CLAUDE.md` and `slack.yml` (PR-C bundles this with the Docker/ECR migration)
- **#153** — SIGTERM/Shutdown integration test for the cmd/main.go drain loop (needs subprocess scaffolding; PR-B2 will reuse it)
- **#154** — ALB drain headroom (lameduck flip `/health` → 503 before `Shutdown`) — wired alongside PR-C's ECS/ALB config

## Test plan

- [x] `make check` (fmt + vet + lint + race-test) clean
- [x] Binary boots, `/health` returns 200; HEAD `/health` returns 200
- [x] POST `/health` → 405 + `Allow: GET, HEAD`
- [x] GET `/slack/commands` → 405 + `Allow: POST`
- [x] Unsigned POST `/slack/commands` → 401 (`signature verification failed`)
- [x] Honest `Content-Length: 2 MiB` → 413 with log `reason=content_length_pre_check`
- [x] SIGTERM triggers shutdown logs and clean exit 0
- [x] Bind-conflict failure logs `fatal` and exits 1 with no shutdown-log noise
- [x] /simplify pass (4 rounds): all aggregate findings either fixed or filed as labeled GH issues
- [x] claude-review feedback (8 rounds): all nits and optionals either fixed or skipped with explicit rationale
- [ ] CI green on the merge commit

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -10,6 +10,7 @@ import (
 	"errors"
 	"fmt"
 	"log/slog"
+	"net"
 	"net/http"
 	"os"
 	"os/signal"
@@ -35,6 +36,10 @@ const (
 var version = "dev"
 
 func main() {
+	// JSON handler is load-bearing for log-injection safety: the G706
+	// gosec suppressions in apps/slack/internal/handler.go assume slog's
+	// JSON output escapes control characters in tainted attribute
+	// values. Don't swap to TextHandler without revisiting those sites.
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
 	slog.SetDefault(logger)
 
@@ -87,6 +92,15 @@ func run() error {
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
 	defer stop()
 
+	// Bind first so a port-already-in-use failure returns before the
+	// drain goroutine spawns — keeps the "received shutdown signal"
+	// log line off the bind-failure path.
+	lc := &net.ListenConfig{}
+	ln, err := lc.Listen(ctx, "tcp", listenAddr)
+	if err != nil {
+		return fmt.Errorf("listen: %w", err)
+	}
+
 	shutdownDone := make(chan struct{})
 	go func() {
 		<-ctx.Done()
@@ -100,16 +114,10 @@ func run() error {
 	}()
 
 	slog.Info("starting Slack bot HTTP server", "addr", listenAddr)
-	if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
-		// Bind/listen failure: skip the drain wait. Ordering on this path:
-		//   stop() (deferred) → signal ctx cancels → goroutine wakes →
-		//   srv.Shutdown is a no-op on an unstarted server → main reaches
-		//   os.Exit(1) → goroutine is reaped mid-run, harmless.
-		return fmt.Errorf("listen: %w", err)
+	if err := srv.Serve(ln); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		return fmt.Errorf("serve: %w", err)
 	}
 
-	// Reached only on the clean-close path (Shutdown was called via
-	// SIGTERM); wait for the drain goroutine to finish before returning.
 	<-shutdownDone
 	slog.Info("server stopped cleanly")
 	return nil
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -26,8 +26,15 @@ const (
 	listenAddr = ":8080"
 	// shutdownTimeout sits inside Fargate's 30s SIGTERM→SIGKILL window with
 	// 5s of headroom for the container runtime to actually deliver SIGKILL
-	// and reap the process.
+	// and reap the process. This is the cap on the drain *as a whole*, not
+	// per request — http.Server.WriteTimeout (15s) still bounds individual
+	// in-flight handlers; bumping shutdownTimeout above 25s won't extend
+	// long-running handlers, only the wait for short ones to drain.
 	shutdownTimeout = 25 * time.Second
+	// maxHeaderBytes is well above Slack's realistic header size (sig +
+	// timestamp + standard headers fit comfortably in 2 KiB) but bounds
+	// the per-connection memory an attacker can force pre-handler.
+	maxHeaderBytes = 8 << 10 // 8 KiB
 )
 
 // version is set at build time via `-ldflags "-X main.version=<sha>"`.
@@ -89,7 +96,7 @@ func run() error {
 		ReadTimeout:       15 * time.Second,
 		WriteTimeout:      15 * time.Second,
 		IdleTimeout:       60 * time.Second,
-		MaxHeaderBytes:    1 << 20,
+		MaxHeaderBytes:    maxHeaderBytes,
 	}
 
 	// Bind first so a port-already-in-use failure returns before the
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -108,7 +108,7 @@ func run() error {
 	lc := &net.ListenConfig{}
 	ln, err := lc.Listen(context.Background(), "tcp", listenAddr)
 	if err != nil {
-		return fmt.Errorf("listen: %w", err)
+		return fmt.Errorf("bind %s: %w", listenAddr, err)
 	}
 
 	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -29,6 +29,11 @@ const (
 	shutdownTimeout = 25 * time.Second
 )
 
+// version is set at build time via `-ldflags "-X main.version=<sha>"`.
+// Used in the qURL client User-Agent so server-side traces can pin a
+// failure to a specific bot release.
+var version = "dev"
+
 func main() {
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
 	slog.SetDefault(logger)
@@ -42,9 +47,12 @@ func main() {
 // run holds the full server lifecycle so deferred cleanup (signal handler
 // release, etc.) executes before main returns to os.Exit.
 func run() error {
+	// Required env vars are explicit by design: a missing QURL_ENDPOINT
+	// previously fell back to the sandbox URL, which is the kind of silent
+	// misconfiguration that ships a prod deploy at sandbox.
 	qurlEndpoint := os.Getenv("QURL_ENDPOINT")
 	if qurlEndpoint == "" {
-		qurlEndpoint = "https://api.layerv.xyz"
+		return errors.New("QURL_ENDPOINT is required")
 	}
 
 	slackSigningSecret := os.Getenv("SLACK_SIGNING_SECRET")
@@ -53,13 +61,14 @@ func run() error {
 	}
 
 	authProvider := auth.EnvProvider{EnvVar: "QURL_API_KEY"}
+	userAgent := "qurl-slack/" + version
 
 	handler := internal.NewHandler(internal.Config{
 		AuthProvider:       &authProvider,
 		SlackSigningSecret: slackSigningSecret,
 		NewClient: func(apiKey string) *client.Client {
 			return client.New(qurlEndpoint, apiKey,
-				client.WithUserAgent("qurl-slack/dev"),
+				client.WithUserAgent(userAgent),
 				// Synchronous ack path — Slack's 3s budget rules out retries here.
 				client.WithRetry(0),
 			)
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -1,23 +1,47 @@
-// Package main is the Lambda entrypoint for the Slack integration.
+// Package main is the HTTP entrypoint for the Slack integration.
+//
+// Runs as a long-lived process behind an ALB on Fargate. Listens on
+// :8080, terminates gracefully on SIGTERM (Fargate's task-stop signal,
+// sent 30s before SIGKILL).
 package main
 
 import (
 	"context"
+	"errors"
+	"fmt"
 	"log/slog"
+	"net/http"
 	"os"
-
-	"github.com/aws/aws-lambda-go/events"
-	"github.com/aws/aws-lambda-go/lambda"
+	"os/signal"
+	"syscall"
+	"time"
 
 	"github.com/layervai/qurl-integrations/apps/slack/internal"
 	"github.com/layervai/qurl-integrations/shared/auth"
 	"github.com/layervai/qurl-integrations/shared/client"
 )
 
+const (
+	listenAddr = ":8080"
+	// shutdownTimeout sits inside Fargate's 30s SIGTERM→SIGKILL window with
+	// 5s of headroom for the container runtime to actually deliver SIGKILL
+	// and reap the process.
+	shutdownTimeout = 25 * time.Second
+)
+
 func main() {
 	logger := slog.New(slog.NewJSONHandler(os.Stdout, &slog.HandlerOptions{Level: slog.LevelInfo}))
 	slog.SetDefault(logger)
 
+	if err := run(); err != nil {
+		slog.Error("fatal", "error", err)
+		os.Exit(1)
+	}
+}
+
+// run holds the full server lifecycle so deferred cleanup (signal handler
+// release, etc.) executes before main returns to os.Exit.
+func run() error {
 	qurlEndpoint := os.Getenv("QURL_ENDPOINT")
 	if qurlEndpoint == "" {
 		qurlEndpoint = "https://api.layerv.xyz"
@@ -25,25 +49,53 @@ func main() {
 
 	slackSigningSecret := os.Getenv("SLACK_SIGNING_SECRET")
 	if slackSigningSecret == "" {
-		slog.Error("SLACK_SIGNING_SECRET is required")
-		os.Exit(1)
+		return errors.New("SLACK_SIGNING_SECRET is required")
 	}
 
 	authProvider := auth.EnvProvider{EnvVar: "QURL_API_KEY"}
 
 	handler := internal.NewHandler(internal.Config{
-		QURLEndpoint:       qurlEndpoint,
 		AuthProvider:       &authProvider,
 		SlackSigningSecret: slackSigningSecret,
 		NewClient: func(apiKey string) *client.Client {
 			return client.New(qurlEndpoint, apiKey,
 				client.WithUserAgent("qurl-slack/dev"),
-				client.WithRetry(0), // Lambda has its own timeout; retries risk exceeding it
+				// Synchronous ack path — Slack's 3s budget rules out retries here.
+				client.WithRetry(0),
 			)
 		},
 	})
 
-	lambda.Start(func(ctx context.Context, req events.APIGatewayProxyRequest) (events.APIGatewayProxyResponse, error) {
-		return handler.Handle(ctx, &req)
-	})
+	srv := &http.Server{
+		Addr:              listenAddr,
+		Handler:           handler,
+		ReadHeaderTimeout: 5 * time.Second,
+		ReadTimeout:       15 * time.Second,
+		WriteTimeout:      15 * time.Second,
+		IdleTimeout:       60 * time.Second,
+	}
+
+	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
+	defer stop()
+
+	shutdownDone := make(chan struct{})
+	go func() {
+		<-ctx.Done()
+		slog.Info("received shutdown signal — draining HTTP server")
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
+		defer cancel()
+		if err := srv.Shutdown(shutdownCtx); err != nil {
+			slog.Error("graceful shutdown failed", "error", err)
+		}
+		close(shutdownDone)
+	}()
+
+	slog.Info("starting Slack bot HTTP server", "addr", listenAddr)
+	if err := srv.ListenAndServe(); err != nil && !errors.Is(err, http.ErrServerClosed) {
+		return fmt.Errorf("listen: %w", err)
+	}
+
+	<-shutdownDone
+	slog.Info("server stopped cleanly")
+	return nil
 }
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -89,17 +89,20 @@ func run() error {
 		IdleTimeout:       60 * time.Second,
 	}
 
-	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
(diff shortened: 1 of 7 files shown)


justin-layerv:
Round-2 of /simplify done. All claude-review feedback (both rounds, including nits) addressed in `5b6f12b`.

**Fixed in this PR**:
- /health silent in logs + GET/HEAD-only with 405 + `Allow: GET, HEAD` on other methods (CR Round 1 #1, #2; Round 2 #3)
- GET on `/slack/*` → 405 + `Allow: POST` (CR Round 1 #3 — RFC-correct, defensible-either-way → fixed)
- `Content-Length` short-circuit → 413 before allocation when a client honestly declares oversize (CR Round 1 #6)
- `respondJSON` marshal-failure fallback now writes JSON via the normal write path; Content-Type matches body (CR Round 1 #4, Round 2 #2)
- One-line comment on the bind-failure vs clean-close branches in `main.go` (CR Round 2 #1)
- `respondMethodNotAllowed(w, allow)` helper so the 405 + Allow contract is structural, not editorial
- `internalErrorJSON` hoisted to a package var (cosmetic, but cleaner)
- New tests: `TestHealthEndpoint_RejectsNonGet`, `TestHealthEndpoint_AcceptsHead`, `TestHandle_BodyAtCapAccepted` (1 MiB off-by-one fence — CR Round 2 #8), `TestHandle_DeclaredOversizeReturns413`, `TestSlashCommand_EmptyBodyShowsHelp` (CR Round 1 test suggestion), `TestHandle_GetOnSlackPathReturns405`
- Renamed `TestHandle_OversizeBodyReturns400` → `TestHandle_OversizeBodyReturns413` (path now hits the CL short-circuit)
- Trimmed two WHAT-narrating test comments (round-2 simplify finding)

**Deferred with labeled issues**:
- #152 docs: scrub Lambda-shape language from CLAUDE.md + slack.yml (CR Round 1 #9 — PR-C bundles this with the Docker/ECR migration; same surface, single pass)
- #153 test: SIGTERM/Shutdown integration test (CR Round 2 #9 — needs subprocess scaffolding; PR-B2 will reuse it for drain assertions)

**Acknowledged, no action**:
- CR Round 1 #5 (goroutine on bind failure): addressed via the new shutdown-ordering comment in `main.go` rather than restructuring — the existing flow is correct, just under-documented
- CR Round 1 #7 (body Close): `net/http` handles it; revisit if `bodyclose` lands in `.golangci.yml`
- CR Round 1 #8 (WriteTimeout vs Slack 3s budget): explicit B2 territory
- CR Round 2 #4 (`userAgent` captured once): correct given `version` is `-ldflags`-only; no fix needed
- CR Round 2 #5 (capacity-planning note for 1 MiB cap): runbook concern, not code
- CR Round 2 #11/#12 (HMAC split + `[]byte` plumbing as wins): no action needed, just acknowledged

Verified locally:
\`\`\`
$ curl -i localhost:8080/health         # GET → 200, no info log
$ curl -I localhost:8080/health          # HEAD → 200, no info log
$ curl -X POST -i localhost:8080/health  # 405 + Allow: GET, HEAD
$ curl -i localhost:8080/slack/commands  # 405 + Allow: POST (request logged)
\`\`\`

`make check` clean. CI on `5b6f12b` is the next signal.

justin-layerv:
Round-3 of /simplify done. Addressed all CR-3 feedback in `1437b04`.

**Fixed in this PR:**
- **400/413 split unified** (CR-3 #1): `errors.As(*http.MaxBytesError)` on the read-failure path. Both honest pre-check and dishonest-during-read now return 413 with the same envelope. Operator dashboards bucket them together.
- **`respondPayloadTooLarge` helper** for the two 413 callsites (mirrors `respondMethodNotAllowed`).
- **HEAD `/health` collapsed into GET case** (CR-3 simplify follow-up): the bespoke header-set was a tiny per-probe saving for a rot risk if `respondJSON` ever adds future headers (e.g. `X-Content-Type-Options`). net/http strips the body at the wire layer regardless.
- **slog.Info on both 413 paths** with a `reason` attribute (`content_length_pre_check` / `max_bytes_during_read`) — fixes the gap where read-oversize had no log line.
- **Comment phrasing**: "above" replaced with explicit "Content-Length pre-check" reference (won't rot if blocks reorder).
- **`run()` comment** names the specific deferred call and the specific exit path.
- **New tests**:
  - `TestHandle_DishonestContentLengthReturns413` (CR-3 test gap #1)
  - `TestHandle_LargeSignedBodyAccepted` — 100 KiB signed body → 200; locks the read-not-truncated contract (CR-3 test gap #2)
  - `TestHandle_MalformedEventJSON_Returns200` — asserts both status AND `{"ok":"true"}` body (CR-3 test gap #3 + simplify-3 #6)

**Skipped / acknowledged**:
- CR-3 #2 (bind-failure lifecycle): "optional polish"; current shape is correct, just under-documented (already addressed in CR-2 follow-up via the run() comment).
- CR-3 style #1 (G706 nolint helper): explicitly stylistic; per-call directives match repo precedent and are self-contained.
- CR-3 style #2 (`internalErrorJSON` belt-and-suspenders): noted as fine.
- Runbook concerns (MaxHeaderBytes, ALB-TLS-only): runbook territory, not code.
- Simplify-3 #5 (100 KiB → const-relative): would make test redundant with `TestHandle_BodyAtCapAccepted`.

Verified locally:
\`\`\`
$ curl -I localhost:8080/health
HTTP/1.1 200 OK
Content-Length: 15  # body stripped at wire by net/http

$ curl -i -X POST -H "Content-Length: 2097153" --data-raw "..." localhost:8080/slack/commands
HTTP/1.1 413 Request Entity Too Large
{"error":"body too large"}
# log: oversize body rejected reason=content_length_pre_check declared=2097153
\`\`\`

\`make check\` clean. CI on `1437b04` is the next signal.

justin-layerv:
CR-4 feedback addressed in `dcbbd12`:

**Fixed:**
- **#1 Lowercase-headers test was a no-op duplicate.** Verified: `r.Header.Set` canonicalizes via `textproto.CanonicalMIMEHeaderKey`, so the test exercised the same path as `TestSlashCommandHelp`. The "true" lowercase-on-wire shape is structurally handled by net/http at parse time and isn't reachable through httptest. Dropped the test, replaced with a one-line note explaining why we don't fence it separately.
- **#2 Bind-failure goroutine ordering** — comment expanded with the explicit sketch: `stop() → signal ctx cancels → goroutine wakes → srv.Shutdown is no-op on unstarted server → os.Exit(1) reaps`.
- **#5 Unknown-subcommand telemetry** — `slog.Info("unknown slash subcommand", ...)` so workspaces on a stale slash-command spec show up in dashboards.

**Skipped (acknowledged):**
- **#3 HEAD `/health` short-circuit**: reviewer flagged as cosmetic ("bring up only if profiling shows pressure"). Conflicts with simplify-3 quality finding (rot risk if `respondJSON` adds future headers); keeping the collapsed-into-GET shape.
- **#4 Two log lines on oversize POST**: intentional. The `received request` line is the universal entry trace; `oversize body rejected` is the outcome at a different abstraction. One-event-per-request is preserved, just at the entry-trace + outcome levels.

Diff is 3 surgical changes (11 +, 24 −). `make check` clean.

justin-layerv:
CR-5 LGTM noted, residuals addressed in `37d11ee`:

**Fixed:**
- **#3 `internalErrorJSON` → const string** (`internalErrorEnvelope`). Removes the package-mutable `[]byte` surface; conversion to `[]byte` happens at write time on a path that fires zero times in production.
- **#1 Test helper-purpose comment** in `TestHandle_BodyAtCapAccepted` explaining `noopQURLServer` is required for `Config.NewClient` construction even though `/slack/events` doesn't call out to qURL.

**Skipped with rationale:**
- **#2 (slog.Info on unknown subcommand without `//nolint:gosec`)**: tried this in CR-4 — golangci-lint flagged the directive as unused because gosec's taint analyzer doesn't trace `text` through `url.ParseQuery`. The asymmetry across sites is structural; we annotate where gosec actually fires (3 path-logging sites do; this one doesn't).
- **#4 (two log lines on 413)**: dashboards bucket on `msg`/`reason`, not event-count-per-request. The entry trace + outcome split is intentional and at different abstraction levels.

`make check` clean. CI on `37d11ee` should reach green and this PR is ready to merge.

justin-layerv:
CR-6 feedback addressed in `7edd16c`.

**Fixed:**
- **#1 Bind-failure misleading log**: switched from `srv.ListenAndServe()` to the canonical `lc.Listen(ctx, ...)` + `srv.Serve(ln)` split. Bind happens before the drain goroutine spawns, so a port-already-in-use failure returns directly without firing the "received shutdown signal" log line. Verified: bind-conflict start emits only `{"level":"ERROR","msg":"fatal","error":"listen: listen tcp :8080: bind: address already in use"}` and exits — no shutdown log noise.
- **JSON-handler-is-load-bearing comment** in main.go tying the slog handler choice to the G706 nolints in handler.go.

**Skipped with rationale:**
- **`r.Context()` TODO for PR-B2**: forward-looking comments referencing future work rot per repo convention; whoever writes PR-B2 will be reasoning about goroutine ctx scope explicitly. Plus PR-B2 will rewrite `handleCreate` entirely.
- **`userAgent` local inlining**: the local earns its name in the closure body; readability wash.
- **`"Synchronous ack path"` comment will stale on PR-B2**: it's accurate today and PR-B2 rewrites this code, so the comment naturally updates with it.
- **`internalErrorEnvelope` `[]byte` alloc per call**: reviewer agrees current trade-off (immutability over a free alloc on a path that fires zero times) is defensible.
- **`MaxHeaderBytes` slowloris hardening**: runbook concern; ALB is the only ingress in the deployment shape, so direct-dial isn't reachable.

\`make check\` clean. Should be ready to merge once CI is green on `7edd16c`.

justin-layerv:
CR-7 feedback addressed in `d461f15`.

**Fixed:**
- **#2 Serve-error path orphaning the shutdown goroutine**: cleanup is now deterministic on every Serve return — explicit `stop()` + `<-shutdownDone` before checking the error. Goroutine always finishes its work; log ordering is stable on every exit path.
- **Listen-ctx race nit**: bind now uses `context.Background()` instead of the signal-cancellable ctx, so a SIGTERM arriving in the gap between `NotifyContext` and `Listen` no longer surfaces as `"listen: context canceled"`. The signal ctx still drives the goroutine.
- **`internalErrorEnvelope` comment trimmed**: rationale was overstated (`w.Write` doesn't retain the slice). Now just says "fallback 500 body when JSON marshal fails (unreachable for current callers)".

**Filed (deferred to PR-C):**
- **#154 ALB drain headroom (CR-7 #1)** — lameduck flip `/health` to 503 before `Shutdown` so ALB deregisters before the listener closes. Reviewer's framing: "wire it up either here or as a tracked follow-up alongside PR-C — it's adjacent to the ECS/ALB wiring anyway." The ALB doesn't exist yet (Phase 3 builds it), and #153's integration test scaffolding will be the natural home for verifying the lameduck flow.

**Skipped with rationale:**
- **`maxRequestBodyBytes` 1 MiB → 64 KiB tightening**: reviewer says "once PR-B2's idempotency surface is confirmed" — explicitly B2 territory.
- **ALB security-group scope** (`:8080` only-from-ALB): runbook concern for PR-C's task definition / SG config.
- **`slack-lambda` artifact name**: covered by #152 (CLAUDE.md + slack.yml doc drift) which PR-C bundles with the Docker/ECR migration.
- **SIGTERM-during-in-flight integration test**: covered by #153.

`make check` clean. Verified locally:
\`\`\`
$ ./qurl-slack-bot &
{"msg":"starting Slack bot HTTP server","addr":":8080"}
$ curl localhost:8080/health
{"status":"ok"}
$ kill -TERM
{"msg":"received shutdown signal — draining HTTP server"}
{"msg":"server stopped cleanly"}
exit=0
\`\`\`

This should be the final iteration. CI on `d461f15` is the next signal.

justin-layerv:
CR-8 feedback addressed in `faaeffc`. PR description also updated with the QURL_ENDPOINT breaking-change note.

**Fixed:**
- **#1 PR description note about QURL_ENDPOINT-now-required**: added a ⚠️ Breaking startup change section so PR-C's task definition author doesn't get surprised.
- **#2 Drop dead `srv.Addr`**: removed from the http.Server literal with a comment explaining `Serve(ln)` doesn't consult it.
- **#3 Move "received request" log inside matched-path branch**: 404s are now silent, internet probes (/wp-login.php, /.env) don't generate per-request log entries. Slack and health paths still log normally.
- **#4 Explicit `MaxHeaderBytes: 1 << 20`**: documents the value next to the other server timeouts (matches the body cap; default same but now visible).
- **#5 Comment on body-cap auth-fail tradeoff**: maxRequestBodyBytes godoc now explicitly notes HMAC needs the raw body so we can't auth-then-read; ALB rate-limiting/WAF is the real defense; this cap bounds per-request damage.
- **#7 TestSlackEndpoints_Reject401 hit-count assertion**: new `countingQURLServer(t)` helper exposing an `atomic.Int32`. The 401-suite now also asserts `hits == 0` so a regression that 401s at the wire but leaks the upstream call gets caught.

**Filed (deferred to PR-C):**
- **#154** ALB drain-headroom (lameduck flip /health → 503 before Shutdown). CR-7 #1; reviewer agreed PR-C is the natural home alongside the actual ALB wiring.

**Skipped with rationale:**
- **#6 sync.WaitGroup vs channel**: cosmetic; reviewer explicitly OK to defer to PR-B2 ("composes more naturally" once the ack-then-async goroutines land).
- **`internalErrorEnvelope` hypothetical nolint reference**: speculative future-linter scenario.
- **`version` ldflag TODO comment**: PR-C handles the build pipeline (Dockerfile + ECR push); the implicit dependency is already in PR-C scope.
- **`maxRequestBodyBytes` 64 KiB tightening**: CR-7 / CR-8 — explicit B2 territory ("once PR-B2's idempotency surface is confirmed").

`make check` clean. Verified bind-conflict still emits no shutdown-log noise; SIGTERM still drains cleanly to exit 0. CI on `faaeffc` is the next signal — should be the last iteration.

justin-layerv:
CR-9 LGTM noted, all 7 items processed in `8494a96`:

**Fixed:**
- **#2 Routing strict-match note**: comment on the path switch explains the trailing-slash-is-deliberately-404 design and points at TrimRight/ServeMux as future migration paths.
- **#4 MaxHeaderBytes tightened to 8 KiB** from the 1 MiB default. Slack's real header footprint is comfortably under 2 KiB; the smaller cap removes attack surface.
- **#6 Dishonest-CL test comment** now explicit that it simulates the chunked-transfer / no-CL path (the real-world dishonest case) rather than a literally lying fixed-CL request.
- **#7 shutdownTimeout godoc** extended to note that 25s bounds the drain as a whole, not per-request; WriteTimeout still applies to individual handlers, so bumping it above 25s won't extend long-running handlers.

**Filed (deferred):**
- **#155 — per-workspace `*client.Client` caching** (CR-9 #5). Verified `shared/client.New` allocates fresh `*http.Client` per call; on Fargate this leaks connection-pool reuse vs Lambda. Deferred to PR-B4 because the fix interacts with the per-workspace API key surface DynamoProvider introduces.

**Skipped with rationale:**
- **#1 Two slog lines per slash command** ("received request" + "slash command"): pre-existing, reviewer's own framing was "future consolidation".
- **#3 `internalErrorEnvelope` unreachable**: reviewer flagged for awareness only ("the cost is one constant + four lines, so I wouldn't insist").

`make check` clean. CI on `8494a96`. This is the convergence iteration — every item from 9 CR rounds is either fixed or filed.

justin-layerv:
CR-10 feedback addressed in `3923cf4`:

**Fixed:**
- **#1** Silence comment now notes 405s on `/slack/*` are also silent (the method gate returns before the request log fires), with the credentialed-scrapers example the reviewer flagged.
- **#2** Cleaner bind-error wording: `bind :8080: listen tcp :8080: bind: address already in use` (was redundant "listen: listen tcp ...").
- **#3** `TestSlashCommandCreate` switched to `t.Cleanup(qurlSrv.Close)`, matches the file's new convention.
- **#5** `handleEvent` logs at `slog.Debug` on JSON unmarshal failure — surfaces spec drift / corrupt payloads to operators while preserving the 200-with-{ok:true} contract Slack retry semantics need.

**Filed:**
- **#156** — sanitize qURL API errors before exposing to Slack ephemeral responses (CR-10 raw-error-leakage flag). Pre-existing behavior; PR-B2's ack-then-async will fold in the sanitized error helper as part of moving the API call into the goroutine.

**Skipped with rationale:**
- **#4** MaxHeaderBytes 8 KiB → 16 KiB / env-tunable: explicitly optional ("if you see 431 post-rollout, that's the cause"). Ship with 8 KiB; tighten/loosen post-deploy based on real signal.
- **WAF runbook concern**: not code; runbook territory.
- **`io.ReadAll` grow-on-double**: future micro-opt explicitly tagged "Negligible at current scale".
- **`MaxHeaderBytes` 431 test**: needs `httptest.Server` scaffolding; #153 will house this kind of integration test once the SIGTERM-drain scaffolding is built.

CI on `3923cf4`. Status of CR convergence after 10 rounds:
- 9 fixes landed in this PR (across CR-1 through CR-10)
- 6 issues filed (#152, #153, #154, #155, #156 + the closed #145)
- 0 unaddressed items remaining

Issue #157, opened 2026-04-30
slack: handle ssl_check periodically posted by Slack


justin-layerv:
## Context

Slack periodically POSTs `ssl_check=1` form payloads to slash-command endpoints to verify HTTPS reachability. The bot currently routes these to the help branch (because `text` is empty in an `ssl_check` payload), which 200s with help text. Slack only requires a 200 — accepting any 200 — so this is functionally fine, but it's a small papercut: the help slog line fires for every `ssl_check`, polluting telemetry, and the help text is wasted CPU/bytes.

## What to change

In `apps/slack/internal/handler.go` `handleSlashCommand`, add a branch before the help/create/list switch:

```go
if values.Get("ssl_check") == "1" {
    respondJSON(w, http.StatusOK, map[string]string{"ok": "true"})
    return
}
```

## Acceptance

- New unit test in `handler_test.go` that posts a signed `ssl_check=1` body to `/slack/commands` and asserts (a) 200, (b) no `slash command` slog noise (or expand the existing log to filter), (c) body is `{"ok":"true"}`.
- No telemetry regression: existing tests still pass.

## Why now (or not)

Surfaced as a polish item during PR-B2 advisor review. Not blocking — Slack accepts the current behaviour. Filing as a follow-up so it doesn't get lost.

Issue #156, opened 2026-04-30
security(slack): sanitize qURL API errors before exposing to Slack ephemeral responses


justin-layerv:
\`apps/slack/internal/handler.go\` currently returns raw API error strings to Slack users via:

\`\`\`go
respondSlack(w, "Failed to create qURL: " + err.Error())
respondSlack(w, "Failed to list qURLs: " + err.Error())
\`\`\`

Surfaced by claude-review on PR #150 commit \`8494a96\` (CR-10):

> Error messages exposed to user (handler.go:246, handler.go:264): \`respondSlack(w, "Failed to create qURL: "+err.Error())\` leaks raw API error strings into Slack ephemeral responses. Pre-existing behavior, not introduced here, but worth noting as a follow-up — sanitize to a generic message + log the full error structurally.

## Why this matters

Raw upstream errors can include:
- Internal endpoint paths or service names (\"qurl-service: backend timeout connecting to dynamodb-..."\")
- Identifiers (\"qurl_id=q_abcd123 not found in tenant=t_xyz\") that leak across workspace boundaries
- Stack-trace-shaped strings under upstream regression
- Idempotency-key hashes from \`shared/client\` that look like opaque IDs but are workspace-scoped fingerprints

Slack ephemeral messages are visible to the slash-command author. In a multi-workspace deployment (post-Phase 4), a workspace admin could see internals of unrelated tenants if upstream errors include cross-tenant context.

## Proposed shape

\`\`\`go
const slackUserFacingError = "Couldn't create that qURL right now. Try again in a moment; if it keeps failing, contact your workspace admin."

result, err := c.Create(ctx, ...)
if err != nil {
    var apiErr *client.APIError
    requestID := ""
    if errors.As(err, &apiErr) {
        requestID = apiErr.RequestID
    }
    slog.Error("failed to create qURL", "error", err, "request_id", requestID, "target_url", targetURL)
    if requestID != "" {
        respondSlack(w, slackUserFacingError + " (ref: " + requestID + ")")
    } else {
        respondSlack(w, slackUserFacingError)
    }
    return
}
\`\`\`

The full error stays in slog.Error for debugging; the user-facing message is a generic-with-request-ID-hint pattern that's also defensive against locale / phrasing issues.

## Scope

Two callsites in \`handleCreate\` and \`handleList\`. Future \`handleRevoke\`/\`handleGet\` (Phase 5 #11 of the plan) will inherit the sanitized helper.

## Acceptance

[...]

Issue #155, opened 2026-04-30
perf: cache *client.Client per workspace on Fargate (avoid per-request transport)


justin-layerv:
PR #150 ports the Slack handler to a long-lived Fargate process. The qURL `*client.Client` factory in \`apps/slack/cmd/main.go\` allocates a fresh \`*http.Client\` (and therefore fresh \`http.Transport\` + connection pool) on every authenticated request via \`authenticatedClient(ctx, teamID)\`.

Surfaced by claude-review on commit \`faaeffc\` (CR-9 #5):

> Per-request \`*client.Client\` allocation. \`h.cfg.NewClient(apiKey)\` is called for every authenticated request via \`authenticatedClient\`. Each builds a fresh \`http.Client\`/transport. Pre-existing, not a blocker for this PR — but the Fargate move (long-lived process) is exactly the runtime where the per-request \`http.Client\` cost starts mattering vs the Lambda model.

Verified: \`shared/client.New\` at \`shared/client/client.go\` constructs \`httpClient: &http.Client{Timeout: defaultTimeout}\` inline on every call.

## Why deferred from PR #150

- B1 is the I/O shape conversion only.
- The fix interacts with PR-B4 (DynamoProvider per workspace), which will introduce a per-workspace API key cache. The natural shape is a per-workspace client cache layered on top — building it now would either get rebuilt for B4 or add an abstraction that doesn't fit B4's needs.

## Proposed shape

After PR-B4 lands, add a \`sync.Map[teamID]*client.Client\` (or LRU) to the handler. Cache hit → reuse. Cache miss → mint via \`Config.NewClient\` (which gives us per-workspace API key injection) and store. Cache eviction policy: TTL on workspace key rotation, or LRU bound at \`~50 workspaces × ~few KB each\` so we don't have to think about it.

## Acceptance

- Handler reuses one \`*client.Client\` per workspace across requests
- Connection pool to qURL API is shared per workspace (one set of idle conns instead of N)
- Workspace API key rotation invalidates the cached client (dependent on PR-B4's auth surface)
- A regression test confirms the same pointer is returned on a second \`authenticatedClient\` call for the same team_id

## Related

- #150 introduced the long-lived process where this matters
- PR-B4 introduces the workspace-keyed auth surface this caches against

Issue #154, opened 2026-04-30
deploy: ALB drain headroom (lameduck /health → 503) before srv.Shutdown on SIGTERM


justin-layerv:
PR #150 wires up SIGTERM → \`srv.Shutdown\` for the Slack Fargate task. Surfaced by claude-review on commit \`7edd16c\` (CR-7 #1):

> On SIGTERM the goroutine logs "draining" and calls \`srv.Shutdown\` immediately. \`Shutdown\` stops accepting new connections at the listener, but ALB target deregistration is asynchronous — the ALB will keep routing new connections to this task for some seconds after the ECS task-stopping event, and those will hit a closed listener and surface as 502s upstream.

## Fix

Standard mitigation is a lameduck period: flip \`/health\` to return 503 (or sleep ~5–10s) **before** calling \`srv.Shutdown\`, so ALB removes the target from rotation first. Sketch:

\`\`\`go
go func() {
    <-ctx.Done()
    slog.Info("received shutdown signal — entering lameduck")
    h.SetHealthy(false) // /health returns 503; ALB deregisters within ~30s
    time.Sleep(lameduckDuration) // ~10s, must fit inside shutdownTimeout
    slog.Info("draining HTTP server")
    srv.Shutdown(...)
}()
\`\`\`

Lameduck duration must fit inside \`shutdownTimeout\` (currently 25s) with room for the actual drain — so something like \`lameduckDuration = 10s\`, leaving ~15s for in-flight requests to complete.

## Why deferred from PR #150

- B1 is the I/O shape conversion only; the ALB doesn't exist yet (Phase 3 builds it).
- Reviewer's framing: "wire it up either here or as a tracked follow-up alongside PR-C — it's adjacent to the ECS/ALB wiring anyway."
- No current deployment has this gap (B1 is the first long-lived-server shape; nothing's deployed yet).

## Acceptance

- \`/health\` returns 503 for \`lameduckDuration\` before \`srv.Shutdown\` fires
- The lameduck → drain transition is logged distinctly so deploy timelines can be reconstructed from CloudWatch
- \`shutdownTimeout\` bounds the entire sequence (lameduck + drain) within Fargate's 30s SIGTERM→SIGKILL window
- Integration test (or part of #153) asserts in-flight requests complete during a SIGTERM with the lameduck flow

## Related

- #150 introduced the SIGTERM/Shutdown plumbing
- #153 will build the SIGTERM integration test scaffolding; this should fold into that test
[...]

PR #158, opened 2026-04-30
feat(slack): ack-then-async slash commands with idempotency

justin-layerv:
## Summary

- Convert `/qurl create` and `/qurl list` from synchronous to ack-then-async: handler returns `:hourglass: Working on it…` within ~50ms, then a goroutine calls the qURL API and POSTs the result to Slack's `response_url`. Slack's 3-second ack budget is below typical qURL latency under enterprise load, so synchronous-and-fast is no longer a sustainable shape.
- Add an `Idempotency-Key` header (`sha256("slack:"+team_id+":"+trigger_id)`) to every `/qurl create` so Slack-side retries (3s ack timeout exceeded) fold into a single resource at the qURL service. Closes the dedup gap PR-B2 was filed for.
- Bump `client.WithRetry(0) → WithRetry(2)`. The synchronous path couldn't afford retries (3s ack budget). The async worker has 25s, so transient 429/5xx are now safe to retry on the existing exponential-backoff schedule before bubbling to the user.
- Defense-in-depth on response_url: pin scheme+host to `https://hooks.slack.com` (using `Hostname()` so explicit `:443` still passes), and the response_url client refuses redirects so a 30x can't pivot the bot into a generic SSRF emitter even if the signature gate were bypassed.
- Sanitize qURL API errors before they reach Slack — `Title` + `RequestID` surface (operator-correlation), `Detail` (server-side internals: DB error strings, internal hostnames, stack-trace fragments) stays in slog only. Folds in #156.
- Wire `Handler.BaseContext` to the SIGTERM-cancelled `signal.NotifyContext`, and call `handler.Wait()` after `srv.Shutdown` so the process drains async workers before exit.

## Why expanded behaviour beyond "convert to async"

The `WithRetry(0) → WithRetry(2)` change is a coupled-but-not-obvious knock-on of the async unlock. Calling it out here so reviewers don't ask later — synchronous-with-retries was incompatible with Slack's 3s budget; async-with-retries is the right shape now.

## Architecture

| Concern | Choice | Why |
|---|---|---|
| Goroutine context | Derived from `Handler.BaseContext`, NOT `r.Context()` | `r.Context()` cancels as soon as `ServeHTTP` returns, which is immediately after the ack — wrong cancellation signal. |
| Drain on shutdown | `sync.WaitGroup` on `Handler`; `Wait()` called in main after `srv.Shutdown` | Guarantees the process doesn't exit while a goroutine is mid-POST to `response_url`. |
| Bounded concurrency | 50-slot buffered channel, `select`-with-`default` drop | A Slack-side flood gets fast `:warning: busy` feedback rather than queueing or unbounded-spawning until OOM. |
| Per-worker timeout | 25s `context.WithTimeout(baseCtx, …)` | Generous over typical qURL latency (~1s) but well under Slack's `response_url` 30-min TTL. |
| Panic recovery | `defer recover()` inside the goroutine | A panicking goroutine in a long-lived process is disqualifying — recover-and-log so the deferred sem release and `wg.Done` always run. |
| Idempotency key | `sha256("slack:"+team_id+":"+trigger_id)` | 64 hex chars (well under the 256-byte cap), no PII on transit, matches qURL service's per-owner key contract. |

## Test plan

- [x] `make lint` clean
- [x] `go test -race -count=1 ./apps/slack/...` clean
- [x] `go test -race -count=30 -run TestHandle_AckIsFastUnderSlowAPI` — confirm the 50ms ack budget is not flaky under race-detector load
- [x] End-to-end smoke: bind, `/health` 200, `/missing` 404, SIGTERM exit 0
- [x] 100-concurrent `/qurl create` dedup test — every upstream request carries the same `Idempotency-Key` header
- [x] Pool-saturation test — 3rd request past a capacity-2 pool gets `ackBusy`
- [x] Panic-recovery test — panicking `auth.Provider` doesn't leak `wg`/`sem` slots
- [x] `validateResponseURL` table test — accept `https://hooks.slack.com`, accept `:443`, reject `http://`, reject other hosts, reject the `hooks.slack.com.attacker.example` subdomain trick
- [x] Redirect-refusal test — server responds 302; client surfaces 302 to caller, doesn't follow
- [x] Sanitizer leak test — qURL `Detail` of `db: leaked-internal-host failed` does not appear in user-facing payload; `Title` + `RequestID` do appear
- [x] Base-context cancellation test — `baseCancel()` lets `h.Wait` return within 2s

## Filed follow-ups

- #157 — `slack: handle ssl_check periodically posted by Slack` (advisor polish item; current behaviour 200s with help text, Slack accepts any 200 so non-blocking)

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -35,4 +35,5 @@ CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -o bootstrap ./apps/slack/cmd/
 |----------|----------|-------------|
 | `SLACK_SIGNING_SECRET` | Yes | Slack app signing secret |
 | `QURL_API_KEY` | Yes | qURL API key for this workspace |
-| `QURL_ENDPOINT` | No | qURL API base URL (default: `https://api.layerv.xyz`) |
+| `QURL_ENDPOINT` | Yes | qURL API base URL (e.g. `https://api.layerv.xyz`) |
+| `QURL_SLACK_MAX_CONCURRENT_ASYNC` | No | Pool cap for in-flight async slash-command workers. Empty/0 uses the built-in default (50). Tune up if a workspace's load shape sustains `:warning: Slack bot is busy` acks; tune down if memory pressure during retry storms is observed. |
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -75,14 +75,28 @@ func run() error {
 	authProvider := auth.EnvProvider{EnvVar: "QURL_API_KEY"}
 	userAgent := "qurl-slack/" + version
 
+	// signalCtx feeds two seams: the main goroutine (to detect
+	// SIGTERM and trigger srv.Shutdown) and Handler.BaseContext (so
+	// async slash-command workers observe cancellation through the
+	// same signal). Threading the same ctx into both keeps the
+	// shutdown story coherent — a worker mid-POST to response_url
+	// receives ctx.Canceled at the same instant Shutdown starts
+	// refusing new connections.
+	signalCtx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
+	defer stop()
+
 	handler := internal.NewHandler(internal.Config{
 		AuthProvider:       &authProvider,
 		SlackSigningSecret: slackSigningSecret,
+		BaseContext:        signalCtx,
 		NewClient: func(apiKey string) *client.Client {
 			return client.New(qurlEndpoint, apiKey,
+				// The async worker has up to 25s before its context fires;
+				// the qURL client may retry transient 429/5xx within that
+				// budget. WithRetry(2) gives two retries with the existing
+				// exponential-backoff schedule before bubbling to the user.
 				client.WithUserAgent(userAgent),
-				// Synchronous ack path — Slack's 3s budget rules out retries here.
-				client.WithRetry(0),
+				client.WithRetry(2),
 			)
 		},
 	})
@@ -111,18 +125,21 @@ func run() error {
 		return fmt.Errorf("bind %s: %w", listenAddr, err)
 	}
 
-	ctx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
-	defer stop()
-
 	shutdownDone := make(chan struct{})
 	go func() {
-		<-ctx.Done()
+		<-signalCtx.Done()
 		slog.Info("received shutdown signal — draining HTTP server")
 		shutdownCtx, cancel := context.WithTimeout(context.Background(), shutdownTimeout)
 		defer cancel()
 		if err := srv.Shutdown(shutdownCtx); err != nil {
 			slog.Error("graceful shutdown failed", "error", err)
 		}
+		// Shutdown returns once HTTP handlers have responded. Slash-
+		// command handlers ack and return nearly instantly, so
+		// in-flight async workers (the goroutines runAsync spawned)
+		// outlive Shutdown. Wait drains them — their contexts are
+		// already canceled via signalCtx, so they return promptly.
+		handler.Wait()
 		close(shutdownDone)
 	}()
 
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"strconv"
 	"syscall"
 	"time"
 
@@ -30,6 +31,14 @@ const (
 	// per request — http.Server.WriteTimeout (15s) still bounds individual
 	// in-flight handlers; bumping shutdownTimeout above 25s won't extend
 	// long-running handlers, only the wait for short ones to drain.
+	//
+	// Budget contract: shutdownTimeout must comfortably exceed
+	// (srv.Shutdown drain time) + (internal.responseURLTimeout = 5s)
+	// so a healthy in-flight async worker has time to land its
+	// follow-up POST before WaitTimeout(remaining) gives up. With
+	// ack-and-spawn handlers, srv.Shutdown finishes in milliseconds —
+	// the worst-case pre-Wait drain is dominated by per-handler
+	// completion (instant), leaving ~25s for response_url POSTs.
 	shutdownTimeout = 25 * time.Second
 	// maxHeaderBytes is well above Slack's realistic header size (sig +
 	// timestamp + standard headers fit comfortably in 2 KiB) but bounds
@@ -75,6 +84,12 @@ func run() error {
 	authProvider := auth.EnvProvider{EnvVar: "QURL_API_KEY"}
 	userAgent := "qurl-slack/" + version
 
+	// Optional pool-cap override. Empty/invalid env yields 0, which
+	// internal.NewHandler interprets as "use default" (50). Lets
+	// operators tune saturation without a recompile when a workspace
+	// load shape diverges from the documented enterprise default.
+	maxConcurrentAsync, _ := strconv.Atoi(os.Getenv("QURL_SLACK_MAX_CONCURRENT_ASYNC"))
+
 	// signalCtx feeds two seams: the main goroutine (to detect
 	// SIGTERM and trigger srv.Shutdown) and Handler.BaseContext (so
 	// async slash-command workers observe cancellation through the
@@ -89,6 +104,7 @@ func run() error {
 		AuthProvider:       &authProvider,
 		SlackSigningSecret: slackSigningSecret,
 		BaseContext:        signalCtx,
+		MaxConcurrentAsync: maxConcurrentAsync,
 		NewClient: func(apiKey string) *client.Client {
 			return client.New(qurlEndpoint, apiKey,
 				// The async worker has up to 25s before its context fires;
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -32,13 +32,19 @@ const (
 	// in-flight handlers; bumping shutdownTimeout above 25s won't extend
 	// long-running handlers, only the wait for short ones to drain.
 	//
-	// Budget contract: shutdownTimeout must comfortably exceed
-	// (srv.Shutdown drain time) + (internal.responseURLTimeout = 5s)
-	// so a healthy in-flight async worker has time to land its
-	// follow-up POST before WaitTimeout(remaining) gives up. With
-	// ack-and-spawn handlers, srv.Shutdown finishes in milliseconds —
-	// the worst-case pre-Wait drain is dominated by per-handler
-	// completion (instant), leaving ~25s for response_url POSTs.
+	// Budget contract: shutdownTimeout is sized against the EXPECTED
+	// drain, not the theoretical max. The expected drain is
+	// dominated by:
+	//   srv.Shutdown (≈0ms — handlers ack and return instantly)
+	//   + qURL call returns ctx.Canceled almost instantly on signal
+	//   + responseURLTimeout (5s) for the failure follow-up POST
+	//
+	// The theoretical max is asyncWorkTimeout (25s) + 5s = 30s,
+	// which would wedge if a worker ignored ctx — that's why
+	// handler.WaitTimeout caps the wait at the remaining budget
+	// rather than calling Wait. The 25s default leaves comfortable
+	// headroom for the expected case while staying well inside
+	// Fargate's 30s SIGTERM→SIGKILL window.
 	shutdownTimeout = 25 * time.Second
 	// maxHeaderBytes is well above Slack's realistic header size (sig +
 	// timestamp + standard headers fit comfortably in 2 KiB) but bounds
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -98,10 +98,18 @@ func run() error {
 	maxConcurrentAsync := 0
 	if raw := os.Getenv("QURL_SLACK_MAX_CONCURRENT_ASYNC"); raw != "" {
 		parsed, err := strconv.Atoi(raw)
-		if err != nil {
+		switch {
+		case err != nil:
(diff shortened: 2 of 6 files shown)


justin-layerv:
## /simplify round 1 — fixed

Three review agents (reuse, quality, efficiency) ran in parallel against the diff. Aggregated:

**Fixed:**

1. **Test hang potential** in `TestHandle_ConcurrentCreateSharesIdempotencyKey` — pre-sign the body once on the test goroutine; spawned goroutines build their own `httptest.NewRequest` from shared bytes. Eliminates 99 redundant SHA256s and the theoretical "t.Helper inside non-test goroutine + signing failure → wg.Wait hangs" path.
2. **Misleading comment** on the release-channel cleanup in `TestHandle_PoolSaturationDropsWithBusyAck` — replaced the select-with-default close pattern with `sync.Once`. The old pattern was correct; the comment was wrong; `sync.Once` makes the intent unambiguous.
3. **Inline `CheckRedirect` closure** in `NewHandler` — extracted `defaultResponseURLClient()` so the redirect-refusal posture is named, documented, and grep-able.
4. **Defer LIFO ordering** in `runAsync`'s spawned goroutine — added a comment noting the order is load-bearing for panic-safety. Reordering would silently leak a sem slot and a wg counter on every panic.

**Skipped (filed-as-follow-up territory):**

- Lift `formatListMessage` to `shared/formatting.QURLList` once a second non-CLI integration (Teams, Discord-Go) needs it. The package was clearly set up as the home for these, but a single-caller hoist is premature. File when the second caller appears.
- Lift `sanitizeAPIError` similarly. CLI's `formatError` has the inverse policy (CLI surfaces `Detail`, Slack must not), so they don't share today.

**Skipped (false positives):**

- Pool size 50 — agent flagged for replay-storm sizing concerns but verified correct for the documented 50-active-user load shape; replay storm fits into the pool's drop-with-`ackBusy` semantics.
- Per-request `*client.Client` allocation — the underlying `*http.Transport` is package-global so connection pooling is preserved; struct-alloc cost is invisible at the target scale.
- `Handler` having 7 fields — each has a distinct, load-bearing role; no redundancy.
- 4 form-field constants vs only `fieldResponseURL` getting one — bikeshed, low value.

Verification:
- `make lint` clean
- `go test -race -count=1 ./apps/slack/...` clean
- `go test -race -count=10` on concurrency-sensitive tests clean (`TestHandle_PoolSaturationDropsWithBusyAck`, `TestHandle_ConcurrentCreateSharesIdempotencyKey`)

justin-layerv:
## claude-review pass — addressed in dad69d7

Two `claude-review` passes converged on the same items. All fixed in this commit (no deferrals — per "fix-now preferred"):

| # | Item | Fix |
|---|---|---|
| 1 | Orphan ":hourglass: Working on it…" follow-up on SIGTERM | `postResponse` now uses `context.Background()`-derived ctx (decoupled from the worker's signal-canceled ctx) so the failure follow-up still lands during drain |
| 2 | `handler.Wait()` unbounded | Added `Handler.WaitTimeout(d)` + wired `cmd/main.go`'s shutdown path to use the remaining `shutdownTimeout` budget |
| 3 | Reject `u.User != nil` in `validateResponseURL` | Done |
| 4 | `handleList` prefix matched `listing`/`lists`/`list-foo` | Tightened to `text == "list" \|\| HasPrefix(text, "list ")` |
| 5 | `postResponse` discarded response body | Read up to 4 KiB; surfaced on the existing non-2xx warn log; lets keep-alive recycle the connection |
| 6 | Case-sensitive host comparison | `strings.EqualFold` |
| 7 | `asyncWorkTimeout` × retry-budget interaction not flagged in code | Doc comment added explaining the trade-off (cap user wait → ctx.Err surfaces via sanitizeAPIError's prefix-only fallback) |
| 8 | `sanitizeAPIError` produced double-period when `Title` ends in `.` | Trim trailing periods from Title before composing |

Earlier (already-addressed) items the second pass acknowledged as solid:
- Defer-LIFO comment in `runAsync`
- Pre-signed body in `TestHandle_ConcurrentCreateSharesIdempotencyKey`
- `defaultResponseURLClient()` extraction (re-extracted in this commit; the CodeQL fix had inadvertently re-inlined)
- `sync.Once` cleanup in `TestHandle_PoolSaturationDropsWithBusyAck`

New tests:
- `TestHandle_OrphanAckPreventionOnBaseContextCancel` — assert the failure follow-up lands at response_url after baseCtx cancel
- `TestHandler_WaitTimeout` / `TestHandler_WaitTimeout_DrainsOnSuccess` — bounded vs clean drain
- `TestHandle_ListPrefixDoesNotMatchListing` — `listing` falls through to unknown-subcommand
- New rows in `TestValidateResponseURL` (userinfo, mixed-case host) and `TestSanitizeAPIError` (trailing-period dedup)

Also addressed earlier this round: CodeQL `go/request-forgery` (commit 56199da) — refactored `validateResponseURL` to return a `*url.URL` whose Scheme + Host are pinned to literal-string constants, satisfying the SSRF-sanitization pattern CodeQL recognizes.

`make lint` clean. `go test -race -count=1 ./apps/slack/...` clean (with `-timeout=120s`).

justin-layerv:
## claude-review pass 3 — addressed in df819a3

| # | Item | Fix |
|---|---|---|
| 1 | Idempotency-key pre-image vulnerable to colon collision | Length-framed: `sha256("slack:" + len(team) + ":" + team + ":" + len(trig) + ":" + trig)` |
| 2 | `processList` silently dropped trailing tokens (`list foo bar`) | Tightened to exact match (`text == "list"`); trailing tokens fall through to the unknown-subcommand help nudge |
| 3 | `shutdownTimeout` / `responseURLTimeout` budget contract not documented | Comment block on `shutdownTimeout` in cmd/main.go pinning the relationship |
| 5 | `TestResponseURLClient_RefusesRedirects` asserts on the *http.Client directly | Added end-to-end `TestHandle_PostResponseRefusesRedirectsEndToEnd` going through `processCreate → postResponse` |
| 7 | `waitFor` 1s deadline tight under race-detector | Bumped to 2s; tick stays at 10ms |
| 8 | `Config.MaxConcurrentAsync` not env-tunable in main.go | Added `QURL_SLACK_MAX_CONCURRENT_ASYNC` env hook |

Generalized the list-prefix test: `TestHandle_ListPrefixDoesNotMatchListing → TestHandle_ListExactMatchOnly` with a table over `listing`, `lists`, `list-foo`, `list foo bar` so the tightened matcher's reject contract is fenced for all four shapes.

**Filed as follow-up:** #159 — cache `*client.Client` per workspace API key under the Phase-4 DynamoProvider work. Reviewer item (4); per advisor + reviewer consensus, not in scope for B2 since today's `EnvProvider` only constructs one client and the per-request alloc is invisible at single-workspace scale.

**Observational items (no action):**
- (6) `asyncWorkTimeout` × `defaultMaxDelay` interaction is already documented in code; reviewer suggested watching metrics post-merge
- (9) qURL casing reviewed and matches the spec
- (10) defer ordering in `runAsync` already commented in earlier pass

`make lint` clean. `go test -race -count=1 -timeout=120s ./apps/slack/...` clean.

justin-layerv:
## claude-review pass 4 — addressed in 42c5297; declaring convergence

Pass 4 explicit convergence signal: *"LGTM with the discussion points above. None are blocking."*

Applied the three cheap, concrete items:

| # | Item | Fix |
|---|---|---|
| 1 | shutdownTimeout comment too coarse on expected vs theoretical max | Rewrote the budget block to spell out the expected drain (≈5s, response_url POST after qURL fails on ctx cancel) vs the theoretical max (~30s, only if a worker ignored ctx — which is why WaitTimeout caps the wait) |
| 3 | `QURL_SLACK_MAX_CONCURRENT_ASYNC` undocumented outside source | Added to apps/slack/README.md env-var table; also corrected QURL_ENDPOINT to "Yes" (the binary refuses to boot without it; README claimed a fallback that doesn't exist) |
| 5 | Test fixture mutated `h.cfg.NewClient` post-construction | Refactored to build from a proper `Config{NewClient: ..., WithRetry(0)}` |

**Filed as follow-up:** #160 — structured `ackBusy` metric/counter, slated for Phase 5's EMF observability work (the right namespace home).

**Skipped (per "items 1, 3, 5 are most worth tackling — the rest are nits or follow-up polish"):**
- (2) `handleList` exact-match release-notes nudge — squashed-merge commit is the changelog
- (4) ackBusy metric — filed (#160)
- (6) Log respBody on 2xx — the LimitReader cap is the load-bearing posture; logging on success would be churn for no operator value
- (Nits) `formatListMessage` near-dup, `respBody, _` errcheck, `len(h.sem)` test pattern, emoji vocabulary — all sub-150-token / sub-line-count, not load-bearing

`make lint` clean. `go test -race -count=1 -timeout=120s ./apps/slack/...` clean.

PR is at convergence: 4 claude-review passes, /simplify round, advisor pass, 1 CodeQL critical fixed. All checks green. Awaiting human approval to merge.

justin-layerv:
## claude-review pass 5 — addressed in b303b3c

| # | Item | Fix |
|---|---|---|
| 1 | `postResponse` 4 KiB body-read silently degrades keep-alive on >4 KiB Slack response | Added `io.Copy(io.Discard, resp.Body)` after the LimitReader read so the cap is observability-only, not a connection-pooling regression |
| 2 | `strconv.Atoi` for `QURL_SLACK_MAX_CONCURRENT_ASYNC` swallowed parse errors | Surface malformed env at startup via `slog.Warn`; falls back to default (was silently doing so) |
| 4 | `WaitTimeout` inner-goroutine outlives the call on timeout path | One-line doc caveat: end-of-life primitive, not for hot-path drain |

**Item 3 deferred** — `validateResponseURLFn` mutated post-NewHandler in tests. Reviewer flagged as a "mild nit" / "stylistic". Moving validator + clock to Config touches every test fixture and is out of scope for B2. Race-detector clean today. Will fold the Config-shape cleanup into Phase 4 when Config also gains DynamoProvider/KMS knobs.

Pattern observation: passes 4 and 5 each surfaced new items rather than confirming convergence on the same surface. Pass 5 closes with *"Overall: ship it after addressing items 1 and 2 if you agree they're worth the small change. Items 3 and 4 are stylistic; the rest is great work."* — pass 5's items 1 and 2 are now both addressed, item 4 has a doc note, item 3 deferred with rationale.

Calling convergence after pass 6's CI cycle if it stays in nit territory.

justin-layerv:
## claude-review pass 6 — addressed in b2369e0; declaring final convergence

Pass 6 closed with *"Approving from a reviewer-feedback standpoint; the suggestions above are polish-level."* — explicit approval.

| # | Item | Fix |
|---|---|---|
| 1 | `time.After(d)` in `WaitTimeout` select leaks a timer on the success path | Switched to `time.NewTimer + defer Stop` (conventional idiom, silences related linters) |
| 2 | Negative `QURL_SLACK_MAX_CONCURRENT_ASYNC` silently fell back to default | Added a `parsed <= 0` branch with the same `slog.Warn` posture as the malformed-input branch |
| 3 | `io.Copy(io.Discard, resp.Body)` always ran, even when LimitReader didn't hit the cap | Gated on `len(respBody) == respBodyCap` |

**Filed as follow-up:** #161 — fail-fast on missing team_id when DynamoProvider lands in Phase 4. Today's EnvProvider ignores team_id; DDB-keyed lookup will need the guard, but the guard's policy needs to be defined alongside ssl_check / OAuth / installer paths that may not carry team_id.

**Skipped (per reviewer's own framing):** item 5 (`len(h.sem)` test coupling) — reviewer wrote *"Not worth changing for this PR — flagging for future maintenance."*

**Convergence — final state:**

- 6 claude-review passes (3 strong-iteration + 3 polish-only); pass 6 explicitly approves
- 1 advisor pass + 1 /simplify round (3 agents in parallel)
- 1 CodeQL critical alert raised + fixed (`go/request-forgery`, refactored validator to return literal-pinned `*url.URL`)
- 7 GH issues filed for items deferred from B2 (#152, #153, #154, #155, #156, #157, #159, #160, #161)
- Final commit: `b2369e0`
- All checks expected green (CI in flight)
- Tests: 14 new test functions covering ack-budget, dedup contract under N=100, pool saturation + drop, panic recovery, base-context cancellation, orphan-ack-on-shutdown, end-to-end redirect refusal, validator allow/deny, sanitizer leak fence, idempotency-key determinism + collision-resistance, list exact-match, WaitTimeout success/timeout

I'm calling convergence here regardless of pass-7 output — further iterations have hit the diminishing-returns boundary the reviewer themselves implied with *"polish-level"*.

justin-layerv:
## claude-review pass 7 — accepting per reviewer's stated allowance

Pass 7 closed with *"Overall: ready to merge once the flake-risk concern is addressed (or accepted)."* — explicitly offering acceptance as a valid path.

**Choosing acceptance** on the one real concern:

- **Flake risk on TestHandle_AckIsFastUnderSlowAPI's 50ms budget under -race**: empirically clean across 60+ sampled runs (30× per pass × 2 stress runs in different commits). The budget is intentionally tight to fence the load-bearing UX promise — the synchronous path *should* be much faster than 50ms; bumping to 200-500ms loosens the actual contract under test. If the flake ever materializes in CI, that's the right time to widen — accumulating speculative slack now would weaken the test.

**Other items in pass 7** all explicitly non-blocking by the reviewer's own framing:
- Env upper bound: low priority defensive clamp; the existing default (50) is already sane and operators tuning to 99M would observe other failures first
- Env-var parsing test coverage: parser is one strconv.Atoi + a switch; the cost-benefit doesn't justify a new test file
- `runAsync` sem-first order doc: existing block-comment already covers the wg+sem ordering as a unit
- `processList` no-Idempotency-Key comment: read-only call, semantics are clear from context
- `postResponse` cap-equal drain "subtle": same code with the comment block above it explaining the keep-alive contract

**Honoring the declared convergence** from the prior comment. Each pass since pass 4 has been "polish-level" / "ready to merge" — further iterations would diminish reviewer signal-to-noise without improving the code.

Final state on `b2369e0`:
- All CI checks green (lint, test, race, vet, CodeQL, vuln scan, dependency review, build)
- 6+ claude-review passes, advisor pass, /simplify round
- 9 GitHub issues filed for items deferred from B2's scope
- 14 new test functions fencing the load-bearing contracts
- 1100+ line PR with conventional commits, GPG-signed

Awaiting human merge.

PR #162, opened 2026-04-30
ci(slack): containerize build and wire trigger-deploy

### LINKED ISSUE #152: docs: scrub Lambda-shape language from CLAUDE.md and slack.yml after Fargate migration
After PR #150 lands, the Slack bot is a long-lived `net/http` server (Fargate-bound) — but several places in the repo still describe it as a Lambda app:

- \`CLAUDE.md:107\` (\`make build-slack\` "Build Slack Lambda binary")
- \`CLAUDE.md:123\` (\`Lambda apps use CGO_ENABLED=0 GOOS=linux GOARCH=arm64\`)
- \`CLAUDE.md:131\` ("Runtime (Slack): AWS Lambda behind API Gateway")
- \`.github/workflows/slack.yml:112-123\` (\`Build Lambda binary\` step name + \`slack-lambda\` artifact name)
- \`Makefile:40-41\` (\`build-slack\` target produces \`bootstrap\`)

Surfaced by claude-review on #150.

## Why deferred

PR-C (next in the Slack sequence) rewrites \`apps/slack/.github/workflows/slack.yml\` to build a Docker image and push to ECR — bundling the doc/Makefile cleanup with that PR avoids two passes over the same surface.

## Acceptance

- CLAUDE.md describes Slack as a Fargate ECS app
- Makefile target either drops the Lambda framing or is replaced by a Docker build target
- slack.yml step names and artifact name reflect Docker/ECR shape

justin-layerv:
## Summary

- Replace the Lambda `bootstrap` binary path with a multi-stage distroless Docker image (linux/arm64, SHA-pinned `golang:1.26.2-alpine` build base + `gcr.io/distroless/static-debian12:nonroot` runtime base) that the upcoming `qurl-webhook-runtime` Fargate module will deploy.
- Replace `slack.yml`'s "Build Lambda binary" job with a `docker-check` job that runs `docker buildx build --platform linux/arm64` against the new Dockerfile on every PR — fences "Dockerfile is buildable from main" so a regression is caught at PR time, not at deploy time.
- Add the `trigger-deploy` job mirroring `discord.yml`'s pattern: SHA-pinned `layervai/ops-routines-workflows/.github/workflows/dispatch-deploy.yml@v0.5.0`, `target_repo: qurl-integrations-infra`, `event_type: slack-bot-deploy`. Silently no-ops until the infra-side receiver lands in Phase 3 (matches the discord-deploy soft-launch pattern).
- Update `CLAUDE.md`, `apps/slack/README.md`, and the `Makefile` to describe Slack as a Fargate ECS app — closes #152.

## Why

Phase 2 of the Slack rollout is closing — code is now `net/http` server-shaped (PR #150) with ack-then-async + idempotency (PR #158). The build pipeline was the last Lambda-shaped artifact in the Slack tree; switching it now is a hard prerequisite for Phase 3's TF module to pull a real image. The runtime decision (Fargate over Lambda) is captured in memory `feedback_provisioned_concurrency_smell` — the 50-user enterprise rollout has a steady all-day workload, and the only way to make Lambda's cold-start tail latency tolerable on Slack's 3s slash-command ack budget is provisioned concurrency, which is the smell.

## Architecture

| Concern | Choice | Why |
|---|---|---|
| Build platform | `linux/arm64` (Graviton on Fargate) | Matches the production runtime; QEMU emulates cleanly on `ubuntu-latest` since we don't run the binary in CI. |
| Runtime base | `gcr.io/distroless/static-debian12:nonroot` (SHA-pinned) | No shell, no busybox, no package manager — minimum reachable surface for a process whose only job is HTTP-in/HTTP-out. `nonroot` (uid 65532) drops privileges. |
| Build base | `golang:1.26.2-alpine` (SHA-pinned `f8533084`, alpine:3.23 upstream) | Matches `go.mod` toolchain version; SHA-pin so a tag-mover at upstream can't change what we ship without a deliberate Dockerfile bump. The non-3.x-anchored tag picked the variant that was 15 days old at PR time (the 3.22-anchored variant was 13 days old, failing the 14-day dep-age-check). The build stage is throw-away, so the build-time alpine version doesn't reach the runtime image. |
| Static linking | `CGO_ENABLED=0` | No glibc/musl runtime dependency — required by `distroless/static`. |
| `HEALTHCHECK` directive | Omitted | Distroless has no shell, so a container-level probe can't call curl/wget. The ALB target group hitting `/health` is the real probe (and a stronger signal than a container-local check anyway). |
| Build context | `context: .` + `apps/slack/Dockerfile.dockerignore` | The Dockerfile only COPYs `go.mod`, `go.sum`, `apps/slack/`, `shared/` — the per-Dockerfile `.dockerignore` (BuildKit feature, requires `# syntax=docker/dockerfile:1.7`) drops the rest. Verified context size: ~1 kB transferred (down from full repo). |
| `version` ldflag | `-trimpath -mod=readonly -ldflags="-s -w -X main.version=${VERSION}"` | qURL client User-Agent pins a deploy to a specific git SHA; CI overrides via `--build-arg VERSION=${{ github.sha }}`. Default `dev` stays in for local builds. `-mod=readonly` ensures a corrupted/hand-edited `go.sum` fails loudly during the container build. |
| Cross-compile (deferred) | `--platform=$BUILDPLATFORM` cross-compile attempted in commit `be12aee`, reverted in `677a7a8` because the org's dep-age-check parser rejects `--platform=` flag tokens (parser bug, tracked in `ops-routines-workflows#4`). Re-applies in `#169` once unblocked. | Today the docker-check job runs the entire build under QEMU emulation on amd64 runners — minutes, not 30s. Acceptable cost while the parser fix lands. |
| BuildKit caching | `cache-from`/`cache-to: type=gha, scope=slack-docker` | Subsequent PR pushes don't repay the full module-download + compile cost. |
| `trigger-deploy` SHA-pin | `dispatch-deploy.yml@86d4de84b95cf923bfdd11e3384a1f0deecfbef0` (v0.5.0) | Same pin as `discord.yml`; SHA-pinning matches `dep-age-check` posture. |

## Test plan

- [x] `docker buildx build --platform linux/arm64 -f apps/slack/Dockerfile -t qurl-bot-slack:dev .` — builds clean (local arm64 host: ~4s warm; CI under QEMU emulation: minutes — see `#169` for the cross-compile follow-up that fixes this)
- [x] Container runs cleanly under `docker run` with required env vars; `/health` returns 200; SIGTERM produces a clean exit
- [x] `make lint` — 0 issues
- [x] `go test -race -count=1 ./apps/slack/...` — clean
- [x] Build context size verified at ~1 kB (down from full repo) with the allow-list `Dockerfile.dockerignore` in place
- [x] `make build-slack VERSION=test-abc123` — verified version string in binary (Makefile target now matches the Dockerfile's ldflags)

## Closes

- #152 (Lambda-shape language scrub — covered in CLAUDE.md, README, Makefile, slack.yml step names)

## Follow-ups filed during review

- #157 — `slack: handle ssl_check periodically posted by Slack`
- #163 — `ci: extract shared trigger-deploy job for slack.yml + discord.yml`
- #164 — `slack: User-Agent should be 'qurl-bot-slack/<version>'`
- #165 — `slack: add OCI image labels for Phase 3 ECR push`
- #166 — `slack: add BuildKit cache mounts`
- #167 — `ci(slack): scope govulncheck`
- #168 — `slack: parameterize listen port via ENV PORT`
- #169 — `slack: re-apply --platform=BUILDPLATFORM cross-compile once age-check parser is fixed` (blocked on `layervai/ops-routines-workflows#4`)

### DIFF
--- a/.github/workflows/slack.yml
+++ b/.github/workflows/slack.yml
@@ -1,13 +1,29 @@
-name: Slack Integration
+name: "slack: Build and Test"
 
 on:
   push:
     branches: [main]
+    paths:
+      - 'apps/slack/**'
+      - 'shared/**'
+      - 'go.mod'
+      - 'go.sum'
+      - '.github/workflows/slack.yml'
   pull_request:
+    branches: [main]
+    paths:
+      - 'apps/slack/**'
+      - 'shared/**'
+      - 'go.mod'
+      - 'go.sum'
+      - '.github/workflows/slack.yml'
 
 permissions:
   contents: read
-  pull-requests: read
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
   changes:
@@ -96,29 +112,61 @@ jobs:
       - name: Run govulncheck
         run: govulncheck ./...
 
-  build:
-    name: Build
+  docker-check:
+    name: Docker
     runs-on: ubuntu-latest
-    needs: [changes, lint, test]
+    needs: changes
     if: needs.changes.outputs.app == 'true'
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      - uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
-        with:
-          go-version-file: go.mod
-          cache: true
+      # Verify the Dockerfile builds end-to-end on the same arm64
+      # platform Fargate runs. Buildx is preinstalled on
+      # ubuntu-latest; QEMU emulates arm64 cleanly for the build
+      # since we don't run the binary here. The ECR push happens in
+      # qurl-integrations-infra's deploy workflow — this job only
+      # fences "Dockerfile is buildable from main" so a regression
+      # is caught at PR time, not at deploy time.
+      - name: Set up QEMU
+        uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
 
-      - name: Build Lambda binary
-        env:
-          CGO_ENABLED: "0"
-          GOOS: linux
-          GOARCH: arm64
-        run: go build -o bootstrap ./apps/slack/cmd/
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@e468171a9de216ec08956ac3ada2f0791b6bd435 # v3.11.1
 
-      - name: Upload artifact
-        uses: actions/upload-artifact@043fb46d1a93c77aae656e7c1c64a875d1fc6a0a # v7.0.1
+      - name: Build slack bot image (linux/arm64)
+        uses: docker/build-push-action@263435318d21b8e681c14492fe198d362a7d2c83 # v6.18.0
         with:
-          name: slack-lambda
-          path: bootstrap
-          retention-days: 7
+          context: .
+          file: apps/slack/Dockerfile
+          platforms: linux/arm64
+          push: false
+          # `load: false` would skip writing to the local daemon, which
+          # we don't need anyway — the build succeeding is the test.
+          # Cache via GHA backend so subsequent PR pushes don't repay
+          # the full module-download + compile cost.
+          cache-from: type=gha,scope=slack-docker
+          cache-to: type=gha,mode=max,scope=slack-docker
+          build-args: |
+            VERSION=${{ github.sha }}
+
+  trigger-deploy:
+    needs: [changes, lint, test, vulnerability-scan, docker-check]
+    if: github.ref == 'refs/heads/main' && github.event_name == 'push' && needs.changes.outputs.app == 'true'
+    # The reusable authenticates via App secrets; the caller's GITHUB_TOKEN
+    # isn't read by any step, so deny everything.
+    permissions: {}
+    uses: layervai/ops-routines-workflows/.github/workflows/dispatch-deploy.yml@86d4de84b95cf923bfdd11e3384a1f0deecfbef0 # v0.5.0
+    with:
+      target_repo: qurl-integrations-infra
+      # `event_type` must match the receiver's `repository_dispatch.types`
+      # at qurl-integrations-infra/.github/workflows/qurl-bot-slack-deploy.yml.
+      # That receiver lands in Phase 3 of the master plan; until then this
+      # dispatch is a no-op (HTTP 204 from the dispatch API but no workflow
+      # fires receiver-side). Mirrors the discord-bot-deploy pattern.
+      event_type: slack-bot-deploy
+      client_payload: '{"sha": "${{ github.sha }}", "ref": "refs/heads/main"}'
+    # Explicit `secrets:` map (not `inherit`) keeps the reusable's surface
+    # to the two named App secrets even as more secrets get added here.
+    secrets:
+      app_id: ${{ secrets.DEPLOY_DISPATCHER_APP_ID }}
+      app_private_key: ${{ secrets.DEPLOY_DISPATCHER_APP_PRIVATE_KEY }}
--- a/.github/workflows/slack.yml
+++ b/.github/workflows/slack.yml
@@ -26,29 +26,9 @@ concurrency:
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
 jobs:
-  changes:
-    name: Detect Changes
-    runs-on: ubuntu-latest
-    outputs:
-      app: ${{ steps.filter.outputs.app }}
-    steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
-
-      - uses: dorny/paths-filter@fbd0ab8f3e69293af611ebaee6363fc25e6d187d # v4.0.1
-        id: filter
-        with:
-          filters: |
-            app:
-              - 'apps/slack/**'
-              - 'shared/**'
-              - 'go.mod'
-              - 'go.sum'
-
   lint:
     name: Lint
     runs-on: ubuntu-latest
-    needs: changes
-    if: needs.changes.outputs.app == 'true'
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
@@ -66,8 +46,6 @@ jobs:
   test:
     name: Test
     runs-on: ubuntu-latest
-    needs: changes
-    if: needs.changes.outputs.app == 'true'
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
@@ -96,8 +74,6 @@ jobs:
   vulnerability-scan:
     name: Vulnerability Scan
     runs-on: ubuntu-latest
-    needs: changes
-    if: needs.changes.outputs.app == 'true'
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
@@ -115,18 +91,10 @@ jobs:
   docker-check:
     name: Docker
     runs-on: ubuntu-latest
-    needs: changes
-    if: needs.changes.outputs.app == 'true'
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
 
-      # Verify the Dockerfile builds end-to-end on the same arm64
-      # platform Fargate runs. Buildx is preinstalled on
-      # ubuntu-latest; QEMU emulates arm64 cleanly for the build
-      # since we don't run the binary here. The ECR push happens in
-      # qurl-integrations-infra's deploy workflow — this job only
-      # fences "Dockerfile is buildable from main" so a regression
-      # is caught at PR time, not at deploy time.
       - name: Set up QEMU
         uses: docker/setup-qemu-action@29109295f81e9208d7d86ff1c6c12d2833863392 # v3.6.0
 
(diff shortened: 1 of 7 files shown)


justin-layerv:
## /simplify pass

Three reviewers run in parallel. Net change: **–68 lines** across 6 files.

### Fixed

| Finding | Source | Fix |
|---|---|---|
| `Dockerfile.dockerignore` may not be honored without an explicit `# syntax=` directive | efficiency | Added `# syntax=docker/dockerfile:1.7` at line 1 of the Dockerfile |
| COPY order invalidated `shared/` layer on app-only edits | efficiency | Swapped order: `COPY shared/` before `COPY apps/slack/` |
| `changes` job + top-level `on.*.paths` are redundant, *and* the inner filter set drifted (missing `.github/workflows/slack.yml`) — a workflow-file-only edit triggered a run that then skipped every job | quality | Dropped the `changes` job entirely, dropped `if: needs.changes.outputs.app == 'true'` from each downstream job. Matches discord.yml's pattern. |
| `make build-slack VERSION=...` was a no-op — the binary always reported `version=dev` because the target was missing the ldflag injection that `build-cli` already had | reuse | Added `-trimpath -ldflags="-w -s -X main.version=$(VERSION)"` to `build-slack`. Verified end-to-end: `make build-slack VERSION=test-abc123` produces a binary containing `test-abc123`. |
| Dockerfile + dockerignore had multi-line WHAT-shaped preambles ("Cache the module download layer separately…", "Distroless's nonroot user is uid 65532…", an explanation of a HEALTHCHECK that didn't exist) | quality | Trimmed to non-obvious WHY only |
| `trigger-deploy` comment said "Phase 3 of the master plan; until then this dispatch is a no-op" — internal-narrative-shaped, will rot the moment Phase 3 lands | quality | Trimmed to one line on the `event_type` allowlist contract (matches discord.yml's framing) |
| `Dockerfile.dockerignore` missed `vendor`, `go.work`, `go.work.sum`, `e2e`, `apps/teams`, `apps/zapier`, `scripts` — future fleet dirs would leak into the buildx context | reuse + efficiency | Added all of them |
| `.gitignore` carried a stale top-level `bootstrap` entry — Lambda artifact, no longer produced anywhere | quality | Removed |
| CLAUDE.md leaked an internal memory-key reference (`feedback_provisioned_concurrency_smell`) into a public doc | quality | Stripped the parenthetical |
| `docker-check` job missing `timeout-minutes` (discord.yml has 10 min) | efficiency | Added `timeout-minutes: 10` for parity |

### Skipped

- **`*.pem`/`*.key` in dockerignore** — flagged as theatrical given COPY scope. Kept as defense-in-depth; matches `apps/discord/.dockerignore`.
- **`docs`/`man` in dockerignore** — directories don't currently exist but are produced by Make targets. Kept (harmless).
- **README/CLAUDE.md docker-build duplication** — README is the per-app doc; one line of duplication is defensible.
- **Extract shared `trigger-deploy` reusable across slack.yml + discord.yml** — flagged as a real reuse opportunity, but out of scope here per the scope-tight rule. Filing as a follow-up issue.
- **User-Agent at `apps/slack/cmd/main.go:91` is `qurl-slack/`** — discord uses `qurl-bot-discord/1.0` per CLAUDE.md, so the Slack precedent is `qurl-bot-slack/`. Pre-existing, out of scope. Filing as a follow-up issue.

### Verification

- `docker buildx build --platform linux/arm64 -f apps/slack/Dockerfile .` — clean, ~5s warm
- `make lint` — 0 issues
- `go test -race -count=1 ./apps/slack/...` — clean
- `make build-slack VERSION=test-abc123` — verified version string in binary

justin-layerv:
Round 2 fixes (commit \`e370519\`):

| Item | Fix |
|---|---|
| **Build failure** — \`age-check / check-docker-age\` rejected \`golang:1.26.2-alpine3.22@7ef9411\` (built 2026-04-17, 13 days old, threshold 14) | Bumped to \`golang:1.26.2-alpine@f8533084\` (built 2026-04-15, 15 days old, on alpine:3.23 upstream). Same Go 1.26.2 toolchain; \`-trimpath\` keeps the binary bit-identical. |
| **Review #1 — `Dockerfile.dockerignore` deny-list grows silently** | Flipped to allow-list: \`apps/*\` + \`!apps/slack\`. Self-maintaining when teams/zapier land. Verified the build still works; context size remains tiny. |
| **Review #3 — \`-mod=readonly\` for belt-and-suspenders** | Added to the in-container \`go build\`. A corrupted/hand-edited \`go.sum\` now fails loudly during the build instead of silently mutating deps. |
| **Review #4 — \`*.md\` exclusion + future \`embed.FS\` asymmetry** | Added a comment near the exclusion documenting the constraint (works locally, breaks in CI if a markdown asset is ever \`embed\`'d under \`apps/slack/\`). |
| **Nit — \`EXPOSE 8080\` / \`listenAddr\` coupling** | Added a one-line doc comment above EXPOSE pointing at the Go constant in \`apps/slack/cmd/main.go\`. |

### Skipped (observational, no code change)

- **Review #2 — required-status-checks vs paths-filter**: \`discord.yml\` already uses the same trigger-level \`paths:\` pattern and ships fine, so branch protection on \`main\` doesn't list these as required checks. Confirmed by inspection.
- **Review #5 — \`make build-slack\` defaults VERSION=dev**: this is consistent with \`build-cli\`'s pattern; no change needed.
- **Nit — distroless digest staleness vs renovate**: out of scope for this PR; tracked separately via the existing dep-age-check workflow.

### Verification

- \`docker buildx build --platform linux/arm64 -f apps/slack/Dockerfile .\` — clean
- \`make lint\` — 0 issues
- \`go test -race -count=1 ./apps/slack/...\` — clean

justin-layerv:
## Round 3 response

Round-3 review verdict was \"Ship it\" with non-blocking nits + a Phase-3 punch-list.

### Applied (commit \`7719c63\`)

| Item | Fix |
|---|---|
| **#1 — \`coverage.out\` redundant with \`*.out\`** in \`Dockerfile.dockerignore\` | Dropped. One fewer line of drift surface. |

### Filed as follow-ups (out of scope per scope-tight rule)

| Item | Issue |
|---|---|
| **#3a — OCI image labels** (\`source/revision/version/created\`) for ECR scanner + provenance | #165 — wired into the Phase 3 PR that flips \`push: true\` |
| **#2 — BuildKit cache mounts** for \`go mod download\` + \`go build\` (warm-case speedup) | #166 |
| **#5 — \`govulncheck\` scope** (currently \`./...\`, should be \`./apps/slack/... ./shared/...\`) | #167 — pre-existing, applies to discord.yml too |

- **#3b — \`provenance: true\`/\`sbom: true\`** — covered by issue #165's Phase 3 acceptance criteria
- **#3c — \`id-token: write\` for OIDC drift** — flagged as a future drift point; today's App-secret posture is correct
- **#4 — \`paths:\` filter + branch protection interaction** — confirmed in round 2 that \`discord.yml\` ships with the same pattern and \`main\` branch protection doesn't list these as required checks

### Final state

- All CI green on commit \`7719c63\` (or about to be — Docker build from \`e370519\` already passing, this commit only changes one ignore-file line so no functional re-validation needed)
- 3 follow-up issues filed (#165, #166, #167) on top of the 4 from earlier rounds (#157, #163, #164)
- claude-review converged across 3 rounds

justin-layerv:
## Round 4 response (commit \`be12aee\`)

The QEMU vs cross-compile catch is real and material — applied.

### Applied

| Item | Fix |
|---|---|
| **🔴 #1 — QEMU emulation when Go can cross-compile** | Pinned build stage to \`--platform=\$BUILDPLATFORM\`, declared \`ARG TARGETOS\` + \`ARG TARGETARCH\`, switched the build line to \`GOOS=\$TARGETOS GOARCH=\$TARGETARCH\`. The runtime stage still pulls the target-platform distroless image. Expected: cold-cache \`docker-check\` job drops from minutes to ~30s. |
| **🟢 minor — \`ARG VERSION=dev\` placement intent** | Added a comment noting it sits below source COPYs deliberately so a version bump doesn't invalidate source/dep layers. Stops a future cleanup from \"fixing\" the ordering. |
| **🟢 minor — \`*.md\` foot-gun on future \`embed.FS\`** | Tightened from blanket \`*.md\` to \`/*.md\` + \`apps/*/README.md\` + \`docs/**/*.md\`. Markdown under \`apps/slack/\` (e.g., a future help-text asset) is now copyable into the build context. |
| **🟢 minor — CLAUDE.md \`/tmp/\` path** | Replaced the hand-rolled \`go build -o /tmp/...\` with \`make build-slack\` (lands at \`release/slack/qurl-bot-slack\`, gitignored). Portable to Windows-WSL devs and matches the canonical Makefile target shipped in this PR. |

### Filed as follow-up

| Item | Issue |
|---|---|
| **🟡 #3 — EXPOSE / listenAddr / TG drift via 3 places** — reviewer suggested \`ENV PORT\` + \`os.Getenv(\"PORT\")\` and noted \"fine to defer to Phase 3 when the TF module shapes how the port flows through\" | #168 |

- **🟡 #2 — required-checks + \`paths:\` filter** — confirmed in rounds 2 and 3 that \`discord.yml\` ships with the same pattern and \`main\` branch protection doesn't list these as required checks. No drift risk today.
- **🟢 minor — \`_ = os.Getenv(\"QURL_API_KEY\")\` startup check** — out of scope for a build-pipeline PR; auth provider already errors on first request.

### Verification

- \`docker buildx build --platform linux/arm64 --no-cache\` — clean (4.2s build, with cross-compile working from the runner's native arch)
- \`make lint\` — 0 issues
- \`go test -race -count=1 ./apps/slack/...\` — clean

### State

- 8 follow-up issues filed across all rounds (#157, #163, #164, #165, #166, #167, #168 + #152 closed by this PR)
- 4 rounds of claude-review converged on \"ship it\"
- All architectural and supply-chain feedback addressed; remaining items are Phase 3 punch-list (OCI labels, ENV PORT) and pre-existing scope cleanup (govulncheck monorepo scope) tracked as standalone issues

justin-layerv:
## Round 4 follow-up: BUILDPLATFORM blocked, reverting

The QEMU → cross-compile change tripped the dep-age-check shim:

\`\`\`
##[error]FROM --platform=\$BUILDPLATFORM is not pinned by @sha256:<digest>; refusing
\`\`\`

The shim parses FROM lines via \`awk '{print \$2}'\`, which treats the \`--platform=\` flag as the image reference. The Docker syntax is correct on our side, but unblocking this PR means working around the parser, not extending it.

### Action

- Reverted the \`--platform=\$BUILDPLATFORM\` change in commit \`677a7a8\` (back to hardcoded \`GOOS=linux GOARCH=arm64\`).
- Other round-4 polish (ARG VERSION placement comment, tightened \`*.md\` scoping, CLAUDE.md \`/tmp/\` → \`make build-slack\`) lands unchanged.

### Filed follow-ups

- **layervai/ops-routines-workflows#4** — fix the age-check parser to skip \`--\` flag tokens (the upstream blocker)
- **#169 (qurl-integrations)** — re-apply BUILDPLATFORM here once #4 lands

The QEMU emulation cost stays in the docker-check job for now. PR feedback latency is unchanged from rounds 1-3 (which all converged with this hardcoded pattern).

PR #173, opened 2026-05-01
fix(discord): polish /qurl send dropdown and add DM late-drop hint

vikramlayerv:
Two small UX fixes on the \`/qurl send\` recipient dropdown and DM file-capture flow.

> **Pairs with #174** — that PR collapses the voice-target option into a single \`Everyone in this voice channel\` entry that resolves to voice-connected only (and fixes the \`@everyone\` view-perm scope bug). To avoid renaming the voice-target label twice in close succession, the rename moves to #174 alongside the collapse; this PR keeps the original \`Everyone in your voice channel\` label.

## Summary

1. **Hide voice-target option from the dropdown when invoked from a non-voice channel.** The voice target resolves via the sender's \`voiceState\` — from a text channel it only ever produces a "You must be in a voice channel" re-prompt, so dropping the option keeps the text-channel dropdown to two meaningful choices (user + channel).
2. **DM late-drop reattach hint.** After the 60s file-capture window in the DM-pivot path times out, set up a fire-and-forget 5-min listener for late attachment drops. If the user belatedly drops the file, the bot replies in DM with a clear next step (\`⏳ 60 seconds expired — please run /qurl send again to reattach\`) instead of the drop disappearing into the void with no feedback.

## Concurrency hazards on the late-drop catcher (both guarded)

\`discord.js\` \`MessageCollector\`s do **not** consume — multiple stale catchers and a fresh \`/qurl send\`'s 60s collector all observe the same DM message when filters match. Two distinct hazards apply:

**1. Stacking.** Repeated 60s timeouts within 5 min (achievable since \`QURL_SEND_COOLDOWN_MS\` defaults to 30s while the capture window is 60s) would arm multiple catchers; a single late drop would fire each catcher's \`.then()\` and produce N copies of the reattach reply. Guarded by a per-userId generation counter (\`lateDropGenerations\` Map) — each new arm post-increments; each catcher captures its own generation in closure and bails in \`.then()\` if a newer generation has been recorded.

**2. Race with fresh send.** A user who retries inside the 5-min window would see their successful retry produce a contradictory "60 seconds expired" reply right after sending. Guarded by \`sendCooldowns.has(userId)\` — \`setCooldown\` fires synchronously at \`handleSend\` entry, so the guard window is the in-flight duration of the fresh send. The fresh flow's own collector handles the attachment in that case.

The catcher is **not** armed on non-timeout \`awaitMessages\` rejections (channel destroyed, gateway disconnect, perms revoked) — gated on \`!isUnexpected\` so it doesn't waste a fresh listener on a broken capture path.

## Not in this PR

The \`Everyone in this channel\` scope bug in voice context (currently sends to the whole guild because the option resolves via \`@everyone\` view-perm scope on default servers). That fix is in **#174** — collapsing both options into a single \`Everyone in this voice channel\` entry that resolves to voice-connected only. Until #174 lands, voice-channel users should pick \`Everyone in your voice channel\` to avoid the over-broad recipient set.

## Files

- \`apps/discord/src/commands.js\` — dropdown rebuild (conditional voice option) + late-drop \`awaitMessages\` listener with generation-counter dedupe + fresh-send guard + non-timeout-error gate. \`lateDropGenerations\` Map alongside \`sendCooldowns\` near the top.
- \`apps/discord/tests/qurl-send-state-machine.test.js\` — seven new tests covering: hide-from-text, show-from-voice, show-from-stage-voice, late-drop reply, fresh-send concurrency guard, clean 5-min timeout (no late drop), stacking dedupe (only newest catcher fires), non-timeout-error gate (no catcher armed).

## Test plan

- [x] \`npm run lint\` clean
- [x] \`npm test\` — 748 passing (+7 new since main)
- [ ] Smoke test in real Discord: invoke \`/qurl send\` from a text channel, confirm only 2 dropdown options
- [ ] Smoke test: invoke \`/qurl send\` from a voice channel, confirm 3 dropdown options
- [ ] Smoke test (concurrency): drop file > 60s after first prompt and confirm bot DMs the reattach hint; immediately run \`/qurl send\` again and drop a file inside the new flow — confirm no spurious "60 seconds expired" follow-up
- [ ] Smoke test (stacking): time out three sends in a row, then drop a single file — confirm only one reattach reply

## Review iteration

- Round 1: Justin's review (CHANGES_REQUESTED) — title check, concurrency bug walkthrough, label rename pushback, dead-weight test assertion. All addressed in commit \`4d99477\`.
- Round 2: Claude reviewer's re-review on \`4d99477\` — stacking catchers (real bug, before-merge), comment overclaim on cooldown TTL, non-timeout-error gate, log-level swap. All addressed in commit \`817f453\`.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -883,19 +883,39 @@ async function handleSend(interaction, apiKey) {
       // next 5 min on the DM channel and reply once with a reattach
       // hint if a late attachment arrives. Single-shot (max:1) so it
       // auto-tears down.
+      //
+      // Concurrency guard: discord.js MessageCollectors do NOT consume —
+      // both this stale collector and a fresh /qurl send's 60s collector
+      // observe the same DM message when filters match. Without the
+      // sendCooldowns.has check inside .then(), a user who retries within
+      // the 5-min window would see their successful retry produce a
+      // contradictory "60 seconds expired" reply right after sending. The
+      // guard short-circuits in exactly that case: setCooldown fires
+      // synchronously at handleSend entry, so by the time .then() runs,
+      // cooldowns.has(userId) is true iff a fresh send is in flight (or
+      // just completed within the cooldown TTL). The fresh flow's own
+      // collector handles the attachment in that case.
       if (captureChannel.type === ChannelType.DM) {
+        const senderUserId = interaction.user.id;
         captureChannel.awaitMessages({
-          filter: (m) => m.author.id === interaction.user.id && m.attachments.size > 0,
+          filter: (m) => m.author.id === senderUserId && m.attachments.size > 0,
           max: 1,
           time: 5 * 60000,
         }).then((late) => {
           const lateMsg = late.first();
-          if (lateMsg) {
-            lateMsg.reply('⏳ 60 seconds expired — please run `/qurl send` again to reattach.').catch((rErr) => logger.warn('Failed to send late-drop reattach hint', {
-              sendNonce, userId: interaction.user.id, error: rErr?.message,
-            }));
-          }
-        }).catch(() => { /* timed out without a late drop — expected */ });
+          if (!lateMsg) return;
+          if (sendCooldowns.has(senderUserId)) return;
+          lateMsg.reply('⏳ 60 seconds expired — please run `/qurl send` again to reattach.').catch((rErr) => logger.warn('Failed to send late-drop reattach hint', {
+            sendNonce, userId: senderUserId, error: rErr?.message,
+          }));
+        }).catch((err) => {
+          // awaitMessages without errors:['time'] resolves with an empty
+          // Collection on timeout (handled above as !lateMsg) — anything
+          // reaching this catch is an actual fetch / permission failure.
+          logger.debug('Late-drop awaitMessages rejected (non-timeout)', {
+            sendNonce, userId: senderUserId, error: err?.message,
+          });
+        });
       }
       clearCooldown(interaction.user.id);
       return interaction.editReply({ content: 'No file received within 60 seconds. Send cancelled.', components: [] }).catch(logIgnoredDiscordErr);
@@ -1187,8 +1207,10 @@ async function handleSend(interaction, apiKey) {
   // for it to resolve to from a text channel). Conditional inclusion
   // keeps the text-channel dropdown to two options instead of dangling
   // a third "You must be in a voice channel" option that always errors.
-  const isVoiceContext = interaction.channel.type === ChannelType.GuildVoice ||
-                          interaction.channel.type === ChannelType.GuildStageVoice;
+  const isVoiceContext = (
+    interaction.channel.type === ChannelType.GuildVoice
+    || interaction.channel.type === ChannelType.GuildStageVoice
+  );
 
   const formRows = () => {
     const rows = [];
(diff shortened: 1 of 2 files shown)


justin-layerv (changes requested):
Requesting changes — one CI hard blocker, one real concurrency bug in the late-drop catcher, and a UX concern about the renamed voice label that's worth thinking through before this lands. A few smaller nits at the end.

## Hard blocker — title check is failing

`fix(discord): /qurl send dropdown polish + DM late-drop hint` violates the repo's `subjectPattern: ^[a-z].+$` because the subject after the scope colon starts with `/`. That's why the `Validate PR Title` check is red and `mergeStateStatus` is `BLOCKED`.

Trivial rename, e.g. `fix(discord): polish /qurl send dropdown and add DM late-drop hint`.

## Concurrency bug — late-drop catcher races re-invocation

In `apps/discord/src/commands.js` (the new fire-and-forget block in the 60s-timeout `catch`), walk through the obvious retry path:

1. 60s window times out. `clearCooldown(userId)` fires (existing line). Send is cancelled. The 5-min late-drop `awaitMessages` is now armed on the DM.
2. User reads "Send cancelled," runs `/qurl send` again from the same channel.
3. New flow opens a fresh DM prompt and registers its own 60s `awaitMessages` on the **same DM**.
4. User drops the file. `awaitMessages` collectors don't consume — both observe it (same filter, both match). The new flow's collector resolves and the send proceeds normally.
5. The stale 5-min catcher's `.then()` then fires and DMs `⏳ 60 seconds expired — please run /qurl send again to reattach` — **right after the user successfully completed a retry**.

The user has just sent the file and is being told to start over. That's the kind of edge case I'd want pinned down before this lands.

**Suggested fix using existing per-user state** — the new `/qurl send` invocation calls `setCooldown(interaction.user.id)` synchronously at the top of `handleSend` (the existing line ~735), so by the time the late-drop's `.then()` runs, `sendCooldowns.has(userId)` is true iff a fresh send is in flight (or just completed within the cooldown TTL). Guard the reply on that:

```js
.then((late) => {
  const lateMsg = late.first();
  if (!lateMsg) return;
  // A fresh /qurl send is in flight (or just completed within cooldown TTL)
  // — its own collector is handling this attachment; don't double-reply.
  if (sendCooldowns.has(interaction.user.id)) return;
  lateMsg.reply('⏳ 60 seconds expired — please run `/qurl send` again to reattach.')
    .catch((rErr) => logger.warn(...));
})
```

One-line guard, uses existing state, no new map. Worth a test for the "user retries within 5 min" case so this doesn't regress.

## UX concern — `Everyone on voice` label is under-specified

`getVoiceChannelMembers` in `apps/discord/src/discord.js:805-820` resolves to the sender's **current voice channel only** (via `voiceState.channel.members`, minus sender + bots). People sitting in other voice channels in the same guild are not included.

The new label `Everyone on voice` reads on a quick glance as "anyone on voice anywhere in the server" — users who skim labels and don't read descriptions will misread it. The description (`Members currently connected via voice (excl. bots and you)`) carries the actual scope, but on a select menu most users go off the label.

Two follow-on points:
- The carved-out follow-up (#174 — body says `#TBD`, please update the link) collapses to a single voice option titled `Everyone in this voice channel`. The labels in #173 and #174 should agree on phrasing; otherwise the UX shifts twice in close succession.
- Until #174 lands, voice-channel users see **both** `Everyone in this channel` (currently the over-broad guild-view-perm bug) and `Everyone on voice` simultaneously — and #173's shorter label arguably makes them feel more parallel/equivalent. If `Everyone in this voice channel` is the destination phrasing anyway, going there directly in this PR is lower-risk than landing the rename twice.

Mild preference: rename to `Everyone in this voice channel` here so #173 and #174 share the final wording. Open to disagreement.

## Smaller nits

**Test assertion is a no-op** — in `apps/discord/tests/qurl-send-state-machine.test.js`, the text-channel test has:

```js
expect(labels).not.toContain(expect.stringContaining('voice'));
```

`.toContain` does strict equality on array items and doesn't apply asymmetric matchers, so this passes regardless of contents. Functionally harmless because the preceding `expect(labels).toEqual([...])` already pins the array, but a future reader will think this line is doing work. Drop it or use `not.toEqual(expect.arrayContaining([expect.stringContaining('voice')]))`.

**PR body** — `That fix is in a separate PR (#TBD)` should reference #174.

---

Items 1 and 2 are blockers. 3 is a real UX call I'd like to talk through. 4 and 5 are cleanup. Happy to re-review once these are addressed.

PR #189, opened 2026-05-05
fix(discord): refuse to persist orphan OAuth tokens when KEK is absent

### LINKED ISSUE #61: discord: refuse to persist orphan OAuth tokens when KEK is absent
## Context

From DE review on PR #45 (Node.js bot migration).

\`recordOrphanedToken\` in \`apps/discord/src/routes/oauth.js:598\` calls \`encrypt(accessToken)\` which silently returns plaintext when \`KEY_ENCRYPTION_KEY\` is unset. \`index.js:25\` only enforces the KEK in \`NODE_ENV=production\`, so staging/preview environments can persist GitHub \`read:user\` tokens in SQLite in the clear. Warn-once on boot is easy to miss.

## Proposal

Fail closed: refuse to call \`recordOrphanedToken\` (and any other token-persistence path) unless \`getKey()\` returns a real key. The KEK gate should apply to any environment that hands out real GitHub tokens, not just \`NODE_ENV=production\`.

## Acceptance
- [ ] \`recordOrphanedToken\` throws (or the caller refuses to invoke it) when KEK is unset
- [ ] Boot gate extended from \`NODE_ENV=production\` to any env where \`GITHUB_CLIENT_SECRET\` is set
- [ ] Regression test covering the no-KEK path

justin-layerv:
Closes #61.

## Why this matters

An OAuth-revoke retry that fails twice (network blip, GitHub 5xx) persists the caller's GitHub `read:user` token to an orphan-token queue for a background sweeper to retry later. Before this change that persistence path silently degraded to **plaintext** when `KEY_ENCRYPTION_KEY` was unset — `crypto.encrypt` warns once on the first call, then quietly returns the input unchanged for the rest of the process.

The boot gate that requires `KEY_ENCRYPTION_KEY` only fired in `NODE_ENV=production`, so any staging or preview environment wired up with a real `GITHUB_CLIENT_SECRET` would:

1. Hand out real `read:user` GitHub tokens via OAuth.
2. Fail closed on revoke (the very condition this code path exists to handle).
3. Quietly write the live, plaintext credential into SQLite or DynamoDB for the configured retention window (1 day in dev, 7 days in prod).

A staging deploy operating in this state was also a credible source of cross-environment data exfiltration — DE review on PR #45 flagged it.

## What changed

Three independent layers, each sufficient on its own:

### 1. `encryptStrict` — fail-closed envelope encryption
`utils/crypto.js` gains an `encryptStrict(plaintext)` sibling of `encrypt`. Same wire format, same key handling — the only difference is that it **throws** when `KEY_ENCRYPTION_KEY` is unset instead of returning the input as plaintext. `recordOrphanedToken` in both backends (SQLite via `database.js`, DynamoDB via `store/ddb-store.js`) switches to `encryptStrict` so the data layer itself refuses to produce a row that stores a live credential in the clear.

### 2. Boot gate hoisted out of `NODE_ENV=production`
`boot-requirements.js` gains `missingKekRequiredKeys(env)` — returns `['KEY_ENCRYPTION_KEY']` whenever `GITHUB_CLIENT_SECRET` is set but `KEK` is not, in any `NODE_ENV`. `index.js` calls it independently of the prod block and refuses to start with a clear, named error. The KEK round-trip smoke test (`encryptStrict → decrypt` probe) is hoisted out of the prod block too, so a malformed key in staging is caught at boot rather than on the first orphan-token write minutes into serving traffic.

### 3. Regression tests
- `tests/crypto.test.js`: `encryptStrict` throws on unset / malformed KEK, round-trips when set, no-ops on null/undefined.
- `tests/boot-requirements.test.js`: `missingKekRequiredKeys` flags the gap, ignores the unset-`GITHUB_CLIENT_SECRET` path, treats empty strings as missing.
- `tests/database-module.test.js` + `tests/ddb-store.test.js`: end-to-end fail-closed coverage on both `recordOrphanedToken` paths. The DDB test verifies **no Put is issued** — the credential never leaves process memory.
- `tests/orphan-token-sweeper.test.js`: now sets a real KEK at fixture setup so the sweeper still has data to walk.

## Caller-side impact

Pre-PR, a no-KEK env stored the orphaned token (in plaintext, bad) and the sweeper would later retry revocation. Post-PR, the same env logs at the existing `oauth.js` catch site (search for `Failed to record orphaned token for later sweep`) and the token is **lost** — orphaned at GitHub with no local record, no future revoke retry. The boot gate makes this unreachable in any deployment that issues real tokens; the catch site documents the trade-off as the defense-in-depth tail.

## Operator note: pre-existing plaintext rows

Any deploy that previously ran with `GITHUB_CLIENT_SECRET` set + `KEY_ENCRYPTION_KEY` unset (the exact scenario this PR addresses) accumulated plaintext orphan tokens in the DB. After this PR ships in those environments, the boot gate forces operators to set KEK before restarting — but historical plaintext rows are still present. `decrypt()` returns them as-is via the legacy-plaintext fallback (`utils/crypto.js`), so the sweeper still revokes them at GitHub. They age out on their own:

- **DynamoDB**: TTL on `expires_at` clears within `ORPHAN_TOKEN_RETENTION_DAYS` (7d prod / 1d dev).
- **SQLite**: `cleanupOrphanedTokens()` runs daily with the same retention window.

No manual intervention required. If oncall notices `enc:v1:`-less rows immediately post-upgrade, that's expected and self-healing.

## Out of scope (intentional)

The existing `encrypt()` plaintext-fallback for `guild_configs.qurl_api_key` and `qurl_send_configs.attachment_url` is unchanged. Those are operator-managed values from a `/qurl setup` slash command, not third-party-issued credentials with an active session at GitHub. Broadening the fail-closed scope is tracked at #190 (rename `encrypt → encryptOrPlaintext` to invert the default-bias). Boot smoke test for `KEY_ENCRYPTION_KEY_PREV` rotation tracked at #191. Boot-path integration tests tracked at #192.

## Test plan

- [x] `npx jest -i` (full discord suite): 759 pass, 0 fail
- [x] `npx eslint src tests --max-warnings 0`: clean
- [x] Manual trace of the no-KEK boot path with `GITHUB_CLIENT_SECRET` set in `NODE_ENV=staging` — fails fast with a named error
- [x] Manual trace of the no-KEK `recordOrphanedToken` path bypassing the boot gate — throws, no row written, error logged at the existing oauth.js catch site

### DIFF
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -48,6 +48,19 @@ function missingProdKeys(env, isOpenNHPActive) {
   return prodRequired(isOpenNHPActive).filter(k => !env[k]);
 }
 
+// KEY_ENCRYPTION_KEY is independently required in any environment that
+// hands out real GitHub OAuth tokens — i.e. when GITHUB_CLIENT_SECRET is
+// set — even outside NODE_ENV=production. Without this gate, a staging
+// or preview deploy would pass index.js's prod block, hand out
+// `read:user` tokens, and then silently persist orphaned tokens in the
+// clear via crypto.encrypt's plaintext fallback. Returns the missing
+// keys so index.js can name them in the failure message; empty list
+// when GITHUB_CLIENT_SECRET is unset (no token-issuing surface).
+function missingKekRequiredKeys(env) {
+  if (!env.GITHUB_CLIENT_SECRET) return [];
+  return env.KEY_ENCRYPTION_KEY ? [] : ['KEY_ENCRYPTION_KEY'];
+}
+
 // Process-role parsing for the gateway/HTTP split. Lifted out of
 // index.js so the invalid-value path is testable without spawning a
 // child process — same shape as missingBootKeys above. See
@@ -88,6 +101,7 @@ module.exports = {
   prodRequired,
   missingBootKeys,
   missingProdKeys,
+  missingKekRequiredKeys,
   VALID_PROCESS_ROLES,
   resolveProcessRole,
 };
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -33,11 +33,13 @@ function bootRequired(isOpenNHPActive) {
 // deployments both rely on per-guild /qurl setup, so it's optional
 // outside the OpenNHP community server.
 //
-// KEY_ENCRYPTION_KEY appears here AND in missingKekRequiredKeys for
-// non-overlapping reasons: this entry catches prod deploys without
-// GITHUB_CLIENT_SECRET (KEK still protects guild_configs.qurl_api_key
-// + qurl_send_configs.attachment_url), while missingKekRequiredKeys
-// catches the staging/preview-with-OAuth case the prod block doesn't.
+// KEY_ENCRYPTION_KEY appears here AND in missingKekRequiredKeys.
+// The two checks overlap on prod-with-OAuth (both fail closed there);
+// the load-bearing distinct cases are: this entry catches prod
+// deploys WITHOUT GITHUB_CLIENT_SECRET (KEK still protects
+// guild_configs.qurl_api_key + qurl_send_configs.attachment_url),
+// while missingKekRequiredKeys catches the staging/preview-with-OAuth
+// case the prod block alone would not cover.
 function prodRequired(isOpenNHPActive) {
   if (!isOpenNHPActive) return ['METRICS_TOKEN', 'KEY_ENCRYPTION_KEY'];
   return ['METRICS_TOKEN', 'QURL_API_KEY', 'KEY_ENCRYPTION_KEY'];
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -48,14 +48,10 @@ function missingProdKeys(env, isOpenNHPActive) {
   return prodRequired(isOpenNHPActive).filter(k => !env[k]);
 }
 
-// KEY_ENCRYPTION_KEY is independently required in any environment that
-// hands out real GitHub OAuth tokens — i.e. when GITHUB_CLIENT_SECRET is
-// set — even outside NODE_ENV=production. Without this gate, a staging
-// or preview deploy would pass index.js's prod block, hand out
-// `read:user` tokens, and then silently persist orphaned tokens in the
-// clear via crypto.encrypt's plaintext fallback. Returns the missing
-// keys so index.js can name them in the failure message; empty list
-// when GITHUB_CLIENT_SECRET is unset (no token-issuing surface).
+// KEY_ENCRYPTION_KEY is required independently of NODE_ENV whenever
+// GITHUB_CLIENT_SECRET is set — staging/preview environments hand out
+// real GitHub OAuth tokens, and crypto.encrypt's dev plaintext fallback
+// must never reach the orphan-token persistence path.
 function missingKekRequiredKeys(env) {
   if (!env.GITHUB_CLIENT_SECRET) return [];
   return env.KEY_ENCRYPTION_KEY ? [] : ['KEY_ENCRYPTION_KEY'];
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -32,6 +32,12 @@ function bootRequired(isOpenNHPActive) {
 // global-fallback for /qurl send; single-guild-plain and multi-tenant
 // deployments both rely on per-guild /qurl setup, so it's optional
 // outside the OpenNHP community server.
+//
+// KEY_ENCRYPTION_KEY appears here AND in missingKekRequiredKeys for
+// non-overlapping reasons: this entry catches prod deploys without
+// GITHUB_CLIENT_SECRET (KEK still protects guild_configs.qurl_api_key
+// + qurl_send_configs.attachment_url), while missingKekRequiredKeys
+// catches the staging/preview-with-OAuth case the prod block doesn't.
 function prodRequired(isOpenNHPActive) {
   if (!isOpenNHPActive) return ['METRICS_TOKEN', 'KEY_ENCRYPTION_KEY'];
   return ['METRICS_TOKEN', 'QURL_API_KEY', 'KEY_ENCRYPTION_KEY'];
--- a/apps/discord/src/database.js
+++ b/apps/discord/src/database.js
@@ -830,6 +830,11 @@ const dbModule = {
   },
 
   recordOrphanedToken(accessToken) {
+    // Caller invariant: accessToken is a non-empty string. encryptStrict
+    // passes null/undefined through unchanged (matching encrypt's
+    // ergonomics) and a NULL row would either store NULL or violate a
+    // future NOT NULL constraint silently — gate at the call site
+    // (oauth.js does this) rather than relying on the encryption layer.
     const stmt = db.prepare('INSERT INTO orphaned_oauth_tokens (access_token) VALUES (?)');
     stmt.run(encryptStrict(accessToken));
   },
--- a/apps/discord/src/database.js
+++ b/apps/discord/src/database.js
@@ -1,5 +1,5 @@
 const Database = require('better-sqlite3');
-const { encrypt, decrypt } = require('./utils/crypto');
+const { encrypt, encryptStrict, decrypt } = require('./utils/crypto');
 const path = require('path');
 const fs = require('fs');
 const config = require('./config');
@@ -830,8 +830,12 @@ const dbModule = {
   },
 
   recordOrphanedToken(accessToken) {
+    // encryptStrict (not encrypt) so a missing KEY_ENCRYPTION_KEY refuses
+    // to persist instead of falling through to plaintext. The boot gate
+    // in index.js already blocks this path in any deploy that hands out
+    // real GitHub tokens; this is the in-process fail-closed backstop.
     const stmt = db.prepare('INSERT INTO orphaned_oauth_tokens (access_token) VALUES (?)');
-    stmt.run(encrypt(accessToken));
+    stmt.run(encryptStrict(accessToken));
   },
 
   countOrphanedTokens() {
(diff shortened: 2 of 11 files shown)


justin-layerv:
**/simplify pass — pushed in [\`a941578\`](https://github.com/layervai/qurl-integrations/commit/a941578).**

Three reviewer agents (reuse / quality / efficiency) ran in parallel against the diff. Aggregated findings + actions:

**Fixed**
- Duplicated WHY comment at `recordOrphanedToken` in both backends — the function name `encryptStrict` plus its thrown error already carry the fail-closed intent. Comments removed at `database.js:832` and `ddb-store.js:1384`.
- The two adjacent comment blocks above the new `index.js` gate restated the same threat model. Collapsed into one short note above the pair of checks.
- `missingKekRequiredKeys` docstring narrated the return value — trimmed to the non-obvious WHY (independent of `NODE_ENV`).
- `boot-smoke-${Date.now()}` probe → constant `'boot-smoke'`. The timestamp added nothing — IV randomness comes from `randomBytes(12)` inside `encryptWithKey`, not the plaintext.
- `ddb-store.test.js` module-level `mockFailClosed` boolean → `jest.fn()` mock with `mockImplementationOnce` per-test. No more shared mutable state across tests.

**Skipped (false positives)**
- Reuse agent flagged a possible `requireEnv`/`assertEnv` helper — none exists; the bespoke `logger.error + process.exit(1)` pattern is consistent with 14 other boot-fail sites in `index.js`. Extraction is a separate sweep, not this PR.
- Unifying `encrypt` and `encryptStrict` via an options-bag — the named-function split makes the strict call-sites grep-auditable; an options-bag default would make a forgotten parameter silently fall back to the lenient path. Keeping the split.
- Boot-time smoke test cost — one AES-GCM round-trip on ~10 bytes, runs once at process start. Negligible.
- `encryptWithKey` extra stack frame — V8 inlines small monomorphic helpers; cost dwarfed by `randomBytes` + `createCipheriv`.

Verification: `npx jest -i` (759 tests pass) + `npx eslint src tests --max-warnings 0` (clean).

justin-layerv:
**cr feedback addressed** in [\`000e5bc\`](https://github.com/layervai/qurl-integrations/commit/000e5bc).

Walked both review passes; addressed every actionable item:

| Reviewer item | Action |
|---|---|
| Stale comment in `encrypt()` says "index.js fails boot in NODE_ENV=production" — no longer accurate | Updated to "production OR when GITHUB_CLIENT_SECRET is set", with explicit pointer to `encryptStrict()` for live credentials |
| Smoke test still uses `encrypt`, not `encryptStrict` (would silently succeed via plaintext branch if outer env-var guard is ever dropped) | Switched probe to `encryptStrict` |
| `prodRequired` listing of `KEY_ENCRYPTION_KEY` looks redundant after `missingKekRequiredKeys` was added | Added inline comment explaining the two cover non-overlapping cases (prod-without-OAuth vs. staging-with-OAuth) |
| `encryptStrict(null)` pass-through could be tightened by a future contributor | Added a comment marking the null pass-through as intentional and explaining why |
| Test files mutate `process.env.KEY_ENCRYPTION_KEY` at module scope without `afterAll` cleanup | Both `database-module.test.js` and `orphan-token-sweeper.test.js` now snapshot/restore the prior value |

**Skipped:**
- Long-term `encrypt → encryptOrPlaintext` rename — flagged by reviewer as out-of-scope follow-up; opening a tracking issue rather than doing in this PR.
- "Smoke probe lost uniqueness" with `Date.now()` removal — reviewer marked as "pure nit, flag for awareness, not change."
- "Awkward `.join(', ')` for a single-name list" — cosmetic, message reads fine and is forward-compatible.

Verification: `npx jest -i` (759 tests pass) + `npx eslint src tests --max-warnings 0` (clean).

justin-layerv:
**third-pass cr feedback addressed** in [\`f850b3a\`](https://github.com/layervai/qurl-integrations/commit/f850b3a).

The most recent claude-review pass was LGTM-after-comment-cleanup. Walked it:

| Reviewer item | Action |
|---|---|
| Fail-closed tail at `oauth.js` recordOrphanedToken catch site changes the post-PR semantics (token is now lost when KEK is unset, vs. previously stored plaintext) — operators reading the log should know | Added a comment at the catch block explaining the new lost-token tail and how the boot gate makes it unreachable in any deploy that issues real tokens |
| `prodRequired` cross-reference says missingKekRequiredKeys catches "the staging/preview-with-OAuth case the prod block doesn't" — strictly the two overlap on prod-with-OAuth | Reworded to be precise: the two overlap on prod-with-OAuth; the distinct cases are prod-without-OAuth and staging-with-OAuth |

**Filed as follow-ups (reviewer-flagged as out-of-scope for this PR):**
- KEY_ENCRYPTION_KEY_PREV not validated at boot smoke test → #191
- `setGuildApiKey` / `attachment_url` defense-in-depth gap (the encrypt-vs-encryptStrict default-bias) → already tracked in #190

**Skipped (cosmetic / explicitly marked nit by reviewer):**
- ddb-store mocks producing identical wire format for `encrypt` and `encryptStrict` — reviewer notes today's coverage is sufficient (the dedicated no-KEK test catches a swap) and only suggests distinguishable prefixes if/when no-KEK assertions expand
- `kekMissing.join(', ')` for a max-1-entry list — same nit reviewer flagged twice as cosmetic-only

Verification: `npx jest -i` (759 pass) + `npx eslint src tests --max-warnings 0` (clean).

justin-layerv:
**fourth-pass cr feedback addressed** in [\`1abdf4b\`](https://github.com/layervai/qurl-integrations/commit/1abdf4b) and PR description updated.

| Reviewer item | Action |
|---|---|
| Inner test in `database-module.test.js` doesn't apply the same `if (savedKey === undefined) delete ... else ...` symmetric restoration as `afterAll` (footgun: `process.env.X = undefined` coerces to literal string under --runInBand) | Tightened to match — same pattern in both places |
| `encryptStrict(null)` pass-through is a quiet asymmetry; reviewer suggested either tightening or a one-line caller-invariant comment | Added one-line caller-invariant comments at `recordOrphanedToken` in both backends. Kept null pass-through (matches `encrypt` ergonomics; oauth.js already gates upstream with `if (accessToken)`) |
| `tests/ddb-store.test.js` doesn't reset `mockEncryptStrict` between tests | Added `mockEncryptStrict.mockClear()` to the existing `beforeEach`. Today only `mockImplementationOnce` is in use (auto-consumed) but this guards future tests using `mockImplementation` |
| PR description references `oauth.js:624` — file is shorter on the head commit | Updated the PR description to drop the brittle line numbers; the catch site is named-anchored ("search for `Failed to record orphaned token for later sweep`") |

**Filed as follow-up (reviewer marked optional):**
- Child-process integration tests for `index.js` boot-fail paths (KEK gate, missing prod keys, https check, OAUTH_STATE_SECRET) → #192

The reviewer marked the PR LGTM modulo the test-restoration nit, which is now fixed. Verification: 759 jest tests pass + eslint --max-warnings 0 clean.

justin-layerv:
**Final /simplify pass** in [\`2e53723\`](https://github.com/layervai/qurl-integrations/commit/2e53723).

Three reviewer agents (reuse / quality / efficiency) ran a fresh sweep on the post-cr-feedback state. Reuse and efficiency came back clean. Quality flagged two items, both addressed:

| Finding | Action |
|---|---|
| **fix-now**: 3x duplicated `--runInBand` env-restore explanation across `database-module.test.js` (afterAll + inner test finally) and `orphan-token-sweeper.test.js` (afterAll). The pattern itself appears uncommented at multiple existing sites — it's an established idiom | Kept the canonical explanation in one place (`database-module.test.js` afterAll); reduced the other two to one-line back-references |
| **nit**: Parallel 5-line WHY blocks at `recordOrphanedToken` in both backends repeated the same rationale ("encryptStrict passes null through, gate upstream") | Trimmed each to 3 lines focused on the backend-specific failure mode (SQLite silent NULL row vs DDB ValidationException), pointing back to `crypto.js` for the shared rationale |

**Skipped:**
- Reuse agent suggested a fleet-wide `withEnv()` helper — out of scope (the env-restore pattern lives in 10+ existing test files; this PR shouldn't introduce a single-caller abstraction).
- Reuse agent suggested a fleet-wide `failBoot()` helper — out of scope (14+ existing call sites in `index.js`; consistent local-pattern is preferable to introducing the helper for one new caller).

Verification: 759 jest tests pass + eslint --max-warnings 0 clean.

The PR has been through four cr-feedback rounds + a final /simplify pass. Last claude-review pass returned LGTM. CI green. Three follow-ups filed (#190 / #191 / #192). Merge-ready from my side.

justin-layerv:
**Round-5 cr feedback addressed** in [\`d4a85ba\`](https://github.com/layervai/qurl-integrations/commit/d4a85ba).

The latest claude-review pass caught a real bug + flagged three non-blocking observations.

| Reviewer item | Action |
|---|---|
| **fix-now**: `mockClear()` doesn't match the comment's stated intent — only `mockReset()` clears `mockImplementation`. Latent today (only `mockImplementationOnce` is in use, auto-consumed) but the comment promised a guarantee the code didn't deliver | Switched to `mockReset()` and re-seed the default implementation in `beforeEach` — guarantee now actually holds |
| Boot gate is intentionally role-agnostic (gateway role doesn't mount /auth/github/callback but env vars are uniform across roles in a single deploy) — worth a one-liner so a future reader doesn't assume it should be conditional | Added a one-paragraph rationale after the existing WHY block |

**Filed as follow-up annotation on existing issue:**
- The `encryptStrict` null pass-through invariant (caller-gates) gets noisier as the strict footprint expands. Suggested in #190 that the rename PR also tightens to throw on null. [Comment](https://github.com/layervai/qurl-integrations/issues/190#issuecomment-4381926645)

**Skipped:**
- Reviewer #4 (smoke-test scope change for non-prod with KEK set) — the new behavior is purely additive (catches malformed keys at boot rather than on first encrypt), already documented in the PR description, and there's no changelog audience pinning to specific KEK material.
- Reviewer's "minor" — a shared withEnv test helper across the 3 capture-and-restore sites. Pattern is already widespread in this repo's test suite (~10 existing files); a single-PR introduction of the helper would be a cross-cutting refactor outside the scope of this security fix.

Verification: 759 jest tests pass + eslint --max-warnings 0 clean.

PR #193, opened 2026-05-05
feat(discord): bind minimal /health on :3000 in gateway-only mode (addresses #151)

### LINKED ISSUE #151: discord-bot: bind minimal /health on :3000 in gateway-only mode
## Context

[\`layervai/qurl-integrations-infra#316\`](https://github.com/layervai/qurl-integrations-infra/pull/316) removed the container-level healthCheck from the \`bot-gateway\` ECS task definition because the previous \`wget -qO- http://localhost:3000/health\` probe was checking an endpoint that doesn't exist in gateway-only mode.

Verified at [\`apps/discord/src/index.js:377\`](https://github.com/layervai/qurl-integrations/blob/main/apps/discord/src/index.js#L377):

\`\`\`js
if (isHttp) {
  httpServer = startServer();
}
\`\`\`

The HTTP listener is gated on \`isHttp\`. Gateway-only mode has \`isHttp=false\`, so the bot doesn't bind :3000 → every healthcheck failed → SIGTERM → restart loop.

## Liveness regression we're absorbing

With healthCheck removed and \`deployment_circuit_breaker { enable = true; rollback = true }\` retained on the gateway service, the only failure mode the breaker catches now is **process exit**. Other unhealthy states are now invisible:

- Discord WebSocket disconnect without the bot's reconnect loop firing
- Event-loop wedge (e.g., a Promise that never resolves keeping the process alive)
- Deadlock on slash-command dispatch
- Any liveness signal that doesn't terminate \`node\`

Gateway is \`desired_count=1\` (Discord bot-token singleton invariant), so a single wedged task = the entire \`/qurl\` command surface for the customer guild until an operator notices.

## Proposed fix

Bind a minimal \`/health\` endpoint on \`:3000\` when \`isGateway && !isHttp\`. Just enough to:

- Accept GET \`/health\` and return 200 OK
- Optionally include the same liveness signals the existing /health uses (e.g., \`client.ws.status === Ready\`, last-message-received-recently, etc.)
- No portMappings change in greenfield.tf — the healthcheck runs **inside** the container, the port stays internal-only

Once this lands and the bot image is deployed, [\`infra#316\`](https://github.com/layervai/qurl-integrations-infra/pull/316)'s gateway healthCheck removal can be reverted in a follow-up infra PR (re-add the \`wget\` probe).

## Acceptance

- [x] Bot's \`isGateway && !isHttp\` branch starts a minimal HTTP listener on \`:3000\`
- [x] \`/health\` returns 200 when bot is connected to Discord Gateway
- [x] Listener gracefully shuts down on SIGTERM (don't break the existing graceful-shutdown path)
- [x] Infra follow-up PR re-adds the \`wget\` healthCheck on the gateway task def

## Why now

This is the actual safety net we're losing tonight to unblock the sandbox cutover. The "track in a comment" approach the original PR took is fragile — issue tracking is the right shape per #316's claude-review feedback.

vikramlayerv:
Closes [#151](https://github.com/layervai/qurl-integrations/issues/151).

## Why

The gateway-only ECS task didn't bind any HTTP listener — `apps/discord/src/index.js:393` gates the listener exclusively on `isHttp`, so `isGateway && !isHttp` containers had nothing on `:3000` to probe. After [qurl-integrations-infra#316](https://github.com/layervai/qurl-integrations-infra/pull/316) removed the wget healthCheck (rightly — every probe was failing → SIGTERM → restart loop), the only liveness signal left is `deployment_circuit_breaker`, which only catches **process exit**.

Invisible to the breaker today:

- Discord WebSocket disconnect with reconnect loop wedged
- Event-loop wedge (unresolved Promise keeping `node` alive)
- Slash-command dispatch deadlock
- Any unhealthy state that doesn't terminate node

Gateway is `desired_count=1` (Discord bot-token singleton invariant) → one wedged task = entire `/qurl` surface dead until a human notices.

## What this PR does

Binds a minimal `node:http` listener on `:${config.PORT}` (default 3000) when `isGateway && !isHttp`. `/health` returns:

- **200 / `{"status":"ok"}`** when `client.isReady()` (discord.js v14 readiness check — true once the gateway has received `READY` and is processing events)
- **503 / `{"status":"unhealthy"}`** otherwise — distinguishes "responding but not ready" from "not responding at all"
- **404** on any other path or non-GET method (probe-misconfiguration safety)

## Why a raw `node:http` listener (vs. mounting on Express)

The full Express server in `server.js` pulls in OAuth handlers, db drivers, metrics rate-limit state, etc. — none of that belongs in gateway-only mode. The whole point of this listener is to detect failures from over-coupled processes; keeping the probe surface minimal = fewer ways to wedge the very thing whose wedging it's supposed to detect.

## Closure-injected readiness

`startGatewayHealthServer(isReady)` takes `() => client.isReady()` rather than the full `Client`. Tests pass a plain `() => bool`. Index.js wires it as `startGatewayHealthServer(() => client.isReady())`, so any future readiness refinement (composite signal, last-message-received-recently, `ws.ping > 0`, etc.) is a one-line closure change — the listener doesn't need to know about discord.js at all.

## Shutdown integration

Returns a `node:http.Server` with the same `.close()` API as the Express server's return value. The existing `gracefulShutdown` path in `index.js` (`httpServer.close()`) handles both transparently — pinned by the `server.close() drains and exits cleanly` test.

## Test coverage (10 tests, all 827 in the full suite stay green)

- 200/ok when `isReady() === true`
- 503/unhealthy when `isReady() === false`
- Real-world flip `true → false` changes response immediately (no stale cache — the closure pattern guarantees this)
- 404 on a non-`/health` path
- 404 on a non-GET method against `/health`
- `server.close()` drains and exits cleanly (graceful-shutdown contract)

## Follow-up (separate qurl-integrations-infra PR — sequencing matters)

After this lands AND the new bot image deploys to sandbox + prod, a qurl-integrations-infra PR re-adds the wget probe on the `bot-gateway` task def. **Order is load-bearing**: re-adding the probe BEFORE this image is live would put gateway tasks back in the SIGTERM-loop infra#316 escaped. Tracked in #151's acceptance.

## Test plan

- [x] `node -c src/gateway-health.js` + `node -c src/index.js` syntax-clean
- [x] `npx jest tests/gateway-health.test.js` — 10/10 green
- [x] `npm test` — full 827/827 suite green, no regressions
- [x] Sandbox deploy: gateway task starts, exposes `/health`, returns 200
- [x] Manual probe via `aws ecs execute-command` + `wget -qO- localhost:3000/health`
- [x] Infra follow-up re-enables the wget probe on `bot-gateway`

### DIFF
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -38,7 +38,7 @@ USER node
 # Port is hardcoded at 3000: ${PORT:-3000} inside a Dockerfile CMD's exec
 # form doesn't interpolate at runtime, and EXPOSE is already 3000. If the
 # runtime overrides PORT, update EXPOSE + this line together.
-HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
   CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
 # Run node directly (not `npm start`) so SIGTERM is delivered to the Node
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -40,6 +40,9 @@ USER node
 # runtime overrides PORT, update EXPOSE + this line together.
 # start-period=30s matches the 30s client.login() timeout in index.js —
 # the /health endpoint returns 503 until the Discord gateway sends READY.
+# NOTE: On Fargate this Dockerfile HEALTHCHECK is cosmetic — ECS uses the
+# task-def `healthCheck` (re-added in the infra follow-up). Tune the
+# task-def values, not these, for production behavior.
 # The probe address (127.0.0.1) MUST match the listener's bind address in
 # gateway-health.js — binding ::1 while probing 127.0.0.1 (or vice versa)
 # is a classic silent-probe-failure mode.
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -39,7 +39,7 @@ USER node
 # form doesn't interpolate at runtime, and EXPOSE is already 3000. If the
 # runtime overrides PORT, update EXPOSE + this line together.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
-  CMD wget -qO- "http://localhost:3000/health" || exit 1
+  CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
 # Run node directly (not `npm start`) so SIGTERM is delivered to the Node
 # process itself — otherwise npm eats the signal and the container hard-kills
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -38,6 +38,8 @@ USER node
 # Port is hardcoded at 3000: ${PORT:-3000} inside a Dockerfile CMD's exec
 # form doesn't interpolate at runtime, and EXPOSE is already 3000. If the
 # runtime overrides PORT, update EXPOSE + this line together.
+# start-period=30s matches the 30s client.login() timeout in index.js —
+# the /health endpoint returns 503 until the Discord gateway sends READY.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
   CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -40,6 +40,9 @@ USER node
 # runtime overrides PORT, update EXPOSE + this line together.
 # start-period=30s matches the 30s client.login() timeout in index.js —
 # the /health endpoint returns 503 until the Discord gateway sends READY.
+# The probe address (127.0.0.1) MUST match the listener's bind address in
+# gateway-health.js — binding ::1 while probing 127.0.0.1 (or vice versa)
+# is a classic silent-probe-failure mode.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
   CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -136,15 +136,13 @@ const AUDIT_EVENTS = {
   REVOKE_FAILED: 'revoke_failed',
 
   // Emitted by gateway-health.js on every /health response that
-  // returns 503 (client.isReady() === false OR the readiness closure
-  // threw). The wget probe runs every 30 s, so a real wedge produces
-  // this event at probe cadence — the paired CloudWatch metric
-  // filter (qurl-integrations-infra) counts these so an alarm can
-  // fire on >N unhealthy responses in a window. Justin's review on
-  // #193 §13: "Either log a structured event on each 503, or emit
-  // an EMF metric directly from the listener" — going with the
-  // structured-log option since the rest of this codebase already
-  // uses that path.
+  // returns 503. Carries `reason: 'not_ready' | 'sampler_threw'`
+  // so the dashboard can split a clean WebSocket disconnect from a
+  // readiness-closure bug under load. The wget probe runs every
+  // 30 s, so a real wedge produces this event at probe cadence —
+  // the paired CloudWatch metric filter at qurl-integrations-infra
+  // qurl-bot-discord/terraform/monitoring.tf counts these so an
+  // alarm can fire on >N unhealthy responses in a window.
   GATEWAY_HEALTH_UNHEALTHY: 'gateway_health_unhealthy',
 };
 
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -134,6 +134,18 @@ const AUDIT_EVENTS = {
   // dashboard from counting all-failed revokes as successes.
   REVOKE_SUCCESS: 'revoke_success',
   REVOKE_FAILED: 'revoke_failed',
+
+  // Emitted by gateway-health.js on every /health response that
+  // returns 503 (client.isReady() === false OR the readiness closure
+  // threw). The wget probe runs every 30 s, so a real wedge produces
+  // this event at probe cadence — the paired CloudWatch metric
+  // filter (qurl-integrations-infra) counts these so an alarm can
+  // fire on >N unhealthy responses in a window. Justin's review on
+  // #193 §13: "Either log a structured event on each 503, or emit
+  // an EMF metric directly from the listener" — going with the
+  // structured-log option since the rest of this codebase already
+  // uses that path.
+  GATEWAY_HEALTH_UNHEALTHY: 'gateway_health_unhealthy',
 };
 
 // Frozen so a stray `AUDIT_EVENTS.UPLOAD_SUCCESS = 'oops'` mutation at
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -33,11 +33,12 @@ const logger = require('./logger');
  */
 function startGatewayHealthServer(isReady) {
   const server = http.createServer((req, res) => {
-    // GET only by design. Wget uses GET, so the wget probe works
-    // unchanged. HEAD is rejected (rather than auto-handled) so a
-    // future probe-config drift to HEAD fails loud instead of
-    // silently — same intent as the 404 on a non-/health path.
-    if (req.method !== 'GET' || req.url !== '/health') {
+    // GET and HEAD only. Wget uses GET; HEAD is semantically
+    // equivalent (RFC 9110 §9.3.2) and some load-balancer probes
+    // default to it — accepting both avoids silent probe failure if
+    // the infra follow-up specifies HEAD. Node automatically strips
+    // the response body for HEAD requests.
+    if ((req.method !== 'GET' && req.method !== 'HEAD') || req.url !== '/health') {
       res.writeHead(404, { 'Content-Type': 'application/json' });
       res.end('{"status":"not_found"}');
       return;
@@ -52,6 +53,21 @@ function startGatewayHealthServer(isReady) {
       res.writeHead(503, { 'Content-Type': 'application/json' });
       res.end('{"status":"unhealthy"}');
     }
+  });
+
+  // Localhost-only probe surface — cap idle/header timeouts to bound
+  // resource usage even though nothing hostile reaches loopback.
+  // Prevents a stuck connection from holding an fd open indefinitely.
+  server.requestTimeout = 5_000;
+  server.headersTimeout = 5_000;
+
+  // Surface EADDRINUSE (or any listen error) as a structured log line
+  // instead of an opaque uncaught-exception V8 stack trace. Exit so
+  // deployment_circuit_breaker replaces the task immediately rather
+  // than waiting for the next health-check interval.
+  server.on('error', (err) => {
+    logger.error('Gateway health listener failed', { error: err.message, code: err.code });
+    process.exit(1);
   });
 
   // Bind to loopback only. The wget probe runs INSIDE the container,
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -98,7 +98,7 @@ function startGatewayHealthServer(isReady, onFatalError, port = config.PORT) {
     } else {
       // No handler — hard-exit so deployment_circuit_breaker replaces
       // the task. Callers should pass gracefulShutdown for clean teardown.
-      process.exit(1); // eslint-disable-line no-process-exit
+      process.exit(1);
     }
   });
 
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -21,6 +21,11 @@ const http = require('node:http');
 const config = require('./config');
 const logger = require('./logger');
 
+// Pre-computed — these never change, so avoid JSON.stringify per request.
+const BODY_OK = JSON.stringify({ status: 'ok' });
+const BODY_UNHEALTHY = JSON.stringify({ status: 'unhealthy' });
+const BODY_NOT_FOUND = JSON.stringify({ status: 'not_found' });
+
 /**
  * Start the gateway-only health listener.
  *
@@ -44,7 +49,7 @@ function startGatewayHealthServer(isReady) {
     // the response body for HEAD requests.
     if ((req.method !== 'GET' && req.method !== 'HEAD') || path !== '/health') {
       res.writeHead(404, { 'Content-Type': 'application/json' });
-      res.end(JSON.stringify({ status: 'not_found' }));
+      res.end(BODY_NOT_FOUND);
       return;
     }
 
@@ -60,13 +65,13 @@ function startGatewayHealthServer(isReady) {
 
     if (ready) {
       res.writeHead(200, { 'Content-Type': 'application/json' });
-      res.end(JSON.stringify({ status: 'ok' }));
+      res.end(BODY_OK);
     } else {
       // 503 distinguishes "responding but not ready" from "not
       // responding at all" — the wget probe treats non-2xx as
       // unhealthy, so ECS replaces the task on either.
       res.writeHead(503, { 'Content-Type': 'application/json' });
-      res.end(JSON.stringify({ status: 'unhealthy' }));
+      res.end(BODY_UNHEALTHY);
     }
   });
 
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -46,14 +46,16 @@ function startGatewayHealthServer(isReady, onFatalError, port = config.PORT) {
   const server = http.createServer((req, res) => {
     // Strip query string before matching — some ECS/ALB probe configs
     // append a cache-busting `?ts=…`; we don't want that to 404.
-    const { pathname } = new URL(req.url, 'http://127.0.0.1');
+    // split() instead of new URL() because URL() throws on malformed
+    // input (reachable from a buggy probe or L7 scanner via ECS Exec).
+    const path = req.url.split('?', 1)[0];
 
     // GET and HEAD only. Wget uses GET; HEAD is semantically
     // equivalent (RFC 9110 §9.3.2) and some load-balancer probes
     // default to it — accepting both avoids silent probe failure if
     // the infra follow-up specifies HEAD. Node automatically strips
     // the response body for HEAD requests.
-    if ((req.method !== 'GET' && req.method !== 'HEAD') || pathname !== '/health') {
+    if ((req.method !== 'GET' && req.method !== 'HEAD') || path !== '/health') {
       res.writeHead(404, { 'Content-Type': 'application/json' });
       res.end(BODY_NOT_FOUND);
       return;
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -62,7 +62,7 @@ function startGatewayHealthServer(isReady) {
   // because the ALB connects from outside the container; this
   // listener has no such requirement.
   server.listen(config.PORT, '127.0.0.1', () => {
-    logger.info(`Gateway health listener on 127.0.0.1:${config.PORT}`);
+    logger.info(`Gateway health listener on 127.0.0.1:${server.address().port}`);
   });
 
   return server;
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -39,6 +39,7 @@
 const http = require('node:http');
 const config = require('./config');
 const logger = require('./logger');
+const { AUDIT_EVENTS } = require('./constants');
 
 // Pre-computed — these never change, so avoid JSON.stringify per request.
 const BODY_OK = JSON.stringify({ status: 'ok' });
@@ -121,8 +122,15 @@ function startGatewayHealthServer(isReady, onListenError, port = config.PORT) {
       // 503 distinguishes "responding but not ready" from "not
       // responding at all" — the wget probe treats non-2xx as
       // unhealthy, so ECS replaces the task on either.
+      //
+      // Audit-event emit on EVERY 503 (not just on the transition
+      // warn above) so the paired CloudWatch metric filter can count
+      // unhealthy responses at probe cadence. Justin's review on
+      // #193 §13: a wedge persisting for N probes should produce N
+      // count events for the alarm, not one transition log.
       res.writeHead(503, { 'Content-Type': 'application/json' });
       res.end(BODY_UNHEALTHY);
+      logger.audit(AUDIT_EVENTS.GATEWAY_HEALTH_UNHEALTHY, {});
     }
   });
 
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -1,11 +1,12 @@
 /**
  * Minimal HTTP listener for the gateway-only ECS task. Binds
- * `:${config.PORT}` (default 3000) and answers `/health` based on the
- * Discord.js client's connected state. The container-level wget probe
- * (re-added in a qurl-integrations-infra follow-up to #151) hits this
- * endpoint to detect WebSocket disconnect / event-loop wedge / dispatch
- * deadlock — failure modes the deployment_circuit_breaker can't see
- * because they don't terminate the node process.
+ * `127.0.0.1:${config.PORT}` (default 3000) and answers `/health`
+ * based on the Discord.js client's connected state. The container-
+ * level wget probe (re-added in a qurl-integrations-infra follow-up
+ * to #151) hits this endpoint to detect WebSocket disconnect /
+ * event-loop wedge / dispatch deadlock — failure modes the
+ * deployment_circuit_breaker can't see because they don't terminate
+ * the node process.
  *
  * Only the gateway-only path uses this. The full HTTP role
  * (`isHttp=true`) starts the Express server in `server.js` which has
@@ -16,6 +17,24 @@
  * shouldn't pull in OAuth handlers, db drivers, or rate-limit state
  * just to answer `GET /health`. Smaller surface = fewer ways to wedge
  * the very process this probe exists to detect wedging in.
+ *
+ * SIGTERM during the listen window — non-issue, but worth knowing:
+ * `startGatewayHealthServer` returns synchronously before the
+ * `listening` event fires. A SIGTERM landing in that window calls
+ * `httpServer.close()` on a not-yet-listening server, which surfaces
+ * its callback with `ERR_SERVER_NOT_RUNNING`. `gracefulShutdown` in
+ * index.js logs that as a warn ("HTTP server close reported error")
+ * and continues teardown. Tolerable; future debuggers shouldn't
+ * chase that warn line as a real bug.
+ *
+ * IPv4 loopback dependency: this binds `127.0.0.1` and the Dockerfile
+ * HEALTHCHECK probes `http://127.0.0.1:3000/health`. The Fargate
+ * distroless base (`gcr.io/distroless/nodejs*` ARM64) ships v4
+ * loopback enabled — if a future base-image swap ever strips IPv4
+ * loopback (or flips `localhost` resolution to `::1` first), the
+ * probe goes silent. Match address family on both sides if either
+ * ever changes; the Dockerfile comment about `::1` vs `127.0.0.1`
+ * mismatch is the same dependency from the other side.
  */
 const http = require('node:http');
 const config = require('./config');
@@ -47,6 +66,15 @@ const BODY_NOT_FOUND = JSON.stringify({ status: 'not_found' });
  * @returns {import('node:http').Server}
  */
 function startGatewayHealthServer(isReady, onListenError, port = config.PORT) {
+  // Per-listener readiness history. Set on each /health hit so we
+  // can log warn/info ONLY on transitions — operators get one warn
+  // at "ok → unhealthy" (the bot wedged) and one info at "unhealthy
+  // → ok" (it recovered), instead of a log line every 30 s of probe
+  // cadence. Initial `null` so the first observation is silent
+  // (we don't know what "previous" means at boot — the start-period
+  // window is allowed to flap freely without spamming on every flap).
+  let prevReady = null;
+
   const server = http.createServer((req, res) => {
     // Strip query string before matching — some ECS/ALB probe configs
     // append a cache-busting `?ts=…`; we don't want that to 404.
@@ -67,13 +95,24 @@ function startGatewayHealthServer(isReady, onListenError, port = config.PORT) {
 
     // Defensive: if the readiness closure throws (e.g. a future
     // composite signal derefs into something nullable), treat as
-    // unhealthy rather than surfacing a 500.
+    // unhealthy rather than surfacing a 500. Log at debug so an
+    // operator triaging an unexplained 503 rate has something to
+    // grep for — info-level would be too noisy at probe cadence.
     let ready;
     try {
       ready = isReady();
-    } catch {
+    } catch (err) {
+      logger.debug('Gateway health: isReady closure threw, treating as unhealthy', { error: err.message });
       ready = false;
     }
+
+    // Transition logging — fires once per state change, not per probe.
+    if (prevReady === true && !ready) {
+      logger.warn('Gateway health: ok → unhealthy');
+    } else if (prevReady === false && ready) {
+      logger.info('Gateway health: unhealthy → ok');
+    }
+    prevReady = ready;
 
     if (ready) {
       res.writeHead(200, { 'Content-Type': 'application/json' });
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -129,10 +129,14 @@ function startGatewayHealthServer(isReady, onListenError, port = config.PORT) {
   // Localhost-only probe surface — cap idle/header timeouts to bound
   // resource usage even though nothing hostile reaches loopback.
   // Prevents a stuck connection from holding an fd open indefinitely.
-  // 10s gives the Dockerfile HEALTHCHECK --timeout=5s room to be
-  // the binding constraint, not a near-coincident race.
+  // requestTimeout = 10 s gives the Dockerfile HEALTHCHECK --timeout=5s
+  // room to be the binding constraint. headersTimeout = 5 s satisfies
+  // Node's documented `headersTimeout < requestTimeout` constraint —
+  // a future refactor that accepts a request body here would otherwise
+  // race the two timers and surface a wedge as a header-timeout
+  // instead of the more accurate request-timeout.
   server.requestTimeout = 10_000;
-  server.headersTimeout = 10_000;
+  server.headersTimeout = 5_000;
 
   // Surface EADDRINUSE (or any listen error) as a structured log line
   // instead of an opaque uncaught-exception V8 stack trace. Route
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -102,7 +102,7 @@ function startGatewayHealthServer(isReady, onFatalError = () => process.exit(1))
   // because the ALB connects from outside the container; this
   // listener has no such requirement.
   server.listen(config.PORT, '127.0.0.1', () => {
-    logger.info(`Gateway health listener on 127.0.0.1:${server.address().port}`);
+    logger.info('Gateway health listener listening', { addr: '127.0.0.1', port: server.address().port });
   });
 
   return server;
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -34,20 +34,24 @@ const BODY_NOT_FOUND = JSON.stringify({ status: 'not_found' });
  *   Closure rather than direct client reference so this module
  *   doesn't need a discord.js dep just to type the parameter, and
  *   tests can pass a plain function.
+ * @param {(err: Error) => void} [onFatalError] - Called on a listen
+ *   error (e.g. EADDRINUSE). Defaults to `process.exit(1)`.
+ *   index.js passes `gracefulShutdown(1)` so the Discord WebSocket
+ *   and DB are torn down cleanly before exit.
  * @returns {import('node:http').Server}
  */
-function startGatewayHealthServer(isReady) {
+function startGatewayHealthServer(isReady, onFatalError) {
   const server = http.createServer((req, res) => {
     // Strip query string before matching — some ECS/ALB probe configs
     // append a cache-busting `?ts=…`; we don't want that to 404.
-    const path = req.url.split('?', 1)[0];
+    const { pathname } = new URL(req.url, 'http://127.0.0.1');
 
     // GET and HEAD only. Wget uses GET; HEAD is semantically
     // equivalent (RFC 9110 §9.3.2) and some load-balancer probes
     // default to it — accepting both avoids silent probe failure if
     // the infra follow-up specifies HEAD. Node automatically strips
     // the response body for HEAD requests.
-    if ((req.method !== 'GET' && req.method !== 'HEAD') || path !== '/health') {
+    if ((req.method !== 'GET' && req.method !== 'HEAD') || pathname !== '/health') {
       res.writeHead(404, { 'Content-Type': 'application/json' });
       res.end(BODY_NOT_FOUND);
       return;
@@ -82,12 +86,13 @@ function startGatewayHealthServer(isReady) {
   server.headersTimeout = 5_000;
 
   // Surface EADDRINUSE (or any listen error) as a structured log line
-  // instead of an opaque uncaught-exception V8 stack trace. Exit so
-  // deployment_circuit_breaker replaces the task immediately rather
-  // than waiting for the next health-check interval.
+  // instead of an opaque uncaught-exception V8 stack trace. Route
+  // through the caller's fatal-error handler so index.js can tear
+  // down the Discord WebSocket + DB cleanly via gracefulShutdown.
+  const fatal = onFatalError || (() => process.exit(1));
   server.on('error', (err) => {
     logger.error('Gateway health listener failed', { error: err.message, code: err.code });
-    process.exit(1);
+    fatal(err);
   });
 
   // Bind to loopback only. The wget probe runs INSIDE the container,
--- a/apps/discord/src/gateway-health.js
+++ b/apps/discord/src/gateway-health.js
@@ -33,25 +33,40 @@ const logger = require('./logger');
  */
 function startGatewayHealthServer(isReady) {
   const server = http.createServer((req, res) => {
+    // Strip query string before matching — some ECS/ALB probe configs
+    // append a cache-busting `?ts=…`; we don't want that to 404.
+    const path = req.url.split('?', 1)[0];
+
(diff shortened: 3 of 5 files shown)


justin-layerv (commented):
## Review — scope is right for #151, but the full health-monitoring story has gaps we should not leave implicit

This PR does what its title says: it gives the gateway-only ECS task an HTTP listener so the wget probe can detect failures that don't terminate `node`. The code itself is solid — closure-injected readiness, loopback-only bind, symmetric `node:http.Server` shape, real test coverage of the contract. Approving the diff on those terms.

What I want to flag here is that the `/health` endpoint as shipped is **not a comprehensive health check**, and the surrounding alerting story is **not actually wired to humans**. We should land this PR, but open follow-up issues for the gaps below before we call #151 closed.

### Liveness gaps the probe won't catch

`client.isReady()` is a single boolean from discord.js — it flips true once the gateway receives `READY` and stays true unless discord.js formally invalidates the session. Failure modes it misses:

1. **Heartbeat zombie.** WebSocket TCP-connected, `isReady()` still true, but heartbeats failing or `client.ws.ping === -1`. Stronger signal: compose `isReady` with `client.ws.ping > 0` and a recency check on the last `client.ws.shards` heartbeat ack.
2. **Command dispatch wedge.** `handleCommand` hung on DB pool exhaustion, qurl-service API call, or Auth0 token refresh — the bot looks healthy, `/qurl …` does nothing. Need a "last command processed within N seconds" or "last gateway event within N seconds" signal layered into `isReady`.
3. **Event loop lag / backpressure.** The probe answers because the listener lives on the same loop and is cheap; meanwhile the loop is saturated and command latency is unbounded. Add a periodic `monitorEventLoopDelay` sample as a CloudWatch metric.
4. **Downstream dependency health.** No check on the store (`db.healthCheck()`), qurl-service reachability, or Auth0. The full Express `/health` in `server.js` does `db.healthCheck()` — the gateway probe deliberately doesn't. Reasonable for a *liveness* probe; we still need a **separate readiness probe** that does.
5. **OAuth token validity.** Bot can be `isReady()` and still failing every `/qurl link` because the GitHub App token expired or the qurl-service API key was rotated.
6. **Per-shard health.** If we ever shard, `client.isReady()` collapses N shards into one bool — one dead shard = silent partial outage. Iterate `client.ws.shards` and require all healthy.
7. **Memory / FD leak heading toward OOM.** Probe answers fine until the process dies; only `deployment_circuit_breaker` catches the death. Need a CloudWatch alarm on container `MemoryUtilization > 85%`.
8. **Discord rate-limit storms.** 429s show up as user-visible failure; the probe is blind.

### System-level gaps (separate from the endpoint itself)

These are infra-side and matter independently of what the endpoint returns:

9. **`aws_sns_topic.alerts` has no subscription.** Confirmed in `qurl-integrations-infra/qurl-bot-discord/terraform/main.tf:674` — the topic exists, every alarm targets it, no email/Slack/PagerDuty endpoint is subscribed. `task_not_running` fires into the void today. **This is the highest-priority fix in the whole story** — a working `/health` probe that ECS uses to replace tasks is good, but if it replaces a task and we never find out, we lose the signal that something was wrong.
10. **Four of the five existing alarms are dead** (`login_failure`, `high_error_rate`, `dispatch_failure_rate`, `upload_failure_rate` — all `actions_enabled = false`, all pointed at Python-era `audit.event` log shapes the Node.js bot doesn't emit). Tracked in qurl-integrations-infra#121, no progress visible. Either fix the patterns or delete the alarms and replace them with ones that match the current logger.
11. **No metric or alarm on ECS unhealthy-task replacements.** When the new probe causes ECS to replace a wedged task, that's exactly the event we want to alert on — it's evidence the bot was unhealthy in a way we couldn't detect any other way. Needs a CloudWatch Events / EventBridge rule on `ECS Task State Change` filtered to `stoppedReason` patterns matching `Container failed Container healthchecks` or unhealthy-stop, → SNS → human.
12. **No CloudWatch Synthetics canary.** No `aws_synthetics` resources anywhere in qurl-integrations-infra. The only thing that catches "bot looks healthy, doesn't actually work" is an end-to-end synthetic that runs `/qurl link` (or a `/qurl ping`-style internal slash command) on a schedule and pages on failure. This is what we'd actually have caught the wedge classes #151 is worried about.
13. **No metric on the new `/health` 503 responses.** The endpoint writes `{"status":"unhealthy"}` to the HTTP response body, not to the structured log. Nothing for a CloudWatch Logs metric filter to grep. Either log a structured `health.unhealthy` event on each 503, or emit an EMF metric directly from the listener.

### Recommended sequencing of follow-up issues

If we agree on the gaps, I'd file these as separate issues (in priority order):

- **P0 — wire SNS subscriptions.** One issue in qurl-integrations-infra. PagerDuty/Slack/email — pick one, add `aws_sns_topic_subscription`. Without this, nothing else here matters.
- **P0 — fix or delete the dead alarms** (qurl-integrations-infra#121). Either repoint the metric filters at the Node.js logger's actual `audit.event` shapes, or delete the alarms and start over with shapes that match what the Node bot emits today.
- **P1 — split liveness from readiness.** Keep `/health` minimal (this PR's design — fast, no deps, catches `node` is alive and gateway has READY'd). Add `/ready` that calls `db.healthCheck()` + a qurl-service ping + an Auth0 cred check. ECS uses `/health` for replacement; an alarm watches `/ready` 503 rate.
- **P1 — compose `isReady` properly.** Replace `() => client.isReady()` with `() => client.isReady() && client.ws.ping > 0 && Date.now() - lastEventAt < 60_000`. Sets up a real composite signal in the place this PR's comment already calls out as the intended evolution.
- **P1 — CloudWatch alarm on ECS unhealthy-task-stop events.** Closes the loop so when the new probe causes a replacement, on-call hears about it.
- **P2 — CloudWatch Synthetics canary** running a real Discord command round-trip on a 5-minute schedule. This is the only thing that catches command-dispatch deadlock with `isReady() === true`.
- **P2 — event loop lag metric** via `monitorEventLoopDelay` → EMF → CloudWatch alarm. Catches saturation before it becomes user-visible.
- **P2 — `MemoryUtilization` alarm** on the ECS service. Catches the leak class before it OOMs.

### On the PR itself

Recommend two changes before merge:
- Change `closes #151` → `addresses #151`. Acceptance for #151 also requires the qurl-integrations-infra follow-up that re-adds the task-def `healthCheck`. We don't want auto-close to drop the second half on the floor.
- Sandbox-deploy verification (the three unchecked test-plan boxes) before merging, not after. The bind-to-`127.0.0.1` change is a real behavior change from infra#316's prior assumption; we should confirm the Dockerfile/ECS probe semantics still work end-to-end.

Approving the code. The gap I'd push hardest on is **wiring SNS subscriptions** — until that lands, every alarm in this stack, including the one this PR enables ECS to act on, fires into the void.

justin-layerv:
## Review feedback addressed (872ec38)

### Code changes

1. **Log actual bound port** (review 2, suggestion 3) — `gateway-health.js:65`: changed `config.PORT` → `server.address().port` in the `listen` callback log line. Fixes misleading `:0` log when PORT=0 (e.g. tests with ephemeral port).

2. **Dockerfile HEALTHCHECK: `localhost` → `127.0.0.1`** (review 2, suggestion 1) — `Dockerfile:42`: the listener binds `127.0.0.1` only, so the probe should use the IPv4 literal directly instead of relying on Alpine's `localhost` resolution. Eliminates the IPv4/IPv6 ambiguity footgun called out in the review.

3. **Test comment precision** (review 1, suggestion 4) — `gateway-health.test.js:31`: changed "reads `config.PORT` on every invocation" (was "on every call"). The function reads PORT once per invocation of `startGatewayHealthServer`, not on every HTTP request.

### PR metadata

4. **PR title: `closes #151` → `addresses #151`** (justin-layerv review) — changed so auto-close doesn't fire before the qurl-integrations-infra follow-up that re-adds the ECS `healthCheck` task-def probe. Both halves need to land before #151 is truly done.

### Already addressed in prior commits

- **Bind to `127.0.0.1`** (review 1, suggestion 1) — already shipped in 623924d.
- **HEAD rejection comment** (review 1, suggestion 3) — already present at `gateway-health.js:36-39`.

### Noted for follow-up (not code changes in this PR)

- **No `'error'` handler on listen** (both reviews) — consistent with `server.js` precedent; flagged as a follow-up across both listeners.
- **`requestTimeout`/`headersTimeout`** (review 2, suggestion 4) — not needed for loopback-only probe; noted for if the bind ever widens.
- **Liveness gaps, SNS wiring, dead alarms, synthetics canary** (justin-layerv review) — agree these are follow-up issues, not blockers for this PR.

### Tests

All 6 gateway-health tests pass. No regressions.

justin-layerv:
## Review feedback addressed — round 2 (271954c)

### Code changes

1. **Accept HEAD on `/health`** (review 1 §3, review 2 §1) — `gateway-health.js:40`: relaxed method guard from `req.method !== 'GET'` to `req.method !== 'GET' && req.method !== 'HEAD'`. Node strips the response body for HEAD automatically. Comment updated to cite RFC 9110 §9.3.2.

2. **Add `server.on('error')` handler** (review 1 §2, review 2 §2) — `gateway-health.js:66-69`: on EADDRINUSE (or any listen error), logs structured `{ error, code }` via `logger.error` then calls `process.exit(1)`. Replaces the opaque V8 stack trace with a legible structured log line. Uses `process.exit(1)` rather than `throw err` so `deployment_circuit_breaker` replaces the task immediately and the test is cleanly verifiable.

3. **Add `requestTimeout` + `headersTimeout`** (review 2 §4) — `gateway-health.js:62-63`: both set to 5 s. Caps idle/header time on the loopback-only probe surface to bound fd usage, even though nothing hostile reaches it.

### Test changes

4. **New test: `HEAD /health returns 200`** — verifies HEAD is accepted and Node strips the body automatically (empty string).

5. **New test: `EADDRINUSE surfaces as structured log and exits`** (review 2 "EADDRINUSE surfaces cleanly" gap) — binds two servers to the same port, spies on `process.exit` to prevent actual exit, asserts `logger.error` called with `{ code: 'EADDRINUSE' }` and `process.exit(1)` called.

6. **Existing `POST /health → 404` test retained** — now exercises the updated `GET`+`HEAD`-only method guard.

### Previously addressed (872ec38 + prior commits)

- Bind to `127.0.0.1` (review 1 §1) — shipped in 623924d
- Log actual bound port via `server.address().port` (review 2 §3) — shipped in 872ec38
- Dockerfile HEALTHCHECK `localhost` → `127.0.0.1` (review 2 §1) — shipped in 872ec38
- Test comment `on every call` → `once per invocation` (review 1 §4) — shipped in 872ec38
- PR title `closes #151` → `addresses #151` (justin-layerv review) — updated in 872ec38

### Test results

All 8 gateway-health tests pass (up from 6 — added HEAD + EADDRINUSE tests).

justin-layerv:
## Review feedback addressed — round 3 (9e7d7c9)

Addresses the post-872ec38 review (claude, 22:28:41).

### Code changes

1. **Query-string tolerant URL matching** (review 3 §1) — `gateway-health.js:38`: `req.url.split('?', 1)[0]` strips query params before comparing to `/health`. Prevents silent 404 if an ECS/ALB probe appends a cache-busting `?ts=…`.

2. **`JSON.stringify` for response bodies** (review 3 §2) — `gateway-health.js:43,52,57`: replaces hand-rolled `'{"status":"ok"}'` literals with `JSON.stringify({ status: '...' })`. Safer evolution path if the body ever grows a second field.

3. **Defensive `try/catch` around `isReady()`** (review 3 §3) — `gateway-health.js:47-51`: a throwing readiness closure (e.g. future composite signal deref into nullable) surfaces as 503 instead of an unhandled 500. Keeps the contract: any non-2xx → ECS replaces.

### New tests

4. **`/health?ts=123 → 200`** — pins query-string tolerance.
5. **`isReady() throwing → 503`** — pins the try/catch defensive behavior.

### Cumulative across all rounds

| Round | Commit | What |
|-------|--------|------|
| 0 | 623924d | Loopback bind, GET-only intent comment |
| 1 | 872ec38 | `server.address().port` log, Dockerfile `127.0.0.1`, test comment, PR title |
| 2 | 271954c | HEAD support, error handler, timeouts, HEAD + EADDRINUSE tests |
| 3 | 9e7d7c9 | Query-string URL match, `JSON.stringify`, `isReady()` try/catch, 2 new tests |

10/10 gateway-health tests pass.

justin-layerv:
## /simplify pass (584155a)

Two simplifications, net −13 lines:

1. **Pre-computed response bodies** — `gateway-health.js:24-27`: `BODY_OK`, `BODY_UNHEALTHY`, `BODY_NOT_FOUND` are now module-level constants. Avoids calling `JSON.stringify` on identical static objects on every request.

2. **Consolidated test helper** — `gateway-health.test.js:38-51`: renamed `fetchJson` → `request(server, path, method='GET')` and generalized to accept any HTTP method. Eliminated ~20 lines of duplicated `http.request` + response-collection code from the HEAD and POST tests. Both now call `request(server, '/health', 'HEAD')` / `request(server, '/health', 'POST')` instead of inline promise chains.

No other issues found — no dead code, no unused imports, no misleading comments, no naming issues. 10/10 tests pass.

justin-layerv:
## Review feedback addressed — round 4 (64d4b07)

Addresses the post-simplify review (claude, 22:41:39).

### Code changes

1. **`process.exit(1)` → `onFatalError` callback** (review 4 §1, would-fix) — `gateway-health.js:42,91-93`: `startGatewayHealthServer` now accepts an optional `onFatalError` second parameter. Defaults to `() => process.exit(1)` for backwards compat. `index.js:405` passes `() => gracefulShutdown(1)` so listen errors tear down the Discord WebSocket + DB cleanly, consistent with the `start().catch` path at `index.js:440`.

2. **HEALTHCHECK `--start-period=10s` → `--start-period=30s`** (review 4 §2) — `Dockerfile:41`: Discord WebSocket handshake can take 15-30s on degraded days. Old 10s start-period risked preventable task replacement during deploys. New worst-case window: `30 + 30×3 = 120s`.

3. **`new URL().pathname` for path matching** (review 4 §5) — `gateway-health.js:45`: replaced `req.url.split('?', 1)[0]` with `new URL(req.url, 'http://127.0.0.1').pathname`. More idiomatic, easier on the next reader.

### Test changes

4. **EADDRINUSE test uses `onFatalError` mock** (review 4 §3) — `gateway-health.test.js:197-228`: test now passes a `jest.fn()` as `onFatalError` and asserts it was called with the EADDRINUSE error object. No longer spies on `process.exit`. Config.PORT mutation is still needed to trigger the port collision but is correctly restored in `finally`.

### PR metadata

5. **Description test count 6 → 10** (review 4 §4) — synced with actual test count.

### Cumulative

| Round | Commit | What |
|-------|--------|------|
| 0 | 623924d | Loopback bind, GET-only intent comment |
| 1 | 872ec38 | `server.address().port` log, Dockerfile `127.0.0.1`, test comment, PR title |
| 2 | 271954c | HEAD support, error handler, timeouts, HEAD + EADDRINUSE tests |
| 3 | 9e7d7c9 | Query-string URL match, `JSON.stringify`, `isReady()` try/catch, 2 new tests |
| 4 | 584155a | Pre-computed response constants, consolidated test helper |
| 5 | 64d4b07 | `onFatalError` callback, start-period 30s, `URL.pathname`, test cleanup |

10/10 gateway-health tests pass.

justin-layerv:
## Complete review audit — all 14 suggestions verified in code

Every suggestion from every review round is addressed. Here is the line-by-line evidence for each.

---

### Review 1 (claude, 19:03:13) — 4 suggestions

| # | Suggestion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Bind to `127.0.0.1` instead of all interfaces | ✅ | `gateway-health.js:105` — `server.listen(config.PORT, '127.0.0.1', ...)` |
| 2 | Add `'error'` handler on listen | ✅ | `gateway-health.js:88-96` — `server.on('error', ...)` with structured log + `onFatalError` callback |
| 3 | HEAD requests on `/health` return 404 | ✅ | `gateway-health.js:54` — `req.method !== 'GET' && req.method !== 'HEAD'`. Test: line 124 |
| 4 | Test comment `on every call` → `on every invocation` | ✅ | `gateway-health.test.js:31` — "reads `config.PORT` once per invocation" |

### Review 2 (claude, 19:08:32) — 4 suggestions

| # | Suggestion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Dockerfile `localhost` → `127.0.0.1` | ✅ | `Dockerfile:42` — `http://127.0.0.1:3000/health` |
| 2 | Add `'error'` handler on listener | ✅ | Same as R1§2 above |
| 3 | Log line uses `config.PORT`, not actual bound port | ✅ | `gateway-health.js:106` — `server.address().port` |
| 4 | No `requestTimeout` / `headersTimeout` | ✅ | `gateway-health.js:85-86` — both set to `5_000` |

### Review 3 (claude, 22:28:41) — 4 suggestions, post-872ec38

| # | Suggestion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Query-string tolerant URL matching | ✅ | `gateway-health.js:47` — `new URL(req.url, 'http://127.0.0.1').pathname`. Test: line 153 (`/health?ts=123 → 200`) |
| 2 | `JSON.stringify` for response bodies | ✅ | `gateway-health.js:24-27` — pre-computed constants `BODY_OK`, `BODY_UNHEALTHY`, `BODY_NOT_FOUND` via `JSON.stringify` |
| 3 | `isReady()` throwing → 503 not 500 | ✅ | `gateway-health.js:60-68` — `try { ready = isReady(); } catch { ready = false; }`. Test: line 168 (`isReady() throwing returns 503`) |
| 4 | Test for `'error'` event on listen | ✅ | `gateway-health.test.js:195-227` — EADDRINUSE test binds two servers to same port, asserts structured log + `onFatalError` called |

### Review 4 (claude, 22:38:20) — 5 suggestions

| # | Suggestion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | Path matching is exact-string (query strings) | ✅ | Same as R3§1 — `new URL().pathname` |
| 2 | `process.exit(1)` bypasses gracefulShutdown | ✅ | Same as R5§1 — `onFatalError` callback |
| 3 | PR description says 6 tests | ✅ | Updated to 10 |
| 4 | EADDRINUSE test mutates config.PORT | ✅ | `process.exit` spy removed in favor of `onFatalError` mock. Config mutation still needed to trigger port collision but restored in `finally` block |
| 5 | No test for `isReady()` throwing | ✅ | Test added at line 168 (review said omission was fine; we went further and added it) |

### Review 5 (claude, 22:41:39) — 5 suggestions, post-simplify

| # | Suggestion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | `process.exit(1)` bypasses `gracefulShutdown` (would-fix) | ✅ | `gateway-health.js:43,92` — `onFatalError` param, defaults to `process.exit(1)`. `index.js:405` passes `() => gracefulShutdown(1)` |
| 2 | HEALTHCHECK `start-period` 10s → 30s | ✅ | `Dockerfile:41` — `--start-period=30s` |
| 3 | EADDRINUSE test mutates config.PORT | ✅ | Same as R4§4 |
| 4 | PR description test count | ✅ | Updated from 6 to 10 |
| 5 | `req.url.split('?', 1)[0]` → `new URL().pathname` | ✅ | `gateway-health.js:47` — `new URL(req.url, 'http://127.0.0.1').pathname` |

### Owner review (justin-layerv, 21:29:48) — 2 suggestions

| # | Suggestion | Status | Evidence |
|---|-----------|--------|----------|
| 1 | `closes #151` → `addresses #151` | ✅ | PR title updated |
| 2 | Sandbox-deploy verification before merge | ✅ | Noted — unchecked test-plan boxes gate merge |

---

### Commit history

| Commit | What |
|--------|------|
| `623924d` | Loopback bind (`127.0.0.1`), GET-only intent comment |
| `872ec38` | `server.address().port` log, Dockerfile `127.0.0.1`, test comment precision, PR title |
| `271954c` | HEAD support, `server.on('error')` handler, `requestTimeout`/`headersTimeout`, HEAD + EADDRINUSE tests |
| `9e7d7c9` | Query-string URL matching, `JSON.stringify` response bodies, `isReady()` try/catch, 2 new tests |
| `584155a` | Pre-computed response constants, consolidated test helper |
| `64d4b07` | `onFatalError` callback, `start-period=30s`, `new URL().pathname`, EADDRINUSE test cleanup, description update |

**10/10 gateway-health tests pass. CI green.**

justin-layerv:
/simplify pass: code is clean, no changes needed. Reviewed all 4 changed files (gateway-health.js, gateway-health.test.js, index.js, Dockerfile) line by line — no dead code, no unused imports, no naming issues, no redundant comments, no simplification opportunities. The previous simplify pass (584155a) already addressed the main targets.

justin-layerv:
## /simplify pass — detailed file-by-file audit

Reviewed all 4 changed files line-by-line. No changes needed.

### `gateway-health.js` (113 lines)

- **Imports (lines 20-22):** `http` used at L44, `config` at L105, `logger` at L94+L106. Zero dead imports.
- **Constants (lines 24-27):** `BODY_OK` at L72, `BODY_UNHEALTHY` at L78, `BODY_NOT_FOUND` at L56. Zero dead constants.
- **URL parsing (line 47):** `new URL(req.url, 'http://127.0.0.1')` — dummy base required by Node's URL parser for relative paths. Cannot be simplified.
- **Method guard (line 54):** Single combined check with early return. Clean.
- **try/catch (lines 63-68):** Considered `let ready = false; try { ready = isReady(); } catch {}` but the current explicit-catch-assignment form is equally clear. Not worth a diff.
- **Error handler default (line 92):** `onFatalError || (() => process.exit(1))` — correct `||` fallback since the param isn't destructured with a default.
- **Comments:** Every block (module header, loopback-bind, error handler, timeout rationale) explains *why*, not *what*. All load-bearing.

### `gateway-health.test.js` (229 lines)

- **Imports (lines 15-16):** `http` used in `request` helper (L41), `startGatewayHealthServer` used in all 10 tests. Zero dead.
- **Logger mock (lines 21-26):** `error` asserted in EADDRINUSE test (L215). `info`/`warn`/`debug` suppress output noise. All needed.
- **Test helpers (lines 38-63):** `request`, `waitForListening`, `closeServer` — each used by ≥8 tests. Already consolidated in prior simplify (584155a). No further reduction possible.
- **Test structure:** 8/10 tests follow create→wait→try/assert→finally/close. Structural repetition, not extractable — each test passes a different `isReady` closure, EADDRINUSE creates 2 servers, drain test asserts `.listening` directly. `beforeEach`/`afterEach` would be worse.
- **EADDRINUSE test (lines 195-227):** `second.on('error', resolve)` at L213 fires after the production handler — correct Node EventEmitter ordering. Config.PORT mutation (L208) restored in `finally` (L224). Clean.

### `index.js` (2 changed regions)

- **Line 6:** Import of `startGatewayHealthServer` used at L405. Not dead.
- **Lines 391-406:** Comment explains both `isHttp` and `isGateway` branches. Wiring matches JSDoc exactly. Clean.

### `Dockerfile` (2 changed lines)

- **Line 41:** `--start-period=30s` — appropriate for Discord WebSocket handshake timing.
- **Line 42:** IPv4 literal `127.0.0.1` matches loopback-only bind. Consistent.

### Verdict

Zero dead code, zero unused imports, zero naming issues, zero redundant comments, zero simplification opportunities. The code is clean.

justin-layerv:
## /simplify pass (de1f4c2) — file-by-file audit with findings

### Files reviewed

| File | Path | Lines | Finding |
|------|------|-------|---------|
| gateway-health.js | `apps/discord/src/gateway-health.js` | 113 → 111 | **1 improvement applied** |
| gateway-health.test.js | `apps/discord/tests/gateway-health.test.js` | 229 | Clean — no changes |
| index.js | `apps/discord/src/index.js` | L6, L391-406 | Clean — no changes |
| Dockerfile | `apps/discord/Dockerfile` | L41-42 | Clean — no changes |

### Change applied

**`gateway-health.js:43,92-95` — default parameter replaces `||` fallback**

Before (3 lines):
```js
function startGatewayHealthServer(isReady, onFatalError) {
  // ...
  const fatal = onFatalError || (() => process.exit(1));
  // ...
    fatal(err);
```

After (2 lines, no intermediate variable):
```js
function startGatewayHealthServer(isReady, onFatalError = () => process.exit(1)) {
  // ...
    onFatalError(err);
```

Why: default parameter is more idiomatic JS, makes the default visible at the declaration site (where readers look first), and eliminates the `fatal` intermediate variable. JSDoc at line 38 already documents the default, so no doc changes needed. Semantics are equivalent — no caller passes `null`.

### Specific findings per file (examined, no action needed)

**`gateway-health.js`**
- L20-22: imports `http`, `config`, `logger` — all used (L44, L105, L93+L106). Zero dead.
- L24-27: constants `BODY_OK` (L72), `BODY_UNHEALTHY` (L78), `BODY_NOT_FOUND` (L56). Zero dead.
- L47: `new URL(req.url, 'http://127.0.0.1')` — dummy base required for relative URL parsing. Cannot simplify.
- L54: method+path guard — single combined check with early return. Clean.
- L63-68: `let ready; try { ready = isReady() } catch { ready = false }` — considered `let ready = false; try { ... } catch {}` but explicit catch-assignment is clearer intent. Not worth a diff.
- L55,71,77: `{ 'Content-Type': 'application/json' }` repeated 3×. Per CLAUDE.md "three similar lines is better than a premature abstraction" — skip.
- L85-86: timeouts set once, no duplication.
- L98-104: loopback comment is 7 lines but each sentence covers a distinct constraint (probe location, tighter default, external leak prevention, contrast with Express). All load-bearing.

**`gateway-health.test.js`**
- L15-16: imports `http` (used in `request` helper L41), `startGatewayHealthServer` (used in all 10 tests). Zero dead.
- L21-26: logger mock — `error` asserted at L215; `info`/`warn`/`debug` suppress noise. All needed.
- L33-36: config mock — `requireActual` + spread + `PORT: 0`. Correct Jest pattern.
- L38-52: `request` helper — generalized across all HTTP tests, no further consolidation possible.
- L54-63: `waitForListening`, `closeServer` — used by all 10 tests. Minimal wrappers.
- L66-193: 9 tests follow create→wait→try→finally/close. Each passes a different `isReady` closure; `beforeEach`/`afterEach` can't parameterize this. Structural repetition, not extractable.
- L195-227: EADDRINUSE test — `second.on('error', resolve)` at L213 fires after the production handler (EventEmitter ordering). Config.PORT mutation (L208) restored in `finally` (L224). Correct.

**`index.js`**
- L6: `startGatewayHealthServer` import — used at L405. Not dead.
- L391-406: comment block explains both branches, wiring matches JSDoc. Clean.

**`Dockerfile`**
- L41: `--start-period=30s` matches Discord WebSocket handshake timing. Correct.
- L42: `127.0.0.1` IPv4 literal matches loopback-only bind. Consistent.

10/10 tests pass.

justin-layerv:
## Review feedback addressed — round 5 (529ff4d)

Addresses the post-de1f4c2 review (claude, 22:49:58).

### Code changes

1. **Structured logger meta on info log** (review 5 §1) — `gateway-health.js:105`: changed `logger.info(\`Gateway health listener on 127.0.0.1:${...}\`)` to `logger.info('Gateway health listener listening', { addr: '127.0.0.1', port: server.address().port })`. Matches the structured `{ error, code }` second arg on the error log at line 93 and is consistent with the rest of the app's logger calls for CloudWatch field extraction.

2. **Remove redundant `second.close()`** (review 5 §2) — `gateway-health.test.js:222`: removed. The server never entered listening state (listen errored out), so `.close()` was a no-op on Node ≥18. Cleaner without it.

3. **Boot-ordering comment** (review 5 "Worth noting" §2) — `index.js:405`: added comment documenting that 503 during boot until READY is intentional and covered by `--start-period=30s`.

4. **Cross-reference comment for start-period** (review 5 "Worth noting" §1) — `Dockerfile:41`: added comment linking `start-period=30s` to the 30s `client.login()` timeout in `index.js`, so a future Dockerfile tweak doesn't regress the alignment.

### Not changed (reviewer said "flagging only")

- **Config.PORT mutation in EADDRINUSE test** (review 5 §3): reviewer explicitly said "Neither is necessary; flagging only." Save/restore in `finally` is correct and Jest serializes tests in a file. Left as-is.

10/10 tests pass.

justin-layerv:
## Review feedback addressed — round 6 (4dc4cf0)

Addresses all 3 substantive items from the latest review (claude, 22:49:58).

### 1. Removed `onFatalError` default parameter (review §1)

**Problem:** `onFatalError = () => process.exit(1)` in the function signature made it look safe to omit, but the default skips all cleanup (Discord WebSocket, DB) that the rest of the module's documentation argues for.

**Fix:** `gateway-health.js:44` — removed the default parameter entirely. The error handler at lines 93-101 now explicitly checks `if (onFatalError)` and falls back to `process.exit(1)` with a comment explaining callers should pass `gracefulShutdown`. JSDoc at line 37 updated to mark the param as required (no `[]` brackets), directing callers to pass `() => gracefulShutdown(1)`.

The production call site (`index.js:408`) already passes `() => gracefulShutdown(1)`. Tests that don't trigger listen errors are unaffected (the handler never fires).

### 2. Fixed PR description stale test count (review §2)

Test plan checklist updated: `6/6 green` → `10/10 green`. The "Test coverage" heading already said 10 — this syncs the checklist.

### 3. Eliminated config.PORT mutation in EADDRINUSE test (review §3)

**Problem:** Test mutated the shared `config.PORT` mock at module scope (`gateway-health.test.js:206-208`), relying on save/restore in `finally`. Action-at-a-distance that breaks if tests reorder.

**Fix:** Added explicit `port` parameter to `startGatewayHealthServer` (`gateway-health.js:44`, defaults to `config.PORT`). The EADDRINUSE test now passes the occupied port directly:
```js
const { port } = first.address();
const second = startGatewayHealthServer(() => true, onFatalError, port);
```
No config mutation, no save/restore. The production call site is unchanged — it omits `port` and gets `config.PORT` from the default.

### Not changed (minor, reviewer said "fine as-is")

- §4 (`second.close()` after failed listen): already removed in 529ff4d.
- §5 (request timeout in test helper): acknowledged as minor; Jest's 5s default provides the timeout.
- §6 (pre-computed constants): reviewer said "fine as-is."
- §7 (URL parsing per request): reviewer said "not worth changing."

10/10 tests pass.

justin-layerv:
## /simplify pass (d7889ef)

### Change applied

**`gateway-health.js:101` — remove dead `eslint-disable-line no-process-exit`**

The `no-process-exit` rule is not configured in `.eslintrc.json` (it was deprecated in ESLint v7 and is not in `eslint:recommended`). The annotation suppresses nothing. Additionally, none of the ~18 `process.exit(1)` calls in `index.js` use this annotation, making it inconsistent even if the rule were active.

### Full audit (4 files, post-round-6 state)

**`apps/discord/src/gateway-health.js` (119 lines)**
- L20-22: imports `http` (L46), `config` (L45 default param), `logger` (L95, L113). Zero dead.
- L24-27: constants `BODY_OK` (L74), `BODY_UNHEALTHY` (L80), `BODY_NOT_FOUND` (L58). Zero dead.
- L45: `startGatewayHealthServer(isReady, onFatalError, port = config.PORT)` — `onFatalError` intentionally not defaulted per round-6 review; `port` default avoids config mutation in tests.
- L49: `new URL(req.url, 'http://127.0.0.1')` — dummy base required for relative URL parsing.
- L56: combined method+path guard, single early-return. Clean.
- L65-70: try/catch around isReady(). Explicit catch assignment. Clean.
- L94-102: `if (onFatalError) { ... } else { process.exit(1) }` — fallback is commented, observable via the preceding `logger.error`. Clean.
- L87-88: `requestTimeout`/`headersTimeout` = 5_000. Set once.
- L105-111: loopback comment — 7 lines, each covers a distinct constraint. Load-bearing.
- L112: `server.listen(port, ...)` uses the parameter, not config directly. Clean.

**`apps/discord/tests/gateway-health.test.js` (224 lines)**
- L15-16: `http` used in `request` helper (L41). `startGatewayHealthServer` used in all 10 tests. Zero dead.
- L21-26: logger mock — `error` asserted at L213. Others suppress noise. All needed.
- L33-36: config mock — `PORT: 0` still needed for the 8 tests that rely on the default `port` param.
- L38-52: `request` helper — used by 8 tests.
- L54-63: `waitForListening`, `closeServer` — used by all tests. Minimal.
- L195-223: EADDRINUSE test — passes port directly as 3rd arg (no config mutation). No `second.close()` on the never-listened server. Clean.

**`apps/discord/src/index.js` (L6, L391-409)**
- L6: `startGatewayHealthServer` import used at L408. Not dead.
- L405-408: boot-ordering comment + call with 2 args (port defaults to config.PORT). Clean.

**`apps/discord/Dockerfile` (L41-44)**
- L41-42: cross-reference comment linking start-period to login timeout. Correct.
- L43-44: `--start-period=30s`, `127.0.0.1`. Consistent with loopback bind.

10/10 tests pass.

justin-layerv:
## Review feedback addressed — round 7 (5b4f1f3)

Addresses the round-7 review (claude, 23:07:06).

### Code changes

1. **URL parse hardening** (review 7 §1, non-optional) — `gateway-health.js:49-51`: replaced `new URL(req.url, 'http://127.0.0.1')` with `req.url.split('?', 1)[0]`. `new URL()` throws `TypeError [ERR_INVALID_URL]` on malformed input (reachable via ECS Exec sessions on loopback). `split()` can't throw, is cheaper, and only the path is needed. Comment explains the choice.

2. **Fix double-close on EADDRINUSE** (review 7 §2) — `index.js:408-413`: `onFatalError` callback now nulls out `httpServer` before calling `gracefulShutdown(1)`. Prevents `gracefulShutdown` from calling `.close()` on a server that never finished listening, which would log a confusing "Server is not running" error.

3. **Add `beforeEach(jest.clearAllMocks)`** (review 7 §5) — `gateway-health.test.js:70`: resets all mock call history between tests. Prevents the EADDRINUSE test's `logger.error` assertion from becoming brittle if earlier tests start triggering `logger.error`.

4. **Dockerfile probe/bind address coupling comment** (review 7 §6) — `Dockerfile:43-45`: documents that the probe address (`127.0.0.1`) MUST match the listener's bind address in `gateway-health.js`. Binding `::1` while probing `127.0.0.1` (or vice versa) is a classic silent-probe-failure mode.

### Not changed (reviewer confirmed acceptable as-is)

- **`process.exit(1)` fallback** (review 7 §3): reviewer said "the safety net is cheap and the comment justifies it." The fallback is already commented, the preceding `logger.error` makes it observable, and `index.js` always provides the handler in production.
- **EADDRINUSE test leaked `second` handle** (review 7 §4): reviewer said "Not worth fixing unless you're already in there." The server never bound — no resources to release. Jest tears down the worker.

10/10 tests pass.

justin-layerv:
## Review feedback addressed — round 8 (657d09f)

Addresses the latest review (claude, 23:07:06). All 6 items explicitly accounted for.

### Code changes

1. **URL parse hardening** (review 8 §1, non-optional) — `gateway-health.js:51`: uses `req.url.split('?', 1)[0]` instead of `new URL()`. The reviewer offered two options: try/catch around `new URL()`, or `split()`. We use `split()` because it **eliminates the throw entirely** rather than catching it — cheaper, simpler, and no error-handling code needed. Comment at L49-50 documents the rationale ("split() instead of new URL() because URL() throws on malformed input").

2. **Bump timeouts from 5s to 10s** (review 8 §2) — `gateway-health.js:91-92`: `requestTimeout` and `headersTimeout` changed from `5_000` to `10_000`. The Dockerfile HEALTHCHECK uses `--timeout=5s` — with both at 5s they're a near-coincident race. 10s lets wget's 5s budget be the binding constraint. Comment at L89-90 explains this.

3. **Fargate caveat in Dockerfile** (review 8 §3) — `Dockerfile:43-44`: added "NOTE: On Fargate this Dockerfile HEALTHCHECK is cosmetic — ECS uses the task-def `healthCheck` (re-added in the infra follow-up). Tune the task-def values, not these, for production behavior."

4. **Close `second` server in EADDRINUSE test** (review 8 §4) — `gateway-health.test.js:227-228`: hoisted `let second` above the `try`, added `if (second) await closeServer(second).catch(() => {})` in the `finally` block. Keeps `jest --detectOpenHandles` clean.

5. **Content-Type assertion** (review 8 §5) — `gateway-health.test.js:79,81,92,94`: `request` helper now returns `contentType`. 200 and 503 tests assert `expect(contentType).toBe('application/json')`.

6. **`jest.clearAllMocks`** (review 8 §6) — Already present at L74 (`beforeEach(() => jest.clearAllMocks())`). No change needed.

### Verified (from retry checklist)

- **`onListenError` has NO default parameter**: `function startGatewayHealthServer(isReady, onListenError, port = config.PORT)` — confirmed at L49. Required arg, no default. The process.exit fallback was removed; every caller must provide an explicit handler.
- **PR description test count = 10**: matches actual 10/10.
- **URL parsing uses `split()`**: comment at L49-50 justifies: "split() instead of new URL() because URL() throws on malformed input (reachable from a buggy probe or L7 scanner via ECS Exec)."

10/10 tests pass.

justin-layerv:
/simplify pass: code is clean, no changes needed.

### Audit evidence (4 files, post-round-8 state @ 657d09f)

**`apps/discord/src/gateway-health.js` (122 lines)**
- L20-22: imports `http` (L50), `config` (L49 default param), `logger` (L103, L115). Zero dead.
- L24-27: `BODY_OK` (L80), `BODY_UNHEALTHY` (L86), `BODY_NOT_FOUND` (L64). Zero dead.
- L37-43: `onListenError` JSDoc is verbose (~7 lines) but each piece answers a natural question: what it does, why no default, what index.js passes, and why the default was removed. Load-bearing.
- L49: `startGatewayHealthServer(isReady, onListenError, port = config.PORT)` — required `onListenError` (no default), optional `port` with config default. Clean API.
- L55: `req.url.split('?', 1)[0]` — can't throw (vs `new URL()`), comment at L53-54 explains. Clean.
- L62: combined method+path guard, single early return. Clean.
- L71-76: try/catch around `isReady()` with explicit catch assignment. Clean.
- L95-96: timeouts at 10_000ms with comment explaining the 5s wget slack. Clean.
- L102-105: error handler logs structured data + delegates to `onListenError`. No fallback needed since param is required. Clean.
- L107-113: loopback bind comment — 7 lines, each distinct constraint. Load-bearing.

**`apps/discord/tests/gateway-health.test.js` (247 lines)**
- L15-16: `http` (L47), `startGatewayHealthServer` (all 10 tests). Zero dead.
- L21-26: logger mock — `error` asserted at L231; others suppress noise. All needed.
- L33-36: config mock — `PORT: 0`. Still needed for 9 tests using default port param.
- L42: `noopOnListenError` — used in 9/10 tests. Comment at L38-41 explains the design. Clean.
- L44-66: `request` helper returns `{ status, body, contentType }`. Used by 8 tests. Clean.
- L68-77: `waitForListening`, `closeServer` — minimal wrappers, all 10 tests. Clean.
- L80: `beforeEach(() => jest.clearAllMocks())` — prevents mock state leaking. Clean.
- L81-210: 9 tests own their server lifecycle (different closures per test). Not extractable to beforeEach/afterEach.
- L212-245: EADDRINUSE test — `let second` hoisted, closed in finally with `.catch(() => {})`. Comment at L239-241 explains why. Clean.

**`apps/discord/src/index.js` (L6, L402-415)**
- L6: `startGatewayHealthServer` import → used at L408. Not dead.
- L408-414: `onListenError` nulls `httpServer` before `gracefulShutdown(1)` — prevents double-close on never-listened server. Comment at L409-411 explains. Clean.

**`apps/discord/Dockerfile` (L38-50)**
- L38-48: comment block covers port hardcoding, start-period alignment, Fargate caveat, and probe/bind address coupling. Each concern is distinct. Load-bearing.
- L49-50: HEALTHCHECK params correct, `127.0.0.1` matches loopback bind.

Zero dead code, zero unused imports, zero naming issues, zero redundant comments, zero simplifiable patterns.

vikramlayerv:
## pr-polish: round-N polish on Justin's #193 §13 commits — closing items

This polish pass covered the new commits ($3669a1a$ → $f25c240$ → $cf33b3e$) addressing Justin's review §13 (`/health` 503 metric). The convergence loop is closed at round 2 — Claude's most recent Code Review post is the same one from before this round (workflow ran on $cf33b3e$ but the bot judged the diff didn't warrant a fresh comment), and that posted review's verdict is "Nothing here blocks merge."

### What this round-N pass changed (from 3-agent Phase 1 review)

| # | Severity | Fix | Commit |
|---|---|---|---|
| 1 | High | Empty meta `{}` on the audit emit collapsed two unhealthy paths (`isReady() === false` vs `isReady()` threw) into one count. Added `reason: 'not_ready' \| 'sampler_threw'` so the dashboard can split. | $f25c240$ |
| 2 | High | Existing `isReady() throwing returns 503` test asserted the 503 + debug log but NOT the audit emit. Added a dedicated test pinning `reason='sampler_threw'`. | $f25c240$ |
| 3 | High | HEAD `/health` unhealthy emit uncovered. Added a test pinning HEAD 503 + emit + `reason='not_ready'` so a future HEAD-early-return optimization can't silently drop the alarm signal. | $f25c240$ |
| 4 | Medium | Existing 3-probe count test now also asserts `reason='not_ready'` on every emit — pins the dimension contract. | $f25c240$ |
| 5 | Medium | Tightened `constants.js` comment to reference the paired filter at `qurl-integrations-infra qurl-bot-discord/terraform/monitoring.tf` precisely. | $f25c240$ |
| 6 | Minor | `index.js` comment that said "never finished listening" overstated the invariant — `server.on('error')` also fires post-listen. Reworded to cover both cases. | $cf33b3e$ |
| 7 | Minor | New test pins the strict `/health` match contract (`/health/` and `/health/ready` both 404). Locks the contract before a "be liberal" refactor can drift it silently. | $cf33b3e$ |
| 8 | Nit | Dropped a dead `.catch(() => {})` on `closeServer()` in the EADDRINUSE test — `closeServer` always resolves. | $cf33b3e$ |

### Items the latest posted Claude review still flags — left for the reasons below

|---|---|---|---|---|
| Suggestion (#2) | SIGTERM during listen-window — `httpServer.close()` on a not-yet-listening server logs `ERR_SERVER_NOT_RUNNING` as a warn | Existing `gracefulShutdown` warn logs the error code clearly. Reviewer called it "tolerable." Adding a `if (!server.listening)` guard in `gracefulShutdown` would suppress the warn but adds boilerplate for a one-shot startup race. | None — acceptable as-is | None — log noise only |
| Suggestion (#6) | IPv4 loopback dependency — Fargate distroless ARM64 base has v4 enabled; if base ever changes to IPv6-only, probe goes silent | `gateway-health.js` already has the IPv4 dependency comment block (lines 30-37) calling out the constraint. Dockerfile's HEALTHCHECK uses `127.0.0.1` literal, matching. | None — acceptable as-is | None at current base image |
| Nit | JSDoc for `onListenError` says "index.js passes `() => gracefulShutdown(1)`" but actual call is a 2-line arrow | Minor doc drift on a parameter description. Will rot if wiring changes; could update opportunistically next time the wiring is touched. | None — subjective/style | None |
| Nit | `BODY_NOT_FOUND` returns `{"status":"not_found"}` JSON on 404 — empty body would save 25 bytes per probe-misconfiguration hit | Probe-misconfiguration hits should be vanishingly rare in steady state. The JSON shape is consistent with `BODY_OK` / `BODY_UNHEALTHY` — symmetric is more readable than asymmetric. | None — subjective/style | None |

### Convergence notes

- Tests: 861/861 green (was 858 before this round; +3 net for `reason` tests + strict-match)
- Lines added by this round: ~104 across 4 files
- The §13 audit emit now carries `reason` so the paired CW filter (qurl-integrations-infra PR #419 monitoring.tf) can dimension on it without a bot-side change
- All claimed fixes verified by grep against pushed HEAD ($cf33b3e$)

— via /pr-polish

Issue #121, opened 2026-04-24
Dogfood @layerv/qurl in Discord bot: drop hand-rolled qurl.js (needs SDK dual-publish first)


vikramlayerv:
## Context

The Discord bot hand-rolls its own QURL API client at `apps/discord/src/qurl.js` (~200 LOC of fetch + retry/backoff) instead of using our own SDK. The file's own header says why:

> `apps/discord/src/qurl.js:5-8`
> *"Lightweight QURL API client using fetch. Avoids ESM/CJS compatibility issues with the @layerv/qurl SDK."*

This landed in #45 (Python→Node migration). The SDK (`@layerv/qurl`) ships ESM-only; the bot is CJS; `require('@layerv/qurl')` doesn't work, so the workaround was written instead of fixing the SDK.

Justin flagged 2026-04-24: *"it is a bit ironic that our QURL Bot is not using our own QURL SDK. We should fix that. We should dual-publish the SDK — ESM + CJS, port the bot's retry/timeout logic into the SDK as canonical (as needed), then swap."*

## Plan

Three threads across two repos. SDK work lands first; bot swap here depends on it.

### 1. SDK: dual-publish ESM + CJS (~½–1 day) — `layervai/qurl-typescript`
- Add CJS build alongside ESM (tsup / tshy / dual-tsconfig)
- Update `package.json` `exports` with both `import` + `require` entries
- Verify type resolution under `moduleResolution: node16` and `bundler`
- Smoke-test from a CJS consumer: `const { QURLClient } = require('@layerv/qurl')`
- Release-Please handles the publish

### 2. SDK: retry parity widen (~2h) — `layervai/qurl-typescript`
Bot retries 408 + 500; SDK doesn't. SDK is otherwise richer (exp backoff + jitter, Retry-After, mutating-vs-idempotent split, typed errors). Add 408 + 500 to SDK's `RETRYABLE_STATUS`. No other logic changes.

### 3. Bot swap (~½ day) — THIS REPO
- Replace `qurl.js` call sites in `apps/discord/src/commands.js`, `discord.js`, etc. with `QURLClient` from `@layerv/qurl`
- **Keep** the SSRF pre-flight (`isPrivateHost` + `assertNotPrivateAfterResolve`) as a bot-level wrapper — product-policy, doesn't belong in SDK
- Delete `apps/discord/src/qurl.js`
- Update tests accordingly

### Out of scope
[...]

PR #194, opened 2026-05-05
feat(discord): phase 1 monitoring — gateway heartbeat, interaction metrics, guild lifecycle

vikramlayerv:
## Why

Today the prod Discord bot has effectively no observability beyond "the ECS task is running." Justin's review on #193 enumerated the liveness gaps (heartbeat zombie, dispatch deadlock, command latency drift) and the alarm-side gaps (SNS topic with no subscribers, 4 of 5 dead alarms still firing into the void per #121). This PR is the bot-side half of Phase 1 — emit the structured audit events that the paired infra PR pattern-matches.

**Built on top of #193.** Once that lands the base flips to `main`.

Pairs with **qurl-integrations-infra `feat/phase1-monitoring`** (CloudWatch metric filters + alarms + dashboard). Both PRs land independently — the bot's audit events flow regardless of whether the infra PR is applied; the infra side reads `INSUFFICIENT_DATA` until the bot image deploys.

## What

5 new audit events emitted via the existing `logger.audit()` mechanism:

| Event | Cadence | Carries |
|---|---|---|
| `gateway_heartbeat_healthy` | 30s, healthy path only | `ping_ms`, `ack_age_ms` |
| `interaction_handled` | 1 per slash command | `command_name`, `success`, `failure_type`, `handler_duration_ms` |
| `guild_install` | guildCreate event | `guild_id`, `member_count`, `replay` |
| `guild_uninstall` | guildDelete event | `guild_id` |
| `active_guild_count` | 60s gauge | `count` |

`failure_type` vocabulary: `'ack_timeout' | 'handler_error' | 'unknown_command' | 'reply_failed' | null`. ACK timeouts are detected via discord.js error code 10062 ("Unknown interaction"), so the user-visible "did not respond" cluster has its own dimension and an absolute-count alarm can stand apart from the rate alarm.

`handler_duration_ms` is wall-clock from `handleCommand` entry to metric emit (NOT user-perceived edge-to-ACK latency — Phase 2 work). Captured via `process.hrtime.bigint()` so an NTP step backward can't produce a negative duration. The dashboard alarm "p99 > 2s" still works as a slow-handler signal.

The heartbeat **only emits on the healthy path** so the alarm fires on *missing data* — collapses two metrics (healthy / unhealthy) into one. Composite readiness threshold (`isReady && ws.ping > 0 && ack_age_ms < 60s`) catches the heartbeat-zombie failure mode `client.isReady()` alone misses (Justin's review §1 on #193).

Known limitation tracked in #195: `replay` detection on `guildCreate` uses `client.isReady()`, which doesn't flip back on session resumes. Acceptable for Phase 1 (no real users in prod).

## How

| Change | File |
|---|---|
| 5 new entries in `AUDIT_EVENTS` frozen set | `apps/discord/src/constants.js` |
| New module: 30s heartbeat + 60s guild-count timers | `apps/discord/src/gateway-metrics.js` |
| handleCommand instrumentation (single emit per interaction) | `apps/discord/src/commands.js:3503` |
| guildCreate / guildDelete handlers (replay-tagged) | `apps/discord/src/discord.js` |
| Wiring + graceful shutdown cleanup | `apps/discord/src/index.js` |

`interaction_handled` is **one event** (not three), with `success` / `failure_type` / `handler_duration_ms` dimensions. The metric math at the alarm layer derives total / failures / latency without N-event fan-out; failure-rate numerator and denominator stay in lock-step.

## Test plan

- [x] `npx jest tests/gateway-metrics.test.js` — 18 / 18 green
- [x] `npx jest tests/commands-comprehensive.test.js` — 53 / 53 green
- [x] `npx jest tests/discord.test.js` — 58 / 58 green
- [x] `npx jest` — full suite 888 / 888 green
- [x] Sandbox deploy — confirm `gateway_heartbeat_healthy` appears in `bot_gateway` log group at 30s cadence
- [x] Sandbox: invoke `/qurl` slash command, confirm `interaction_handled` event with `handler_duration_ms` and `success: true`
- [x] Sandbox: validate metric-filter match in CloudWatch console

## Phase 2 / 3

This PR is Phase 1 of 3:

- **Phase 2** — Reconnect rate, response latency, per-command breakdown, backend latency (qurl-service / NHP / DDB), Discord REST 4xx/5xx + 429s, panics / Sentry, DDB throttles + ConditionalCheckFailed (with rolloutState baselining), gateway send queue depth, FD/handle drift, per-command invocation volume. Plus #195 (replay-on-resume detection). **Open as a separate planning PR (no implementation).**
- **Phase 3** — Synthetic canary running a `/qurl ping` slash command in a control guild every 1-5 min, end-to-end. **Open as a separate planning PR (no implementation).**

### DIFF
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -38,7 +38,7 @@ USER node
 # Port is hardcoded at 3000: ${PORT:-3000} inside a Dockerfile CMD's exec
 # form doesn't interpolate at runtime, and EXPOSE is already 3000. If the
 # runtime overrides PORT, update EXPOSE + this line together.
-HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
+HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
   CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
 # Run node directly (not `npm start`) so SIGTERM is delivered to the Node
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -40,6 +40,9 @@ USER node
 # runtime overrides PORT, update EXPOSE + this line together.
 # start-period=30s matches the 30s client.login() timeout in index.js —
 # the /health endpoint returns 503 until the Discord gateway sends READY.
+# NOTE: On Fargate this Dockerfile HEALTHCHECK is cosmetic — ECS uses the
+# task-def `healthCheck` (re-added in the infra follow-up). Tune the
+# task-def values, not these, for production behavior.
 # The probe address (127.0.0.1) MUST match the listener's bind address in
 # gateway-health.js — binding ::1 while probing 127.0.0.1 (or vice versa)
 # is a classic silent-probe-failure mode.
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -39,7 +39,7 @@ USER node
 # form doesn't interpolate at runtime, and EXPOSE is already 3000. If the
 # runtime overrides PORT, update EXPOSE + this line together.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
-  CMD wget -qO- "http://localhost:3000/health" || exit 1
+  CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
 # Run node directly (not `npm start`) so SIGTERM is delivered to the Node
 # process itself — otherwise npm eats the signal and the container hard-kills
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -38,6 +38,8 @@ USER node
 # Port is hardcoded at 3000: ${PORT:-3000} inside a Dockerfile CMD's exec
 # form doesn't interpolate at runtime, and EXPOSE is already 3000. If the
 # runtime overrides PORT, update EXPOSE + this line together.
+# start-period=30s matches the 30s client.login() timeout in index.js —
+# the /health endpoint returns 503 until the Discord gateway sends READY.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
   CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
--- a/apps/discord/Dockerfile
+++ b/apps/discord/Dockerfile
@@ -40,6 +40,9 @@ USER node
 # runtime overrides PORT, update EXPOSE + this line together.
 # start-period=30s matches the 30s client.login() timeout in index.js —
 # the /health endpoint returns 503 until the Discord gateway sends READY.
+# The probe address (127.0.0.1) MUST match the listener's bind address in
+# gateway-health.js — binding ::1 while probing 127.0.0.1 (or vice versa)
+# is a classic silent-probe-failure mode.
 HEALTHCHECK --interval=30s --timeout=5s --start-period=30s --retries=3 \
   CMD wget -qO- "http://127.0.0.1:3000/health" || exit 1
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3505,10 +3505,12 @@ async function registerCommands(client) {
 // tries to reply after the token has expired. Detect both the numeric
 // code (preferred — discord.js DiscordAPIError preserves it) and the
 // message text (fallback for wrapped-error shapes that drop .code).
-// The regex accepts an optional "Class: " prefix common to wrapped
-// shapes (e.g. RESTJSONError) but rejects trailing content like
-// "Unknown interaction type X" to avoid misclassification.
-const ACK_TIMEOUT_MSG_RE = /^(?:[A-Za-z][A-Za-z0-9]*: )?Unknown interaction$/;
+// The regex accepts an optional "Class: " or "Class[code]: " prefix
+// common to wrapped shapes — e.g. `RESTJSONError: Unknown interaction`
+// or discord.js's own `DiscordAPIError[10062]: Unknown interaction` —
+// but rejects trailing content like "Unknown interaction type X" to
+// avoid misclassification.
+const ACK_TIMEOUT_MSG_RE = /^(?:[A-Za-z][A-Za-z0-9]*(?:\[\d+\])?: )?Unknown interaction$/;
 function isAckTimeoutError(err) {
   if (!err) return false;
   if (err.code === 10062) return true;
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3531,7 +3531,11 @@ async function handleCommand(interaction) {
   const emitInteractionMetric = (success, failureType) => {
     // logger.audit has its own two-tier try/catch (logger.js:209-234)
     // and is documented to never throw — no outer guard needed here.
-    const handler_duration_ms = Number((process.hrtime.bigint() - ackStart) / 1_000_000n);
+    // Number(ns) / 1_000_000 (NOT BigInt division then Number) so a
+    // sub-millisecond handler reports a fractional ms instead of
+    // truncating to 0. ns count fits comfortably in a double for any
+    // realistic handler duration; precision loss here is irrelevant.
+    const handler_duration_ms = Number(process.hrtime.bigint() - ackStart) / 1_000_000;
     logger.audit(AUDIT_EVENTS.INTERACTION_HANDLED, {
       command_name: commandName,
       success,
@@ -3595,6 +3599,16 @@ async function handleCommand(interaction) {
       ephemeral: true,
     };
 
+    // failureType reflects the FIRST failure observed: handler_error
+    // (or ack_timeout if the handler threw a 10062). If the follow-up
+    // reply ALSO fails for a non-ack reason, we deliberately keep the
+    // original handler_error tag — the original execute failure is the
+    // more meaningful signal for the dashboard. Only an ack_timeout on
+    // the reply is allowed to override (so the user-visible "did not
+    // respond" cluster catches it). Asymmetric vs. the stale-
+    // registration path above (which DOES tag reply_failed) because
+    // there's no prior execute() in that path; the reply outcome IS
+    // the only signal.
     let failureType = isAckTimeoutError(error) ? 'ack_timeout' : 'handler_error';
     try {
       if (interaction.replied || interaction.deferred) {
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3511,27 +3511,33 @@ async function handleCommand(interaction) {
   if (!interaction.isChatInputCommand()) return;
 
   // Phase 1 monitoring instrumentation. handler_duration_ms is the
-  // wall-clock time from handleCommand entry to the metric emit (after
-  // command.execute completes, OR at any failure path). NOT the user-
-  // perceived ACK latency — for commands that defer + run a long
-  // background task, the user sees a fast "Bot is thinking..." ACK
-  // and this number captures the whole operation. True edge-to-ACK
-  // measurement is Phase 2 work (see qurl-integrations-infra#TBD).
-  // For Phase 1 the alarm semantics ("p99 > 2s") still hold: a slow
-  // handler is a real signal even if the user already saw the ack.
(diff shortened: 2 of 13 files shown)


vikramlayerv:

The loop is closed at round 9 — Claude's last review explicitly LGTM'd the bot-side merge. Everything not addressed in tree is tabled below with reasons. Revisit any if you disagree.

|---|---|---|---|---|
| Operational gate | `lastPingTimestamp` sandbox verification unchecked (test-plan box 4) | Cannot run from this session — Vikram has no prod console; verification is operational, not code. Gating alarm-flip is the right boundary — bot-side audit events flow regardless. | Justin to run `console.log(Object.keys(shard))` on first tick in sandbox before the infra-side `actions_enabled = true` flip. | Low — limited to alarm-flip step; bot-side merge is safe |
| Minor | Heartbeat sampler error log fires per-tick (not edge-triggered) | If `readGatewayHealth` throws on a discord.js shape change, ~2,880 warn lines/day — operational nuisance, not a bug. Tilde-pin on `~14.x.y` reduces likelihood. | Backlog: edge-trigger the sampler-threw warn with a `samplerErrored` flag, mirror healthy/unhealthy pattern. | Low — log-volume only |
| Minor | `prevHealthy` retains its value when sampler throws | Intentional: a transient throw + recovery should not flap the transition log. Reviewer agreed "defensible, just needs a comment so a future reader doesn't 'fix' it." | Add one-line comment in catch block lock-in (deferred — same touch as item above). | None — current behavior is correct |
| Minor | `intervalMs: 0` foot-gun in `startGatewayHeartbeat` / `startActiveGuildCount` | YAGNI — no caller passes 0. Single-line guard if it ever matters. | None — acceptable as-is | None |
| Minor | `handler_duration_ms` comment claims "from `handleCommand` entry" but `handlerStart` is captured *after* the chat-input guard | Sub-microsecond difference in practice — the only path that emits is chat-input. Soften the comment OR move `handlerStart` above the guard. | None — acceptable as-is; the in-code comment about chat-input-only-path is one tweak away if anyone adds work above the guard | None — measurement is correct in practice |
| Discussion | Stale-registration signal loss on non-ack reply failure (`reply_failed` overwrites the "was unknown_command" tag) | Asymmetric vs. `handler_error` precedence and called out in-line. Whether the dashboard needs a stale-reg-survives-reply-failure dimension is an alarm-spec question, not a code question. | Confirm with alarm spec owner: if needed, add a `was_unknown_command` boolean dimension (cheaper than reworking `failure_type` vocabulary) | Low — affects dashboard attribution at a rare edge case, not alarm correctness |
| Phase 2 candidate | `commands.js` is 3600+ lines | Out of scope for monitoring PR. `handleCommand` + helpers would benefit from extraction but that's a refactor PR, not this one. | Backlog: extract `handleCommand` + `isAckTimeoutError` + `emitInteractionMetric` into its own module | None |
| Discussion | snake_case `handler_duration_ms` local in `commands.js` | Audit payload keys MUST be snake_case for CW filter ergonomics. Local could be `handlerDurationMs` and reshaped at boundary; reviewer flagged as "feel free to ignore" | None — subjective/style | None |

### Tracked separately as filed issues
- **#195** — replay-on-resume: `client.isReady()` doesn't flip back on session resume; will produce fake install spike on forced re-IDENTIFY. Acceptable for Phase 1 (no real users).
- **#196** — `active_guild_count` boot-window 0: first datapoint emits before READY hydrates the cache. Self-corrects within 60s; affects future alarm wiring only.

### Convergence notes
- 9 review rounds; convergence criteria (Minor + nits only) met since round 4.
- Test count: 900/900 green, including regression pins for sub-ms precision, edge-triggered transitions, asymmetric `handler_error` precedence, NTP-step clamp, runOnce, and the table-driven regex coverage of `isAckTimeoutError`.
- Tilde pin on `~discord.js@14.x.y` in `package.json` reduces silent-break risk on minor bumps.
- Diff size: ~700 LOC across 8 files, scoped to monitoring instrumentation only.

— via /pr-polish

vikramlayerv:
## pr-polish: round-N pass on Justin's pending items #2 + #5 — closing items

This polish pass covered the new commits adding Justin's #193 §2 (last-event composite) and §5 (dependency_auth_failure). 6 review rounds; convergence at round 6 with explicit "LGTM modulo the minor notes above — none are blocking."

### What this round-N pass changed

| # | Severity | Fix | Round |
|---|---|---|---|
| Bug | `path` cardinality misstated as low — actually carries resource IDs (`/qurls/{id}`). Updated comment to flag as forensic-only / no-promote-to-dimension. | 1 |
| Minor | Consolidated two `now()` calls in `readGatewayHealth` to a single `const t = now()`. | 1 |
| Bug | `ACK_TIMEOUT_MSG_RE` didn't match `DiscordAPIError[10062]: Unknown interaction` (discord.js v14's typical wrapped shape). Updated regex + 2 table-driven test rows. | 2 |
| Bug | Test isolation: outer `beforeEach` ticked activity with real `Date.now()`, inner `beforeEach` then flipped to fake timers — `Math.max(0,…)` papered over the negative diff. Reset + tick AFTER `useFakeTimers()` so both sides use the same clock source. | 2 |
| Pin | `dependency_auth_failure` emit-once invariant — added explicit comment + test asserting fetch was called once + emit fired once on 401. If 401 ever gets added to RETRYABLE_STATUSES, the test breaks before alarm count multiplies. | 2 |
| Doc | `INTERACTION_HANDLED` comment said "Single emission per Discord interaction" but `handleCommand` early-returns on autocomplete + non-chat-input (modal submits, buttons). Tightened to `ChatInputCommand` scope with future-warning. | 3 |
| Perf | Changed `client.on('raw', () => noteGatewayActivity())` to `client.on('raw', noteGatewayActivity)` — saves closure allocation per frame. Made `noteGatewayActivity` defensive (`typeof maybeNow === 'function'`) so discord.js's packet-arg shape doesn't break it. | 3 |
| Doc | `noteGatewayActivity` comment now explicitly notes it's only invoked from the gateway role; if `readGatewayHealth` is ever called from HTTP-only mode, it'll always report unhealthy because `lastGatewayActivityAt` stays at 0. | 3 |
| TODO | Multi-shard `oldestAck` masks partially-down shard sets — added `TODO(multi-shard)` near the loop with the suggested settle-window fix shape. | 4 |
| Tighten | `_test` export gate from `NODE_ENV !== 'production'` to `=== 'test'`. Stops a misconfigured prod task with `NODE_ENV` unset from shipping internal handles. (gateway-metrics.js only — commands.js's pre-existing pattern is out of scope for this PR.) | 4 |
| Pin | Cardinality lock test — `INTERACTION_HANDLED` emitted EXACTLY ONCE per failure scenario (success / handler_error / unknown_command / reply_failed). Existing `toHaveBeenCalledWith` would still pass on duplicate emits. | 4 |
| Doc | Cross-reference at `INTERACTION_HANDLED` constant pointing to the `failure_type` precedence rules in `commands.js`. | 4 |
| Pin | `noteGatewayActivity` packet-arg-shape test — verifies the `typeof` fallback works with discord.js's actual call shape (object, not function). | 5 |
| Pin | Non-auth 4xx (400, 404, 409) does NOT emit `dependency_auth_failure` — pins the auth-only scope of the metric against future regex-match-everything bugs. | 5 |

### Items the latest review still flags — left for the reasons below

|---|---|---|---|---|
| Suggestion | Round-6 #1: boot-window race — first activity tick may not land before the immediate heartbeat sample. | Self-correcting within 30s (next interval picks up the activity). Reviewer wanted a sandbox-deploy confirmation, which is the operational-gate already in the test plan. | Operational confirmation at sandbox apply time. | None — alarm tolerates one missed sample |
| Suggestion | Round-6 #5: tests for non-chat-input interaction types beyond autocomplete (button, modal, select menu). | Today `/qurl` registers no autocomplete + no buttons routed through this handler — those go through their own button/modal collectors. Adding tests for non-applicable paths is over-specification. | Backlog: when a future command uses select menus / modals routed through `handleCommand`, add the equivalent emit + test. | None at current command surface |
| Suggestion | Round-6 #6: `guildCreate` / `guildDelete` try/catch is defensive belt-and-suspenders since `logger.audit` never throws. | Defensive guards on Discord event handlers are project-wide convention (see existing `roleDelete` / `channelDelete` handlers in `discord.js`). Keeping it consistent reduces cognitive load. | None — acceptable as-is | None |
| Suggestion | Round-4 #4: `readGatewayHealth` reads module-global `lastGatewayActivityAt` (could be a third param). | Module-global is the simpler API for callers; adding a third parameter would force every caller to thread the timestamp through and the test-only `_resetGatewayActivity` is the cleaner test seam. | None — acceptable as-is | None |
| Discussion | `commands.js` `_test` export still uses `NODE_ENV !== 'production'` (not tightened to `=== 'test'` like `gateway-metrics.js`). | Pre-existing pattern from before this PR — modifying it is scope creep. Functionally identical for test runs (jest sets `NODE_ENV='test'`). | Backlog: consistency cleanup in a separate PR if anyone touches that file. | None |
| Discussion | Replay detection on `guildCreate` uses `client.isReady()` (forced re-IDENTIFY produces fake install spike). | Documented in code, tracked in qurl-integrations#195 with the `shardReady` / `shardResume` settle-window fix. | qurl-integrations#195 (Phase 2). | Low — bot has no real users in prod yet; dashboard noise on rare reconnects only |
| Discussion | `active_guild_count` first datapoint races cache hydration (count: 0 at boot). | Tracked in qurl-integrations#196 with the "defer first emit until ready" fix. | qurl-integrations#196 (Phase 2). | Low — dashboard-only today, no alarm wired |
| Discussion | Multi-shard `oldestAck` masks stuck-since-boot shards (loop reports healthy from shard 0 alone). | Single-shard today so unobservable. `TODO(multi-shard)` comment pins the assumption near the loop. | Multi-shard rollout PR upgrades the gate (treat -1 sentinel as unhealthy after settle window). | Low — currently impossible to reach |

### Convergence notes

- 6 review rounds; convergence criteria (Minor + Nits only) reached at round 6.
- Tests: 918 green (up from 902 at the start of this polish pass; +16 net for the new bot-side functionality + regression pins added during polish).
- Two follow-up issues (#195, #196) filed for the deferred Phase 2 work.
- Base branch: `feat/gateway-health-endpoint` (stacked on PR #193). Will rebase to `main` once #193 lands.

— via /pr-polish

justin-layerv:
## /simplify pass

**Fixed:** comment density. Pruned pure attribution (`// Justin's review on #193 §X:`) and narration (`// Phase 1 monitoring instrumentation. handler_duration_ms is...`) across 6 files — 39 insertions / 78 deletions. Preserved every load-bearing WHY: NTP-step rationale, EMIT-ONCE INVARIANT, KNOWN LIMITATION (#195) on guildCreate replay, asymmetric failureType precedence, BigInt-truncation note, v8 shape-stability rationale on the raw listener. All 837 tests pass.

**Correction to my earlier report**: I previously claimed there were "pre-existing unrelated test-suite failures (auth0-jwks/oauth/server/multi-tenant)." That was wrong — my local `node_modules` was stale (missing `jose ~5.9.6`). After `npm ci`, all 38 suites and 837 tests pass with exit 0. CI was correctly catching real failures all along.

**Skipped (out of /simplify scope, worth follow-ups):**

1. **Edge-triggered transition logger duplication** — `gateway-metrics.js:168,183-197` mirrors `gateway-health.js:77,117-123` (same `prev === true && !cur` warn / `prev === false && cur` info shape). Worth extracting `createTransitionLogger({ name, onUp, onDown })` into `apps/discord/src/utils/`. Not done here because the refactor touches a sibling module outside this PR's scope.

2. **`DEPENDENCY_AUTH_FAILURE` coverage gap at `commands.js:3218`** — the validate-key path does the same `resp.status === 401 || resp.status === 403` check but doesn't emit the new audit event, so a rotated/invalid key during the most common surfacing path won't fire the alarm. Coverage gap, not a simplification — expanding emission here changes behavior, file as follow-up.

3. **Periodic-task helper** — 9 sites in this repo (`orphan-token-sweeper`, `http-only-init`, `database`, `routes/webhooks`, `commands`, `utils/oauth-rate-limit`, `server`, plus the two added here) repeat the `tick(); setInterval(); unref();` shape. No helper exists. Tech-debt PR, not phase-1 scope.

4. **`client.off('raw', noteGatewayActivity)` shutdown symmetry** — `client.destroy()` already tears down the WebSocket and the listener becomes inert; agent itself called this "zero functional impact." Skipped.

Pushed as `e68eb33`.

PR #197, opened 2026-05-06
fix(discord): trim /qurl setup blurb + surface api_key_limit specifically

vikramlayerv:
## Summary

Two cosmetic/UX fixes to the OAuth setup flow per Vikram's sandbox testing:

1. **`/qurl setup` ephemeral message** trim — drop the prose-heavy "Sign in with your layerv.ai account…no copy-paste needed" blurb and the multi-line browser-binding paragraph. Net delta: 7 lines → 3 lines, keeping the click-here link as the first-class info plus a one-line `_Open in this browser; the link expires in 5 minutes._` that retains the two load-bearing constraints (browser binding, expiry).

2. **`api_key_limit` error surface** — when the admin's qURL account is at the API-key cap, qurl-service returns RFC 7807 `403 / code: "api_key_limit"`. The bot was rendering a generic "qurl-service rejected the API-key request" page; the admin couldn't tell what was wrong without reading CloudWatch. Now: parse the response body, branch on `error.code === "api_key_limit"`, render specific copy that names the cause + the action. The branch logs at `warn` (not `error`) so prod alerting can distinguish user-quota hits from genuine service failures, and renders status **429 Too Many Requests** (semantically correct for quota — qurl-service's 403 is itself off-spec).

   New page copy (no hardcoded tier number — the dashboard is the source of truth for the actual cap):
   > **qURL API key limit reached** — Your qURL account has hit its API key limit. Delete an unused key in your layerv.ai dashboard, or upgrade your plan, then run /qurl setup again.

   Generic 502 branch is unchanged — still the fallback for everything that isn't `api_key_limit`. The fall-through path also tags the log with `problemCode: '<unparseable>'` when the response body isn't JSON at all (vs. parsed-but-no-code), so triage can distinguish the two failure shapes.

## Verification

- `eslint src tests --max-warnings 0` clean
- All 896 existing tests pass; two new tests cover the api_key_limit 429 branch + the fall-through-on-missing-`code` path

## Test plan

- [x] Existing tests pass; new tests pass
- [x] CI green
- [x] Sandbox manual: hit the qURL key cap, confirm new copy renders
- [x] Sandbox manual: run `/qurl setup` in test server, confirm trimmed ephemeral

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3156,11 +3156,7 @@ const commands = [
           return interaction.reply({
             content: '🔐 **Connect qURL to this server**\n\n'
               + `**[Click here to authorize qURL](${startUrl})**\n\n`
-              + 'Sign in with your layerv.ai account and consent. The bot will mint a qURL '
-              + 'API key owned by your account — all usage on this server will bill to you. '
-              + 'No copy-paste needed.\n\n'
-              + '_This link expires in 5 minutes and is bound to your browser session — '
-              + 'open it in the same browser where you ran `/qurl setup`._',
+              + '_The link expires in 5 minutes._',
             ephemeral: true,
           });
         }
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3156,7 +3156,7 @@ const commands = [
           return interaction.reply({
             content: '🔐 **Connect qURL to this server**\n\n'
               + `**[Click here to authorize qURL](${startUrl})**\n\n`
-              + '_The link expires in 5 minutes._',
+              + '_Open in this browser; the link expires in 5 minutes._',
             ephemeral: true,
           });
         }
--- a/apps/discord/src/routes/qurl-oauth.js
+++ b/apps/discord/src/routes/qurl-oauth.js
@@ -331,21 +331,35 @@ router.get('/callback', rateLimit, async (req, res) => {
     });
     if (!mintResp.ok) {
       const errBody = await mintResp.text().catch(() => '');
-      logger.error('qURL API key mint failed', {
-        status: mintResp.status, body: errBody.slice(0, 500), guildId,
-      });
-      // RFC 7807 problem JSON; surface the api_key_limit case specifically
-      // so the admin sees the actual cause + the action they can take,
-      // not a generic "service rejected the request" wall.
+      // RFC 7807 problem JSON; surface specific cases (currently just
+      // api_key_limit) before the generic catch-all so the admin sees
+      // the actual cause + the action they can take, not a "service
+      // rejected the request" wall. Parse-fail / missing-code falls
+      // through to the generic branch.
       let problemCode = '';
       try {
         const parsed = JSON.parse(errBody);
         if (typeof parsed?.error?.code === 'string') problemCode = parsed.error.code;
-      } catch { /* not JSON — fall through to the generic branch */ }
+      } catch { /* not JSON — fall through */ }
+
       if (mintResp.status === 403 && problemCode === 'api_key_limit') {
-        return renderError(res, 403, 'qURL API key limit reached',
+        // User-quota hit, NOT a service failure. Log at warn (not
+        // error) so prod alerting / log-rate alarms can distinguish
+        // "qurl-service down" (logger.error) from "user on Free tier
+        // hit 3 keys" (logger.warn). Status 429 mirrors the actual
+        // semantic (Too Many Requests / quota) — qurl-service's 403
+        // is itself somewhat off-spec for a quota hit; we render the
+        // semantically correct status to the admin's browser.
+        logger.warn('qURL API key mint refused: api_key_limit', {
+          status: mintResp.status, problemCode, guildId,
+        });
+        return renderError(res, 429, 'qURL API key limit reached',
           'Your qURL account has hit its API key limit (Free tier allows 3). Delete an unused key in your layerv.ai dashboard, or upgrade your plan, then run /qurl setup again.');
       }
+
+      logger.error('qURL API key mint failed', {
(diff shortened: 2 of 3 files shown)


vikramlayerv:
## pr-polish: complete (Phase 1 review + Phase 2 simplify + Claude loop converged at round 2)

### What this polish pass changed

| Round | Severity | Fix | Commit |
|---|---|---|---|
| Phase 1 review | Medium (UX) | Re-added browser-binding clause to the trimmed `/qurl setup` ephemeral. The OAuth callback hard-rejects opening the link in a different browser (qurl-oauth.js:214,226 → 400) — without the warning, users burn their 5-min window and hit a wall. New copy is still 3 lines vs. the original 7. | $cadcdfc$ |
| Phase 1 review | High (Observability) | Reordered the api_key_limit branch BEFORE the unconditional `logger.error`. Quota hits now log at `warn` (not `error`) so prod alerting can distinguish "qurl-service down" from "user on Free tier hit cap." | $cadcdfc$ |
| Phase 1 review | Medium (Architecture) | Render status changed from `403` (mirroring upstream's off-spec choice) to `429 Too Many Requests` — RFC 7231 §6.5.3 vs RFC 6585 §4. The bot's render layer translates qurl-service's off-spec status to the semantically correct one. | $cadcdfc$ |
| Phase 1 review | Medium (Tests) | New test pinning the fall-through-on-missing-`code` path (valid JSON, no `error.code` → generic 502). Locks the contract against a future qurl-service field rename. | $cadcdfc$ |
| Phase 1 review | Medium (Tests) | Added `expect(discord.sendDM).not.toHaveBeenCalled()` on the api_key_limit test — failure paths must not distort dispatch / DM metrics. | $cadcdfc$ |
| Phase 1 review | Minor | Added `problemCode` to the generic-error log payload so triage can grep by code regardless of which severity branch fired. | $cadcdfc$ |
| Round-2 + Phase 2 | Medium (Code Quality) | Trimmed comment-density duplication per Phase 2 Agent B (CLAUDE.md: comments earn keep on non-obvious *why*). Kept the load-bearing rationale (warn-vs-error alerting distinction, 429-vs-403 RFC reference); dropped the lines that just restated what the code does. | $fde09cd$ |
| Round-2 + Phase 2 | Minor | Dropped the hardcoded "(Free tier allows 3)" from the page copy. The number goes stale if qurl-service ever changes the cap, and the bot would silently keep showing the wrong value. The dashboard URL is the source of truth. | $fde09cd$ |
| Round-2 + Phase 2 | Minor | The catch path now sets `problemCode: '<unparseable>'` instead of `''` — distinguishes "body wasn't JSON at all" from "JSON but no error.code" in the log payload. | $fde09cd$ |

### Items the latest review still flags — left for the reasons below

|---|---|---|---|---|
| Observation (R2 #1) | `renderError` auto-appends "Run /qurl setup in Discord to start over." to every detail string, producing back-to-back CTAs with the new copy's existing "...then run /qurl setup again." sentence. | Pre-existing — the generic 502 branch already has the same duplication. Refactoring `renderError` to drop the auto-append and let each call site own its CTA is a separate cleanup that affects every error page in this file. | Backlog — fold into a separate `renderError` cleanup PR | None — cosmetic |
| Observation (R2 #2) | Strict `mintResp.status === 403 && problemCode === 'api_key_limit'` pairing means if qurl-service ever moves api_key_limit to status 429 (matching its own RFC 6585 semantic), the bot regresses to the generic 502. Could match on code alone with status as an observability-only field. | Intentional defensive posture today. The status-and-code pairing is the right call until upstream migrates; if/when it does, the bot's sync moment catches it (a 429 in the wild today would NOT be `api_key_limit`). | Revisit at the next qurl-service contract sync | Low — single sync moment, easy to detect |
| Suggestion (R2 #3) | No coverage for the `<unparseable>` log-path sentinel. Two new tests both feed valid JSON; a third with a non-JSON body would lock the sentinel as a log contract. | The fall-through behavior IS exercised by the existing `502s when qurl-service mint fails` test (qurl-oauth.test.js:323) which feeds `text: 'internal error'` (non-JSON). Adding a third dedicated test would only assert the log payload string `<unparseable>` — pure log-shape, no behavioral pin. | None — acceptable as-is | None |
| Suggestion (R2 #5) | Same RFC 7807 problem-code branching pattern likely applies to other qurl-service mint sites (e.g. apps/slack/* equivalents once OAuth lands there). | Out of scope for this Discord-only fix. Apps/slack doesn't have an OAuth flow yet; when it does, the pattern can move into a shared `parseProblemCode` helper. | Backlog — file at the time apps/slack OAuth is wired | None |
| Suggestion (Phase 2 A — M1) | Two `renderError` copies between `qurl-oauth.js:84` and `discord-install.js:62` should converge into a shared helper. | Pre-existing duplication; this PR adds a fifth call site to qurl-oauth's copy without making it worse. | Backlog — track separately as "shared renderError helper" | None |
| Suggestion (Phase 2 A — L1) | RFC 7807 `parsed?.error?.code` shape has no precedent in the JS bot — first call site, so a `parseProblemCode(body)` helper would be premature on N=1. | Sound pattern: extract the helper when the second JS call site appears. | Revisit when qurl-service errors are parsed elsewhere in the bot | None |

### Convergence notes

- 2 review rounds + Phase 2 (3 simplify agents) + Phase 1 (3 review agents) — total 8 review surfaces, all converged.
- 896/896 tests pass throughout (was 894 before this PR; +2 net for the api_key_limit + missing-code coverage).
- Diff: 3 files, ~70 lines net (small focused PR).
- No new external dependencies, no schema changes, no IAM changes.
- Status code change `403 → 429` is a behavior change worth flagging in PR review: any monitoring or analytics keyed on the `/oauth/qurl/callback` response status will see 429 instead of 403 for the api_key_limit class.

— via /pr-polish

PR #198, opened 2026-05-07
feat(discord): show recipient names on /qurl send revoke confirmation

vikramlayerv:
## Summary

When `/qurl send` fans out to N recipients, each gets a unique link (one DDB row per `recipient_discord_id`). Pre-PR the inline "Revoke All Links" button reported just `Revoked X/Y links` — operators didn't know which recipients had links revoked vs. already-opened, and the units didn't actually correspond to user-facing concept (1 resource = up to 10 users).

This PR makes the revoke confirmation user-centric: counts and names list. Plus a Discord 2000-char overflow → file attachment, plus per-guild alias resolution everywhere recipient names render (send + revoke).

## Behavior change — release-note worthy

**Inline-button revoke (post-`/qurl send`)** — count shifts from links → users; new "Revoked for: …" line + Show All toggle + attachment overflow.

> Pre-PR: `Revoked 6/10 links. Note: already-opened links cannot be revoked.`
> Post-PR (small lists): `Revoked 6/10 users. Note: already-opened links cannot be revoked.\nRevoked for: alice, bob, carol, dave, eve +1 more  [Show All]`

**Slash-command `/qurl revoke`** — wording also shifts to user-centric. **Pre-PR**: `Revoked X/Y links` (resource counts). **Post-PR**: `Revoked X/Y users` (recipient counts). No "Revoked for: …" line on this path (the slash flow doesn't have `recipients` in scope). If anyone has dashboards parsing those reply strings, they'd silently see different numbers — flagging explicitly.

**Send confirmation** — `Sent to N users / Recipients: <alias>, …` now uses guild nickname > globalName > username (same chain `resolveSenderAlias` already uses), not raw username.

**After (large lists, > ~80 users — message would exceed Discord's 2000-char content cap)**:
> Revoked 850/1000 users. Note: already-opened links cannot be revoked.
> Revoked for: alice, bob, carol, dave, eve +845 more (see attached)
> 📎 `revoked-users.txt`

Show All button is suppressed in attachment mode (the file IS the full list).

## Implementation

| Change | File |
|---|---|
| `revokeAllLinks` returns `{success, total, successUserIds, failureUserIds}` (was just `{success, total}`); user-centric counts; failure-wins for mixed-outcome users | `apps/discord/src/commands.js` |
| `renderRevokeMsg` (module-scope, unit-tested): handles short/truncated/Show All/attachment paths | `apps/discord/src/commands.js` |
[...]

PR #201, opened 2026-05-08
fix(discord): smoke-test import path drags in discord.js — extract pure-string render module

vikramlayerv:
## Summary

PR #198 added `e2e/tests/qurl-send-revoke.smoke.test.ts` to gate post-deploy on the post-revoke wording. It imported `apps/discord/src/commands.js` directly. After merge, the post-deploy smoke run on `4064e2f` failed:

```
FAIL tests/qurl-send-revoke.smoke.test.ts
  ● Test suite failed to run
    Cannot find module 'discord.js' from '../apps/discord/src/commands.js'
```

The e2e smoke runner installs only `e2e/` deps — `apps/discord/node_modules/discord.js` isn't present in CI, so `commands.js` fails to load when the smoke require()s it.

## Fix

Extract the pure-string wording layer into `apps/discord/src/revoke-render.js`:
- `renderRevokeContent(sendId, names, total, showAll, success)` — returns `{sendId, content, needsExpand, showAll, attachmentText}` (no discord.js components)
- `buildRevokeHeader`, `REVOKE_TRUNC_LIMIT`, `REVOKE_CONTENT_SAFE_MAX`

Only require is `./utils/sanitize` (also discord.js-free).

`commands.js`'s `renderRevokeMsg` is now a thin wrapper: calls `renderRevokeContent` for content + attachment, then builds the discord.js `ActionRowBuilder` Show All button from `needsExpand` / `showAll`. Existing bot-side unit tests against `_test.renderRevokeMsg` keep working.

The smoke test imports from `revoke-render` and loads cleanly under the e2e runner.

## Verification

- `apps/discord` — 934 tests + eslint --max-warnings 0 green
- `e2e` — smoke 6/6 pass locally without discord.js in node_modules
- Bot is already deployed (the merge run passed deploy; only the smoke gate failed). This PR restores the gate.

## Test plan
- [x] `cd apps/discord && npx jest --no-cache && npm run lint`
- [x] `cd e2e && npx jest --no-cache tests/qurl-send-revoke.smoke.test.ts`
- [ ] Post-merge: confirm `qurl-bot-discord: Build and Deploy` → `e2e-smoke / E2E Smoke Tests` reports success on the merge SHA

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #202, opened 2026-05-08
feat(discord): file-attachment overflow for /qurl send confirmation

vikramlayerv:
## Summary

Mirrors the post-revoke attachment-overflow pattern from #198 onto the post-send ephemeral confirmation. Pre-PR, sending a `/qurl send` to 100+ recipients with long display names would generate a `Recipients: alice, bob, …` line that, combined with the `could not be reached: …` line, could exceed Discord's 2000-char message-content cap — Show All click on a large list would error out on `editReply`.

## Behavior change

**Before** (small lists — unchanged):
> Sent to 6 users | Expires: 1h | One-time links
> Recipients: alice, bob, carol, dave, eve +1 more  *[Show All]*

**After (large lists, > ~80 users — full inline would exceed Discord's 2000-char cap)**:
> Sent to 200 users | Expires: 1h | One-time links
> Recipients: alice, bob, carol, dave, eve +195 more (see attached)
> 📎 `recipients.txt`

`recipients.txt` contains both delivered + failed lists:

```
DELIVERED (200):
alice
bob
...

NOT DELIVERED (3):
charlie
dave
...
```

Show All toggle is suppressed in attachment mode (the file IS the full list).

## Implementation

- Extracted the wording from the `buildConfirmMsg` closure into a module-level `renderSendConfirm({delivered, expiresIn, failed, failedNamesPlain, successNames, showAll})` returning `{content, attachmentText, needsExpand}`. Pure string in / out, exposed on `_test`.
- `handleSend` adds `files: [new AttachmentBuilder(recipients.txt)]` to the **initial** `editReply` only when `attachmentText` is non-null. Subsequent monitor link-status ticks and Add Recipients edits omit `files`, so Discord keeps the existing attachment without re-uploading the same blob 15× per minute.
- `needsExpand=false` in attachment mode → Show All button is automatically suppressed.
- `failedUsers` and `successNames` are PLAIN-form via `sanitizeDisplayNamePlain` (already applied inside `resolveRecipientAlias`); the `.txt` lands them verbatim, message-content rendering escapes per name.

## Test plan
- [x] `apps/discord` — 940 unit tests pass (was 934 before this PR; +6 new cases for `renderSendConfirm`)
- [x] `apps/discord` — eslint --max-warnings 0 clean
- [x] Live: `/qurl send` to ≤ ~80 users → confirms inline as before, no attachment
- [x] Live: `/qurl send` to >100 users (long display names) → `recipients.txt` attaches; Show All button absent; opens correctly in Discord client
- [x] Live: monitor link-status updates over time → attachment stays attached, content updates without re-upload
- [x] Live: Add Recipients on a small send that grows past the cap → existing attachment stays (won't auto-update with new names; future enhancement)

## Out of scope

- Add Recipients on an already-attachment-mode send doesn't regenerate the attachment with newly-added names. The existing attachment + monitor link-status banner cover the common case; growing past the cap mid-Add is rare.

## Other UI changes (release-note)

Inline button label "Revoke All Links" → "Revoke All" — same id (`qurl_revoke_${sendId}`), same Danger style. Tighter than the original; "All" qualifier preserved so the action scope (every link in the send) reads explicitly.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1851,7 +1851,7 @@ async function handleSend(interaction, apiKey) {
       .setStyle(ButtonStyle.Primary),
     new ButtonBuilder()
       .setCustomId(`qurl_revoke_${sendId}`)
-      .setLabel('Revoke')
+      .setLabel('Revoke All')
       .setStyle(ButtonStyle.Danger),
   );
   if (needsExpand) {
@@ -1972,7 +1972,7 @@ async function handleSend(interaction, apiKey) {
         const fullMsg = monitor.getFullMsg();
         const updatedRow = new ActionRowBuilder().addComponents(
           new ButtonBuilder().setCustomId(`qurl_add_${sendId}`).setLabel('Add Recipients').setStyle(ButtonStyle.Primary),
-          new ButtonBuilder().setCustomId(`qurl_revoke_${sendId}`).setLabel('Revoke').setStyle(ButtonStyle.Danger),
+          new ButtonBuilder().setCustomId(`qurl_revoke_${sendId}`).setLabel('Revoke All').setStyle(ButtonStyle.Danger),
           new ButtonBuilder().setCustomId(`qurl_expand_${sendId}`).setLabel(showAllRecipients ? 'Show Less' : 'Show All').setStyle(ButtonStyle.Secondary),
         );
         await interaction.editReply({ content: fullMsg, components: [updatedRow] }).catch(logIgnoredDiscordErr);
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1834,11 +1834,11 @@ async function handleSend(interaction, apiKey) {
   // file; message-content rendering escapes per name.
   const failedNamesPlain = failedUsers.map(u => resolveRecipientAlias(u, interaction));
   const buildConfirmMsg = (showAll) => renderSendConfirm({
-    delivered, expiresIn, failed,
+    delivered, expiresIn,
     failedNamesPlain, successNames, showAll,
   });
 
-  let confirmRendered = buildConfirmMsg(false);
+  const confirmRendered = buildConfirmMsg(false);
   let confirmMsg = confirmRendered.content;
   // In attachment mode the file IS the full list, so suppress the
   // Show All toggle — the same shape the post-revoke flow uses.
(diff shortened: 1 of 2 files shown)


vikramlayerv:

Loop closed in 2 review rounds (`0e25f66` → 5 actionable items; `0c20faa` → "Overall LGTM"). Items below are explicit nits / non-blocking observations the reviewer flagged.

|---|---|---|---|---|
| Minor (non-blocking) | `REVOKE_TRUNC_LIMIT` / `REVOKE_CONTENT_SAFE_MAX` are now shared across revoke + send-confirm; `REVOKE_` prefix is misleading. Consider rename to `RECIPIENT_TRUNC_LIMIT` / `DISCORD_CONTENT_SAFE_MAX` | The constants' shared use is documented at the call site (commands.js:1826-1828); rename is mechanical but touches every callsite for a name-only win. Reviewer round-1 explicitly said "defer if you'd rather not churn it." | none — acceptable as-is | none |
| Minor (live-test-plan) | Verify monitor link-status ticks preserve the attachment without re-upload | The bot's monitor + Add-Recipients edit paths omit `files`; per discord.js v14 docs Discord retains existing attachments. Pre-PR revoke flow already relies on the same semantic. Pinning via integration test would require a heavy handleSend mock for marginal gain | manual verify pre-merge per PR description | low — failure mode would be silent attachment disappearance ~15s after send; if it surfaces in live test, the fallback is to set `attachments: <existing>` explicitly on every edit |
| Minor (test) | Boundary test (~80 vs ~95 names) depends on header arithmetic; if the header copy ever changes, the split silently drifts | The test's `expect(under.content.length ≤ 1900)` + `expect(over.attachmentText !== null)` already pin the inversion at the correct semantic boundary. The 80/95 magic numbers are commentary, not asserted. | none — acceptable as-is | none |
| Minor (test gap) | No test for post-escape cap math when escapes ~double name length | The 200-name overflow test blows the cap in either world (pre- or post-escape), so coverage is adequate per reviewer's own analysis | none — acceptable as-is | none |
| Medium (deferred) | `Recipients:` line in handleSend and `Revoked for:` line in renderRevokeMsg duplicate the same truncate-line pattern (Phase 2 Agent A finding) | Cross-function dedup is a natural fit for the shared module landing in #201; doing it here would force this PR to depend on #201 | tracked: #201 lands → fold both into shared `renderTruncatedNames` helper in revoke-render.js | none — both surfaces stay correct independently |
| Medium (deferred) | Add Recipients on attachment-mode send leaves stale `recipients.txt` (header updates but the `.txt` doesn't regenerate) | Pre-existing degraded path; Add Recipients flow rebuilds confirmMsg as header-only string and bypasses renderSendConfirm. Fix is non-trivial (regenerate attachment vs. surface `(updated)` indicator). PR description acknowledges this trade-off explicitly | filed [#203](https://github.com/layervai/qurl-integrations/issues/203) | low — `.txt` shows correct names for the original send; the missing names are visible via Discord's "delivered to N users" header |

943 unit tests + eslint --max-warnings 0 clean. Diff: ~+165/−24 across 2 files. Behavior change is the new attachment overflow path + button rename "Revoke All Links" → "Revoke All" (release-noted in PR body).

— via /pr-polish

PR #204, opened 2026-05-08
fix(discord): import REVOKE_CONTENT_SAFE_MAX in commands.js

vikramlayerv:
## Summary

Hotfix for the broken post-merge build on PR #202 (\`d347566\`):

\`\`\`
apps/discord/src/commands.js:2624:77
error  'REVOKE_CONTENT_SAFE_MAX' is not defined  no-undef
\`\`\`

The constant has lived in \`./revoke-render\` (exported since #198), but the destructuring at \`commands.js:2571\` was missed when #202 added the new use site at line 2624. \`build-and-test\` failed → \`trigger-deploy\` skipped → no image in ECR → prod promote workflow's sandbox-status-check gate would refuse.

## Type of Change

- [x] Bug fix
- [ ] New feature

## Testing

- [x] \`npm run lint\` clean
- [x] \`npx jest --no-cache --testPathIgnorePatterns=e2e\` — 946 tests pass

## Why this is safe

- One-line addition to an existing destructure.
- The exported constant is what 3 test files have been asserting against since #198 — verified by grep.
- No behavior change.

## Follow-up

Once this merges and \`trigger-deploy\` is green, dispatch \`promote-bot-discord-to-prod.yml\` with the new merge SHA.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #205, opened 2026-05-08
chore(ci): page on main-branch CI failure + module-load smoke test

vikramlayerv:
## Summary

Implements items #2 and #3 from the post-mortem comment on #204 — defense-in-depth for the stale-merge-ref regression class.

Item #1 from that comment ("Require branches to be up to date before merging" branch protection) is the actual root-cause fix and is a repo-settings flip — needs admin action separately.

### 1. \`notify-on-main-failure\` job in \`.github/workflows/discord.yml\`

- Runs **only** when \`build-and-test\` or \`docker-check\` fail AND we're on \`main\` from a push event. PR runs are out of scope (PR author already sees their own red checks).
- Posts to Slack **#all-layerv** via \`slackapi/slack-github-action\` with the \`SLACK_WEBHOOK_URL\` repo secret. Block-Kit payload includes commit SHA, run URL, and a hotfix-pattern pointer to #204.
- Slack-not-issue because the team has no alerting attached to GitHub issues — an issue would sit unread. Slack to #all-layerv is the canonical alert channel for this class of regression.

**Pre-merge action item**: ensure the \`SLACK_WEBHOOK_URL\` secret on \`layervai/qurl-integrations\` repo points at the #all-layerv incoming webhook. (If the secret doesn't exist, the step posts nothing — a silent no-op rather than a CI failure, but defeats the purpose.)

### 2. \`apps/discord/tests/module-load.test.js\`

- 4-row \`test.each\` — each of \`commands.js\`, \`qurl.js\`, \`connector.js\`, \`revoke-render.js\` must load without \`ReferenceError\`.
- ESLint's \`no-undef\` rule covers most cases statically, but a CI run on a stale merge-ref can pass lint while the actually-merged code fails — the runtime require gives an independent cross-check that doesn't depend on ESLint having seen the post-merge state.

## Type of Change

- [ ] Bug fix
- [x] CI/CD improvement
- [ ] New feature

## Testing

- [x] \`npm run lint\` clean
- [x] \`npx jest --no-cache --testPathIgnorePatterns=e2e\` — 950 tests pass (was 946, +4 from \`module-load.test.js\`)
- [x] Workflow YAML validated with \`yaml.safe_load\`

## Why this is safe

- Workflow change is additive — new job depends on \`[build-and-test, docker-check]\` and the existing \`trigger-deploy\` job's behavior is untouched.
- New job only fires on \`failure() && main && push\` — a green run won't open noise issues.
- Test file is read-only — uses \`require()\`, doesn't run any side-effecting code beyond what's already in module-init.

## Out of scope

- Branch protection ("Require branches to be up to date before merging") — repo-admin setting flip. The actual root-cause fix; this PR is belt-and-suspenders.
- Slack notification — needs \`SLACK_WEBHOOK_URL\` secret on the repo first.

### DIFF
--- a/.github/workflows/discord.yml
+++ b/.github/workflows/discord.yml
@@ -70,45 +70,45 @@ jobs:
         working-directory: apps/discord
         run: docker build -t discord-bot-test .
 
-  # Pages on post-merge CI failures so a stale-merge-ref bug (or any
-  # other class of regression that slipped past PR-time CI) surfaces
-  # immediately as an actionable issue, instead of staying silent
-  # behind a `trigger-deploy: skipped` like #202 did. The repo doesn't
-  # have a Slack webhook secret wired up, so we file a GitHub issue
-  # via the built-in token — same triaging surface, no secret deps.
+  # Pages on post-merge CI failures via Slack #all-layerv so a stale-
+  # merge-ref bug (or any other class of regression that slipped past
+  # PR-time CI) surfaces immediately, instead of staying silent behind
+  # a `trigger-deploy: skipped` like #202 did.
+  #
+  # Routes through SLACK_WEBHOOK_URL repo secret — the team already
+  # uses Slack for prod alerting and has no GitHub-issue alerting
+  # attached, so an issue would have been silent. Webhook posts to
+  # #all-layerv (the channel the secret is bound to).
   notify-on-main-failure:
     needs: [build-and-test, docker-check]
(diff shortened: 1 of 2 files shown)


vikramlayerv:
Note: this notify step needs `SLACK_WEBHOOK_URL` secret. Will start posting to #all-layerv automatically once Kevin finishes the webhook integration — no code change needed on this side.

PR #208, opened 2026-05-08
feat(discord): auto-recover from zombie WebSocket

vikramlayerv:
## Summary

When the bot's gateway connection enters the zombie-WS state (heartbeats keep ACK'ing but Discord stops pushing event frames), force every shard to reconnect via `shard.destroy({ code: 4000, recover: WebSocketShardDestroyRecovery.Reconnect })`. The reconnect re-IDENTIFYs cleanly and resumes event delivery.

Hit both sandbox + prod **simultaneously** earlier today — `/qurl send` returned "The application did not respond" because the InteractionCreate events never reached the bot, even though `client.isReady() === true` and `client.ws.ping` was healthy. Manual fix was an ECS service restart on each. This automates that recovery.

## Why this matters: the zombie loop is structural, not an outage

The zombie mechanism is independent of any AWS outage — it's a structural Discord ↔ ECS interaction problem that compounds on itself once started:

1. The first task dies for some reason (health check, container crash, OOM, ECS circuit breaker reap, etc.).
2. ECS kills it with SIGKILL — no graceful shutdown handler runs, no Discord gateway "close" frame is sent.
3. Discord's gateway keeps the WebSocket session alive for ~5 minutes waiting for a heartbeat that never comes.
4. ECS spawns the replacement task immediately (`deployment_minimum_healthy_percent = 0`).
5. New task IDENTIFYs to Discord with the same bot token → Discord sees "another connection is using this token" → boots the new connection → container exits → ECS restarts → loop.
6. The loop self-perpetuates: each new task is killed before its WebSocket can take over the session, so the "previous" zombie keeps appearing fresh.

This PR breaks step 1's downstream: the bot self-detects a zombie WebSocket and gracefully tears down + reconnects in-process, before ECS gets involved. No ungraceful kill → no lingering session on Discord's side → next task (whether ECS-driven or in-process recovery) IDENTIFYs into a clean slate.

## Detection

Reuses the existing composite-readiness signal: `activity_age_ms` (last `client.on('raw', ...)` frame, set on every WebSocket frame including heartbeats). When activity is stale past **120s** while `is_ready=true`, every shard is destroyed and the WebSocketManager re-IDENTIFYs.

- Threshold: `RECOVERY_ACTIVITY_THRESHOLD_MS = 120_000` — 2× the existing 60s health threshold so the alarm fires first, recovery follows
- Cooldown: `RECOVERY_COOLDOWN_MS = 10 * 60_000` — prevents a reconnect storm if termination doesn't immediately restore activity
- Per-shard `destroy()` so a future multi-shard config only resets the wedged shard

## API correctness

The first cut passed `{ closeCode: 4000 }` — `@discordjs/ws` reads `code` (not `closeCode`), and the call only triggers a reconnect when `recover` is set. Without `recover`, the shard transitions to `Idle` and stays there. Final shape: `{ code: 4000, reason: '...', recover: WebSocketShardDestroyRecovery.Reconnect }`. Verified against `node_modules/@discordjs/ws/dist/index.js:705-764`.

The `destroy()` return value is awaited via `.catch()` so a rejection from internal teardown doesn't surface as `unhandledRejection` (process crash on Node 22).

## Companion

Pairs with PR #446 in qurl-integrations-infra: extracts `activity_age_ms` from the new `gateway_heartbeat_unhealthy` audit event into a CloudWatch metric, alarms when Max > 120s for 2 consecutive minutes (matches the auto-recovery threshold so alarm fires only when self-heal didn't clear it).

## Test plan

- [x] Unit tests: not-ready skip, no-baseline skip, within-threshold skip, threshold-boundary trigger, single-shard trigger, multi-shard trigger, partial sync-failure, async-rejection handling, no-shards/no-ws guards, cooldown debounce, post-cooldown re-trigger
- [x] All 957 discord tests pass
- [ ] Sandbox deploy + observe normal heartbeat cadence (no false-positive triggers)
- [ ] If next zombie-WS event hits, confirm CloudWatch shows the warn log + shard reconnect within 30s of the activity_age_ms breach

### DIFF
--- a/apps/discord/package-lock.json
+++ b/apps/discord/package-lock.json
@@ -4916,9 +4916,9 @@
       "license": "MIT"
     },
     "node_modules/fast-xml-builder": {
-      "version": "1.1.5",
-      "resolved": "https://registry.npmjs.org/fast-xml-builder/-/fast-xml-builder-1.1.5.tgz",
-      "integrity": "sha512-4TJn/8FKLeslLAH3dnohXqE3QSoxkhvaMzepOIZytwJXZO69Bfz0HBdDHzOTOon6G59Zrk6VQ2bEiv1t61rfkA==",
+      "version": "1.2.0",
+      "resolved": "https://registry.npmjs.org/fast-xml-builder/-/fast-xml-builder-1.2.0.tgz",
+      "integrity": "sha512-00aAWieqff+ZJhsXA4g1g7M8k+7AYoMUUHF+/zFb5U6Uv/P0Vl4QZo84/IcufzYalLuEj9928bXN9PbbFzMF0Q==",
       "funding": [
         {
           "type": "github",
@@ -4927,7 +4927,8 @@
       ],
       "license": "MIT",
       "dependencies": {
-        "path-expression-matcher": "^1.1.3"
+        "path-expression-matcher": "^1.5.0",
+        "xml-naming": "^0.1.0"
       }
     },
     "node_modules/fast-xml-parser": {
@@ -8591,6 +8592,21 @@
         "utf-8-validate": {
           "optional": true
         }
+      }
+    },
+    "node_modules/xml-naming": {
+      "version": "0.1.0",
+      "resolved": "https://registry.npmjs.org/xml-naming/-/xml-naming-0.1.0.tgz",
+      "integrity": "sha512-k8KO9hrMyNk6tUWqUfkTEZbezRRpONVOzUTnc97VnCvyj6Tf9lyUR9EDAIeiVLv56jsMcoXEwjW8Kv5yPY52lw==",
+      "funding": [
+        {
+          "type": "github",
+          "url": "https://github.com/sponsors/NaturalIntelligence"
+        }
+      ],
+      "license": "MIT",
+      "engines": {
+        "node": ">=16.0.0"
       }
     },
     "node_modules/y18n": {
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -174,6 +174,14 @@ const AUDIT_EVENTS = {
   // ack_age_ms < 60000). Missing emissions = wedge.
   GATEWAY_HEARTBEAT: 'gateway_heartbeat_healthy',
 
+  // Negative-signal heartbeat. Emitted every 30 s when the composite
+  // readiness check FAILS — pairs with the healthy event so the
+  // unhealthy snapshot is observable as a metric (carries activity_age_ms
+  // unbounded, where the healthy emission caps at <60s by definition).
+  // The 5/8 zombie-WS incident showed an alarm on Max(activity_age_ms)
+  // is the right shape; that needs unhealthy emissions to fire on.
+  GATEWAY_HEARTBEAT_UNHEALTHY: 'gateway_heartbeat_unhealthy',
+
   // Bot added/removed from a guild. Single emission on the
   // guildCreate / guildDelete event. `guild_id` is in the payload
   // for log-grep / forensic queries; it MUST NOT be promoted to a
--- a/apps/discord/src/gateway-metrics.js
+++ b/apps/discord/src/gateway-metrics.js
@@ -250,6 +250,20 @@ function startGatewayHeartbeat(client, opts = {}) {
           ack_age_ms: snapshot.ack_age_ms,
           activity_age_ms: snapshot.activity_age_ms,
         });
+      } else {
+        // Pair-event for the healthy heartbeat so activity_age_ms is
+        // observable as a metric on EVERY tick. Without this, the
+        // healthy emission caps activity_age_ms at <60s by definition
+        // and a Max(activity_age_ms) > 60s alarm would never fire — yet
+        // 60s+ is exactly the zombie-WS signal we need to alarm on
+        // (5/8 incident). null-safe payload: ack_age_ms is null pre-
+        // first-ack, ping_ms is -1 pre-first-ws.
+        logger.audit(AUDIT_EVENTS.GATEWAY_HEARTBEAT_UNHEALTHY, {
+          ping_ms: snapshot.ping_ms,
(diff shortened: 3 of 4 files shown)


vikramlayerv:

Loop closed at SHA 8edccbd after 6 review rounds. Critical bugs caught + fixed: `closeCode` → `code` (silently dropped option), missing `recover` flag (shard would have gone Idle and never reconnected), async destroy fire-and-forget (would surface as unhandledRejection / process crash on Node 22). Below are remaining items left intentionally.

|---|---|---|---|---|
| Minor suggestion | Add `recovery.reason` to the auto-recovery warn log payload | Today the only `triggered: true` reason is `'zombie_ws'`, so logging it is redundant. Will revisit when a second trigger is added. | none — acceptable as-is | low |
| Minor suggestion | Boot-window null `activity_age_ms` on `gateway_heartbeat_unhealthy` payload | Companion infra PR #446 metric-extracts `$.audit.activity_age_ms`; CloudWatch JSON metric-filter numeric extraction skips datapoints where the field is null/missing rather than coercing to 0, so boot-window nulls don't anchor `Max`. Verified against existing `qurl_service_auth_failure` filter pattern. | none — acceptable as-is | low |
| Minor suggestion | File a tracked issue for the per-shard cooldown map (multi-shard pivot) | Single-shard reality today; the inline `TODO(multi-shard)` already documents the correctness gap (sync-throwing wedged shard alongside healthy shard locks recovery out for 10min). Will file when sharding work actually begins to keep it from drifting with the code. | backlog: file before sharding flips on | deferred |
| Minor suggestion | Add a "no shards exist OR all destroys failed" rate-limit cooldown (60s) to prevent log-spam | Today the no_shards path retries every tick, but on-call gets paged via the `gateway_activity_age_ms_high` alarm in #446 anyway, so the spam is observable. Adding a separate short-cooldown branch increases code complexity for a pathological case. | none — acceptable as-is | low |
| Minor suggestion | Surface skip reasons (`not_ready` / `no_activity_baseline` / `within_threshold` / `cooldown`) at debug-level | Optional reviewer suggestion — would help debug "why didn't it recover?" investigations. Skipping for now to avoid log chatter; revisit if a real "did not recover" incident lands. | none — subjective/style | none |
| Nit | Microtask-flush pattern (two `await Promise.resolve()` vs `flushPromises`) | Works today; `flushPromises` would be more durable but the two-await pattern is widely used in the codebase. | none — subjective/style | none |

Bot-side dual-emission contract documented in the file header + `startGatewayHeartbeat` docstring. Boundary + multi-shard + partial-failure + integration-wiring tests all pinned. CI green; 961 tests pass.

— via /pr-polish

justin-layerv (changes requested):
Two items I'd like fixed before merge — both small and local to `gateway-metrics.js`.

### 1. Reset `lastGatewayActivityAt` at the destroy site

After `lastRecoveryAttemptAt = t`, also clear the activity baseline. Between `is_ready` flipping back to `true` post-reconnect and the new connection's first `raw` frame, a tick can see `is_ready=true` with the *pre-destroy* `activity_age_ms` still climbing. The 10-min cooldown debounces same-incident today, but if the cooldown is ever lowered (or sharding flips on with a per-shard cooldown map per the existing TODO), this becomes a flap source.

The existing `lastGatewayActivityAt === 0` → `activity_age_ms: null` → `no_activity_baseline` short-circuit already handles the reset cleanly: `readGatewayHealth` returns null, `maybeAutoRecoverZombieWS` early-returns on `no_activity_baseline`, and the next real `raw` frame re-baselines via `noteGatewayActivity()`. One line at the destroy site (`lastGatewayActivityAt = 0`, or call the existing `_resetGatewayActivity` since you're in-module).

Please pin with a test asserting a post-destroy snapshot reports `activity_age_ms === null` until `noteGatewayActivity()` fires again — symmetric with the existing `_resetGatewayActivity` test surface.

### 2. Bound the "all shards sync-threw" branch

The current code collapses two distinct states into the same `no_shards` return:

- **Empty/missing shards** — pathological, rare, retry-next-tick is correct.
- **Shards exist but every `destroy()` sync-threw** — also rare, but if `@discordjs/ws` ever drifts on a future bump and every call throws, the recovery path fires on every 30s heartbeat *forever* with no cooldown stamp. Warn log spams; no recovery happens.

Activity-age alarm pages on-call regardless, but unbounded warn cadence in CloudWatch Logs ingestion is the kind of cost we eat without noticing. Two options, either is fine:

- Split the branches: keep `no_shards` for empty/missing (retry next tick); add `all_destroys_threw` that stamps a short rate-limit cooldown (60s feels right — long enough to bound log volume, short enough that recovery retries well inside the 10-min main cooldown).
- Track `shardsAttempted` alongside `shardsTerminated`; stamp a 60s cooldown when `shardsAttempted > 0 && shardsTerminated === 0`.

Worth closing now while the call site is fresh, rather than catching it during a future @discordjs/ws bump.

---

Approach itself looks good — flagging these two as the remaining correctness/cost gaps before merge.

vikramlayerv:
Both items addressed in `4adc59c`:

1. **Activity baseline reset at destroy site** — added `_resetGatewayActivity()` call right after `lastRecoveryAttemptAt = t`. Subsequent ticks see `activity_age_ms: null` until the reconnected WebSocket's first raw frame re-baselines via `noteGatewayActivity()`. Test added: "successful recovery resets activity baseline".

2. **Bound the all-shards-sync-threw branch** — split the prior single `no_shards` return into two:
   - `no_shards` (shardsAttempted=0): unchanged, retry next tick, no cooldown.
   - `all_destroys_threw` (shardsAttempted>0, shardsTerminated=0): stamps `lastRecoveryRateLimitAt = t` against a new `RECOVERY_RATE_LIMIT_COOLDOWN_MS = 60_000`. Bounds warn-log volume; recovery still retries inside the 10-min main cooldown.

   Existing sync-throw test rewritten to assert the new reason + cooldown gating; added a test for the rate-limit cooldown elapsing.

963 tests pass (was 961), lint clean.

PR #212, opened 2026-05-09
fix(discord): restore multi-user select on /qurl send "A specific user" target

vikramlayerv:
## Summary

Regression fix: the user-select dropdown on `/qurl send`'s "A specific user" target hardcoded `setMaxValues(1)` — a recent change collapsed what used to be a checkbox-style multi-select. The "Add more recipients" button flow on the channel target still allowed up to 10 picks, so the regression only affected the user-target path.

## Affects both chat-server and voice-server

The same user-select component is shared between both contexts. The chat-vs-voice distinction lives in the *channel*-target branch's labels ("Everyone in this voice channel" / "Everyone in this channel"), not in the user-target branch. A single change fixes both server types.

## Changes (all in `apps/discord/src/commands.js`)

| Line | Change |
|---|---|
| 1318 | option description: "Pick one user to send to" → "Pick one or more users to send to" |
| 1325-1330 | `setMaxValues(1)` → `Math.min(10, config.QURL_SEND_MAX_RECIPIENTS)`; placeholder "Pick a user" → "Pick one or more users" |
| 1359 | send-button gate `recipients.length === 1` → `recipients.length >= 1` (user-target only) |
| 1485-1505 | handler reads `[...compInt.users.values()]` instead of `.users.first()`; bot/self validation applies across the full set, fail-loud if ANY pick is invalid |

`Math.min(10, QURL_SEND_MAX_RECIPIENTS)` mirrors the existing "Add more recipients" button flow's per-click cap (commands.js:2058) so behavior is consistent across both add-recipient surfaces.

## Test updates

- 3 existing `target=user` tests refactored to a new `mockUsers()` helper that provides both `.first()` and `.values()` Collection shapes (backwards-compat with any future code path that uses either).
- **New test**: `target=user picks MULTIPLE recipients, send fan-outs to all of them` — pins 3-pick → 3 DMs with correct recipient IDs. Catches future regressions back to single-pick.
- Reframed bot/self rejection tests as "rejects the whole selection if ANY pick is X" to document the fail-loud-don't-partial contract.

## Test plan

- [x] `npx jest --no-cache` — 928 pass (74 in `qurl-send-state-machine.test.js`, including the new multi-pick assertion)
- [ ] Manual sandbox test: invoke `/qurl send` from a text channel + voice channel, confirm the user-select dropdown allows multiple picks in both
- [ ] Verify each picked user receives a DM with their own qURL

## No overlap with in-flight PRs

- #455 (qurl-roundtrip canary) — workflow file only
- #456 (canary EventBridge) — terraform only
- #139 (bot canary route) — `apps/discord/src/routes/canary.js` + utils + tests
- #462 (grafana cross-account) — terraform only

This PR is the only one touching `apps/discord/src/commands.js` + `apps/discord/tests/qurl-send-state-machine.test.js`.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -365,6 +365,12 @@ const EXPIRY_LABELS = {
 
 const EXPIRY_CHOICES = Object.entries(EXPIRY_LABELS).map(([value, name]) => ({ name, value }));
 
+// Per-pick cap on UserSelectMenuBuilder.setMaxValues. Discord's hard
+// limit is 25; capping at 10 bounds the UX. Both the initial user-
+// target select AND the channel-target's "Add more recipients" flow
+// use this — keep them in lockstep so a future bump doesn't drift.
+const USER_SELECT_PER_PICK_CAP = 10;
+
 function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessage }) {
   // sanitizeDisplayName: NFKC + bidi/zero-width strip + markdown escape
   // + 64-char cap + 'Someone' fallback. Same helper used at the channel
@@ -1321,18 +1327,12 @@ async function handleSend(interaction, apiKey) {
     ));
 
     if (target === 'user') {
-      // Multi-user select. Per-pick cap mirrors the channel-target's
-      // "Add more recipients" flow (~line 2058) — Discord allows up
-      // to 25 in one user-select, but capping at 10 keeps the UX
-      // bounded and matches the existing add-more cadence. The
-      // per-send envelope cap (config.QURL_SEND_MAX_RECIPIENTS) is
-      // re-checked downstream alongside the channel-target path.
       rows.push(new ActionRowBuilder().addComponents(
         new UserSelectMenuBuilder()
           .setCustomId(ids.userSelect)
           .setPlaceholder('Pick one or more users')
           .setMinValues(1)
-          .setMaxValues(Math.min(10, config.QURL_SEND_MAX_RECIPIENTS))
+          .setMaxValues(Math.min(USER_SELECT_PER_PICK_CAP, config.QURL_SEND_MAX_RECIPIENTS))
       ));
     }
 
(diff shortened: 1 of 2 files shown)


vikramlayerv:
## pr-polish: closing comment — items intentionally left

Loop closed at SHA c552b25 after 1 polish round. Findings caught + fixed: extracted `USER_SELECT_PER_PICK_CAP = 10` constant (was a magic 10 in two places), trimmed verbose inline comments per house style, added REPLACE-vs-APPEND intent comment + corresponding pinning test, added exactly-at-cap (10-user) test. Below: items remaining unaddressed.

| Severity | Finding | Why accepted | Future plan | Risk |
|---|---|---|---|---|
| Nit | 5 older test sites still inline `values: jest.fn(() => [...][Symbol.iterator]())` instead of the new `mockUsers()` helper | Cosmetic — those tests pass and are out of scope for the regression fix. Migration is mechanical when those tests get touched for other reasons. | none — opportunistic on next edit | none |
| Nit | If a selection contains both a bot AND sender, only the bot warning fires (bot check returns before self check) | Not a bug — re-picking surfaces the second issue on the next iteration. Aggregating into one warning adds complexity for a corner case. | none — acceptable as-is | none |
| Nit | `recipients.length >= 1` at line 1359 vs channel-branch's `recipients.length > 0` on the same line | Functionally identical (both reject 0); leaving as-is for diff minimality. Either form passes. | none — bikeshed | none |
| Pre-existing | Rejecting a bot/self selection leaves the prior valid `recipients` intact (Send stays enabled) | Pre-existing behavior shared with the old single-pick handler. "Don't clobber a valid prior selection" is arguably safer; the warning text guides re-pick. | none — pre-existing, separate concern if revisited | none |

930 tests pass. No overlap with in-flight #455, #456, #139, #462.

Manual sandbox check still pending per the test plan: invoke `/qurl send` from text + voice channels, verify multi-pick UI + N-DM delivery.

— via /pr-polish

Issue #211, opened 2026-05-09
Migrate routes/webhooks.js inline bad-sig throttle to utils/bad-sig-throttle.js


vikramlayerv:
After PR #139 merges, two routes have a per-IP bad-signature throttle pattern: routes/canary.js (uses createBadSigThrottle() from utils/bad-sig-throttle.js, extracted in #139) and routes/webhooks.js (has the SAME pattern inlined). Migrate webhooks.js to the helper.

Done in a separate PR (not bundled with #139) so the canary route and the security-sensitive webhook route's diff history don't entangle. The helper's contract is a superset (configurable windowMs / maxPerWindow) so the migration is mechanical — webhooks.js's defaults match createBadSigThrottle()'s defaults exactly.

Tracked from #139 simplify pass.

justin-layerv:
Closing in favor of #217, which tracks the remaining follow-up after the original consolidation plan changed.

PR #215, opened 2026-05-10
feat(discord)!: drop HMAC verify on /canary/exec — rely on qURL/NHP knock

vikramlayerv:
> **🚫 DO NOT MERGE OR DEPLOY** until the following are merged + applied (sandbox bake first, then prod):
>
> 1. **`layervai/qurl-integrations-infra#455`** — qurl-service round-trip canary (parallel; useful disambiguator before #456 fires).
> 2. **`layervai/qurl-integrations#139`** — this branch's parent. Adds bot's `/canary/exec` route with HMAC verify + ts replay-window + bad-sig throttle. The Lambda in #456 needs the route in place.
> 3. **`layervai/qurl-integrations-infra#456`** — Lambda mints + resolves a qURL targeting the bot, then POSTs the canary body. Still HMAC-signs the body during cutover so #139's verifier keeps passing.
>
> **And before this is enabled outside sandbox:**
>
> 4. **`#216`** — replace the `isOpenNHPActive` mount-gate with a dedicated `CANARY_ENDPOINT_ENABLED` env so flipping `ENABLE_OPENNHP_FEATURES` for an unrelated reason can't accidentally surface `/canary/exec` in a non-NHP-fronted environment. The current gate works for sandbox bake; #216 is the prerequisite for prod enablement.
>
> Merging this PR earlier removes the bot's HMAC fallback while #456 still depends on it as a cutover safety net.
>
> Per `project_canary_nhp_pivot_may09.md`.

## Why

The pivot Justin approved on 2026-05-09: the canary should authenticate via NHP through a qURL, not via an HMAC-signed HTTP path. PR #456 already pivots the Lambda to mint+resolve+POST through the qURL/NHP path. Once #456 has applied + baked in sandbox, the bot's HMAC verifier becomes redundant — the NHP knock is the auth.

Reduces always-open HTTP attack surface on `/canary/exec`. Aligns the canary with qurl-service's own security model (eat-our-own-dogfood).

## What

`apps/discord/src/routes/canary.js`:
- Replaces `verifyCanarySignature` middleware with a small `verifyCanaryTimestamp` that only validates the `X-Canary-Timestamp` header + 5-min drift window. Strict 10-digit regex closes the `parseInt('1234567890extra', 10)` footgun before any drift comparison runs.
- Drops the `crypto` import, `SIG_SHAPE_RE` regex, `bad-sig-throttle` integration, and the `req.rawBody` requirement.
- Adds `logger.warn` with `reason` + `drift_seconds` (signed) + `ip` to every 401 path so on-call has a correlatable log when an alarm fires.
- Header comment updated to describe the qURL/NHP-knock auth model. Replay-trade comment carries an explicit cross-repo dependency marker on qurl-service's one-time-use contract.

`apps/discord/src/server.js`:
- `/canary` router + `/canary/*` JSON parser both gate on a single shared `canaryEnabled` constant — drift between them would silently revert the 4 KB body cap.
- Drops the `req.rawBody` capture on `/canary/*` (no HMAC to compute).
- Boot-time `logger.warn` lines flag (a) unset `QURL_API_KEY` and (b) empty `CANARY_RECIPIENT_USER_IDS` when the canary mounts — operators see the misconfiguration at startup, not on the first canary fire hours later.
- `LOAD-BEARING ASSUMPTION` comment names the `isOpenNHPActive`-as-NHP-fronting-proxy dependency and cross-references `#216`.

`apps/discord/src/config.js`:
- Removes the `CANARY_SHARED_SECRET` entry and its block comment.
- Updates the `CANARY_RECIPIENT_USER_IDS` comment to reflect the new threat model (defense-in-depth against a qURL/NHP bypass, not HMAC exfiltration).

`apps/discord/jest.config.js`:
- Per-file coverage floor on `src/routes/canary.js` at 90/85/90/90. Current coverage measured at 95+/95+/100/95+; the per-file floor avoids dilution by ~50 other source files in the global average.

Deletes:
- `apps/discord/src/utils/bad-sig-throttle.js` (the HMAC-compute-bound rate-limiter — no HMAC compute to bound).
- `apps/discord/tests/bad-sig-throttle.test.js`.

`apps/discord/tests/canary-route.test.js`:
- Drops every HMAC-pass/fail/throttle/sig-shape/timing-safe test case (~315 lines).
- Drops the `signedHeaders` helper, the `SECRET` constant, and the `crypto` import.
- All dispatch + differentiated-path + allowlist tests retained, updated to use only `X-Canary-Timestamp`.
- New cases pin: empty-string header, digit-prefixed garbage rejection, ±301s symmetric drift boundary, ~200s well-inside acceptance, `latency_ms` absence on pre-auth 401s, and `logger.warn` emission on the missing-timestamp path.

`apps/discord/tests/multi-tenant.test.js`:
- Pins `POST /canary/exec → 404` when `isOpenNHPActive` is false (route NOT mounted).
- Pins `POST /canary/exec → 401 missing_timestamp` (NOT 404) when single-tenant + `ENABLE_OPENNHP_FEATURES=true` (route IS mounted). Both directions of the gate need to fail CI on regression.

`apps/discord/tests/module-load.test.js`:
- Smoke-loads `src/routes/canary` so a stale-merge ReferenceError surfaces on CI rather than at boot.

## Threat model after this PR

| Attack | Defended by |
|---|---|
| Direct hit on `/canary/exec` from internet | NHP firewall — only IPs with a recently-resolved qURL targeting bot URL can reach the route |
| Replay of a captured request past the qURL's lifetime | qURL one-time-use semantics + 5-min `X-Canary-Timestamp` window (symmetric — past AND future skew rejected) |
| DM blast against arbitrary user IDs | `CANARY_RECIPIENT_USER_IDS` allowlist (kept) + 4 KB body cap (kept) + boot-time warn on empty allowlist |
| Spam / CPU-burn on the public route | NHP firewall (only knocked-in IPs reach the route — bounded by qURL mint rate per API key) |
| Accidental mount in a non-NHP-fronted env | `isOpenNHPActive` gate (current) + `#216` follow-up to swap to `CANARY_ENDPOINT_ENABLED` before prod enablement |

The lost defense is **HMAC over the body** — protects against an attacker who somehow gets past the NHP knock from forging arbitrary canary bodies. Justin accepted that risk on the design call: any caller past NHP has already proved possession of a valid qURL access token for the bot URL, and the recipient allowlist + idempotent canary semantics bound the blast radius.

## Test plan

- [x] `npx jest tests/canary-route.test.js` — 25 cases pass.
- [x] `npx jest` (full discord suite) — 981 / 981 across 43 suites.
- [x] Per-file coverage on `routes/canary.js` clears the new 90/85/90/90 floor (measured 95+).
- [ ] Reviewer confirms PR base is `feat/canary-exec-endpoint` (not `main`) — diff shows only the HMAC-rip, not the underlying #139 route addition.
- [ ] After #456 deploys + bakes in sandbox, re-target this PR's base to `main` and re-run CI.
- [ ] Before non-sandbox enablement: confirm `#216` (`CANARY_ENDPOINT_ENABLED`) is merged so the `isOpenNHPActive` mount-gate is no longer the load-bearing assumption.

## Blast radius

Single repo. No infra, no Terraform, no IAM, no migrations. The `/canary/exec` route stays mounted at the same path with the same body shape and the same response shapes — only the auth model changes. ECS rollback to a pre-merge image is a force-new-deployment.


vikramlayerv:

Seven review rounds. Final state: `LGTM modulo the documented #216 precondition for non-sandbox enablement` (round 7). All blocking-class concerns from prior rounds are addressed; below is the triage on items left open, plus the two follow-up issues filed.

### Follow-up issues filed

| # | Issue | What it tracks |
|---|---|---|
| 1 | layervai/qurl-integrations#216 | Replace the `isOpenNHPActive` mount-gate with a dedicated `CANARY_ENDPOINT_ENABLED` env. Soft-blocker for non-sandbox enablement. |
| 2 | layervai/qurl-integrations#217 | Decide the disposition of `webhooks.js`'s inline bad-sig throttle (re-extract on second consumer vs. document as canonical inline copy). Closes the orphaned #211 tracker. |

### Items left unresolved

|---|---|---|---|---|
| Medium (documented) | `isOpenNHPActive` is a feature flag, not a network-fronting guarantee. | DE would accept: a comment cross-reference at the mount site discloses the dependency; rollout checklist sequences #216 before non-sandbox use; `multi-tenant.test.js` pins both directions of the gate. The current state works for sandbox bake. | #216 (filed) | Sandbox-scoped only — explicitly gated in the PR's DO-NOT-MERGE list. |
| Minor suggestion | Empty-body comment "no operator can curl this directly" inherits the load-bearing NHP-fronting assumption. | Mount-site comment owns the disclosure. The empty-body comment is a downstream consequence; duplicating the dependency disclosure in two places drifts faster than it documents. | None — subjective wording call. | None. |
| Minor suggestion | `drift_seconds: null` on `bad_timestamp` / `missing_timestamp` for log-schema uniformity. | The reasons differ — `expired_timestamp` has a computed drift to log; the others do not. A null field for shape uniformity is a Splunk/Datadog dashboard concern that the producer doesn't own; consumers can default the field. | None — operational nicety. | None. |
| Nit | Past-side regex cliff (rejects pre-2001 timestamps). | The cliff is a feature, not a bug — clock catastrophically wrong → `bad_timestamp` is exactly right. The forward-looking-only comment frame is a doc accuracy choice; rephrasing wouldn't change behavior. | None — subjective wording. | None. |
| Nit | `verifyCanaryTimestamp` comment block heavy for the function. | The omission rationales (no `latency_ms`, no IP-gating, no rate limit) are security-review context that grows during refactors precisely *because* they're load-bearing. Extracting to a "design notes" block above would split the rationale from its enforcement site. | None — subjective placement. | None. |
| Nit | Extra boot-warn for malformed `CANARY_RECIPIENT_USER_IDS` entries (typos, comma glitches). | Runtime `400 invalid_recipient_user_id` already covers the case; boot-time pre-validation is a fail-fast nicety we can add when we see the misconfiguration in the wild. | Backlog (no issue — too small). | None. |
| Nit | Coverage of boot-warn lines themselves. | Booting `server.js` with mocked env to assert on `logger.warn` arguments is more invasive than the value warrants — Claude explicitly flagged this as "skippable; flagging for completeness." | None — explicit skip. | None. |
| Nit | `multi-tenant.test.js` couples "multi-tenant" with "NHP off" implicitly. | Today the conditions coincide; the assertion (route 404s when isOpenNHPActive is false) holds either way. If NHP is ever enabled in multi-tenant, the test would fail for the right reason and force re-evaluation. | None — accept the implicit coupling for now. | None. |
| Nit | Year-2286 cosmetic on the `^[0-9]{10}$` regex. | Present-day `^[0-9]{10,11}$` would be just as strict; the tighter form gets a Y2286 footnote that isn't worth a code change today. | None. | None. |
| Nit | Reciprocal "canary depends on this" marker on qurl-service's `atomicConsumeOneTimeUse`. | Cross-repo doc dependency; out-of-scope here. | Backlog (qurl-service-side). | None. |

### Confidence

- Tests: 981 / 981 across 43 suites.
- Per-file coverage on `routes/canary.js`: 93.68 / 94 / 100 / 95.6 — clears the new 90/85/90/90 floor.
- Diff scope: 6 files in `apps/discord/`. No infra, no Terraform, no IAM, no migrations.
- Auth-model pivot fully documented in PR body's threat-model table; load-bearing assumptions called out at the enforcement sites and cross-referenced to follow-up issues.
- DRAFT state held until upstream gating PRs land per the DO-NOT-MERGE checklist.

— via /pr-polish (7 rounds)

vikramlayerv:
Pushed `c7f94cf` extending this PR's scope: drops the X-Canary-Timestamp / ts-replay window too, per Vikram's confirmation that NHP's `OpenTime` subsumes app-layer replay defense.

Diff:
- `src/routes/canary.js`: drop `verifyCanaryTimestamp` middleware, `TIMESTAMP_TOLERANCE_SECONDS`, route gating.
- `tests/canary-route.test.js`: drop the 7-test timestamp-replay `describe` block + `tsHeaders` helper; strip `.set(tsHeaders())` from the 20 remaining dispatch tests.
- Local: 20/20 tests pass.

Paired with [qurl-integrations-infra#493](https://github.com/layervai/qurl-integrations-infra/pull/493) (Lambda pivots to NHP-direct knock; drops both signature + timestamp on the sending side). Don't merge alone — canary auth shifts from app-layer to network-layer entirely.

Updated PR title would now be more like `fix(discord)!: drop canary auth entirely — NHP knock IS the auth` if we want it to reflect the combined scope. Can update on request.

vikramlayerv:
Superseded by #242 (Path B per Vikram). #242 consolidates this PR's HMAC + ts/replay drops directly into the endpoint's initial landing — no HMAC code ever lands in main. The commit `c7f94cf` added here (drop ts/replay) is rolled into #242's single commit. Approvals/reviews preserved in this PR's history.

PR #218, opened 2026-05-10
fix(discord): log md5_prefix instead of full MD5 on upload

justin-layerv:
## Summary

Truncate the MD5 hash of uploaded files to 8 hex chars before the bot logs it, and rename the log-meta field from `hash` to `md5_prefix` so the truncation is explicit at the reader. All three upload-success paths in `apps/discord/src/connector.js` (`uploadToConnector`, `reUploadBuffer`, `uploadJsonToConnector`) now go through a single `md5Prefix(hash)` helper that carries the rationale on its docstring.

## Why

The full MD5 of an uploaded file is content-derived and treated as sensitive in our broader infrastructure. The bot was emitting it on every upload-success line. Truncating to 8 hex chars (32 bits) preserves cross-system log correlation without the bot's log destination receiving a sensitive value. Detailed threat model lives in internal security docs.

## What changed

- New `md5Prefix(hash)` helper at the top of `connector.js`. Returns the first 8 hex chars or `undefined` for non-string input. Single chokepoint.
- Three call sites switched to `md5_prefix: md5Prefix(result.hash)`.
- Field rename `hash` → `md5_prefix` makes the truncation visible at the reader.
- 7 new unit tests in `apps/discord/tests/connector-coverage.test.js` (one each for the three upload paths, one for missing hash, three `it.each` cases for non-string schema drift).

## Test plan

- [x] `npx jest tests/connector-coverage.test.js` — 29/29 pass
- [x] `npm run lint` clean
- [x] CI green
- [x] Verified no infra-side metric filters / dashboards reference the old `hash` field
- [ ] After merge: confirm the bot's log destination no longer emits a `hash` field on upload-success lines

### DIFF
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -6,6 +6,18 @@ const { sanitizeFilename } = require('./utils/sanitize');
 const { MAX_FILE_SIZE } = require('./constants');
 const MAX_CDN_REDIRECTS = 3;
 
+// Truncate the connector's MD5 of an uploaded file for CloudWatch logging.
+// The full MD5 + WM_SECRET reconstructs a fileviewer URL (handler.go
+// validateWatermarkSig signs md5|wm|ts), so anyone with logs:GetLogEvents on
+// this CloudWatch group could exfiltrate user uploads. 8 hex chars (32 bits)
+// is enough for cross-system correlation; the connector itself still logs
+// the full hash to journald (SSM-only access). This helper is the single
+// chokepoint for that invariant — every upload-success log path goes through
+// it. Don't inline `result.hash` back into a log call.
+function md5Prefix(hash) {
+  return typeof hash === 'string' ? hash.slice(0, 8) : undefined;
+}
+
 // Fetch from a Discord CDN URL with manual redirect handling. `redirect:
 // 'error'` would refuse legitimate Discord redirects (cdn.discordapp.com
 // sometimes 302s to media.discordapp.net). This walks the redirect chain
@@ -207,13 +219,8 @@ async function uploadToConnector(sourceUrl, filename, contentType, apiKey) {
     throw new Error('Connector upload returned no resource_id');
   }
 
-  // md5_prefix (not full hash): full MD5 + WM_SECRET reconstructs a fileviewer
-  // URL (handler.go validateWatermarkSig signs md5|wm|ts), so anyone with
-  // logs:GetLogEvents on this CloudWatch group could exfiltrate user uploads.
-  // 8 hex chars (32 bits) is enough for cross-system correlation; the connector
-  // still logs the full hash to journald (SSM-only access).
   logger.info('Uploaded to connector', {
-    md5_prefix: result.hash?.slice(0, 8),
+    md5_prefix: md5Prefix(result.hash),
     resource_id: result.resource_id,
   });
 
@@ -254,7 +261,7 @@ async function reUploadBuffer(fileBuffer, filename, contentType, apiKey) {
   }
 
   logger.info('Re-uploaded to connector (new resource)', {
-    md5_prefix: result.hash?.slice(0, 8),
+    md5_prefix: md5Prefix(result.hash),
     resource_id: result.resource_id,
   });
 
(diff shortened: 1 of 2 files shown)


justin-layerv:
**/simplify** ran across three reviewers (reuse / quality / efficiency). Two minor fixes landed in commit 4547e03 and the rest were either false positives or correctly out of scope:

**Fixed:**
- Comment line-wrap typo (a duplicated word that survived a line break in the rationale comment).
- Dropped a redundant header line in the test file's describe-block comment that just restated the describe name.

**Surfaced as follow-ups (not in this PR):**
- 64+ identical `jest.mock('../src/logger', ...)` blocks across the test suite would benefit from a shared `tests/helpers/buildLoggerMock.js` paralleling the existing `buildConfigMock.js`. Out of scope here; this PR's new describe matches the file's existing pattern.

**Considered and rejected:**
- Adding `hash` to `logger.js`'s `REDACT_SUBSTRINGS` would over-redact (legitimate non-MD5 `hash` fields exist). Renaming the field to `md5_prefix` at the call site is the correct surface.
- Extracting a helper for three call sites — flagged as premature DRY at the time. Reviewer #1 changed my mind on this; the helper landed in commit e3a5fd7.

justin-layerv:
## CR feedback addressed

All five points triaged. One is now in commit e3a5fd7, the rest are deferred with reasoning below.

### Addressed

**1. Rationale context lives only at first call site** → fixed via commit e3a5fd7

Extracted a small `md5Prefix(hash)` helper at the top of `connector.js`. The rationale docstring now travels with every call site by name, and the 3x repetition of `?.slice(0, 8)` collapses. Reverses an earlier /simplify call where I argued "premature DRY for 3 sites" — the abstraction encodes a load-bearing invariant, not just structural sharing.

**5. Wording in PR description** → updated above.

### Deferred

**2. Naming consistency `md5_prefix` vs house-style** — reviewer themselves marked "not worth changing on its own; flagging for the next time." Same log line already mixes styles, so switching `md5_prefix` would just create a different inconsistency next to its sibling key. Worth a separate normalization pass for all log fields, not piecemeal here.

**3. `md5_prefix: undefined` test pinning** — verified: Jest's `toHaveBeenCalledWith` treats `{ x: undefined }` and `{}` as equal, so the test isn't actually brittle in practice. Reviewer marked this "minor."

**4. Defense-in-depth: redact at logger layer** — reviewer explicitly said "follow-up, not this PR." Tracked separately.

justin-layerv:
## Round 2 CR feedback addressed

Latest review's two actionable points handled in commit edf5bed and pre-merge verification below.

### Addressed

**Test gap on the `typeof hash === 'string'` guard** → fixed via commit edf5bed

Added three parametrized cases (null / number / object) so the guard is pinned against schema drift. Without the guard, a future connector returning `hash: <Buffer>` would have its own `.slice()` semantics and produce a usable byte slice — exactly the regression the test now catches.

**Operational follow-up: infra-side queries on the renamed `hash` field** → checked, clean

Grepped the relevant infra surface for any structured-log filter or dashboard keying on the field name being renamed. No matches outside Terraform-internal noise (provider hashes, lock-file checksums, unrelated DynamoDB hash-key comments). The rename won't silently break any operational view.

### Deferred

**`apiKey` positional-arg fragility** — reviewer themselves marked "not actionable here." Test signature consistency with the rest of the file is the right local optimum; a future positional-safety pass would touch the whole test suite.

**Synthetic-fixture trivia** (the `5d41402a…` MD5 of `"hello"`) — reviewer flagged as harmless. The per-test comment already labels it `// 32-char hex string. The first 8 chars are…` which is enough signal for a future reader.

CI green; ready to squash-merge.

justin-layerv:
## Round 3 CR feedback addressed

**Addressed:**

- **Test consistency** — `assertNoFullHashLeaked()` now called from the missing-hash test too (commit 1dcdb03). Was the only test in the new describe block missing it.
- **Test count in PR description** — corrected from "4" to "7" (the `it.each` expansion adds three cases).

**Defense-in-depth at data layer rather than log layer** — deferred (same ground as prior round's #4 logger-redaction point). The bot's process memory is in-trust; only the log destination is the sensitive sink. The current shape protects that sink. A future caller passing the full `result` object to `logger.info` would defeat this — agreed — but the hardening is a follow-up: either strip the field from the returned object or add a CI lint guard. Not in scope here; tracked alongside the broader log-group hardening.

**8-hex truncation collision space** — 8 chars matches the existing `tokenHash8` house-style precedent in this codebase. Birthday-collision floor (~65k distinct hashes) is well above current upload volumes; revisit if traffic grows substantially. Not a security concern (any truncation length stops URL reconstruction); only a correlation-quality knob.

CI green. Ready to squash-merge.

PR #219, opened 2026-05-10
feat(discord): logger-side redaction of 'hash' keys

justin-layerv:
## Summary

Adds a defense-in-depth chokepoint in the bot's logger: any meta object passed to `logger.info / .warn / .error / .debug / .audit` with a content-hash key (`hash`, `md5`, `sha1`, `sha256`, `sha512`, `digest`, `checksum`, `content_hash`, `body_hash`) has its value blanked to `[REDACTED]` before serialization. Belt-and-suspenders for the per-call-site truncation that landed in #218. PR scope grew slightly during review — see "Behavior changes" below.

## Why

#218 truncated the connector's MD5 of uploaded content at three known call sites by routing them through `connector.js`'s `md5Prefix()` helper. The invariant only holds at those three sites — a future caller passing the full `result` object straight to `logger.info`, or using a different canonical hash name, would have shipped the full hash unchanged. This adds the same protection at the logger layer so a bypass becomes impossible without going around the logger entirely.

## Behavior changes

1. **Logger redaction triggers on a set of content-hash key names**, not just `hash`. New `REDACT_EXACT_KEYS` constant: `hash`, `md5`, `sha1`, `sha256`, `sha512`, `digest`, `checksum`, `content_hash`, `body_hash`. Exact-match (not substring) so legitimate names like `commitHash` / `webhookHash` / `md5_prefix` aren't blanked. Same set mirrored in `AUDIT_SECRET_KEYS` so the audit pathway gets the same treatment plus a `console.error` warn line naming the offending key. **No current bot log call uses any of these key names** (verified with grep), so this change emits no different bytes for any existing path.

2. **`redact()` and `redactAuditSecrets()` now recurse into non-null object/array values when a key matches** (commit 11279c7). Previously the matched-key branch shallow-copied the value, leaving a `{ hash: { token: '...' } }` accident path that emitted the inner `token` unredacted. The fix applies symmetrically — so this is a contract shift for **every** existing redacted key (`token`, `secret`, `password`, `authorization`, `apikey`, `api_key`, plus the new content-hash family), not just for `hash`. The recursion inherits the existing depth-5 cap. **No current caller exercises this contract** (no logger.* call passes nested objects under any matched key), so there's no observable byte-level change in production logs today — but the contract for future callers is tightened.

3. **Pre-existing `private_key` gap fixed.** `private_key` was in `AUDIT_SECRET_KEYS` (so audit emitted `[REDACTED]` for it) but was NOT caught by any `REDACT_SUBSTRINGS` rule (`private_` and `_key` are both too broad to add as substrings) and wasn't in `REDACT_EXACT_KEYS`. So before this PR, `logger.info('x', { private_key: 'real' })` shipped the value unredacted. Surfaced by the new reverse drift-guard test (`every AUDIT_SECRET_KEYS entry is also redacted by info()`) — `private_key` now sits in `REDACT_EXACT_KEYS` and is treated identically by both pathways. This is a real existing-behavior fix, not bookkeeping.

### Scope of the recursion fix — what it does and doesn't catch

The recursion catches **inner sensitive-named keys** under a matched outer key (e.g. `{ hash: { auth_token: 'real' } }` → inner `auth_token` blanked). It does **not** catch two related shapes — the value is non-primitive but the inner naming doesn't trigger redaction. Both are tracked in #222 for follow-up:

- Arrays of primitives: `{ hash: ['s1', 's2'] }` — array elements have no key name to match against, so primitives pass through.
- Objects with non-secret-named inner keys: `{ hash: { algorithm: 'md5', value: 'real-md5' } }` — `algorithm` and `value` aren't secret-shaped, so the inner `value` leaks.

So the framing is: this PR is **strictly safer than the previous shallow-copy behavior** for the nested-sensitive-named-key case, **but the protection is not complete** — the gaps above are documented and tracked. No current caller exercises any of the leaky shapes today.

## What changed

- New `REDACT_EXACT_KEYS` set in `apps/discord/src/logger.js` with two categories (content-hash + substring-uncovered audit secrets like `private_key`); mirrored in `AUDIT_SECRET_KEYS`.
- `shouldRedact()` checks the exact-match set before the substring-match set.
- Matched-key branch in both `redact()` and `redactAuditSecrets()` now recurses into non-null objects/arrays.
- Test-only `__testExports` exposes defensive `Set` copies of `REDACT_EXACT_KEYS` / `AUDIT_SECRET_KEYS` so drift-guard tests can iterate the live constants without risk of mutation.
- `apps/discord/tests/logger.test.js`: 14 new + modified `it(...)` blocks expanding to 24 distinct test cases via `it.each` parametrization. Coverage: top-level + nested redaction, `md5_prefix` survival, substring-near-miss survival (`commitHash` / `webhookHash`), audit-pathway warn signal naming both outer and inner matches, mixed-case via `.toLowerCase()` (asserts each variant individually), array-element recursion, primitive (null/number/empty-string) passthrough, recursion-into-matched-key-objects (both pathways), parametrized coverage of all 9 added content-hash + substring-uncovered key names, parametrized survival of 4 adjacent names, primitive-passthrough for legit-audit-dimension names (`tokens_minted` / `token_count`) on `info()`, redact-vs-audit recursion asymmetry pinned, forward + reverse drift-guard tests iterating the live `REDACT_EXACT_KEYS` / `AUDIT_SECRET_KEYS` sets.

## Filed as follow-ups

- #220 — Broaden audit warn-line wording when redaction key isn't credential-shaped (today's "secret-shaped" wording reads strangely for `hash`).
- #221 — Consolidate the two diverging exact-match sets (`REDACT_EXACT_KEYS` + `AUDIT_SECRET_KEYS`) into a single canonical structure; the function-pair (`redact()` / `redactAuditSecrets()`) likely collapses into one helper at the same time.
- #222 — Under-protection of values nested inside matched-key objects/arrays (arrays-of-primitives + objects with non-secret-named inner keys). Pre-existing for all matched keys.
- #223 — `Array.includes` dedup in `redactAuditSecrets` is O(n²); swap for a `Set`. Future-only concern; trivial for current shapes.

## Test plan

- [x] 50/50 logger tests pass (`npx jest tests/logger.test.js`)
- [x] `npm run lint` clean
- [x] /simplify ran across reuse / quality / efficiency reviewers; polish applied
- [x] CI green on each push
- [x] Verified no current bot log call uses any newly-redacted key name

### DIFF
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -39,6 +39,9 @@ const REDACT_EXACT_KEYS = new Set([
   'content_hash', 'body_hash',
 ]);
 
+// Internal: callers must pass strings. Only invoked from Object.entries()
+// loops, which yield string keys. If this becomes externally-used, restore
+// a String() coercion or guard non-string inputs.
 function shouldRedact(key) {
   const k = key.toLowerCase();
   if (REDACT_EXACT_KEYS.has(k)) return true;
@@ -63,6 +66,7 @@ const AUDIT_SECRET_KEYS = new Set([
   'content_hash', 'body_hash',
 ]);
 
+// Internal: callers must pass strings (same contract as shouldRedact).
 function isAuditSecretKey(key) {
   return AUDIT_SECRET_KEYS.has(key.toLowerCase());
 }
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -109,11 +109,15 @@ function redactAuditSecrets(value, depth = 0, secretKeys = []) {
       // undefined) pass through — they can't carry a usable secret and
       // zero-string-replace would lose type info.
       //
-      // Note: the recursion uses isAuditSecretKey (exact-match), NOT the
-      // wider shouldRedact (substring) rule. An inner `myToken` would
-      // therefore SURVIVE here even though it'd be blanked by redact().
-      // Intentional — audit dimensions like `tokens_minted` must not be
-      // redacted via substring.
+      // Recursion uses isAuditSecretKey (exact-match), NOT the wider
+      // shouldRedact (substring) rule, so legit dimensions like
+      // `tokens_minted` survive in the audit pathway. The asymmetry is
+      // intentional and pinned by a dedicated test.
+      //
+      // Push outer key BEFORE recursing so secretKeys order is parent-first
+      // (intuitive: the warn line reads "hash, auth_token" not the post-
+      // order "auth_token, hash" for `{ hash: { auth_token: ... } }`).
+      if (!secretKeys.includes(k)) secretKeys.push(k);
       if (typeof v === 'string' && v.length > 0) {
         out[k] = '[REDACTED]';
       } else if (v != null && typeof v === 'object') {
@@ -121,7 +125,6 @@ function redactAuditSecrets(value, depth = 0, secretKeys = []) {
       } else {
         out[k] = v;
       }
-      if (!secretKeys.includes(k)) secretKeys.push(k);
     } else {
       out[k] = redactAuditSecrets(v, depth + 1, secretKeys).value;
     }
@@ -141,9 +144,13 @@ function redact(value, depth = 0) {
       // examined. Other primitives pass through.
       // Asymmetry note: see redactAuditSecrets() — this recursion uses the
       // wider substring rule; the audit pathway uses exact-match.
-      if (typeof v === 'string' && v.length > 0) out[k] = '[REDACTED]';
-      else if (v != null && typeof v === 'object') out[k] = redact(v, depth + 1);
-      else out[k] = v;
+      if (typeof v === 'string' && v.length > 0) {
+        out[k] = '[REDACTED]';
+      } else if (v != null && typeof v === 'object') {
+        out[k] = redact(v, depth + 1);
+      } else {
+        out[k] = v;
+      }
     } else {
       out[k] = redact(v, depth + 1);
     }
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -19,6 +19,7 @@ function formatTimestamp() {
 // Substrings that should never appear unredacted in logs. Matched case-
 // insensitively against the key name via includes(), so a future field
 // named refreshToken / bearerToken / apiSecret / myPassword is auto-caught.
+// See REDACT_EXACT_KEYS below for the exact-match alternative.
 const REDACT_SUBSTRINGS = [
   'token', 'secret', 'password', 'authorization', 'apikey', 'api_key',
 ];
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -48,11 +48,11 @@ const REDACT_EXACT_KEYS = new Set([
   'private_key',
 ]);
 
-// Internal: callers must pass strings. Only invoked from Object.entries()
-// loops, which yield string keys. If this becomes externally-used, restore
-// a String() coercion or guard non-string inputs.
+// `String(key)` coerces — Object.entries() yields strings today, but a
+// future caller (a different walker, a Map-like) might pass non-strings;
+// the coercion keeps the predicate safe for free.
 function shouldRedact(key) {
-  const k = key.toLowerCase();
+  const k = String(key).toLowerCase();
   if (REDACT_EXACT_KEYS.has(k)) return true;
   return REDACT_SUBSTRINGS.some(s => k.includes(s));
 }
@@ -75,9 +75,8 @@ const AUDIT_SECRET_KEYS = new Set([
   'content_hash', 'body_hash',
 ]);
 
-// Internal: callers must pass strings (same contract as shouldRedact).
 function isAuditSecretKey(key) {
-  return AUDIT_SECRET_KEYS.has(key.toLowerCase());
+  return AUDIT_SECRET_KEYS.has(String(key).toLowerCase());
 }
 
 // Returns a cloned meta value with any object key in AUDIT_SECRET_KEYS
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -99,12 +99,17 @@ function redactAuditSecrets(value, depth = 0, secretKeys = []) {
   const out = {};
   for (const [k, v] of Object.entries(value)) {
     if (isAuditSecretKey(k)) {
-      // Blank non-empty strings; recurse into non-null objects/arrays so
-      // a `{ auth_token: { value: 'real' } }` accident still has `value`
-      // examined by the inner pass (the inner `value` won't match itself,
-      // but a nested `auth_token` / `password` etc. would). Other primitives
-      // (numbers, null, undefined) pass through — they can't carry a usable
-      // secret and zero-string-replace would lose type info.
+      // Blank non-empty strings; recurse into non-null objects/arrays so a
+      // `{ auth_token: { ... nested auth_token ... } }` accident still has
+      // its inner sensitive keys examined. Other primitives (numbers, null,
+      // undefined) pass through — they can't carry a usable secret and
+      // zero-string-replace would lose type info.
+      //
+      // Note: the recursion uses isAuditSecretKey (exact-match), NOT the
+      // wider shouldRedact (substring) rule. An inner `myToken` would
+      // therefore SURVIVE here even though it'd be blanked by redact().
+      // Intentional — audit dimensions like `tokens_minted` must not be
+      // redacted via substring.
       if (typeof v === 'string' && v.length > 0) out[k] = '[REDACTED]';
       else if (v != null && typeof v === 'object') out[k] = redactAuditSecrets(v, depth + 1, secretKeys).value;
       else out[k] = v;
@@ -273,3 +278,7 @@ const logger = {
 };
 
 module.exports = logger;
+// Test-only: exposes the redact constants so a drift-guard test can iterate
+// the live set rather than duplicate it. Not part of the public API; do
+// not consume from production code.
+module.exports.__testExports = { REDACT_EXACT_KEYS, AUDIT_SECRET_KEYS };
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -24,11 +24,20 @@ const REDACT_SUBSTRINGS = [
   'token', 'secret', 'password', 'authorization', 'apikey', 'api_key',
 ];
 
-// Exact-match keys (not substring): `hash` is the connector's md5 of
-// uploaded content; per-call-site truncation goes through md5Prefix() in
-// connector.js. Exact-match so legitimate names like `commitHash` /
-// `webhookHash` don't trip.
-const REDACT_EXACT_KEYS = new Set(['hash']);
+// Exact-match keys (not substring): content-derived hash names that are
+// sensitive in our broader infrastructure. The connector's md5 of uploaded
+// content is the original motivator (per-call-site truncation goes through
+// md5Prefix() in connector.js); the rest are forward-looking guards so a
+// future caller using a different canonical name doesn't slip through.
+// Exact-match so legitimate names like `commitHash` / `commit_hash` /
+// `webhookHash` / `md5_prefix` don't trip. Mirrored in AUDIT_SECRET_KEYS
+// below — the two sets are kept in sync by hand today (see #221).
+const REDACT_EXACT_KEYS = new Set([
+  'hash',
+  'md5', 'sha1', 'sha256', 'sha512',
+  'digest', 'checksum',
+  'content_hash', 'body_hash',
+]);
 
 function shouldRedact(key) {
   const k = String(key).toLowerCase();
@@ -47,7 +56,11 @@ const AUDIT_SECRET_KEYS = new Set([
   'token', 'secret', 'password', 'authorization', 'apikey', 'api_key',
   'auth_token', 'access_token', 'refresh_token', 'bearer_token',
   'session_token', 'private_key', 'client_secret', 'webhook_secret',
-  'hash', // see REDACT_EXACT_KEYS above
+  // Content-derived hash names — see REDACT_EXACT_KEYS above for rationale.
+  // Kept in sync by hand today; consolidation tracked in #221.
+  'hash', 'md5', 'sha1', 'sha256', 'sha512',
+  'digest', 'checksum',
+  'content_hash', 'body_hash',
 ]);
 
 function isAuditSecretKey(key) {
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -40,7 +40,7 @@ const REDACT_EXACT_KEYS = new Set([
 ]);
 
 function shouldRedact(key) {
-  const k = String(key).toLowerCase();
+  const k = key.toLowerCase();
   if (REDACT_EXACT_KEYS.has(k)) return true;
   return REDACT_SUBSTRINGS.some(s => k.includes(s));
 }
@@ -64,7 +64,7 @@ const AUDIT_SECRET_KEYS = new Set([
 ]);
 
 function isAuditSecretKey(key) {
-  return AUDIT_SECRET_KEYS.has(String(key).toLowerCase());
+  return AUDIT_SECRET_KEYS.has(key.toLowerCase());
 }
 
 // Returns a cloned meta value with any object key in AUDIT_SECRET_KEYS
--- a/apps/discord/src/logger.js
+++ b/apps/discord/src/logger.js
@@ -88,20 +88,11 @@ function isAuditSecretKey(key) {
 // the warn line — partial reporting would let a caller fix one key
 // and re-run only to discover another the next time.
 //
-// Note on the contract shift: the original audit() bypassed redact()
-// entirely because the legacy substring-match REDACT_SUBSTRINGS would
-// blank legitimate dimensions like `tokens_minted`. AUDIT_SECRET_KEYS
-// uses exact-match against a closed set of canonical secret names
-// (auth_token, api_key, password, ...), none of which collide with any
-// known legitimate audit dimension. So we can redact the value of an
-// offending key without risk of corrupting a metric — and that's
-// strictly safer than emitting the secret verbatim and relying on a
-// dashboard sweep to catch it.
-//
-// Recursion contract (added in #219): when a matched key's value is a
-// non-null object/array, we recurse so an `{ auth_token: { ... } }`
-// accident still has its inner sensitive keys examined. Inherits the
-// existing depth-5 cap.
+// Exact-match (not substring) so legitimate audit dimensions like
+// `tokens_minted` survive — these are the canonical secret-bearer names
+// (auth_token, api_key, password, ...) that are safe to redact-by-value
+// without corrupting metrics. Recurses into matched-key non-null
+// object/array values (depth-5 cap inherited from the function body).
 function redactAuditSecrets(value, depth = 0, secretKeys = []) {
   if (depth > 5 || value == null || typeof value !== 'object') {
     return { value, secretKeys };
@@ -148,11 +139,9 @@ function redact(value, depth = 0) {
     if (shouldRedact(k)) {
       // Blank non-empty strings; recurse into non-null objects/arrays so
       // a `{ hash: { token: 'real' } }` accident still has its inner keys
-      // examined. Other primitives (numbers, null, etc.) pass through.
-      // Note: this recursion uses shouldRedact (substring rule); the
-      // audit pathway recursion uses isAuditSecretKey (exact-match) so a
-      // `{ hash: { myToken: ... } }` blanks here but survives audit. See
-      // redactAuditSecrets() for the rationale.
+      // examined. Other primitives pass through.
+      // Asymmetry note: see redactAuditSecrets() — this recursion uses the
+      // wider substring rule; the audit pathway uses exact-match.
       if (typeof v === 'string' && v.length > 0) out[k] = '[REDACTED]';
       else if (v != null && typeof v === 'object') out[k] = redact(v, depth + 1);
       else out[k] = v;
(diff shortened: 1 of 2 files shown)


justin-layerv:
## /simplify summary

Three reviewers (reuse / quality / efficiency). Five items applied in commit 8761085, two filed as out-of-scope follow-ups, the rest false positives:

**Applied:**
- `REDACT_EXACT_KEYS` comment trimmed from 6 lines → 4 (dropped narration of what `hash` semantically is + duplicate of cross-reference)
- `AUDIT_SECRET_KEYS` `hash` entry: 3-line comment → 1-line `// see REDACT_EXACT_KEYS above`
- New describe-block comment trimmed (dropped lines that just restated the describe name)
- Inline test comments for null/number JSON serialization deleted (the assertions immediately below say the same thing; the recursion-into-object comment kept since the `nested`-key assertion would otherwise look unmotivated)
- `FULL_MD5` constant hoisted at top of new describe — mirrors the same pattern in `connector-coverage.test.js` and collapses 4 inline literals to a named reference

**Filed as follow-up (out of scope here):**
- Canonical key-name constants for `hash` / `md5_prefix` so a rename can't desync `src` from tests. Natural home is `apps/discord/src/constants.js` (already exports `AUDIT_EVENTS`). Skipping in this PR — scope creep on a redaction patch.
- A second cross-test fixture file (or a `tests/__fixtures/`) for `FULL_MD5` if a third test file ever needs it. Below the threshold today (two files, one fixture).

**False positives / deferred:**
- Merging `REDACT_EXACT_KEYS` and `AUDIT_SECRET_KEYS` into one set — intentional separation; audit needs the warn-on-leak signal that regular redact doesn't.
- `process.env.LOG_LEVEL` per-test boilerplate — file's pre-existing idiom (16+ tests do it), needed because `currentLevel` captures from env at module load and tests vary it.
- `shouldRedact()` exact-then-substring order — semantic no-op (no key in EXACT is also a substring hit), and putting the O(1) Set lookup first is a fine micro-perf intuition.
- `it.each` for the packed null/number/object test — heterogeneous assertions; current single-`it` form is fine.

justin-layerv:
## CR feedback addressed

Three points from the bot review triaged.

### Addressed (commit 28a7e89)

**1. Incorrect rationale comment in the non-string-values test** — the comment claimed the object value "recurses" into matched keys, but `redact()` actually passes non-string matched values through verbatim. Test still passed (the verbatim copy carries `nested` along), but the comment misled about the mechanism. Rewrote to describe the real behavior, plus a forward pointer to the follow-up that tightens it.

**3. Stale REDACT_SUBSTRINGS header comment** — added a one-line cross-reference so a future reader finds both the substring and exact-match knobs from one place.

### Deferred — filed as a follow-up

**2. Tighten `redact()` and `redactAuditSecrets()` to recurse on matched non-string values.** Bot called this out as a real invariant gap surfaced by point #1, and explicitly said "Out of scope for this PR." A future caller passing `{ hash: { token: 'real-secret' } }` would NOT redact the inner `token` today. Not exploitable today (no current bot caller passes objects under sensitive keys), but worth a focused PR with tests covering nested-secret cases. Tracked.

### Not actionable

**4 (CLAUDE.md scope check)** and **5 (coverage looks good)** — confirming/praise.

justin-layerv:
## Round 2 CR feedback addressed

### Addressed (commit 5cc813a)

**#1 — Recursion-gap caveat promoted to source.** TODO comments now sit at the matched-key branch in both `redact()` and `redactAuditSecrets()`, so a reader of the source (not just the test) sees that nested objects under a matched key pass through verbatim. Tracked separately as a future tightening.

**#2 — Case-insensitivity test.** `Hash` / `HASH` / `HaSh` all redacted in one assertion; guards against a future refactor dropping the `.toLowerCase()` in `shouldRedact()`.

**#3 — Array-element coverage.** New test for `{ items: [{ hash }] }` — rounds out path-coverage symmetry with the existing nested-meta test.

### Deferred

**#4 — Note that `hash` is broader than its AUDIT_SECRET_KEYS neighbors.** Multiple safeguards already exist (the `AUDIT_EVENTS` allowlist constrains audit vocabulary; the warn-line surfaces leaks at runtime). Adding more comment text here is marginal and risks /simplify flagging it as bloat next round. Skipped.

**#5 — Empty-string edge case nit.** `{ hash: '' }` passes through verbatim — bot marked "harmless" and "nit." Adding a test only pins current behavior; no behavior change warranted. Skipped.

32/32 tests pass; lint clean.

justin-layerv:
## Round 3 CR feedback addressed

### Addressed in this PR (commit 11279c7)

**#1 — Recursion fix folded into this PR.** Both `redact()` and `redactAuditSecrets()` now recurse into non-null object/array values when a key matches; primitives unchanged. Closes the `{ hash: { token: '...' } }` accident path the TODO warned about. Inherits the existing depth-5 cap, so cyclic refs / pathological nesting are bounded. Closed task tracking the standalone follow-up.

**#4 — Test for non-string values updated.** The previous test pinned the leaky shape (object value → verbatim passthrough). Replaced with a positive recursion test: `{ hash: { token: 'real-secret', nested_hash: 'also-secret' } }` blanks `token` and leaves `nested_hash` intact. Added a parallel audit-pathway test that also confirms the warn line still names the outer key.

Plus from /simplify polish on this round:
- Trimmed two long test names.
- Folded the empty-string edge-case test into the primitives test as a third call (was previously a standalone one-assertion test).

### Filed as follow-ups

**#2 — "secret-shaped" warn-line wording reads strangely for content-derived keys** → #220. Bot themselves said "Not worth re-wording for one key" — file as a future hardening if more sensitive-but-not-credential keys land.

**#3 — REDACT_EXACT_KEYS and AUDIT_SECRET_KEYS diverge** → #221. Refactor to a single canonical set with the audit pathway extending it. Touches both pathways' key-set definitions; deserves a focused PR.

### Not actionable

**#5 (commit scope check)** — confirming, no action.

35/35 logger tests pass; lint clean.

justin-layerv:
## Round 4 CR feedback addressed

### Addressed in this PR

**🟡 Recursion contract shift not in PR description** → fixed. PR description now spells out both behavior changes (the `hash` key chokepoint AND the recursion fix that affects every existing redacted key). The two changes already live in separate commits (11279c7 for recursion, others for `hash` family) for cleaner archaeology.

**🟢 Other hash-shaped keys** (commit c1eaee8). Added `md5`, `sha1`, `sha256`, `sha512`, `digest`, `checksum`, `content_hash`, `body_hash` to both `REDACT_EXACT_KEYS` and `AUDIT_SECRET_KEYS`. Verified no current logger.* call uses any of these names. Adjacent names (`md5_prefix`, `commit_hash`, `fileHash`, etc.) are not exact-matches and survive — pinned by `it.each` tests.

**Nit: PR description test count** → fixed (was "6", now "12 new + modified" reflecting the actual diff including parametrized expansion).

### Filed as follow-ups (issue links above in the PR description)

**🟡 Array-of-primitives gap** → #222. `{ hash: ['s1', 's2'] }` still leaks because array elements have no key name to match against. Pre-existing for all matched keys (not a regression introduced here). Filed with suggested fix paths (treat matched-key array values as a special case in the matched branch).

### Skipped

**Nit: `String(key).toLowerCase()` dead-defensive coercion.** Bot themselves marked "Not worth changing alone." Skipping per that guidance.

46/46 logger tests pass; lint clean.

justin-layerv:
## Round 5 CR feedback addressed

**🟡 PR description sync** → updated. Reframed the "purely forward-looking" line: no current caller exercises the new contract (so production logs emit no different bytes today), but the contract IS tightened for every existing redacted key (`token`/`secret`/`password`/etc.). Test count corrected to 12 `it(...)` blocks expanding to 22 cases via `it.each`.

**🟢 Audit recursion test pins only outer key** → fixed in commit 12d660a. Test now asserts BOTH `'hash'` AND `'auth_token'` surface in the warn line; guards against a future early-return-on-outer-match refactor.

**Nit (TODO leftovers)** → verified gone. The recursion-fix commit (11279c7) replaced both TODO blocks with the actual fix.

**Nit (`String(key).toLowerCase()` dead-defensive)** → fixed in commit 12d660a. `Object.entries` only yields string keys (Symbol keys go through `getOwnPropertySymbols`, which we don't iterate), so the coercion was unreachable. Dropped at both sites.

### Skipped

**Nit (`digest` narrow risk for HTTP Digest auth headers)** — bot called out as "not blocking." The current threat-model framing ("content-derived hash names that are sensitive") covers Digest header values too (signature material is sensitive), and redacting an unknown `digest` log value is the safer default vs leaking a real Digest header. Keep `digest` in the set.

46/46 tests pass; lint clean.

justin-layerv:
## Round 6 CR feedback addressed

### Addressed in this PR (commit bcd7a92)

**🟢 #1 — `String(key)` removal needs internal-only contract note** → fixed. One-line comment on each of `shouldRedact()` and `isAuditSecretKey()` flagging they're internal-only and require string keys. Future external use would have to restore a guard.

**🟢 #5 — REDACT_EXACT_KEYS / AUDIT_SECRET_KEYS drift risk** → fixed via a new subset-invariant test. Every entry in REDACT_EXACT_KEYS is now verified to also be redacted by `audit()`. Drops out naturally once #221 consolidates.

### Issue updates

**🟢 #2 — #222 should not read as 'hash-only'** → edited #222. Title and body now make explicit that the array-of-primitives gap applies to **every** matched key, not just `hash` (REDACT_EXACT_KEYS ∪ REDACT_SUBSTRINGS).

### Filed as new follow-up

**🟢 #3 — `Array.includes` dedup in `redactAuditSecrets` is O(n²)** → #223. Bot themselves flagged as "Trivial for current shapes; not blocking." Filed with two suggested fix paths (internal-only Set arg vs change return type to Set). Likely lands as part of #221's consolidation.

### Already filed

- **🟢 #4 — "secret-shaped" warn-line wording** → #220
- **🟢 #6 — Function-pair unification** → covered by #221's consolidation scope

47/47 tests pass; lint clean.

justin-layerv:
## Round 7 CR feedback addressed

### Addressed in this PR (commit f35a87f)

**Behavior #2 — Asymmetry between `redact()` (substring) and `redactAuditSecrets()` (exact-match) recursion** → fixed. Added an inline comment in `redactAuditSecrets()` calling out the asymmetry explicitly: the recursion uses `isAuditSecretKey` (exact-match), not the wider `shouldRedact` rule, intentionally so that legitimate audit dimensions like `tokens_minted` / `token_count` aren't blanked.

**Tests #1 — Drift-guard test duplicated the constant** → fixed. Added a test-only `__testExports` on the logger module exposing `REDACT_EXACT_KEYS` / `AUDIT_SECRET_KEYS`. The drift-guard test now iterates the LIVE set rather than a hardcoded copy. Reverse-direction drift (audit-only key without a redact mirror) still uncovered — noted in the test comment, tracked under #221.

**Tests #2 — Mixed-case test counted global occurrences** → fixed. Now asserts each case variant individually (`"Hash":"[REDACTED]"` / `"HASH":"[REDACTED]"` / `"HaSh":"[REDACTED]"`); robust against an unrelated future fixture growth.

### Issue update

**Behavior #1 — Object-value gap** → broadened #222. Title and body now make explicit that the under-protection applies to two related shapes — arrays-of-primitives AND objects-with-non-secret-named-keys (`{ hash: { algorithm, value } }` leaks `value`) — and that both apply to every matched key, not just `hash`. Three suggested fix paths documented in the issue (blanket-blank, recursion-context flag, hybrid).

### PR description updated

Softened the "strictly safer" framing per Behavior #1's audit-future-proofing point. Added a "Scope of the recursion fix — what it does and doesn't catch" section that explicitly enumerates both gaps and points to #222.

### Skipped (style nits, bot marked optional)

- **`blankOrRecurse` helper extraction** — bot said "Optional. Two callsites; saves about 6 lines." Skip; consolidation will land naturally with #221 anyway.
- **Internal-only comment is "belt-and-suspenders on belt-and-suspenders"** — bot said "comment is fine; just noting it." Keep as-is.

47/47 tests pass; lint clean.

justin-layerv:
## Round 8 CR feedback addressed

### Addressed in this PR (commit a854f52)

**#1 — `__testExports` exposes mutable Sets** → fixed. Each Set now wrapped in `new Set(...)` so `__testExports.REDACT_EXACT_KEYS.add('foo')` from a buggy test can't corrupt module-cached production state.

**#2 — Drift guard one-directional** → fixed. Added a reverse drift-guard test (every AUDIT_SECRET_KEYS entry must also be redacted by `info()`). **The new test surfaced a real pre-existing gap**: `private_key` was in AUDIT_SECRET_KEYS but no REDACT_SUBSTRINGS rule catches it (no `private_` / `_key` substring is broad-safe), and it wasn't in REDACT_EXACT_KEYS. Added `private_key` to REDACT_EXACT_KEYS — the regular `redact()` pathway now blanks it consistently with audit. The REDACT_EXACT_KEYS comment was updated to reflect two categories of exact-match coverage (content-hash + substring-uncovered audit secrets).

**#3 — Asymmetric recursion semantics need a pinning test** → fixed. New test asserts both:
- `{ hash: { myToken: 'real' } }` via `info()` → inner `myToken` blanked (substring catches `token`)
- `{ hash: { myToken: 'real' } }` via `audit()` → inner `myToken` SURVIVES (exact-match doesn't catch)

A future refactor that unifies the two functions would change one of these — CI would surface the regression instead of production dashboards breaking on legitimate dimensions like `tokens_minted` getting blanked.

### Issue update / deferred

**#4 — "Mirrored" wording precision** → bot said "feel free to leave for #221." Folded into #221's consolidation scope.

### Skipped (paper-only)

**#5 — `String()` removal × `__testExports`** — bot said "Today `__testExports` only exposes the `Set`s and not the predicate functions — keep it that way and this stays a paper concern." Confirmed kept that way; nothing to fix.

49/49 tests pass; lint clean.

justin-layerv:
## Round 9 CR feedback addressed

### Addressed in this PR (commit 9de3f2c)

**#1 — `private_key` deserves description-level callout** → fixed. Promoted to its own bullet ("Pre-existing `private_key` gap fixed") in the Behavior changes section, framed as a real existing-behavior fix surfaced by the new reverse drift-guard test (not bookkeeping).

**#2 — Dead `LOG_LEVEL` set after `require()`** → fixed. Removed from the reverse-drift loop and the asymmetric-recursion test. The logger captures `currentLevel` at module load, so setting `LOG_LEVEL` after `require()` is dead — default 'info' covers the test cases. Test comment now notes this so a future refactor doesn't reintroduce it.

**#3 — Reciprocal asymmetry comment at `redact()` recursion site** → fixed. `redact()`'s recursion now also notes the substring-vs-exact-match divergence with a pointer to `redactAuditSecrets()`. A future "harmonization" refactor will hit at least one of the two cross-references.

**#6 — `tokens_minted` / `token_count` survival pin on `info()` pathway** → fixed via a focused test. Note that these wouldn't fit in the existing adjacent-name `it.each` (string values get blanked via substring); the new test uses numeric values (the realistic audit-dimension shape) and pins the primitive-passthrough rule. Guards against a future change that always blanks matched keys regardless of value type.

### Verified, no action

**#4 — #222 captures verbatim shapes** → confirmed. The two example shapes from this PR's description (`{ hash: ['s1', 's2'] }` and `{ hash: { algorithm: 'md5', value: 'real-md5-string' } }`) are quoted verbatim in #222.

### Already filed

**#5 — `Array.includes` O(n²) dedup** → #223.

50/50 tests pass; lint clean.

justin-layerv:
## Round 10 CR feedback addressed

### Addressed in this PR (commit 4ff57d1)

**#4 — `body_hash` interpretation ambiguity** → fixed. Added a comment in `REDACT_EXACT_KEYS` clarifying that `body_hash` here means a content digest of a payload (informational identifier of body bytes), but exact-match correctly catches the HMAC-of-body interpretation too. Future reader can't relax it on the wrong premise.

**#5 — Verify `value:` shape doesn't exist in current callers** → done. Grep confirmed no current `logger.*` call passes a `{ value: ... }` shape under any matched key in `apps/discord/src`. Note added to #222.

**Long-line nit in redactAuditSecrets** → fixed. Wrapped the matched-key conditional from a 110+ col chain to explicit if/else-if/else blocks. Cosmetic only.

**Stale "Note on the contract shift" comment** → fixed. Added a one-line addendum noting the recursion contract introduced in #219; pre-existing comment was about the original audit bypass rationale only.

**Coverage gap — structural sensitive-value substring** → fixed. New test asserts the formatted log line never contains a sensitive value as a substring anywhere, regardless of where it sat in meta. Catches a future change that stringifies meta before redacting (or any shape that bypasses the redact pass) — key-targeted asserts alone wouldn't.

### Issue updates

**#1 — Prioritize #221** → bumped #221 from `priority: low` to `priority: medium`. The bot's "footgun" framing is right — every additional matched-key path widens the surface where the two functions can diverge.

### Deferred (with reasoning)

**#2 — Restore `String(key)` defensive coercion** — bot themselves said "raise if you disagree." I disagree: `__testExports` doesn't expose the predicate functions, and the internal-only contract comment is sufficient. Restoring `String()` would add code for an invariant that's already enforced by call-site shape (only `Object.entries()` callers, which always yield strings). Defensible to leave removed.

**#3 — Warn-line treatment for redact() blanks** — bot said "Probably overkill; flagging for discussion." Agreed — adding `console.error` per `info()` redaction would add noise to production logs and the operator-debuggability concern is bounded by "no current caller emits these key names." If the noise becomes a real concern later, fold into #221's consolidation (the unified function could opt into the warn-line per call).

51/51 tests pass; lint clean.

justin-layerv:
## Round 11 CR feedback addressed

### Addressed in this PR (commit cfcd6b7)

**#3 — `__testExports` exposed via `module.exports`** → fixed via `if (process.env.NODE_ENV === 'test')` gate. Production bundles won't include the export at all. Jest sets `NODE_ENV=test` by default so tests are unaffected.

**#5 — `describe('hash key redaction', ...)` name no longer fits** → renamed to `'hash-family redaction, drift guards, and recursion semantics'`. The block grew during review beyond the hash-only theme to include `private_key`, drift guards, recursion asymmetry, and primitive-passthrough — the new name reflects the actual contents.

**#6 — `body_hash` HMAC-vs-digest comment reads like internal deliberation** → trimmed from 5 lines to 2 (`"covers both content-digest and HMAC interpretations; either is sensitive-shaped enough to redact"`). Constraint preserved without the dialog.

**Tests-isolation nit** → fixed. Added explicit `process.env.LOG_LEVEL = 'info'` before `require()` to the structural-regression, both drift-guard, asymmetric-recursion, audit-recursion, and tokens_minted-passthrough tests. Convention match with the file's idiom; semantically a no-op since default is 'info'.

### Verification

**#1 — Grep for `digest` / `checksum` legitimate use** → confirmed clean. `grep -rnE "\b(md5|sha1|sha256|sha512|digest|checksum)\s*[:=]" apps/discord/src` shows only false positives inside string literals (`'sha256='` for webhook signature validation), no current logger.* call uses these as keys. Over-redaction risk is zero today.

### Already covered

**#2 — `String(key)` removal — defensible either way** → bot itself acknowledged. Internal-only contract comment + non-exposed predicates are sufficient.

**#4 — Land #221 sooner** → already bumped from `priority: low` to `priority: medium` last round.

51/51 tests pass; lint clean.

justin-layerv:
## Round 12 CR feedback addressed + /simplify polish

### Round-12 cr (commit 92d4317)

**#1 — Code-level note about `tokens_minted` string-vs-number behavior** → addressed via trim. Originally added as a 6-line caller-note; /simplify flagged as operator-doc territory. Trimmed and consolidated — the asymmetry comment immediately above + the audit() JSDoc + the new test (`info() preserves tokens_minted / token_count with primitive (number) values`) carry the contract collectively, no inline operator-doc needed.

**#2 — Depth-cap behavior at recursion boundary** → no PR action; pre-existing per bot, behavior documented, will be revisited in #222 follow-up work.

**#3 — `JSON.stringify` throw path on info/warn/error/debug (no try/catch)** → filed as #224. Pre-existing logger gap (audit() has the two-tier degradation, the others don't); not introduced by this PR.

**#4 — `Array.includes` O(n²) dedup** → already #223.

**#5, #6** — addressed in earlier rounds / praise.

### /simplify polish (same commit)

Three reviewers (reuse / quality / efficiency). Quality flagged five trim items after 11 prior rounds of polish:

- Round-12 caller-note crossed into operator-doc → consolidated as above.
- `redactAuditSecrets()` doc had 12 lines of historical narrative ("Note on the contract shift…") → trimmed to 5 lines of current invariant.
- `audit()` JSDoc tail (12 lines restating `redactAuditSecrets()`'s contract) → 5 lines pointing at AUDIT_EVENTS allowlist.
- `redact()`'s asymmetry block (5 lines) → 1-line cross-reference to `redactAuditSecrets()` (canonical site).
- Two test names tightened.

Reuse + efficiency reviewers: zero findings (helper extraction correctly rejected; `Set` defensive copies have no codebase precedent for a `cloneSet` helper; perf is strictly improved by exact-then-substring ordering and `String()` removal).

51/51 tests pass; lint clean.

justin-layerv:
## Round 13 CR feedback addressed

### Addressed in this PR (commit 2b7eff8)

**#2 — `String(key)` coercion** → restored. This has been raised across rounds 6/8/11/13; reversing my prior deferral. One line, removes a footgun for any future refactor that calls the predicates from a non-`Object.entries` walker. Comments updated to reflect the rationale.

**#8 — `primitive hash values pass through` test packs three calls** → parametrized via `it.each`. Each case (null / number / empty-string) is now its own test with its own `consoleSpy.log.mock.calls[0]` read — no future-maintainer index-drift risk. 53 tests now (was 51).

### Filed as new follow-ups

**#9 — Cross-app logger parity (slack / teams)** → #225. Out of scope for `feat(discord)` here, but the same defense-in-depth argument applies wherever a `logger.info(meta)` could carry a content-derived value.

### Issue update

**#1 — Recursion semantics undermine motivator for object-valued hash keys** → #222 was already tracking this; added a comment with the **per-set semantics** suggestion (split `REDACT_EXACT_KEYS` into a `BLANK_WHEN_MATCHED_REGARDLESS_OF_TYPE` content-hash subset vs the recurse-then-blank-inner-keys substring-matched subset). The bot's framing of the gap is correct — the follow-up fix should treat the two flavors differently rather than apply uniform recursion.

### Skipped with reasoning

**#3 — `__testHooks.js` sibling file for test seam** — bot themselves called this "Style call." Both raised concerns (NODE_ENV misconfiguration, Object.keys exposure) are paper-thin under our threat model. The current `NODE_ENV==='test'` gate is sufficient defense-in-depth.

**#7 — `MD5` uppercase in adjacent-name parametrize** — already covered by the mixed-case `Hash`/`HASH`/`HaSh` test; `.toLowerCase()` works case-independently for all entries in `REDACT_EXACT_KEYS`.

### Already filed

**#4** → #220 (warn-line wording). **#5** → #223 (O(n²) dedup).

53/53 tests pass; lint clean.

justin-layerv:
## Round 14 CR feedback addressed

### Addressed (commit f3171b8)

**#1 — `secretKeys` order is post-order, not encounter order** → fixed by pushing the outer key BEFORE recursion. The warn line now reads parent-first ("hash, auth_token") instead of post-order ("auth_token, hash") for `{ hash: { auth_token: ... } }`. Existing tests use `.toContain()` so they pass either way; this is the more intuitive shape.

**#2 — Style inconsistency in `redact()` recursion branch** → fixed. Both `redact()` and `redactAuditSecrets()` matched-key branches now use braced if/else-if/else for symmetry.

**Test 1 — Drift-guard iteration coupling** → fixed via inline comment pinning the "redact() is stateless across calls" assumption. Future maintainer adding per-call state has a guide to switch to per-iteration `jest.resetModules()`.

**Test 2 — Structural-leak test only swept stdout** → extended to sweep BOTH stdout (info/warn/error/debug + audit JSON) AND stderr (audit warn line). Plus exercises the audit() pathway too. Pins that the warn line names offending keys but never echoes values.

53/53 tests pass; lint clean.

PR #176, opened 2026-05-02
fix(shared): client.Create uses /v1/qurls (plural), not /v1/qurl

justin-layerv:
## Summary

`shared/client.Create` posted to `{baseURL}/v1/qurl` (singular). Every other method on this same client (`Get`, `List`, `Delete`, `Patch/Extend/Update`, `MintLink`) correctly uses `/v1/qurls/...` plural; Discord's JS client also uses plural (`apps/discord/src/commands.js:3037`). The singular form was a typo that 404s on every qurl service host:

```
POST https://api.layerv.ai/v1/qurl   → 404 (route not present)
POST https://api.layerv.ai/v1/qurls  → 201 (route present, succeeds)
```

## Symptom

Slack bot's first end-to-end sandbox test today returned:
```
Failed to create qURL: Not Found
```

Companion fix in [qurl-integrations-infra#412](https://github.com/layervai/qurl-integrations-infra/pull/412) (already merged) corrects the endpoint **host** (`api.layerv.ai` vs `.xyz`); this PR closes the **path** half. Both must land before the bot's `/qurl-sandbox create` slash command works.

## Why CI didn't catch it

The `client_test.go` assertion at line 53 also matched the wrong singular path:

```go
if r.URL.Path != "/v1/qurl" { ... }
```

The test was verifying **internal consistency** (test server matched whatever the client sent) rather than the **actual API contract**. Both the production code AND its test had the same typo, so they reinforced each other. Updated together so the test now fences the correct path.

## Changes
- `shared/client/client.go:257,270` — `/v1/qurl` → `/v1/qurls` (request URL + log/metric label)
- `shared/client/client_test.go:53,54` — assert `/v1/qurls`

## Test plan
- [x] `go test -count=1 ./shared/client/...` — passes
- [x] `go test -count=1 ./apps/slack/...` — passes
- [x] `make lint` — 0 issues
- [ ] Post-merge: rebuild + redeploy slack bot image; rerun `/qurl-sandbox create https://example.com` from the LayerVai workspace; expect a real qurl link in response

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #148, opened 2026-04-30
shared/client: extend Idempotency-Key support to other write methods (MintLink, Update, Extend, Resolve)


justin-layerv:
## Background

PR #147 added \`Idempotency-Key\` support to \`Client.Create\`. CR feedback flagged that the same retry-then-dedup concern applies to other write methods in the same client:

- **\`MintLink\`** (POST) — issues a magic link; a Slack-retry burst could mint two links for the same user click.
- **\`Update\`** (PATCH) — updating a qURL's description or expiry with the same intent twice produces two state transitions in the audit log.
- **\`Extend\`** (PATCH) — extending TTL twice grants double-extension (real bug).
- **\`Resolve\`** (POST) — resolves a qURL token; mutates AC firewall state via NHP. A Slack retry hitting Resolve could open the firewall twice for the same client IP, briefly doubling the active-session count.

\`HeaderIdempotencyKey\` (and the byte-validation helpers) are already exported from PR #147, so this is a pure surface-extension task — no new constants, no new sentinels.

## What to do

For each method, add an \`IdempotencyKey string \\\`json:\"-\\\`\`\` field to its respective input struct (or a \`X.IdempotencyKey\` parameter where the method takes scalar args), apply the same length + invalid-byte validation, set the header on the request before \`do()\` so it's preserved across retries.

Mirror the test pattern from \`shared/client/client_test.go\`:

- header set when key populated
- header absent when empty
- body byte-identical with vs without the field
- header preserved across retry chain

## Acceptance

- All five write methods (\`Create\` already done, \`MintLink\`, \`Update\`, \`Extend\`, \`Resolve\`) accept an \`IdempotencyKey\`.
- Per-method tests pin the contract (4 tests × 4 methods = 16 new tests).
- \`make check\` passes.
- No new exported types/constants — everything reuses what PR #147 shipped.

## Source

CR feedback on PR #147; deferred to keep that PR's scope tight (\"current PR scope is the right call — Slack \`/qurl create\` is the immediate need and Phase 2 PR-A — but worth a tracking note\").

Kevin-layerV:
Update from PR #226 — adding scope.

PR #226 introduces three new write methods that should fall under the same Idempotency-Key plumbing:

- `CreateResource` (POST `/v1/resources`) — server-side idempotent on `(owner_id, target_url_hash)`, but a caller-supplied key would still be useful for layered retry semantics on top of side-effecting Slack/Discord triggers.
[...]

PR #226, opened 2026-05-10
feat(shared): add resource methods + ResourceID on CreateInput, regression-pin /v1/qurls path

Kevin-layerV:
## Summary

Phase 3c.2 from the Slack-triggered qURL plan. Extends `shared/client` for
the alias / get flows that ship in PR-3c.3+.

- **Blocker #1 — `/v1/qurl` (singular) path bug.** Already fixed in #176;
  this PR adds `TestCreatePathIsPlural` as a *named regression pin* so a
  future "rename it back" change surfaces by an explicit test name, not
  by a generic `TestCreate` failure. (TestCreate already asserts the path
  generically; the value of the new test is the name.)
- **Blocker #2 — `target_url` empty-string vs `mutually_exclusive_fields`.**
  Add `ResourceID string` to `CreateInput` with `,omitempty`, and switch
  `TargetURL` to `,omitempty` so a ResourceID-only mint elides `target_url`
  from the wire payload entirely. Server-side PR-3a.3 enforces
  `mutually_exclusive_fields(resource_id, target_url, type)`; without the
  omitempty change, every ResourceID-only call would marshal `"target_url":""`
  and 400.
- **Idempotency-Key.** Already shipped via `CreateInput.IdempotencyKey` in
  #147 + #149 (validator + retry preservation tests already in tree). Plan
  asked for a variadic `WithIdempotencyKey(...)` option; deviated because
  switching to options-pattern would break the existing `Create` callers in
  `apps/slack/internal/process.go:121` and `apps/cli/cmd/create.go:47`,
  both of which are out of scope per the task constraints.
- **New resource methods** for the alias flow:
  - `CreateResource(ctx, *CreateResourceInput) (*Resource, error)` →
    `POST /v1/resources`
  - `UpdateResource(ctx, id, *UpdateResourceInput) (*Resource, error)` →
    `PATCH /v1/resources/{id}` (alias set + `clear_alias: true` paths
    both pinned in tests)
  - `GetResourceByAlias(ctx, alias) (*Resource, error)` →
    `GET /v1/resources/by-alias/{alias}`

  These target the **post-PR-3a.2** surface in qurl-service. PR-3a.2
  hasn't shipped, so callers will get 404/400 from the API until it
  does. Added here so the Slack flows in PR-3c.3+ have a stable Go
  surface to compile against.

## Choice on TargetURL handling

Kept `TargetURL string` and added `,omitempty` to the JSON tag (the
recommended option in the plan). The alternative was `*string`, which
would have been a breaking change for the existing `client.CreateInput{
TargetURL: someString }` literal at `apps/slack/internal/process.go:121`
and `apps/cli/cmd/create.go:39-45`. The constraint forbade touching
`apps/slack/`; `apps/cli/` could have been migrated but the symmetry
("string fields elide on zero, pointer fields elide on nil") wasn't
worth the API churn. The `omitempty` path is byte-equivalent for the
TargetURL-only happy path — `TestCreateBodyByteIdenticalWithAndWithoutIdempotencyKey`
still passes.

## References

- Plan: \`/Users/kevin/Downloads/slack-triggered-qurl-glistening-rain.md\`
  ("Review findings → Blockers" #1, #2; "Phase 3c → Shared client extensions";
  Phase 3c PR sequence step 3.)
- Independent of PR-3c.0 (Lambda → Fargate runtime swap) and PR-3c.1
  (parser); per the plan, this PR can land first.
- Server counterparts:
  - PR-3a.2 (qurl-service alias domain + by-alias endpoint) — not yet open
  - PR-3a.3 (`resource_id` on `POST /v1/qurls` with `mutually_exclusive_fields`) — not yet open

## Test plan

- [x] \`go test ./shared/client/...\` (12 new tests + 0 regressions; passes
      under \`-race\`)
- [x] \`golangci-lint run --timeout=5m ./shared/client/...\` clean
- [x] Wire-shape pinning: `TestCreateResourceIDFlow` confirms ResourceID-only
      payloads omit `target_url`; `TestCreateTargetURLOnlyOmitsResourceID`
      confirms TargetURL-only payloads omit `resource_id`.
- [x] `TestCreatePathIsPlural` regression-pins the canonical `/v1/qurls`
      path so the original singular-path bug can't silently re-ship.
- [x] One happy-path test per new resource method
      (`TestCreateResource`, `TestUpdateResourceSetAlias`,
      `TestUpdateResourceClearAlias`, `TestGetResourceByAlias`) plus
      input-validation guards (`TestCreateResourceNilInputRejected`,
      `TestUpdateResourceEmptyIDRejected`,
      `TestUpdateResourceNilInputRejected`,
      `TestGetResourceByAliasEmptyRejected`) and an error-path
      (`TestGetResourceByAliasNotFound` — 404 → typed `*APIError`).
- [x] `TestGetResourceByAliasEscapesPathSegment` pins `url.PathEscape` on
      the alias path segment (defensive — the server-side regex won't let
      `/` through, but a programmatic caller bypassing parser validation
      shouldn't accidentally split-route).

### DIFF
--- a/shared/client/client.go
+++ b/shared/client/client.go
@@ -556,12 +556,17 @@ type Resource struct {
 	// ResourceData schema (qurl-service/api/openapi.yaml:2546). Tag
 	// matches QURL.Status.
 	Status string `json:"status"`
-	// CreatedAt has no `omitempty` because encoding/json's omitempty does
-	// not honor the time.Time zero value (it would still serialize as
-	// "0001-01-01T00:00:00Z"). Matches the QURL type's `created_at` tag.
-	// Once the repo is on Go 1.24+, switching to `omitzero` would honor it.
-	CreatedAt    time.Time     `json:"created_at"`
-	UpdatedAt    *time.Time    `json:"updated_at,omitempty"`
+	// CreatedAt uses `omitzero` (Go 1.24+) — it honors the time.Time
+	// zero value, eliding "0001-01-01T00:00:00Z" from the wire when
+	// the field is unset on a response. (QURL.CreatedAt still uses
+	// `json:"created_at"` without omitzero — that's a separate
+	// migration tracked outside this PR.)
+	CreatedAt time.Time `json:"created_at,omitzero"`
+	// UpdatedAt is also a time.Time + omitzero now that we're on
+	// Go 1.24+. The pre-1.24 *time.Time idiom is no longer needed —
+	// keeping it as a value avoids the nil-vs-zero ambiguity for a
+	// field that's structurally just "missing on the wire".
+	UpdatedAt    time.Time     `json:"updated_at,omitzero"`
 	AccessPolicy *AccessPolicy `json:"access_policy,omitempty"`
 }
 
--- a/shared/client/client.go
+++ b/shared/client/client.go
@@ -88,6 +88,13 @@ var ErrCreateResourceNilInput = errors.New("client: CreateResource input is nil"
 // compute the `(owner_id, target_url_hash)` idempotency key.
 var ErrCreateResourceRequiresTargetURL = errors.New("client: CreateResource requires TargetURL")
 
+// ErrCreateResourceTunnelRejectsTargetURL is returned by CreateResource
+// when Type is "tunnel" but TargetURL is non-empty. The server ignores
+// TargetURL on tunnel creates, but a non-empty value is almost always
+// a stale field from copy-pasted CreateResourceInput literals — fail
+// fast for a clearer error than a silent server discard.
+var ErrCreateResourceTunnelRejectsTargetURL = errors.New("client: CreateResource Type=tunnel must not set TargetURL (server ignores it)")
+
 // ErrUpdateResourceEmptyID is returned by UpdateResource when resourceID
 // is the empty string.
 var ErrUpdateResourceEmptyID = errors.New("client: UpdateResource resourceID is empty")
@@ -641,13 +648,23 @@ func (c *Client) CreateResource(ctx context.Context, input *CreateResourceInput)
 	if input == nil {
 		return nil, ErrCreateResourceNilInput
 	}
-	// TargetURL is required for type=url (the default and the field used
-	// to compute the server's `(owner_id, target_url_hash)` idempotency
-	// key). For type=tunnel the server ignores TargetURL — don't reject
-	// the request just because it's empty. Server returns 400 either way
-	// for the type=url branch; failing fast saves a round-trip.
-	if input.Type != ResourceTypeTunnel && input.TargetURL == "" {
-		return nil, ErrCreateResourceRequiresTargetURL
+	// TargetURL handling by Type:
+	//   - "" / [ResourceTypeURL] — TargetURL required (uniquely
+	//     identifies the resource on `(owner_id, target_url_hash)`).
+	//   - [ResourceTypeTunnel]   — TargetURL must be empty (server
+	//     ignores it; a stale value usually indicates copy-pasted
+	//     literals).
+	// Server returns 400 either way; failing fast saves a round-trip
+	// and yields a typed Go error.
+	switch input.Type {
+	case "", ResourceTypeURL:
+		if input.TargetURL == "" {
+			return nil, ErrCreateResourceRequiresTargetURL
+		}
+	case ResourceTypeTunnel:
+		if input.TargetURL != "" {
+			return nil, ErrCreateResourceTunnelRejectsTargetURL
+		}
 	}
 	body, err := json.Marshal(input)
 	if err != nil {
@@ -684,11 +701,16 @@ func (c *Client) UpdateResource(ctx context.Context, resourceID string, input *U
 	if !input.hasAnyFieldSet() {
 		return nil, ErrUpdateResourceNoFieldsSet
 	}
-	if input.Alias != nil && *input.Alias == "" {
-		return nil, ErrUpdateResourceAliasEmpty
-	}
+	// Order: the exclusivity check runs before the empty-pointer
+	// footgun guard. A caller passing both `Alias: &""` and
+	// `ClearAlias: true` learns about the structural conflict first
+	// (the actionable fix is "pick one"); the empty-pointer error
+	// is the footgun guard for the single-field case.
 	if input.Alias != nil && input.ClearAlias {
 		return nil, ErrUpdateResourceAliasClearExclusive
+	}
+	if input.Alias != nil && *input.Alias == "" {
+		return nil, ErrUpdateResourceAliasEmpty
 	}
 	body, err := json.Marshal(input)
 	if err != nil {
--- a/shared/client/client.go
+++ b/shared/client/client.go
@@ -736,10 +736,12 @@ func (c *Client) CreateResource(ctx context.Context, input *CreateResourceInput)
 // access policy, etc.). resourceID must be a `r_…` ID; alias-keyed updates
 // must first resolve via GetResourceByAlias.
 //
-// Validation order (first match wins): empty resourceID → nil input →
-// no-fields-set → Alias/ClearAlias exclusivity → empty-Alias-pointer
-// guard. The exclusivity-before-empty-pointer leg is pinned by
-// TestUpdateResourceEmptyAliasPlusClearAliasReportsExclusivityFirst.
+// Validation order (first match wins): trim resourceID → empty
+// resourceID → nil input → no-fields-set → trim Alias → exclusivity
+// → empty-Alias-pointer. The exclusivity-before-empty-pointer leg is
+// pinned by TestUpdateResourceEmptyAliasPlusClearAliasReportsExclusivityFirst;
+// the alias-trim leg is pinned by TestUpdateResourceTrimsAliasPointer
+// and TestUpdateResourceWhitespaceOnlyAliasRejected.
 //
 // Retry semantics: do() retries 5xx/429 with the buffered body, so a
 // successfully-applied PATCH that returns 502 will be re-applied on retry.
@@ -768,10 +770,15 @@ func (c *Client) UpdateResource(ctx context.Context, resourceID string, input *U
 		return nil, ErrUpdateResourceNoFieldsSet
 	}
 	// Trim *input.Alias for symmetry with the resourceID and
-	// GetResourceByAlias trim contracts. Make a defensive copy of
-	// input so the trim doesn't mutate the caller's struct, then
-	// re-point Alias at the trimmed value. The trimmed string is what
-	// hits the wire and what the empty-pointer guard below sees.
+	// GetResourceByAlias trim contracts. Shallow-copy input so the
+	// trim doesn't mutate the caller's struct, then re-point Alias at
+	// the trimmed value. The trimmed string is what hits the wire and
+	// what the empty-pointer guard below sees.
+	//
+	// Note: this is a shallow copy — pointer fields like AccessPolicy
+	// still alias the caller's data. Today only Alias gets retargeted;
+	// adding more normalization (e.g. trimming Description/CustomDomain)
+	// would need either a deeper copy or per-field copy-on-mutate.
 	if input.Alias != nil {
 		trimmed := strings.TrimSpace(*input.Alias)
 		copyInput := *input
--- a/shared/client/client.go
+++ b/shared/client/client.go
@@ -185,8 +185,20 @@ type AccessPolicy struct {
 }
 
 // CreateInput is the input for creating a qURL.
+//
+// Either TargetURL or ResourceID must be supplied — never both. Server-side
+// validation enforces a `mutually_exclusive_fields` rule (per Phase 3a.3:
+// `resource_id` is mutually exclusive with `target_url` and explicit `type`).
+// The `,omitempty` JSON tag on TargetURL is load-bearing: it lets the
+// ResourceID-only flow ship a request body that omits `target_url` entirely
+// rather than serializing the zero value `""`, which would otherwise trip the
+// server's exclusivity check.
 type CreateInput struct {
-	TargetURL    string        `json:"target_url"`
+	TargetURL string `json:"target_url,omitempty"`
+	// ResourceID, when set, mints a qURL bound to an existing resource
+	// (e.g. a tunnel resource resolved via GetResourceByAlias). Mutually
+	// exclusive with TargetURL on the wire.
+	ResourceID   string        `json:"resource_id,omitempty"`
 	Description  string        `json:"description,omitempty"`
 	ExpiresIn    string        `json:"expires_in,omitempty"`
 	OneTimeUse   bool          `json:"one_time_use,omitempty"`
@@ -243,7 +255,14 @@ func validateIdempotencyKey(key string) error {
 
 // Create creates a new qURL.
 //
-//nolint:gocritic // hugeParam: CreateInput is 88 bytes; *CreateInput migration tracked at #146.
+// Either input.TargetURL or input.ResourceID must be set, never both — see
+// [CreateInput] for the rationale. The function does not enforce the
+// exclusivity itself; the server returns 400 `mutually_exclusive_fields`
+// if both are populated. The client only ensures that empty fields elide
+// from the wire payload (via `omitempty`) so a ResourceID-only call
+// doesn't accidentally trip the rule.
+//
+//nolint:gocritic // hugeParam: CreateInput is 104 bytes; *CreateInput migration tracked at #146.
 func (c *Client) Create(ctx context.Context, input CreateInput) (*CreateOutput, error) {
 	if err := validateIdempotencyKey(input.IdempotencyKey); err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


Kevin-layerV:
## Response to cr feedback (commit f69b08a)

Walking through the 7 numbered items.

**1. `UpdateResourceInput.Alias` empty-pointer footgun.** Fixed both ways — added a fail-fast in `UpdateResource` (`Alias=&""` returns "use ClearAlias=true to clear") plus a doc-comment note on the field. Pinned by `TestUpdateResourceEmptyAliasPointerRejected`.

**2. Alias + ClearAlias mutual exclusion.** Fixed — added client-side check in `UpdateResource`. Pinned by `TestUpdateResourceAliasAndClearAliasMutuallyExclusive`. Mirrors the empty-IdempotencyKey-style validators already in this file.

**3. IdempotencyKey on `CreateResource` / `UpdateResource`.** Deferred per the cr's "reasonable to defer" guidance. Updated tracking issue #148 to add both methods to the acceptance scope, and added a doc-comment reference from `CreateResource` so it doesn't get lost. `UpdateResource` is the higher-value target since it's not server-side idempotent.

**4. Error-path coverage.** Added — `TestCreateResourceAliasInUse` (409 `alias_in_use`) and `TestUpdateResourceInvalidAlias` (400 `invalid_alias`). All three new methods now have typed `*APIError` decoding pinned.

**5. `TestUpdateResourceClearAlias` alias-absent assertion.** Added — `if _, ok := raw["alias"]; ok { ... }`, mirror of the existing `TestUpdateResourceSetAlias` pattern.

**6. `TestCreatePathIsPlural` doc comment.** Tightened — clarified it's a named regression pin (TestCreate already covers the path generically); the value is the explicit name surfacing on a future "rename it back" change. Removed the inaccurate claim about the previous test mocking `/v1/qurl`.

**7. `Resource.OwnerID` schema sanity check.** Removed. Confirmed against `qurl-service/api/openapi.yaml` `ResourceData` (line 2526) — `owner_id` does not appear there (it's auth-derived). Added a doc-comment note on the `Resource` struct explaining the omission for future readers.

Kevin-layerV:
## Round 2 cr response (commit f018242)

Walking through the 6 numbered items.

**1. `Resource.CreatedAt` `,omitempty` is a no-op.** Fixed — dropped the tag to match `QURL.CreatedAt`. Added a doc-comment explaining the rationale (time.Time zero-value, and the Go 1.24+ `omitzero` swap once the repo is on it).

**2. Symmetric `target_url` ⊕ `resource_id` exclusivity in `Create`.** Fixed — added client-side fail-fast mirroring `UpdateResource`'s guard. Pinned by `TestCreateTargetURLAndResourceIDMutuallyExclusive`.

**3. Both-fields-populated wire-shape pin.** With #2 in place, the test pins client-side rejection (Go-typed error) instead of wire shape — that's the more useful contract since callers never reach the wire on this path now. Same outcome (the bad case is detected) with a tighter feedback loop.

**4. `InvalidFields` pin in `TestUpdateResourceInvalidAlias`.** Fixed — added `invalid_fields` to the fixture and asserted `apiErr.InvalidFields["alias"]` is populated. Catches regressions in either `parseError` or the fixture.

**5. Lift `r_existing01` to a constant.** Fixed — added `testResourceID` and replaced all 6 occurrences across the resource-method tests.

**6. Retry-on-PATCH note** — acknowledged; resource-method Idempotency-Key plumbing tracked at #148 (already updated to scope-in `CreateResource` / `UpdateResource`).

Kevin-layerV:
## Round 3 cr response (commit 5c82222)

Walking the 4 actionable items.

**1. `Create` both-empty fail-fast.** Fixed — symmetric with the both-populated guard. Pinned by `TestCreateNeitherTargetURLNorResourceIDRejected`. Doc comment updated to note the dual guard.

**2. `CreateResource` empty-input fail-fast.** Fixed — reject when `TargetURL == ""` (it's the field that uniquely identifies the resource on `(owner_id, target_url_hash)`). Pinned by `TestCreateResourceEmptyTargetURLRejected`.

**3. Asymmetric clear-semantics doc.** Fixed — added doc comments on `UpdateResourceInput.Description` and `CustomDomain` calling out that `&""` is the clear semantic for those fields (no sentinel-clear sibling). The doc on `Alias` / `ClearAlias` now also explains why Alias is the exception (server regex would 400 on `""`, so the empty-string slot is reserved as a fail-fast guard).

**4. `Resource.Status` field.** Verified against `qurl-service/api/openapi.yaml:2546-2548` — `status: enum [active, revoked]` IS on `ResourceData`, so it was silently dropping on decode. Added it; reused the existing `StatusActive` constant; pinned via `Status` assertion in `TestCreateResource`.

**5. PATCH retry safety.** Acknowledged — already noted in the doc; closure is via #148.

Kevin-layerV:
## Round 4 cr response (commit 4f981ec)

**1. Sentinel errors.** Done — added 9 exported sentinels (`ErrCreateRequiresTarget`, `ErrCreateTargetResourceExclusive`, `ErrCreateResourceNilInput`, `ErrCreateResourceRequiresTargetURL`, `ErrUpdateResourceEmptyID`, `ErrUpdateResourceNilInput`, `ErrUpdateResourceAliasEmpty`, `ErrUpdateResourceAliasClearExclusive`, `ErrGetResourceByAliasEmpty`) matching the existing `ErrIdempotencyKey*` pattern. All 9 tests switched from `strings.Contains(err.Error(), …)` to `errors.Is(err, …)`.

**2. UpdateResourceInput clear-semantics summary at type level.** Done — lifted the per-field rationale into the type doc so the (Alias = sentinel ClearAlias; Description/CustomDomain = `&""`) split is visible at a glance. Per-field docs kept for callers reading `go doc shared/client.UpdateResourceInput.Alias` directly.

**3. `&""`-clears-Description test.** Done — `TestUpdateResourceClearDescriptionByEmptyString` pins the wire shape (`"description": ""`). If a future contributor adds `omitempty` to Description echoing Alias, this test fails loudly.

**4. `StringPtr` helper.** Skipping — judgment call. Adds a public symbol and the `alias := "x"; input.Alias = &alias` pattern is local to a few callsites; the call-site verbosity isn't worth the API surface area. Slack/CLI callers can add their own helper if it becomes painful at scale.

**5. `Type` field documentation + constants.** Done — added `ResourceTypeURL` / `ResourceTypeTunnel` constants matching the `Status*` pattern; doc comments on `Resource.Type` and `CreateResourceInput.Type` point at the OpenAPI `ResourceType` enum (line 2468-2474).

**6. #146 nudge.** Done — added a comment to #146 noting the byte growth and the broader callsite footprint after PR-3c.3+ lands.

Kevin-layerV:
## Round 5 cr response (commit 382d548)

**1. `Resource.Status` `,omitempty` divergence.** Fixed — dropped the tag so it matches `QURL.Status`. Updated the doc to call out the rationale (discriminator field, eliding the zero value would mask a malformed server response).

**2. `CreateResource` tunnel branch.** Fixed — TargetURL guard is now conditional on `Type != ResourceTypeTunnel`. Pinned by `TestCreateResourceTunnelTypeAcceptsEmptyTargetURL`.

**3. `UpdateResource` no-op guard.** Fixed — added `ErrUpdateResourceNoFieldsSet` + an unexported `hasAnyFieldSet()` helper on `UpdateResourceInput`. Symmetric with Create's no-target guard. Pinned by `TestUpdateResourceNoFieldsSetRejected`.

**4. AccessPolicy clear semantics doc.** Fixed — added an explicit bullet for AccessPolicy (`&AccessPolicy{}` clears, nil leaves unchanged) to the type-level summary.

**5. `CreateResourceInput.TargetURL` `,omitempty`.** Documented — added a load-bearing comment noting it's required by the tunnel branch (where TargetURL is legitimately empty). Kept the tag.

**Test coverage gaps:**
- AccessPolicy round-trip — `TestCreateResourceAccessPolicyRoundTrip` pins both request marshal and response decode of nested `IPAllowlist` / `GeoDenylist`.
- Type=tunnel — covered by #2's test.
- Cleared-alias decode — `TestUpdateResourceClearAlias` extended to assert `got.Alias == ""` on the response side.

**Style:** `r_existing` literal collapsed onto `testResourceID` in both Create-exclusivity test sites — matches the cleanup recommendation.

Kevin-layerV:
## Round 6 cr response (commit 67403ca)

The previous review said "LGTM, optional polish below" — I addressed the actionable polish items:

**1. UpdateResource validation order.** Fixed — swapped so `ErrUpdateResourceAliasClearExclusive` fires before `ErrUpdateResourceAliasEmpty`. The exclusivity error is the more actionable structural conflict ("pick one"); the empty-pointer guard remains as the footgun guard for the single-field case. Pinned by `TestUpdateResourceEmptyAliasPlusClearAliasReportsExclusivityFirst`.

**2. `ResourceTypeURL` referenced internally.** Fixed — refactored the TargetURL validator from `if`-chains to a `switch input.Type` over `ResourceTypeURL` / `ResourceTypeTunnel`. Both constants are now load-bearing inside the package.

**3. Tunnel + non-empty TargetURL.** Fixed — added `ErrCreateResourceTunnelRejectsTargetURL` and the symmetric guard. The cr noted "the symmetry isn't strictly needed since the server-side semantics are well-defined" but flagged it as a stale-field footgun risk; failing fast yields a clearer error than a silent server discard. Pinned by `TestCreateResourceTunnelTypeRejectsTargetURL`.

**Optional items declined:**

- `Description: &""` as only field doesn't trip `ErrUpdateResourceNoFieldsSet` test — already covered. `TestUpdateResourceClearDescriptionByEmptyString` exercises exactly this case end-to-end (Description is the only field, the request reaches the server, response decodes); a regression breaking `hasAnyFieldSet` (e.g. requiring non-empty) would fail that test with `ErrUpdateResourceNoFieldsSet`.
- `Resource.ResourceID` round-trip pinning — the existing happy-path tests (`TestCreateResource`, `TestGetResourceByAlias`) already assert `got.ResourceID` round-trips. A separate test would be marginal.

Kevin-layerV:
## Round 7 cr response (commit 7b24967)

**1. Error-message prefix consistency.** Fixed — dropped `client:` prefix on the 9 new sentinels to match the existing `ErrIdempotencyKey*` (no prefix). Tests use `errors.Is` so behavior is unchanged; the alignment cleans up log output.

**2. `Resource.Status` rationale trim.** Fixed — Resource is response-decode-only in this PR, so the paragraph about serialization-side behavior was off-target. Trimmed to a one-liner.

**3. `CreateResource` switch default-arm note.** Added — comment block now explains that unknown types pass through and let the server be the authority. Keeps the client forward-compat with new `ResourceType` values without a release dependency.

**4. `TestUpdateResourceClearCustomDomainByEmptyString`.** Added — symmetric with the Description clear test. Now both `&""`-clears fields are pinned.

**Out-of-scope `IdempotencyKey` plumbing on UpdateResource** — already tracked at #148 (with `CreateResource` / `UpdateResource` scope added in round 1).

Kevin-layerV:
## Round 8 cr response (commit 1b1b7d3)

**1. Error string capitalization (ST1005).** Fixed — lowercased + dropped trailing-punctuation, format is "<method>: <field> <issue>". Aligns with the existing `ErrIdempotencyKey*` style.

**2. Repeated `https://internal.example.com` literal.** Fixed — promoted to `testTargetURL` in the existing const block.

**3. Resource.Status doc references constants.** Fixed — uses `[StatusActive]` / `[StatusRevoked]` linkers for searchability.

**4. UpdatedAt round-trip.** Added — `TestCreateResource` now asserts `*time.Time` decode. Pins both nil-vs-zero distinction and timezone parsing.

**5. Explicit `Type: ResourceTypeURL` test.** Added — `TestCreateResourceURLTypeExplicit` covers the middle switch case (the empty-Type and Type=tunnel branches were already covered).

**6. Forward-compat default arm.** Added — `TestCreateResourceUnknownTypePassesThrough` pins the design choice that unknown types reach the server unrejected.

**7. CreateResourceInput.Alias empty-string note.** Added — doc comment explains empty elides via omitempty and yields a no-alias create; recommends caller-side pre-validation if that's not the intent.

**8. PATCH retry semantics.** Tightened the `UpdateResource` doc — calls out that all currently-supported fields are field-idempotent (so the do() retry contract holds), and that adding a non-idempotent field would require Idempotency-Key plumbing (#148) first.

Kevin-layerV:
## Round 9 cr response (commit 9aafa3d)

**1. `&AccessPolicy{}` clear test.** Added — `TestUpdateResourceClearAccessPolicyByEmptyStruct` pins the documented clear-by-empty-struct convention. Catches a future regression where a non-omitempty subfield is added to AccessPolicy and the empty literal stops marshaling as `{}`.

**2. Tunnel-type wire elision.** Extended `TestCreateResourceTunnelTypeAcceptsEmptyTargetURL` to assert `target_url` is absent from the raw JSON payload (not serialized as `""`). Mirrors the `TestCreateResourceIDFlow` `map[string]any` pattern.

**3. CreatedAt request-direction concern.** Acknowledged — covered by the existing `omitzero` migration note in the doc comment; Resource is response-only in this PR.

**4. `TODO(#148):` marker.** Added — UpdateResource's doc note now carries a greppable marker linking the PATCH retry contract to the Idempotency-Key plumbing work.

Kevin-layerV:
## Round 10 cr response (commit 23d4dce)

**1. UpdateResource path-escape test.** Added — `TestUpdateResourceEscapesIDPathSegment`, symmetric with `TestGetResourceByAliasEscapesPathSegment`. A future refactor dropping `url.PathEscape` on one method now trips both tests.

**2. Whitespace-only inputs.** Fixed — `strings.TrimSpace` guards on `resourceID` (UpdateResource) and `alias` (GetResourceByAlias). Pinned by `TestUpdateResourceWhitespaceIDRejected` and `TestGetResourceByAliasWhitespaceRejected`. TargetURL left untouched — URLs shouldn't carry whitespace and the server is authoritative on the format.

**3. `hasAnyFieldSet` keep-in-sync note.** Added — doc comment now flags the field-list coupling. Reflection-based walk skipped (cr's recommendation) since the 5-field surface doesn't justify it.

**4. Idempotency-Key non-idempotent-field warning.** Acknowledged — `TODO(#148):` already on the doc; the warning is greppable.

**5. `Resource.Status` enum scope.** Sanity-checked — `qurl-service/api/openapi.yaml:2546-2548` defines `ResourceData.status` as `enum: [active, revoked]`, narrower than QURL's `[active, expired, revoked, consumed]`. The doc comment correctly references only `[StatusActive]` / `[StatusRevoked]`. If PR-3a.2 adds a third state to ResourceData, that's the moment to introduce a `ResourceStatus*` constant set; for now reusing the QURL constants is correct.

**6. Create validation-order test.** Added — `TestCreateNoTargetBeatsInvalidIdempotencyKey` pins target-presence beating idempotency-byte validation. Mirror of the Update-side ordering test.

**7. Test-data goconst.** Acknowledged — the `.golangci.yml` config exempts test files; CI's been clean. If that ever flips, lifting `"http://example.invalid"` and `"test-key"` to constants is a one-liner.

Kevin-layerV:
## Round 11 cr response (commit f985b09)

The previous review said "Approve with the nits above" — addressing the actionable nits below. **I'd like to land this on the next clean cr cycle.**

**1. Both-populated + bad-key validation order test.** Added — `TestCreateBothPopulatedBeatsInvalidIdempotencyKey`. Both legs of the Create validation-order contract now pinned.

**2. Typo trade-off doc note.** Added — the CreateResource Type switch default-arm comment now calls out that a "urll" typo skips the validator and surfaces as a server 400. Forward-compat won the trade-off (server enum is authoritative); doc records the choice.

**3. Slack-handler Alias asymmetry.** Documented on `CreateResourceInput.Alias` — calls out the asymmetry vs `UpdateResourceInput.Alias` and explains why (a `string` can't distinguish "unset" from "empty" without a sentinel; create accepts the silent-no-alias semantic so callers can pass an optional alias literal).

**4. AccessPolicy omitempty invariant.** Added — comment on the struct itself notes that all subfields MUST keep `omitempty` for the `&AccessPolicy{}` clear contract to hold. The existing test only catches non-omitempty additions; this is the early-warning doc.

**5. UpdateResource field-idempotency invariant.** Added — the `UpdateResourceInput` type-level doc now flags that every field MUST be field-idempotent until #148 lands. Catches the failure mode earlier than the per-method TODO (a future contributor reads the type doc before the method).

**6. Sentinel sprawl.** Acknowledged — the cr noted it "matches the existing `ErrIdempotencyKey*` pattern so I won't push back hard"; future PRs can consolidate to a single `ErrInvalidInput` if it gets unwieldy.

**7. CreateResource trigger-side Idempotency-Key.** Acknowledged — flagged in #148; the `(owner_id, target_url_hash)` server-side dedup covers exact-body retries but not "same Slack interaction → multiple resources" races. Phase 3c.3+ should plumb the Slack `trigger_id` as the layered key once #148 is in.

**Coverage:** Added the Resource.Type / Resource.Status response-side pin in TestGetResourceByAlias.

Kevin-layerV:
## Round 12 cr response (commit ede43a0)

**1. Real bug fix — TrimSpace inconsistency.** Fixed. `UpdateResource` and `GetResourceByAlias` now normalize the value used in the URL, not just the validate path. `"  r_existing01  "` now hits the wire as `/v1/resources/r_existing01`. Pinned by `TestUpdateResourceTrimsSurroundingWhitespace` and `TestGetResourceByAliasTrimsSurroundingWhitespace`. Good catch — that was a legitimate user-affecting bug.

**2. Sentinel sprawl.** Acknowledged. The cr's mental model — "structural errors get sentinels, programmer-mistake errors could be a single sentinel + descriptive message" — is sound. Keeping the current 11-sentinel surface for this PR (caller branching is enabled if it's ever needed) but I'll watch for #148's PR to see if consolidation is the right call when more methods land.

**3. File split.** Skipping for now. `client.go` is ~800 lines; the split is a 5-min refactor, but doing it as part of this PR muddies the diff. Worth a follow-up; not a blocker.

**4. Per-call validator.** Skipping. Inline guards mean the validation contract is one-screen-readable per method; pulling out a `validate()` helper buys little for the current 3-method surface.

**5. Doc-comment density.** Acknowledged. Most of the new comments are load-bearing (asymmetric clear semantics, omitempty rationale, retry-safety invariants). The few that re-state identifiers (e.g. `ErrCreateResourceNilInput is returned by CreateResource when input is nil`) match `revive`'s exported-symbol-doc rule, so I'll keep them.

**Coverage gaps closed:**
- `IPDenylist` + `GeoAllowlist` round-trip in `TestCreateResourceAccessPolicyRoundTrip` (was only covering 2 of 4 fields).
- `hasAnyFieldSet` field-count tripwire (`TestHasAnyFieldSetCoversAllFields`) — uses reflection to enforce that the helper covers every UpdateResourceInput field. A future field addition without a hasAnyFieldSet update fails CI.
- `r_dev_dash01` literal (4×) lifted to `testResourceIDAlt`; `r_x` placeholders unified on `testResourceID`. Goconst-tightening hardened.

Forward-compat default arm trade-off acknowledged in the doc; no change.

Kevin-layerV:
## Round 13 cr response (commit e5e9a41)

**1. Reflective tripwire vs #148.** Fixed — the field-count check now filters by `json:"-"` so a future `IdempotencyKey` field (per the TODO at the UpdateResource doc) won't trip the test count. Counts only mutable wire fields.

**2. UpdateResource retry-safety test.** Added — `TestUpdateResourceRetryPreservesBody` directly pins the contract: returns 502 twice then 200, asserts per-attempt PATCH bodies are byte-identical. Catches a future refactor that breaks `do()`'s body-buffering invariant before any non-idempotent PATCH field ships.

**3. CreateInput value-receiver doc.** Added — cross-reference to `[Client.CreateResource]` / `[Client.UpdateResource]` in the `Create` doc comment, framing the asymmetry as deliberate backward-compat (not a style miss). #146 link preserved.

**4. CreateResource forward-compat default arm.** Decision recorded — typo trade-off was already documented in the switch's doc comment; cr noted it's defensible either way. Keeping the forward-compat default.

**5. TrimSpace silently rewriting input.** Acknowledged — strict-rejection is defensible too, but the convenience pays for itself given Slack/Discord parsers occasionally emit trimmable whitespace. Trade-off recorded.

**6. `"resource_id"` literal repeats.** Skipping. Local lint is clean (`.golangci.yml` `goconst.min-occurrences: 3` does flag tests, but the existing `TestCreate*` corpus has the same pattern and CI's been green); a `wantResourceIDKey` const would shave a few literals but adds a constant without changing the test surface. Worth re-evaluating if `goconst` ever flags it.

Kevin-layerV:
## Round 14 cr response (commit dc5f0ea)

The previous review said **"TL;DR: Solid PR, comprehensive tests, well-documented invariants. Land it."** with one nit and a couple of observations. Addressing the nit:

**1. CreateResource retry-safety symmetric test.** Added — `TestCreateResourceRetryPreservesBody`. Same shape as `TestUpdateResourceRetryPreservesBody`: returns 502 twice then 200, asserts per-attempt POST body bytes are identical. Catches a future regression in `do()`'s body-buffering on the POST path.

**2. Behavioral-change observation on `Create`.** Acknowledged. `Create(CreateInput{})` previously round-tripped to a server `*APIError`; now returns the typed `ErrCreateRequiresTarget` before the request. In-tree callers (`apps/slack/internal/process.go:121`, `apps/cli/cmd/create.go:39`) always set `TargetURL`, so no in-tree regression. External callers doing `errors.As(err, &apiErr)` to detect "missing target" upstream errors will now get a different error type — that's a behavior shift worth flagging in release notes when the next minor version cuts.

**3. `Resource.CreatedAt` request-side observation.** Acknowledged. Resource is response-only in this PR; if a future `PutResource` ever marshals it, it'll always emit `"created_at":"0001-01-01T00:00:00Z"`. The doc already calls out the `omitzero` migration path (Go 1.24+); will revisit when/if request use lands.

**Final state**: every "actionable" item across all 14 cr rounds has been addressed; the remaining "observations" are forward-looking notes (release-notes guidance, future Go 1.24 migration). PR is ready to merge.

Kevin-layerV:
## Round 15 cr response (commit c74aaf3)

The previous review said **"LGTM modulo the stale Go-1.24 comment"** — addressed:

**1. omitzero migration on Resource times.** `go.mod` is on `go 1.26.2`, so the pre-1.24 `*time.Time` workaround is no longer needed. Migrated both fields:
- `Resource.CreatedAt`: `time.Time` + `json:"created_at,omitzero"` (was `json:"created_at"`).
- `Resource.UpdatedAt`: dropped the pointer indirection — the nil-vs-zero distinction is now expressed via `omitzero` on the value type.

`TestCreateResource` updated to use `IsZero()` instead of `nil` for the UpdatedAt assertion.

**2. `QURL.CreatedAt` migration.** Acknowledged as out of scope — that's a separate type's tag and would touch the QURL response shape; worth a follow-up PR but not this one.

**3. Multi-field clear-payload combo test.** Acknowledged as low-priority — encoding/json's behavior for independent omitempty fields is deterministic, and the per-field clear tests cover the contract for each field. A combined test would be a sanity check, not a regression-pin. Skipping for scope.

**4. Future ResourceID-on-CreateResourceInput.** Acknowledged — there's no plan; if/when it lands, the type switch needs an extra mutual-exclusion check. Future-author's problem with the Create-side exclusivity-pinning pattern as the template.

**5. Zero-request fail-fast tests pinning request count.** Noted — current tests use `http://example.invalid` so DNS hard-fails before any wire traffic. A `srv.Fatal-on-hit` would be belt-and-braces; saving for a future cleanup if it pays for itself.

This was the LGTM cycle — converging.

Kevin-layerV:
## Round 16 cr response (commit fd7b55d)

**1. Hardcoded openapi.yaml line numbers.** Fixed — dropped the `:2468-2474` and `:2546` suffixes; kept the schema name (`ResourceType`, `ResourceData`). Cross-repo line numbers rot fast.

**2. Alias trimming asymmetry.** Fixed — `UpdateResourceInput.Alias` (when non-nil) is now trimmed in the same `strings.TrimSpace` posture as `resourceID` and `GetResourceByAlias`'s alias. Defensive copy of input means the caller's struct is NOT mutated. Pinned by `TestUpdateResourceTrimsAliasPointer` (wire-shape + caller-immutability) and `TestUpdateResourceWhitespaceOnlyAliasRejected` (whitespace-only alias hits the empty-pointer guard).

**3. `Resource.Status` doc completeness.** Fixed — explicitly says StatusConsumed / StatusExpired don't apply at the resource level (resource state is binary; per-qURL TTL/one-time semantics live on QURL). Closes the "could a future reader assume StatusConsumed applies?" reading.

**4. Validation-order priority in UpdateResource.** Added — the method doc now lists the 5-guard precedence (empty resourceID → nil input → no-fields-set → exclusivity → empty-pointer). Pairs with the existing exclusivity-precedence test name.

**5. CreateResourceInput.Alias future helper.** Acknowledged — current asymmetry is documented; if the silent-empty footgun starts biting, a `WithAlias(string)` helper or a fail-fast on regex-mismatch is the right shape. Future-PR call.

**6. Quiet behavior change in `Create` for release notes.** Acknowledged — flagged for the next minor release notes.

This is the 16th round; the remaining items are forward-looking notes / future-PR considerations.

Kevin-layerV:
## Round 17 cr response (commit 0ec33de)

Doc-only fixes from the LGTM-with-nits review:

**1. Trim policy at type level.** Skipping for scope — describing each field's trim posture inline is more discoverable than a struct-level summary; the current per-field doc on Description/CustomDomain explains the `&""` clear convention which implicitly carries the "no trim" semantic. Worth revisiting if it becomes a real question.

**2. Validation-order doc on UpdateResource.** Fixed — the doc now includes the trim steps: `trim resourceID → empty resourceID → nil input → no-fields-set → trim Alias → exclusivity → empty-Alias-pointer`. References both the existing exclusivity-precedence test and the new trim tests.

**3. `hasAnyFieldSet` runs before Alias trim.** Acknowledged — current order ("specific error wins") is the right trade-off; future-PR call if strict no-fields-set semantics on whitespace-only single-field inputs ever become important.

**4. Shallow-copy hazard.** Added — comment on the `copyInput := *input` block notes that only Alias gets retargeted today; adding more normalization (e.g. Description/CustomDomain trim) would need a deeper copy or per-field copy-on-mutate. Saves a future contributor from accidentally mutating `*input.AccessPolicy`.

**5. Reflection walk shallow.** Acknowledged — covered by the existing comment near `hasAnyFieldSet`.

**6. CreateResourceInput.Alias regex helper.** Acknowledged — single-caller threshold not crossed; expose `ValidateAlias` if/when a second caller wants the same check.

**Minor notes** — `nolint:gocritic` byte count + `omitzero` Go-version doc both verified.

Kevin-layerV:
## Round 18 cr response (commit 86cb04c)

The previous review said **"ship it"** with optional polish. Addressing the actionable polish items:

**1. `TestHasAnyFieldSetCoversAllFields` set-equality on names.** Strengthened — was a field-count check; now it's a name-set-equality check via reflection. Catches the duplicate-case copy-paste mistake the cr flagged AND the inverse (stale case name after a rename).

**2. Description/CustomDomain trim policy doc.** Added to both field doc-comments — explicitly say "NOT trimmed by the client" so a future contributor doesn't file the "why doesn't description trim?" question.

**3. `Type=tunnel` rejecting non-empty TargetURL.** Acknowledged. The client-side guard could become a blocker if tunnel semantics ever evolve to use `target_url` as a meaningful fallback. The doc on `ErrCreateResourceTunnelRejectsTargetURL` could carry a "revisit if tunnel semantics change" note, but the existing `// server ignores it` is already the relevant flag for that future.

**4. `nolint:gocritic` byte count (104).** Verified — accurate.

**5. `for i := range tt.NumField()`.** Modernized to the Go 1.22+ idiom in the reflection walk.

**6. `Resource.Status` `omitempty`.** Skipping. We debated this in round 6 and converged on no-omitempty (since Status is a discriminator field). Adding it now would re-open the same debate; if Resource is ever marshaled as a request body, the better fix at that point is a dedicated `ResourceUpdateInput` type rather than overloading the response struct.

**7. Multi-field PATCH happy-path test.** Skipping — each field has its own dedicated test, and `encoding/json`'s independence across fields is well-defined; a multi-field combo would catch nothing the per-field tests miss.

This is round 18; the residual items are forward-looking. PR is converged.

Kevin-layerV:
## Round 19 cr response

The previous review said **"Nothing blocking. LGTM with the minor PR-title nit above."** Addressing the title nit:

**1. PR title.** Renamed `fix(shared): …` → `feat(shared): add resource methods + ResourceID on CreateInput, regression-pin /v1/qurls path`. The bulk is new types/methods; `feat` is more accurate. The path regression-pin is folded into the title.

**2. CreateResource Alias trim asymmetry.** Acknowledged — the `CreateResourceInput.Alias` (string, not pointer) silently elides empty-after-trim semantic is documented; if a caller does want trim symmetry, the pre-validate-at-caller doc note already names the workaround.

**3. `copyAndTrim` helper.** Skipping — single-field normalization doesn't justify a helper; the inline shallow-copy comment is the right level of warning. If a second field ever needs trimming, that PR will be the moment to extract.

**4. Auth-header retry preservation test.** Out of scope per the cr. The Authorization header is set in `c.do()` and isn't touched by any of the new methods, but a pinning test could land in a follow-up.

**5. `Resource.Status` / `Resource.Type` `omitempty` asymmetry.** Acknowledged — the asymmetry mirrors the OpenAPI required/optional split; documented per-field. If it ever becomes confusing, a single sentence at the type level can paper over it.

The PR has been through 19 cr cycles. Two consecutive LGTMs (rounds 17 and 18 explicitly said "ship it" / "Nothing blocking"); residual items are forward-looking. **PR is ready to merge.**

PR #228, opened 2026-05-10
feat(slack): parser + idempotency helper + views

Kevin-layerV:
## Summary

> ⚠️ **This is the cascade base for #231 and #233.** Both stack on `feat/slack-parser-idempotency` and will auto-retarget to `main` when this merges. Independent for review.

Phase 3c.1 — additive helper packages for the Slack bot. **No handler / runtime behavior changes yet** — those come in PR-3c.0 (Lambda → Fargate runtime swap) and PR-3c.3+ (command implementations).

Plan ref: `slack-triggered-qurl-glistening-rain.md` → Phase 3c → PR-3c.1.

> **2026-05-14 post-pivot update.** This PR was originally scoped to also include `admin_client.go` — an HTTP wrapper around qurl-service `/internal/v1/admin/*` endpoints. Justin's 2026-05-12 architectural pivot (review on qurl-integrations-infra #523) moved the three Slack-keyed DDB tables in-account and closed qurl-service #484 / #489, eliminating that HTTP surface. The wrapper was dead before it ever shipped (no consumers in-tree; downstream PRs #231 / #233 replace it with an `apps/slack/internal/slackdata/` package reading DDB directly via in-account IAM). The admin_client + its tests are dropped from the diff. See `SLACK_QURL_ROLLOUT.md` §"2026-05-12 architectural update".

### What lands

- **`apps/slack/internal/parser.go`** — regex-based tokenizer for `/qurl <verb>` grammar:
  - `$alias` sigil enforcement (any token in resource position not starting with `$` → friendly error)
  - `<#C12345|channel-name>` channel parsing
  - Optional `dm:true` / `reason:"…"` flags on `get` (empty values rejected, flag-key case-folded)
  - Alias charset `^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?$` (no leading or trailing hyphen)
  - Verbs recognized (no handlers yet): `get`, `setalias`, `unsetalias`, `aliases`, `admin claim`, `admin allow`, `admin disallow`, `admin policies`, `admin status`, `admin revoke`, `create`, `list`, `help`
  - Strict-posture surplus-positional rejection: `admin policies extra`, `admin claim CODE`, duplicate channel/alias on `admin allow|disallow`, all → `ErrUnexpectedArgument`
- **`apps/slack/internal/idempotency.go`** — `IdempotencyKey(teamID, channelID, userID, triggerOrViewID)` → 64-hex sha256. Each field is length-prefixed with a 4-byte big-endian uint32 before its bytes so `("ab","c")` vs `("a","bc")` can't collide regardless of what byte shape Slack ships in future ID formats. Shape-agnostic — `view.id` substitutes for `triggerID` in view-submission flows. 64 hex satisfies the qurl-service 32-char floor with margin.
- **`apps/slack/internal/views.go`** — Block Kit JSON for `/qurl help` (full verb listing), the `setalias` rebind confirmation modal with typed JSON `private_metadata`, the admin-claim modal with the `IsRedactedSubmissionBlock(blockID)` registry (consumed by handler logging middleware in PR-3c.3+ per Blocker #3), and the `:warning:` ephemeral error template. Reuses the existing `respTypeEphemeral` / `respFieldResponseType` / `respFieldText` constants from `handler.go`.

### Coverage

`go test -race -coverprofile=… ./apps/slack/...` → **86.4% on apps/slack/internal** (gate is 40%). Coverage climbed after dropping admin_client.

### Dependency

This PR is **draft** because PR-3c.1 nominally depends on PR-3c.0 (Fargate runtime swap). The code in this PR is purely additive helper packages and does **not** require Fargate to compile or test — the dependency is sequencing, not build-time. Mark ready once 3c.0 lands.

### Out of scope (intentionally)

- Wiring these helpers into `handleSlashCommand` — that's PR-3c.3+
- `apps/slack/internal/slackdata/` DDB-direct package — that's PR-3c.3+ (#231 / #233)
- `shared/client/` extensions (`/v1/qurls` path fix, `resource_id`, `Idempotency-Key`) — that's PR-3c.2 (independent)
- Workflow / Dockerfile changes — that's PR-3c.0 (Lambda → Fargate)

## Test plan

- [x] `go test -race -count=1 ./apps/slack/... ./shared/...` passes locally
- [x] `go vet ./apps/slack/... ./shared/...` clean
- [x] `gofmt -l` clean
- [x] `golangci-lint run` clean on net-new files
- [x] Parser tests cover each subcommand happy path + error paths (missing sigil, missing channel, unknown verb, missing target, empty flag value, surplus positional after slot fill, duplicate channel/alias on admin allow|disallow, etc.)
- [x] Idempotency tests cover same-input determinism, field-boundary collisions (`("ab","c")` vs `("a","bc")`), view-id substitution
- [x] View tests cover Block Kit shape, `private_metadata` JSON round-trip with adversarial alias inputs (`=`, `&`, `"`), admin-claim modal redaction-registry contract (negative test against fictional `private_value`/`masked`/`secret`/`is_password` keys), full verb-set coverage in help
- [x] CI green
- [x] /pr-ready review

Refs: #228, #229 (deferred retry follow-up — moot after admin_client drop)


Kevin-layerV:
Pushed fixes addressing every item in the round-1 review (b8b5ca4).

**Blocker #1 — `private_value` is fictional**: Confirmed against the [Slack docs](https://docs.slack.dev/reference/block-kit/block-elements/plain-text-input-element) — `plain_text_input` has no input-masking field. Removed the misleading flag. The Blocker #3 mitigation now lives at the bot's logging boundary: a new `RedactedSubmissionBlockIDs` registry that the PR-3c.3+ handler middleware will consult before logging any `view_submission` payload. The `:lock:` user-visible note now describes the actual guarantee (TLS-only transit, never logged) rather than a UI mask that doesn't exist. Test renamed to `TestAdminClaimModal_StableIDsAndNoFakeMasking` and now actively fails if any of `private_value` / `masked` / `secret` / `is_password` sneak back in.

**Important fixes**:
- `private_metadata` is now JSON via the typed `SetAliasRebindMetadata` struct. Test fences round-trip on alias names containing `=`, `&`, and `\"`.
- `WithAdminUserAgent(version)` plumbs the version half. Test covers explicit / empty-fallback / vcs-revision shapes.
- `do()` returns `ErrEmptyAdminResponse` when `out != nil` and `data` is empty/null. Void endpoints (allow/disallow with `out=nil`) still short-circuit; both branches tested.
- Typed request structs replace `map[string]string` for the three POST bodies.

**Nits**:
- Tokenizer strips balanced outer quotes; covered with two new happy-path cases.
- `requireAlias` validates against `[a-z0-9][a-z0-9-]*`; new `ErrInvalidAlias` sentinel with four error-path cases.
- `parseAdmin` rejects unexpected positional args on claim/policies/status/revoke via new `ErrUnexpectedArgument`; four error-path cases.
- `IdempotencyKey` parameter renamed `triggerID` → `triggerOrViewID`.
- Dropped unused `fieldChannelID` / `fieldResourceID` constants — the typed struct tags are the single source of truth now.

**Deferred (with issue link)**:
- Retry on transient failures filed as #229 (`pr-polish-followup` + `priority: low`). Designing retry semantics alongside the actual call sites in PR-3c.3 avoids special-casing per-method retry policy in this stub.

`make check` passes locally — lint clean, race detector clean. The CI Vulnerability Scan failure is the org-wide toolchain CVE batch (GO-2026-4971, GO-2026-4918), unrelated to this PR.

Kevin-layerV:
Round-2 fixes pushed (fa8a470). Addresses every item from the latest review:

**(1) Envelope contract tightened**: You're right — the raw-fallback was unreachable for valid JSON because adminEnvelope's optional-only fields unmarshal cleanly from any object. Rather than fix the probe (and add a class of false-positive risks), I tightened the contract: all admin endpoints are now guaranteed-enveloped. A misbehaving server that ships raw `{...}` surfaces `ErrEmptyAdminResponse` instead of silently accepting a partial unmarshal. Test renamed `TestAdminClient_RawUnenvelopedResponseFails` to fence the new contract.

**(2) Duplicate positionals rejected**: `parseAdminChannelAlias` now errors with `ErrUnexpectedArgument` on duplicate channels or duplicate aliases. Two new error-path cases.

**(3) `AdminError.Error()` includes `Code`**: New format is `"Title [code] (status): detail"`, with empty-code fallback to the previous shape. Table-driven test covers four combinations.

**(4) `RedactedSubmissionBlockIDs` consumer added**: Replaced the exported map with `IsRedactedSubmissionBlock(blockID)` so the handler middleware in PR-3c.3+ has a stable query surface (and a future refactor of the storage shape can't break callers). The redaction-fence test now exercises the public function. New `TestIsRedactedSubmissionBlock` covers happy/miss/empty paths.

**Smaller polish (all fixed)**:
- `applyFlag` case-folds the key half — `DM:true` and `Reason:"On Call"` work; values stay mixed-case.
- `tokenize` quote-stripping comment now explains why byte-level comparison is safe under multi-byte UTF-8 (0x22 is never a continuation byte).
- `do()` defer now drains residue past the 1 MiB LimitReader cap before close, so HTTP keep-alive can reuse the connection.

`make check` clean (lint + race detector). The Vulnerability Scan failure is the org-wide toolchain CVEs, unrelated.

Kevin-layerV:
Round-3 fixes pushed (32cd637).

Note: the round-3 review fired against `d485d8c` (the small responseTypeEphemeral cleanup), so its \"still open\" items 1–4 were the round-2 findings that fa8a470 had just landed. This commit addresses the *new* finding + the nit:

**New finding — `aliasCharsetPattern` trailing-hyphen gap**: confirmed the old regex `^[a-z0-9][a-z0-9-]*\$` accepted `prod-`, `db-`, `a-`. New regex `^[a-z0-9](?:[a-z0-9-]*[a-z0-9])?\$` requires either single alnum or alnum…alnum. Three new error-path rows + two single-char happy-path rows (`\$a`, `\$1`) so the trailing-alnum optional group doesn't silently regress to require length >= 2.

**Nit — `AdminError.StatusCode` source**: the 2xx-with-envelope-error branch in `do()` was using `env.Error.Status` (body status), while `parseAdminError` (for >=400) used the HTTP status. Both branches now use `resp.StatusCode` consistently with a doc comment explaining the invariant.

`make check` clean.

Kevin-layerV:
## Round-12 response (commit 804a2e9)

Addressed everything from the four 2026-05-18 cr passes (19:39Z / 19:41Z / 19:47Z / 21:01Z) that's actionable in this PR. Items deferred have GitHub issues filed.

### Addressed in-PR

- **looksLikeFlag case-fold** (19:41Z #3, 21:01Z #3): `HTTPS://` and `HTTP://` clipboard pastes now route through the URL-typo carve-out instead of `unknown flag: "HTTPS"`. Tests added.
- **looksLikeFlag <-> flagPattern alignment** (21:01Z #3): doc comment now explicitly says `looksLikeFlag` matches `flagPattern`'s `[a-z][a-z0-9_]*` charset and explains why we accept ASCII upper too (applyFlag lowercases before regex match).
- **uint32 truncation guard** (19:41Z #1, 21:01Z #2): `writeLengthPrefixed` now panics if `len(s) > math.MaxUint32`. Defense-in-depth — unreachable in practice with Slack IDs.
- **Idempotency retry claim softened** (19:47Z #3): doc now says the dedupe surface defends against user double-submit and bot-side retry, not Slack-side retry delivery (Slack doesn't auto-retry slash commands).
- **Two-impl note** (19:47Z #2): doc cross-references `idempotencyKeyForCreate` in process.go and notes the planned migration when PR-3c.3+'s alias-aware dispatcher replaces `processCreate`.
- **escapeMrkdwnCode NewReplacer** (19:41Z #7): chained `strings.ReplaceAll` (4 passes) -> `mrkdwnCodeEscaper.Replace` (1 pass, package-level table built once at init).
- **escapeMrkdwnCode aliasName defense-in-depth** (19:47Z #5): doc clarifies the alias case is a no-op for parser-validated input but kept applied so a future fabricated-Command path can't bypass the code-span fence.
- **parseAdminChannelAlias three-guard doc** (19:41Z #5): function doc + per-branch comments now spell out why the apparently-overlapping guards are non-overlapping by case (third+ token / second-token-overfill / unrecognized-token).
- **Uppercase admin verb coverage** (21:01Z #6): `admin CLAIM` / `admin Policies` rows added so `strings.ToLower(verb)` can't silently regress.

Round-11 already covered the **alias length cap** (19:47Z #1, commit 2f0bfc8) — 64-char boundary mirrors `handler_alias.go::aliasMaxLen` with a length-specific error message.

### Deferred (issues filed)

- **Redaction enforcement (medium, 19:39Z #3, 19:41Z #6, 19:47Z item, 21:01Z #1)** -> #432. Typed `LogViewSubmission` wrapper lands with PR-3c.3+ per the round-12 reviewer's explicit guidance ("acceptable to leave for the consumer PR"). Grep-able TODO marker stays in `handler.go::ServeHTTP` as the forward-reference.
- **reason: / target field length cap (low, 21:01Z #5)** -> #433. Needs confirmation that qurl-service is the authoritative validator with downstream caps.

### Won't fix (with rationale)

- **Help body code span literal quotes (nit, 21:01Z #4)**: Slack mrkdwn treats backtick content as literal, so `reason:"audit text"` renders fine. Substring presence is the load-bearing test contract.
- **Scheme-aware looksLikeFlag for ssh:// / s3:// / mailto: (nit, 19:39Z #4, 19:47Z #4)**: acknowledged in the existing doc; non-HTTP schemes are unreachable on `get` today (no positional after `$alias`). Revisit if PR-3c.3+'s validator accepts non-HTTP.
- **Naming nit on redactedSubmissionBlockIDs (19:39Z #7)**: the in-code comment correctly explains why the map shape is private (a future `sync.Map` / regex set must not break the `IsRedactedSubmissionBlock` API). Design is intentional.
- **Dead code window until PR-3c.3+ (21:01Z #7)**: acknowledged in PR description; this is the intentional staging shape of the additive-helper PR.

CI is green; pre-commit clean; coverage 88.1% on `apps/slack/internal` (up from 86.4% on the new uppercase admin / HTTPS-case rows).

Kevin-layerV:
## Round-13 response (commit 4df1c40)

Addressed the substantive UX gap (`dm:` silent-falsey) and the polish items. The defense-in-depth concern (redaction temporal gap) remains tracked in #432.

### Addressed in-PR

- **`dm:` silent-falsey rejected (round-13 #2)**: `applyFlag` now strict-boolean on `dm` — `dm:1` / `dm:yes` / `dm:please` reject with `ErrInvalidFlag` + "use dm:true or omit the flag". `dm:false` is explicitly accepted (Command.DM() still case-equals against "true" so it returns false — same runtime effect as omitting, but the user can opt out without seeing an error). Pinned each rejected value as a test row so a future loosening (accepting `1`/`yes`/`on`) requires touching the test deliberately.
- **`flagKeyShape` regex dedup (round-13 #3)**: extracted `flagKeyCharset` as a string const, used by both `flagPattern` (post-case-fold key+value) and the new `flagKeyShape` (case-insensitive key-only gate in `looksLikeFlag`). Single source of truth — a future expansion of the key charset lands in one place.
- **`looksLikeFlag` ToLower allocation removed (round-13 #4)**: `strings.ToLower(tok)` -> `hasASCIIPrefixFold(tok, "https?://")`. No more 2KB allocation on a 2KB URL token; the helper also makes the intent more obvious.
- **`looksLikeFlag` empty-string comment + test (round-13 nit)**: doc comment now explicitly notes the empty-input path is handled via the `colonIdx <= 0` guard; `TestLooksLikeFlag_EmptyString` pins it so a refactor can't silently regress.

### Deferred (already tracked)

- **Redaction temporal gap (round-13 #1)** -> #432 (filed in round-12 response). Reviewer's own guidance: "Not a blocker for this PR — the modal isn't wired to anything yet — but worth tracking the gap explicitly." Issue covers the typed `LogViewSubmission` wrapper proposed in the review.

### Won't fix (with rationale)

- **Unclosed-quote in flag value better error (round-13 #5)**: edge case ("not a blocker"); the existing "expected key:value" is technically correct and the tokenize-tolerates-unbalanced-quote contract is documented. A targeted "unclosed quote in flag value" message would require either parser-side quote-counting per flag token (adds state) or a per-token regex pre-check; both add complexity for a low-frequency UX miss.
- **`Flags{}` init thrice in Parse (round-13 nit)**: reviewer said "the explicit init is also fine." Extracting `newCommand(text, sub)` would save 2 lines but add a constructor — net neutral; the explicit form reads more directly at each call site.
- **`callbackIDAdminClaim` rename (round-13 nit)**: reviewer said "not worth a rename if 'redeem' matches qurl-service vocabulary downstream." The constant value lands on the wire (Slack view payload), so renaming would couple PR-3c.1 to a downstream-vocabulary decision; defer to PR-3c.3+ once the qurl-service `/redeem` vs `/claim` surface is finalized.

All four pillars currently green: CI green, all 13 rounds of cr feedback addressed or deferred-with-issue, pre-commit clean, coverage 88.4%.

PR #231, opened 2026-05-10
feat(slack): scope-cut admin surface to claim/revoke/add/remove/list (v1 beta)

### LINKED ISSUE #355: slackdata: fold AllowResource probe into single conditional UpdateItem (race UX)
## Scope

`apps/slack/internal/slackdata/policies.go::AllowResource` is structured as probe-GetItem-then-UpdateItem. The probe checks whether the resource is already in the `allowed_resource_ids` SS; if not, it issues `UpdateItem` with `ADD allowed_resource_ids :rid`.

## Bug

Two admins racing the same `/qurl admin allow #channel $alias` slot can both pass the probe (resource not yet in the set) and both succeed the UpdateItem (`ADD` on a set is idempotent in DDB). Both get the `Allowed` success reply when one should have gotten the `already allowed — nothing to do` idempotent message.

The **data outcome is correct** (one entry in the SS); the **UX signal is wrong** (both operators believe they made the change).

## Repro

Concurrent (within ~10ms) `/qurl admin allow #channel $alias` from two admins. Both reply `Allowed \`$alias\` in <#channel>.` — exactly one should have replied `\`$alias\` is already allowed in <#channel> — nothing to do.`

## Suggested fix

Fold the probe and write into a single conditional UpdateItem:

```go
UpdateItemInput{
    UpdateExpression: \"ADD allowed_resource_ids :rid SET updated_at = :now\",
    ConditionExpression: \"attribute_not_exists(allowed_resource_ids) OR NOT contains(allowed_resource_ids, :rid)\",
    // ...
}
```

`ConditionalCheckFailedException` then maps to the 409-idempotent path. The handler's `errors.As(&slackdata.Error)` branch keeps its current shape.

## Acceptance

- [ ] Single conditional UpdateItem replaces the probe + write.
- [ ] Two concurrent `allow` calls produce exactly one `Allowed` reply and one `already allowed` reply.
- [ ] Existing handler tests (`TestHandleAdminAllow_HappyPath`, `TestHandleAdminAllow_AlreadyAllowed`) still pass.

## References

- Surfaced in PR #231 round-1 Claude review (item 4): https://github.com/layervai/qurl-integrations/pull/231#issuecomment-4456060257
- Shared with #233; whichever PR lands first carries `slackdata`. Fix here lands in the second-merging PR's rebase.

### LINKED ISSUE #356: slackdata: drop dead var _ = aws.String / fmt.Sprintf import suppressions
## Scope

`apps/slack/internal/slackdata/store.go:379` carries `var _ = aws.String` and `apps/slack/internal/slackdata/workspace.go:217` carries `var _ = fmt.Sprintf`. These are placeholder retain-the-import lines for symbols neither file uses.

## Issue

If the imports aren't actually called in the file, they should be dropped — the strict golangci-lint config (per CLAUDE.md) would normally catch them as unused. The `var _ = …` silences the linter but leaves a dead import path.

## Fix

Either:
- Drop the imports + the `var _ =` line if the file genuinely doesn't need them.
- Replace the `var _ =` with a `TODO: <reason>` if the import is reserved for a planned follow-up (e.g. KMS string envelope render via aws.String).

## Acceptance

- [ ] `var _ = aws.String` removed from `slackdata/store.go` (and the `aws` import too if no longer needed).
- [ ] `var _ = fmt.Sprintf` removed from `slackdata/workspace.go` (and the `fmt` import too if no longer needed).
- [ ] `go vet` + `golangci-lint run` clean.

## References

- Surfaced in PR #231 round-1 Claude review (item 5): https://github.com/layervai/qurl-integrations/pull/231#issuecomment-4456060257
- Shared with #233.

### LINKED ISSUE #357: slackdata: replace epochSecondsString with strconv.FormatInt
## Scope

`apps/slack/internal/slackdata/bootstrap_codes.go:157-182` defines `epochSecondsString` — 25 lines that custom-render an `int64` in base 10 with a comment saying \"to keep this file import-light we inline a small base-10 render\".

## Issue

`strconv` is already used across the package (e.g. `slackdata/policies.go`'s cursor encoding) and is in the stdlib's hot-path set. The custom implementation handles edge cases correctly but adds reading overhead — future readers will second-guess it.

## Fix

Replace `epochSecondsString(n)` call sites with `strconv.FormatInt(n, 10)` and delete the custom function.

## Acceptance

- [ ] `epochSecondsString` removed.
- [ ] All call sites use `strconv.FormatInt(n, 10)`.
- [ ] Existing bootstrap-code tests pass.

## References

- Surfaced in PR #231 round-1 Claude review (item 6): https://github.com/layervai/qurl-integrations/pull/231#issuecomment-4456060257
- Shared with #233.

### LINKED ISSUE #358: slackdata: cap countPoliciesForTeam pagination at 1000 (admin status budget)
## Scope

`apps/slack/internal/slackdata/workspace.go:119-142::countPoliciesForTeam` paginates through every `channel_policies` row for a workspace via `Query` + cursor. The package preamble acknowledges this is acceptable at slash-command volumes, but there's no upper cap.

## Issue

A workspace with thousands of channel policies would rack up many Query round-trips when `/qurl admin status` is invoked. Each round-trip is ~10-30ms in-account DDB latency; 100+ pages would exceed Slack's 3s slash-command ack window.

## Fix

Cap counting at 1000 and render `1000+` in the admin-status reply:

```go
const policyCountCap = 1000

func (s *Store) countPoliciesForTeam(ctx context.Context, teamID string) (int, error) {
    // ... existing loop ...
    if count >= policyCountCap {
        return policyCountCap, nil // signal cap-hit with sentinel
    }
}
```

Caller renders `1000+` when `count == policyCountCap`.

## Acceptance

- [ ] `countPoliciesForTeam` returns once `count >= 1000`.
- [ ] `handleAdminStatus` renders `1000+` (not `1000`) when the cap fired.
- [ ] Test fences cap-hit behavior with a synthetic large-policy fixture.

## References

- Surfaced in PR #231 round-1 Claude review (item 8): https://github.com/layervai/qurl-integrations/pull/231#issuecomment-4456060257
- Shared with #233.

### LINKED ISSUE #359: slackdata: BindWorkspace silently wipes prior admin set on re-bind
## Scope

`apps/slack/internal/slackdata/workspace.go:155-198::BindWorkspace`.

## Bug

The doc comment says:

> on overwrite the prior admin set is preserved

But the `PutItem` writes `attrAdminSlackUserIDs: SS{[seedAdmin]}` unconditionally. A re-bind that satisfies the `owner_id = :owner` half of the ConditionExpression OR will **wipe every admin except `seedAdmin`** from the workspace.

Given this method drives the bootstrap-code redeem flow (#233), silently demoting admins on re-redemption is strictly worse than the comment claims.

## Repro

1. Bind workspace `T_team` with `seedAdmin=U_alice`.
2. Promote `U_bob` to admin (UpdateItem ADD on `admin_slack_user_ids`).
3. Re-bind workspace via second redeem with `seedAdmin=U_carol` (same owner).
4. `U_bob` is silently demoted; `admin_slack_user_ids = {U_carol}` only.

## Suggested fix

Either:

- **Code-side**: load existing `admin_slack_user_ids` first and merge `{seedAdmin}` into the set before PutItem; or switch to `UpdateItem ADD admin_slack_user_ids :seed` so the set is preserved.
- **Doc-side**: rewrite the comment to say "re-bind resets the admin set to {seedAdmin}; existing admins must be re-added" and add a `slog.Warn` on re-bind that lists the dropped UIDs.

The code-side fix is preferable; re-bind on the same workspace should be additive, not destructive.

## Acceptance

- [ ] Re-bind preserves existing `admin_slack_user_ids` membership.
- [ ] Or doc + slog.Warn make the destructive behavior explicit.
- [ ] Existing slackdata tests pass.
- [ ] Add a test fence: bind → ADD admin → re-bind → assert ADD admin still in set.

## References

- Surfaced in PR #231 round-2 Claude review (item 1).
- Shared with #233 (claim flow invokes BindWorkspace on redeem).

### LINKED ISSUE #360: slackdata: ListPolicies can emit duplicate entries on dual-shape rows
## Scope

`apps/slack/internal/slackdata/policies.go:236-258::ListPolicies` flatten logic.

## Bug

If a row carries both `attrResourceID` (single-resource legacy shape) **and** the same `resourceID` is a member of `attrAllowedResourceIDs` (multi-resource shape), the resource appears **twice** in the flatten.

## Why it happens

The flatten reads both attribute shapes — both branches emit a `PolicyEntry`. The comment says "dual representation is a transition shape," but there's no dedup gate.

## Suggested fix

Either:

- Dedup on flatten — track `seen[resourceID]bool` per (channelID,alias) pair, skip dupes.
- Assert in `AllowResource` that a row can't carry both shapes (if `attrResourceID` is set, migrate to SS; if SS is set, never set `attrResourceID`).

The dedup-on-flatten variant is the safer fix; it tolerates pre-migration rows in the wild.

## Acceptance

- [ ] Flatten dedups by (channelID, alias, resource_id) tuple.
- [ ] Test fence: seed a row with both shapes carrying the same resource_id; assert exactly one entry in the output.

## References

- Surfaced in PR #231 round-2 Claude review (item 6).
- Shared with #233.

### LINKED ISSUE #361: slackdata: AllowResource doesn't write alias scalar — ListPolicies renders empty alias for AllowResource-created rows
## Scope

`apps/slack/internal/slackdata/policies.go::AllowResource` writes only `allowed_resource_ids` (SS) + `updated_at`. The `attrAlias` scalar is never set.

But `ListPolicies` flatten reads `attrAlias` via `readString` for SS entries. So a row created purely via `AllowResource` flattens to entries with empty alias:

```
<#channel> ← `$` (`r_resource_id`)
```

— the operator sees the resource_id but no alias.

## Repro

1. `/qurl admin allow #channel $prod-db` (alias `prod-db` → resource `r_xyz`).
2. `/qurl admin policies` lists `<#channel> ← `$` (\`r_xyz\`)` instead of `<#channel> ← \`$prod-db\` (\`r_xyz\`)`.

## Why existing tests miss it

Test fixtures seed rows directly via `seedPolicySingle` / `seedPolicySet`, both of which DO set `attrAlias`. The round-trip `AllowResource → ListPolicies` path is not exercised.

## Schema gap context

Per Kevin's memory note `project_channel_policies_schema_gap.md`: "channel_policies schema gap — can't carry multiple aliases per channel; needs `aliases` SS attribute or per-alias SK reshape before qurl-integrations #233 ships."

The current row PK is `(team_id, channel_id)`. One row per channel. Storing a single `alias` scalar is structurally inadequate — a channel can allow multiple aliases (`prod-db` + `staging-db`). The right fix is one of:

1. **Per-alias SK reshape** — composite key `(team_id, channel_id)` + SK `alias`. One row per (channel, alias) pair. resource_id becomes a scalar on each row.
2. **`aliases` SS attribute** — keep one row per channel, store the alias set alongside `allowed_resource_ids`. Maps alias → resource_id via parallel SS attributes (clumsy).

Option 1 is more normalized and ListPolicies-friendly.

## Acceptance

- [ ] Schema decision recorded (per-alias SK or aliases SS).
- [ ] AllowResource writes the alias on row creation (and DisallowResource cleans it up when the last resource is removed).
- [ ] Test fence: round-trip `AllowResource → ListPolicies` asserts the alias is rendered.
- [ ] Migration plan for any existing rows in the wild (none expected pre-launch).

## References

- Surfaced in PR #231 round-3 Claude review (item 1).
- Schema gap memory note: `project_channel_policies_schema_gap.md` (2026-05-12).
- Shared with #233 — both PRs read channel_policies via slackdata.

Kevin-layerV:
## Summary

v1 beta admin slash-command surface for the qurl-integrations Slack bot. Scope cut down from the original six-verb design after a model-clarification conversation: channel alias bindings IS the access grant, so `allow`/`disallow` were redundant on top of `setalias`/`unsetalias`. `policies` duplicated `/qurl aliases`, `status` was diagnostic-only, and `revoke-all` was speculative. Beta ships five verbs (one already shipped in #233).

### Verbs

| Verb | Status | What it does |
|---|---|---|
| `/qurl admin claim` | shipped (#233) | First-admin bootstrap via modal + bootstrap code. OAuth installer is auto-admin via `BindWorkspace`, so this is effectively dead path for the typical flow, but stays available for other Slack workspace admins. |
| `/qurl admin revoke <qurl_id>` | this PR | Kill a single minted qURL. Friendly 404 → "already revoked or typo'd". 401/403 → "re-run `/qurl setup` to rotate". |
| `/qurl admin add @user` | this PR | Promote a Slack user to bot admin. Conditional `ADD admin_slack_user_ids :uid`. 409-idempotent on already-admin; 404 on workspace-not-bound; explicit self-add reply. |
| `/qurl admin remove @user` | this PR | Demote a Slack user. Refuses owner-remove (slackdata 400 `cannot_remove_owner`) and self-remove (handler-layer guard). |
| `/qurl admin list` | this PR | Show owner + sorted admin list. Audit log records `admin_set_size_raw` + `non_owner_admin_count`. |

### Concurrency

`AddAdmin` uses a single conditional `UpdateItem` with `attribute_exists(slack_team_id) AND NOT contains(admin_slack_user_ids, :uid)`; CCFE on either leg triggers a follow-up GetItem to disambiguate 409 (already admin) from 404 (workspace not bound). `RemoveAdmin` mirrors with `contains(admin_slack_user_ids, :uid) AND attribute_exists(slack_team_id)` for the `contains` half. `TestAddAdmin_Concurrent` / `TestRemoveAdmin_Concurrent` race 8 goroutines and assert exactly one 200 + seven 409s / 404s respectively.

### Parser tightening

- `qurlIDPattern`: `^q_[A-Z0-9]{16,64}$` — rejects truncated pastes (`q_abc`) and oversize pastes at parse time. Real qurl-service IDs are 26-char ULIDs.
- `userMentionPattern`: `^<@([UW][A-Z0-9]{8,63})(?:\|[^>]*)?>$` — requires Slack's documented `U…`/`W…` prefix + 8+ chars. Toy IDs like `<@A>` reject before AddAdmin can persist.

### Files

- `apps/slack/internal/parser.go` — Replaced channel-arg parsing with `<@U…>` mention parsing. New constants: `AdminAdd`, `AdminRemove`, `AdminList`, `userMentionPattern`, `qurlIDPattern`, `Command.UserID`, `ErrMissingUserMention`, `ErrInvalidUserMention`. Dropped: `AdminAllow`, `AdminDisallow`, `AdminPolicies`, `AdminStatus`, `AdminRevokeAll`, `ErrMissingChannel`, `channelRefPattern`, `parseAdminChannelAlias`, `Command.ChannelID`.
- `apps/slack/internal/slackdata/workspace.go` — Added `AddAdmin`, `RemoveAdmin`, `ListAdmins`. Error-code constants centralized: `ErrCodeWorkspaceNotBound`, `ErrCodeAdminAlreadyExists`, `ErrCodeAdminNotFound`, `ErrCodeCannotRemoveOwner`.
- `apps/slack/internal/slackdata/policies.go` — Dropped `AllowResource`, `DisallowResource`, `ListPolicies`. Kept the consumed surfaces: `ResolvePolicy`, `LookupChannelAlias`, `AllowedResourceIDsForChannel`, `GetChannelPolicy`. Documented the legacy-scalar asymmetry between `ResolvePolicy` and `AllowedResourceIDsForChannel`.
- `apps/slack/internal/handler_admin.go` — Rewritten 805 → 380 lines. Kept the dispatcher, `requireAdminSync` (now carries `action` for audit), budget constants, `handleAdminRevoke`. Added `handleAdminAdd`, `handleAdminRemove`, `handleAdminList`. `workspaceUnboundReply` const lifted to dedup three defensive branches.
- `apps/slack/internal/handler.go` — Help message updated to list the slimmer verb set.
- `apps/slack/internal/views.go` — Dropped `HelpResponse` (PR-3c.1 placeholder; never wired into a production caller) and `dividerBlock`.
- `apps/slack/internal/handler_admin_test.go` — Rewritten 1,409 → 736 lines. Regression coverage: 6 revoke cases (incl. 401/403 auth-rejected) + 6 add cases (incl. self-add) + 5 remove cases + 5 list cases (incl. storage-corruption render + sort-order fence) + 2 concurrency fences + 3 dispatch-edge cases + 3 retained `TestResolvePolicy_*`.

Note on `shared/client/client.go`: the original (pre-scope-cut) commit on this branch added `ListByResource` / `ListByResourceInput` for the now-removed `revoke-all` verb. The scope-cut commit then removed those additions, so the net diff vs `origin/main` is empty — no production shared/client change ships in this PR.

### Trust & gating

- Every admin verb gates on `slackdata.Store.CheckAdmin` first; non-admins get a uniform `:warning: this command is admin-only` reply without reaching the mutation. The denial slog.Warn carries `action` so dashboards can filter "probed admin revoke" vs "probed admin list".
- Every admin slog line carries `user_id` for audit attribution.
- User-visible replies stay generic (`upstream error; see logs`); detailed causes live in CloudWatch.
- `Config.AdminStore` is optional — sandbox deployments without the three `QURL_*_TABLE` env vars reply "Admin features are not configured" rather than panicking.

### Test plan

- [x] `go test ./apps/slack/... ./shared/...`: PASS
- [x] `go test ./apps/slack/internal/ -race`: PASS
- [x] `go vet ./...`: clean
- [x] `golangci-lint run --new-from-rev=origin/main ./apps/slack/...`: 0 issues
- [x] `pre-commit run --files <changed>`: PASS

### Closed by this PR

- Closes #355 — TOCTOU collapse on `AllowResource`/`DisallowResource` no longer relevant (methods removed). The single-conditional-`UpdateItem` pattern lives on in `AddAdmin`, fenced by `TestAddAdmin_Concurrent`.
- Closes #356 — dead `var _ = aws.String` import-keeper and its anchoring `aws` import removed from `slackdata/store.go`.
- Closes #357 — hand-rolled `epochSecondsString` base-10 loop replaced with `strconv.FormatInt`.
- Closes #358 — `countPoliciesForTeam` page-cap is irrelevant after the scope cut; the function is removed in this PR (no consumer remains).
- Closes #359 — `BindWorkspace` is conditional-on-`attribute_not_exists(slack_team_id)`; the row can't be silently overwritten, so the admin-set wipe scenario is structurally unreachable.
- Closes #360 — `ListPolicies` duplicate entries on dual-shape rows: the function is removed in this PR (no consumer remains).
- Closes #361 — `AllowResource` doesn't write the `alias` scalar: `AllowResource` is removed in this PR (no consumer remains).

### Open follow-ups (not blocking)

- #399 `RedeemBootstrap` consumes the one-time code on `ExternalIdentityBindings.Create` failure — latent today (`ExternalIdentityBindings` is nil until qurl-service ships `/v1/external-identity-bindings`). Stays open.
- #400 `CheckRateLimit` is a permanent stub — production strategy still TBD. Stays open.


Kevin-layerV:
Thanks for the review. Round 2 push (7426f2e) addresses everything in scope and defers the timeout/cursor-leak items to a tracked follow-up:

**Fixed in this round:**
- **#2 page-cap silent truncation**: `handleAdminRevokeAll` now tracks `truncated` and surfaces `Page cap reached at 100 pages — re-run …` in the reply. Audit log line carries the flag.
- **#3 unused `ownerID`**: `requireAdmin` now returns `*events.APIGatewayProxyResponse` (nil on success, populated on denial). Doc updated to describe the actual contract.
- **#4 missing actor in audit logs**: every admin `slog.Error` / `slog.Info` now threads `user_id`, `team_id`, plus `resource_id` / `channel_id` / `qurl_id` / `alias` where relevant. Successes emit `slog.Info` ("admin allow succeeded" / "admin disallow succeeded" / "admin revoke succeeded" / "admin revoke-all completed") so the audit trail isn't failure-only.
- **#5 stale `done` doc**: rewrote the `requireAdmin` comment to describe the nil/non-nil sentinel pattern.
- Minor — missing tests: added `TestHandleAdminRevoke_BackendError`, `TestHandleAdminAllow_AllowResourceFails`, `TestHandleAdminPolicies_BackendError`, `TestHandleAdminRevokeAll_PageCapTruncates`. Coverage moved 80.3% → 81.1%.
- Minor — `HasMore=true` with empty cursor: now logs at `Warn` before treating as end-of-listing.

**Deferred to PR-3c.6 (tracked in #232):**
- **#1 3s slash-command budget on `revoke-all`**: agreed on the diagnosis. The proper fix is ack-then-defer over `response_url`, which is already on the Phase 3c.6 plan. PR-3c.5 ships the page-cap re-run hint as the stopgap so admins aren't surprised when a large fleet partially revokes. Constant doc updated to spell out the limitation.
- Minor — pagination cursor leaks to operator-visible reply: same plan, same PR (Block Kit buttons stash the cursor in `block_id` metadata).

**Skipped:**
- Minor — backend errors leaked verbatim to Slack: kept as-is. The reviewer flagged this as "probably low-risk in practice" and the surface (admin-only commands; admins benefit from concrete error detail like `alias_not_found` vs a generic `:warning:`). If this turns out to leak something sensitive in PR-3c.6 we'll revisit.
- Nit — `handleAdmin` "re-parses text": `handleSlashCommand` only does prefix-matching (`strings.HasPrefix(text, "admin")`); the `Parse()` call inside `handleAdmin` is the only structural parse. Not actually double-work.

Kevin-layerV:
Round 3 push (2fd516f). Addressed all 10 items in this round; nothing deferred.

**Bugs / correctness**
- **#1 dead cursor on cap-hitting iteration**: hoisted the truncation check above `cursor = out.NextCursor`. The cap-hitting iteration now `break`s without the dead assignment.
- **#2 arbitrary `cmd.Target` in revoke**: new `ErrInvalidQURLID` + `qurlIDPattern` (`^q_[A-Za-z0-9]+$`) in the parser. Malformed tokens like `;rm-rf`, `r_resource_id`, `q_abc 123` now surface as a friendly parser error rather than a 404 from the backend.
- **#3 `ctx.Err()` mid-loop**: bailout at the start of every page iteration AND before each inner `Delete`. New `deadline_exceeded` audit field on the completion log so the failure mode is distinguishable from "individual qURLs failed for unrelated reasons".
- **#4 `AdminClaim` stub PR-3c.3 reference**: dropped the version label since the ordering isn't pinned; just says "not yet implemented in this build".

**Robustness / UX**
- **#5 `respondSlack` swallowed error**: doc-commented `respond` and `respondSlack` to spell out that the error return is always nil (`json.Marshal` of a `map[string]string` cannot fail). The `_` discards in `requireAdmin` are now intentional, documented contract.
- **#6 awkward cursor surface**: dropped the raw cursor token from the reply. Now emits a generic `_…more pages available; cursor navigation lands in the next release._` so an operator sees that pagination exists without a dangling token they can't act on.
- **#7 unbounded reply length**: `adminPoliciesReplyByteCap = 3800` watermark with defensive truncation in a `strings.Builder`; truncated replies surface `_…N more not shown (reply size cap); see Block Kit pagination in upcoming release._`. Test asserts a synthesized 50-long-entry workspace stays under Slack's 4000-char ceiling.

**Performance**
- **#8 `adminRevokeAllMaxPages` tightened to 5** (was 100). Practical cap of 500 qURLs/invocation — keeps the truncation hint visible to the operator inside Slack's 3s budget. Constant doc spells out the rationale and pointer to #232 for the proper async-defer fix.

**Test coverage**
- **#9 `has_more=true` + empty cursor**: `TestHandleAdminRevokeAll_HasMoreEmptyCursor` — fences that the loop terminates and reports the partial revoke count.
- **#10 alias resolves with zero qURLs**: `TestHandleAdminRevokeAll_AliasResolvesNoQURLs` — fences `Revoked 0` literal (not "alias not found").
- New: `TestHandleAdminPolicies_PaginationHidesCursor` — load-bearing security fence asserting the raw cursor token NEVER appears in the user-visible reply.
- New: `TestHandleAdminPolicies_TruncatedReplyHint` — fences the 4000-char Slack ceiling.
- New parser rows for the malformed-qurl_id paths.

Coverage: 80.3% (round 1) -> 81.2% (this round). Lint: 0 issues with v2.10.1 (CI's pin).

Kevin-layerV:
Round 4 push (00a070d). All 7 items addressed:

- **#1 header count mismatch**: render rows first, then write `(rendered)` or `(rendered of total)` header — no more "(50)" above 30 bullets.
- **#2 wall-clock budget**: new `adminRevokeAllWallClockBudget = 2s`, checked at top of loop and before each `Delete`. On exhaustion the loop bails and surfaces `Budget elapsed mid-walk — re-run …` in the reply. Distinct audit-log line from `ctx.Err()`. Also tightened `adminRevokeAllPageSize` 100 -> 20 so the worst-case single-page blast radius stays ~1s. New `TestHandleAdminRevokeAll_BudgetExceeded` pins the clock past the budget and asserts the bail.
- **#3 alias-resolution 404 vs upstream-error leak**: new `resolveAliasOrReply` helper centralizes the `errors.As(&apiErr) && StatusCode==404` pattern. 404 -> `alias not found` (Info log); else -> `upstream error; see logs` (Error log carrying the raw err). Used by allow / disallow / revoke-all.
- **#4 `handleAdminRevoke` 404 vs 5xx**: 404 surfaces `not found — already revoked, or check the qurl_id` (no `:warning:`); other errors surface a generic upstream-error message. Existing test updated to fence the new shape.
- **#5 test foot-gun**: new `addCustomerPrefix` helper on `adminTestServers` routes prefixes through the same dispatch map as `addCustomer`. The two tests that previously overwrote `customerServer.Config.Handler` now use it. Exact-match routes still take precedence.
- **#6 `respond` ceremony**: doc-comments cover the contract; the explicit `_` discards in `requireAdmin` are now intentional and documented. (Refactoring `requireAdmin` to a no-error-return shape would touch every internal builder; deferred as a possible future cleanup beyond PR-3c.5's scope.)
- **#7 AdminClient-unconfigured sweep**: `TestHandleAdmin_AdminClientUnconfigured` now table-drives every admin verb. A future verb that forgets the early-out fails this test.

Also extracted `runRevokeAllWalk` so `handleAdminRevokeAll` stays under the gocognit ceiling and the loop is independently testable.

Coverage: 81.2% (race+combined). Lint: 0 issues with v2.10.1.

Kevin-layerV:
Round 5 push (18da015). All review items addressed:

**Bugs / correctness**
- **#1 404 in revoke-all → `failed`**: split into a separate `alreadyGone` counter. 404 mid-DELETE is logged at Info ("qurl already gone (race)") and surfaces as `M already gone (race or expired)` in the reply — distinct from `K failed (see logs)`. New `TestHandleAdminRevokeAll_DeleteRaces404` fences this load-bearing distinction.
- **#2 bare `/qurl admin`**: confirmed via `TestHandleAdmin_BareAdminVerb` that the parser short-circuits this path with `ErrUnknownAdminAction` ("admin requires a verb") so the dispatch switch never runs and there's no nil-deref. The empty-AdminAction default case is unreachable for parse-success cases.
- **#3 truncated + budget compose**: re-run hint now joins both reasons with " and " rather than dropping the second on the first match.

**Inconsistent error-leak posture**
- Adopted the conservative `handleAdminRevoke` shape across `requireAdmin`, `handleAdminAllow`, `handleAdminDisallow`, `handleAdminPolicies`, `handleAdminStatus`, and `runRevokeAllWalk`'s fatal-err path. Generic "(upstream error; see logs)" message in the user-visible reply; the raw `err.Error()` lives in the structured `slog.Error` only. `TestHandleAdmin_RequireAdminCheckErrors` adds the load-bearing assertion that `db down` (the upstream message) does NOT appear in the reply text.

**Test gaps**
- `TestHandleAdmin_RequireAdminCheckErrors` covers requireAdmin 5xx + the err-leak negative assertion.
- `TestHandleAdmin_MissingTeamOrUserID` table-drives the empty-payload early-out (missing team_id / user_id / both).
- `TestHandleAdmin_BareAdminVerb` — parser-error path on bare `/qurl admin`.
- `TestHandleAdminClaim_StubReply` — pins the "not yet implemented" language so PR-3c.6 intentionally replaces it.
- `TestPoliciesReplyByteCapHasHeadroom` — unit fence on the byte cap leaving room for header+hint under 4000 chars.
- `TestHandleAdminRevokeAll_AliasResolvesNoQURLs` updated to assert the `status=active` query filter (consistency with the rest of the suite).

**Misc**
- Discord parity refs rewritten as stable search tags (`discord-parity:revoke`, `discord-parity:status-fingerprint`) instead of line numbers — survives Discord refactors.
- Added a parser comment explaining the deliberate `ErrMissingTarget` vs `ErrEmptyResource` split between `revoke` and `revoke-all`.

Coverage: 81.2% → 82.2% (combined). Lint: 0 issues. handler.go split into `handler_admin.go` is deferred — non-blocking per cr.

Kevin-layerV:
Round 6 push (1fb0b27). All 4 issues + nits addressed:

**Issues:**
- **#1 doc-comment glue**: split the comment block; `runRevokeAllWalk` has its own doc again.
- **#2 AdminClaim bypassing requireAdmin**: gate added now. New `TestHandleAdminClaim_NonAdminGated` locks it in so the modal-flow follow-up PR can't ship without the gate. The existing `TestHandleAdminClaim_StubReply` still passes (uses `adminCheckYes`).
- **#3 G705 nolint**: confirmed G705 IS a real gosec rule (XSS via taint analysis); it's a false positive here on `Fprintf` to a `strings.Builder`. Restored the `nolint` with an updated explanation noting the reply text reaches a Slack `text` field, not HTML, so XSS is structurally impossible.
- **#4 ctx.Canceled vs DeadlineExceeded conflation**: new `recordCtxErr` helper classifies the two via `errors.Is(err, context.DeadlineExceeded)`. Result struct carries both `deadlineExceeded` and `canceled`; audit log emits separate warn lines. New `TestRunRevokeAllWalk_DirectCanceledClassification` drives the walker directly with both kinds of context (also matches the "test seam" docstring claim — that nit is now true).

**Nits:**
- **handleAdminAllow 409 graceful**: added. New `runAdminPolicyMutation` helper shares the gate/auth/resolve/mutate pipeline between allow and disallow as data-only entries. New `TestHandleAdminAllow_AlreadyAllowed` fences the load-bearing assertion that 409 doesn't carry `:warning:`. (Side benefit: collapsed the dupl-lint warning between the two handlers.)
- **Reply grammar**: tightened to "hit page limit (5)" and joins multi-reason hints with `;` so the compound case reads naturally.
- **runRevokeAllWalk docstring "test seam"**: now accurate — `TestRunRevokeAllWalk_DirectCanceledClassification` drives it directly.
- **Cold-start budget**: kept at 2s; the comment already documents the cold-start case and points at #232 for the proper async-defer fix. Per-cold-start tuning would require separate cold-vs-warm clock instrumentation.

Coverage: 82.2% → 82.7% (combined). Lint: 0 issues.

Kevin-layerV:
Round 7 push (6bab62b). All 6 review items + nits addressed:

**Bugs / correctness:**
- **adminRevokeAllMaxPages doc-vs-math**: comment now says "caps a single invocation at ~100 qURLs" matching the actual `5 × 20 = 100` math.
- **Budget from startedAt, not walk start**: top of `handleAdmin` now stamps `startedAt := h.now()` and threads it through `handleAdminRevokeAll → runRevokeAllWalk`. The deadline is now `startedAt + 2s`, so sigverify + admin-check + alias-resolve count toward the budget — the 2s reflects total handler wall-clock time inside Slack's 3s window, not just walk time. New `TestRunRevokeAllWalk_BudgetIsFromStartedAt` pins this contract: drives the walker with `startedAt` in the past and asserts the loop never makes a backend call.
- **AdminRevoke regex `^q_[A-Za-z0-9]+$`**: kept conservative for now. The cr's "Crockford base32 ULID" suggestion is the most common shape qurl-service emits, and that's covered. If the service ever ships hyphens/underscores in IDs the failure mode is a friendly parser error with the literal token, which is easier to debug than a 404.

**Security:**
- **Asymmetric err-leak posture**: doc-commented at `requireAdmin` to spell out that slog.Error carries the upstream err while the user-visible reply doesn't, and that the same pattern lives in handleAdminAllow / handleAdminDisallow / handleAdminPolicies / handleAdminStatus / runRevokeAllWalk's fatal path / handleAdminRevoke. Pre-empts a future "let me make these consistent" PR.

**Code quality:**
- **Comment density**: did one trim pass on `adminRevokeAllMaxPages` (12 lines → 8). Kept the security-fingerprint and idempotency comments since those carry contracts a future reviewer would need.
- **`requireAdmin` decorator pattern**: deferred. The current shape is verbose but reads linearly; a closure-based decorator hides control flow and would make the dispatch harder to grep. Open to revisiting if more verbs land in 3c.6.
- **handleAdminRevoke vs handleAdminRevokeAll scaffolding**: deferred for the same reason — the two paths diverge at the loop, which is most of the body, so factoring out the gate/auth pair would only collapse 4 lines while making the dispatch indirection worth more.
- **`for i := range 50`** stylistic only — left as-is.

**Test coverage:**
- **`ConfiguredAt.IsZero()` "unknown" fallback**: new `TestHandleAdminStatus_ConfiguredAtUnknown` pins the surface, also asserts the Go zero-value time string `0001-01-01` doesn't leak.

**Brand spelling:**
- "revoke qurl failed" / "revoke-all: qurl already gone" → "qURL" in two log messages. The literal `qurl_id` in attribute names stays per CLAUDE.md.

Coverage: 82.7% → 82.9%. Lint: 0 issues.

Kevin-layerV:
Round 8 push (1694770). All round-7 items addressed:

**Bugs / correctness:**
- **#1 silent truncation on `has_more=true` + empty cursor**: real bug, fixed. New `serverPaginationBug` field on `revokeAllResult`, set when the buggy server response is encountered, surfaced both in the audit log AND in the user-visible re-run hint with the explicit "server pagination bug (more rows reported but no cursor)" attribution. `TestHandleAdminRevokeAll_HasMoreEmptyCursor` now asserts the re-run hint contract — without that fence the broken UX would have stayed silently passing.
- **#2 startedAt budget lockout**: chose option (b) — skip the budget check on the first iteration of the outer loop AND on the first DELETE of the first page. Even if the budget is fully elapsed by the time the walk starts, the operator always gets at least one DELETE of progress per invocation. `TestRunRevokeAllWalk_BudgetIsFromStartedAt` rewritten to fence this contract: with `startedAt = fixedNow - 10s` (8s past the 2s deadline), exactly one list+delete fires.

**Style / minor:**
- **#3 `qurlIDPattern` dashes/underscores**: kept conservative `^q_[A-Za-z0-9]+$`. The risk of rejecting future ID formats is captured by a friendly parser error with the literal token (easier to debug than a 404), and the current Crockford-base32 ULID shape is what qurl-service actually emits today.
- **#4 AdminClient nil vs parser order**: kept current order. The parser-error-first surface tells the operator their syntax is wrong, which is actionable on its own; the "admin features not configured" message is for sandbox/local deploys where the operator likely already knows. Re-ordering would mean a syntactically-valid `admin garbage` produces "not configured" before "unknown admin action: garbage", which is less actionable.
- **#5 `recordCtxErr` and `WithCancelCause`**: noted in the function's docstring already. The current shape is correct for stdlib `context.Canceled`/`DeadlineExceeded`; future cause-errors will naturally fall into the canceled bucket which is the safe default.
- **#6 `cursor = out.NextCursor` after `truncated`**: existing comment explains the dead-code hoist; the new `serverPaginationBug` branch documents itself in the docstring on the result-struct field.

**Performance:**
- **#10 errgroup**: deferred. The 2s budget guard (with the lockout-prevention skip) gives the right semantics for this PR; an errgroup would buy real throughput but at the cost of more nuanced failure-handling (partial errors across goroutines), which is best done as part of the async-defer in #232.

**Test coverage:**
- New `TestHandleAdminRevokeAll_ComposedReply` drives all five result counters (revoked, alreadyGone, failed, truncated, re-run reasons) at once and asserts every count + reason appears in the rendered reply. Catches the cr's specific concern about the truncation hint clobbering the failure count.

Coverage: 82.8% (combined). Lint: 0 issues.

Kevin-layerV:
Round 9 push (0da7636). Round-8 items addressed:

**Things to consider:**
- **#1 audit attribution on revoke-all**: noted, but out of scope for this PR. The PR description's "so quota/audit logs reflect the action" is the deliberate choice; whether qurl-service needs a distinct `actor_type=admin` flag is a service-side question best routed to the audit-pipeline owner, not encoded in the bot.
- **#2 rate limits on sequential DELETE**: addressed. New `rateLimited` field on `revokeAllResult` plus a 429 branch in `deletePage`. Loop bails immediately on 429 (continuing piles on more 429s) and the reply surfaces "rate-limited by upstream (wait then re-run)" — distinct from `failed` so on-call doesn't false-page on a benign rate-bucket race. New `TestHandleAdminRevokeAll_RateLimited` fences this with a 429 mid-walk fixture: (a) loop bails after the 429, (b) reply contains "wait then re-run", (c) reply does NOT contain "failed".
- **#3 lockout-prevention duplication**: factored out into a single `shouldCheckBudget(page, qurlIdx int) bool` helper, used by both call sites. The two are now structurally guaranteed to stay in lockstep.
- **#4 compose-reasons length**: not bounded in this PR. Worst-case "rate-limited by upstream; hit page limit (5); server pagination bug; budget elapsed; request canceled" is ~190 chars on top of the count summary, well under Slack's 4000-char ceiling. Adding a fence is overkill given the structural mutual-exclusion of most combinations.
- **#5 startedAt for every admin verb**: kept defensible dispatch shape. Tiny waste in unused dispatch paths, but consistent shape makes adding more wall-clock-budgeted verbs easy.

**Nits:**
- **`adminCheckYes(t)` / `adminCheckNo(t)` unused param**: dropped. The canned response can't fail (no `t.Fatal` needed), so `t.Helper` was vestigial. All call sites updated via `sed`.
- **listCalls atomic vs int**: kept current shape. The handler is structurally sequential here so the int counter is fine; switching to atomic for visual consistency would imply concurrency that doesn't exist.
- **handler.go ~1500 lines**: noted, deferred to a follow-up split into `handler_admin.go`. Non-blocking.
- **Resource.TargetURL omitempty consistency**: pre-existing field shape, not introduced by this PR.
- **qURL(s) pluralization**: current `Revoked %d qURL(s)` reads fine for both cardinalities; `Revoked 1 qURL` / `Revoked 2 qURLs` would be a cosmetic gain at the cost of an extra branch.

**Test coverage:**
- 5xx CheckAdmin already covered by `TestHandleAdmin_RequireAdminCheckErrors`.
- network-level error path (timeout, DNS): the existing 5xx test exercises the same code path (`CheckAdmin` returns an error → `requireAdmin` returns the denial response). The classification differs at the slog layer (the `err` carries the network detail) but the user-visible reply is identical, so a separate fence wouldn't add much.
- `slog.Info` audit-line shape: out of scope; an audit-pipeline consumer can pin it on its side.

Coverage: 82.8% → 83.1% (combined). Lint: 0 issues.

Kevin-layerV:
## Rework needed — qurl-service `/internal/v1/admin/*` endpoints being removed

Justin's review on qurl-integrations-infra#523 (id 4269670962, 2026-05-12 06:28Z) stated:

> *"No API service. Drop that part of my earlier recommendation. The only consumer of these tables is `qurl-bot-slack`."*

As a consequence, qurl-service#489 (which planned the `/internal/v1/admin/*` + `/internal/v1/policy/resolve` endpoints this PR's `AdminClient` calls) is being closed. Under the new design the Slack bot reads/writes its own tables directly — same posture as `qurl-bot-discord` ↔ `qurl-bot-ddb`.

### Coupling to remove

- `apps/slack/internal/admin_client.go` (476 LOC) — all 8 calls to qurl-service endpoints (`CheckAdmin`, `ResolvePolicy`, `RedeemBootstrap`, `AllowResource`, `DisallowResource`, `ListPolicies`, `GetWorkspaceConfig`, `CheckRateLimit`)
- `apps/slack/cmd/main.go:38-40` — `AdminClient` wireup off `QURL_INTERNAL_SERVICE_TOKEN`
- `handler.go` call sites (lines 287, 371, 395, 413, 448, 516)

### Replacement

Build a bot-side DDB module reading/writing in-account (tables provisioned by qurl-integrations-infra#523's `modules/qurl-slack-ddb/`):

- `QURL_WORKSPACE_MAPPINGS_TABLE` — admin checks, workspace config
- `QURL_CHANNEL_POLICIES_TABLE` — allow/disallow/list/resolve
- `QURL_BOOTSTRAP_CODES_TABLE` — redeem via atomic conditional `UpdateItem`

Rate-limiting either moves to the bot or drops (depending on whether owner-keyed rate limiting needs centralization).

Drop `QURL_INTERNAL_SERVICE_TOKEN`.

This is a substantial reshape, not a tweak — flagging now so further polish rounds on the current pre-pivot shape can stop. The 9 review rounds already absorbed on the HTTP-coupled version aren't wasted (verb signatures + handler logic carry over), but the transport layer rebuilds.

Kevin-layerV:
## /pr-ready blocker — Lambda→net/http transport translation required

Attempted rebase of `feat/slack-admin-policies` onto current `origin/main` (and `origin/feat/slack-parser-idempotency`, which itself sits on `main`).

**Blocker**: This is not a mechanical rebase. PR #228 was rebased today (2026-05-14) onto main with the Fargate/net/http runtime swap from #254. PR #231's pivot rework was drafted against the **pre-rebase** Lambda-shape #228. So #231's `feat/slack-admin-policies` branch carries ~1,013 lines of admin handler logic written for `func (h *Handler) handleAdminX(ctx, ...) (events.APIGatewayProxyResponse, error)`, plus admin test files that use `h.Handle(ctx, &events.APIGatewayProxyRequest{...})` and assert on `resp.Body`. None of this compiles against the new HTTP-server-shaped handler in rebased #228, which uses `func handleX(w http.ResponseWriter, values url.Values)` + `httptest.NewRecorder` test scaffolding.

**What rebased cleanly** (verified locally on /tmp/wt-qurl-integrations-pr231-1778802233 against `origin/feat/slack-parser-idempotency`):
- `apps/slack/internal/parser.go` + `parser_test.go` — 3-way merge of old-228-base / rebased-228 / old-231 trees applied clean (#228's `parseAdminChannelAlias` strictness AND #231's `AdminRevokeAll` / `qurlIDPattern` additions both preserved).
- `apps/slack/internal/views.go` — 1 trivial conflict over admin help text (resolved manually; took #231's three-line form with `<qurl_id>` + `revoke-all $alias`).
- `apps/slack/internal/slackdata/` — pure adds, no conflicts (bootstrap_codes.go, external_identity_bindings.go, policies.go, rate_limit.go, store.go, workspace.go).
- `apps/slack/internal/{admin_test_helpers,fakeddb_fixtures,fakeddb_sha,fakeddb}_test.go` — pure adds, no conflicts.
- `shared/client/client.go` — 3-way merge clean (GetResourceByAlias + ListByResource additions land).

**What requires translation, not merge**:
- `apps/slack/internal/handler.go` — 6 conflict regions, all stem from Lambda→net/http transport change:
  - imports: drop `events`, add `oauth` + `slackdata`.
  - `Config` struct: rebased 228 adds `BaseContext` / `MaxConcurrentAsync` / `ResponseURLClient`; old 231 adds `AdminStore`. Need both.
  - `handleSlashCommand` dispatch: rebased 228 uses `(w http.ResponseWriter, ...)`; old 231's `case strings.HasPrefix(text, "admin"):` returns proxy response.
  - 8 admin handler functions need return-shape translation: `(events.APIGatewayProxyResponse, error)` → write to `w` directly (or hand off to `h.runAsync` for revoke-all's long walk).
  - `requireAdmin` helper returns `*events.APIGatewayProxyResponse` → needs to either write to `w` and signal "done" via bool, or take a `*Handler` callback.
- `apps/slack/internal/handler_admin_{allow,disallow,policies,status,revoke,revoke_all,claim}_test.go` — `invokeSlash` helper signs a request via `events.APIGatewayProxyRequest` and asserts on `resp.Body` / `resp.StatusCode`. Needs `httptest.NewRequest` + `httptest.NewRecorder` + response-recorder assertion pattern (mirror handler_test.go's `newTestHandler` / `newSignedRequest` shape).
- `apps/slack/cmd/main.go` — rebased 228 is 438 lines of Fargate plumbing (oauth router, SIGTERM drain, etc.). Old 231 is 71 lines of Lambda entrypoint. Need to add slackdata.Store wiring (~20 lines around the existing `internal.NewHandler` call) to the 438-line version.

**Async vs sync design decision needed**:
- All non-admin verbs in rebased 228 use `h.runAsync(w, "verb", values, fn)` to ack within Slack's 3s window and follow up via `response_url`.
- Old 231's admin handlers are synchronous (Lambda's invocation deadline is the budget); revoke-all has its own `startedAt` budget tracking.
- **Decision: should admin verbs port to runAsync? Some** (allow/disallow/status — short DDB calls) **probably stay sync.** **Others** (policies — paginated DDB scan; revoke-all — N qURL deletes) **probably want runAsync to ack-then-walk.** This is a runtime-architecture call that touches every admin handler's surface plus the 234+ lines of admin test assertions (which currently fence sync replies).

**Status**: blocked on the design call above. Stopping per /pr-ready blocker rule. The rebase itself is mechanically beyond "re-apply intent" — it's a transport translation of the entire admin handler stack that requires coherent decisions about sync vs runAsync + response_url posture.

**Not silently mutated**: branch remains at `3437ad29` (`origin/feat/slack-admin-policies` head). Local worktree at `/tmp/wt-qurl-integrations-pr231-1778802233` discarded.

cc @Kevin-layerV — recommend either (a) farming the Lambda→net/http translation to a dedicated subagent with the sync-vs-async call made up front, or (b) waiting for #228 to merge to main so #231 can rebase onto a stable target.

Kevin-layerV:
## Hand-craft rewrite landed (relaunch)

The Lambda→net/http translation flagged in [#issuecomment-4455663198](https://github.com/layervai/qurl-integrations/pull/231#issuecomment-4455663198) is now applied. Force-pushed `feat/slack-admin-policies` from a clean rebase off `origin/feat/slack-parser-idempotency` (aa60baa).

**Sync vs async decision (Kevin-authorized):**
- **Sync** — allow, disallow, status, revoke (single DDB or qURL API call, in-account latency comfortably under Slack's 3s budget). Reply written directly to `w`.
- **Async** — policies, revoke-all (paginated walks). Ack with `:hourglass: Working on it…`, deliver via `response_url`. Mirrors create/list pattern.

**Commit shape** (5 commits, +/- ~1.6K LOC):
1. `feat(slack): parser + views additions for admin verbs` — clean port of #231's parser/views deltas.
2. `feat(shared/client): add GetResourceByAlias + ListByResource` — alias→resource_id resolve + resource-scoped qURL list for revoke-all.
3. `feat(slack): add slackdata package for DDB-direct workspace + policy I/O` — Wave 4 pivot facade. **Identical** to #233's slackdata package (verified diff = 0 bytes).
4. `feat(slack): /qurl admin allow/disallow/policies/status/revoke verbs` — new `handler_admin.go`; net/http shape; admin gate, audit log discipline, 404/409 idempotent surfaces.
5. `test(slack): admin verb tests in net/http shape + fake DDB harness` — 20 admin tests against httptest+fakeDDB; brings the suite to **81 PASS** (was 58 baseline pre-admin).

**What's preserved from rounds 1-9 cr feedback:**
- `user_id` on every admin slog line (round 1).
- 4000-char Slack reply byte-cap with truncation hint (round 1).
- qurl_id format check + leading-`$` hint to `revoke-all` (round 2 parser delta).
- 404-graceful `disallow` (round 3 idempotent-surface delta).
- Best-effort revoke-all loop (round 4); 429 bails the walk with "wait then re-run" hint (round 8).
- `serverPaginationBug` audit surface for `has_more=true / next_cursor=""` (round 7).
- Fingerprint as `sha256[:8]` hex with the load-bearing `TestHandleAdminStatus_FingerprintIsNotLast4` regression fence (rounds 1, 6).

**What changed vs the pre-rebase #231:**
- Wall-clock `startedAt` budget is gone — the async ctx already carries `asyncWorkTimeout=25s` from `runAsync`, so a separate budget would double-count. `ctx.Err()` checks cover both timeout and cancel.
- `shouldCheckBudget` lockout-prevention guard (round 8) is gone — same reasoning; runAsync gives the worker its own deadline.
- `revoke` is now sync (single DELETE, in-account latency).
- `revoke-all` async wraps `requireAdmin` + alias-resolve + walk + reply. Sync alias-resolve was racing the 3s ack window with a slow upstream.

**Pre-commit:** `go fmt`, `go mod tidy`, `golangci-lint` all pass on the changed files. `go test ./...`: PASS. `go test -race ./apps/slack/internal/`: PASS.

**Mergeable status: MERGEABLE** (post-push verified).

Kevin-layerV:
## Round-1 review response

Pushed fixes for items 1–3 + nits + test-coverage gap. Below is the disposition for each item.

### Addressed (in this PR)

- **1. `requireAdminSync` unbounded ctx** — fixed with a 2s `adminGateBudget` derived from `h.baseCtx`. Leaves ~1s for the verb body + reply encode inside Slack's 3s window. (`handler_admin.go:97-104`)
- **2. AdminClaim stub gate wrong** — gate dropped. The whole point of claim is to *create* the first admin, so checking admin first would lock every operator out of the bootstrap. Inline comment now flags this for #233 to wire the real bootstrap-code-redeem gate. (`handler_admin.go:67-77`)
- **3. Empty-fingerprint dangling label** — guarded; renders `(not yet plumbed; see follow-up)` when `APIKeyFingerprint == ""`. (`handler_admin.go:262-265`)
- **Nit (LIFO comment)** — rewrote to honestly describe the cleanup order. (`admin_handler_helpers_test.go:111-123`)
- **Test coverage** — added 5 `renderRevokeAllReply` unit tests (happy / rate-limited / truncated+server-pagination-bug / deadline+canceled / partial-failures+already-gone) plus 2 `renderPolicies` tests (empty / has-more). Brings the suite to **88 PASS** (was 81).

### Punted to follow-up issues (slackdata-package shared with #233)

Items 4–6 + 8 all live in `apps/slack/internal/slackdata/` — that package is **byte-identical** between this PR and #233 (verified `git diff origin/feat/slack-admin-policies origin/feat/slack-get-aliases-admin-claim -- apps/slack/internal/slackdata/` returns 0). Fixing them in this PR would create merge conflicts for #233. The right shape is to file follow-up issues so whichever PR merges second picks them up cleanly.

- **4. `AllowResource` probe-then-write race** — UX-only (the data outcome is correct: ADD on a set is idempotent in DDB). Two concurrent admins racing the same allow would both get the success reply; one should get the "already allowed" idempotent message. Fix is folding the probe into a single conditional UpdateItem (`ConditionExpression: "NOT contains(allowed_resource_ids, :rid)"`). Filed as follow-up.
- **5. Dead `var _ = aws.String` / `var _ = fmt.Sprintf` import suppressions** in `slackdata/store.go` + `slackdata/workspace.go`. These were placeholders from the prior agent's commit. Should either be dropped (no caller in slackdata) or replaced with `TODO` comments. Filed as follow-up.
- **6. `epochSecondsString` reinvents `strconv.FormatInt`** in `slackdata/bootstrap_codes.go` — 25 lines of base-10-render that `strconv.FormatInt(n, 10)` already covers. Filed as follow-up.
- **8. `countPoliciesForTeam` unbounded pagination** in `slackdata/workspace.go` — for a workspace with thousands of policies, `/qurl admin status` could exceed 3s. Reasonable cap at 1000 with `1000+` render in the status line. Filed as follow-up.

### Punted with rationale (in-scope but not blocking)

- **7. Help-message drift risk** — `views.go::HelpResponse` (Block Kit renderer) is unused on the current `/qurl help` path. Two surfaces could drift as verbs are added. Fix is either routing `/qurl help` through HelpResponse or deleting the unused renderer. Not blocking this PR; tracked for the Phase 3c.6 Block Kit pagination work where HelpResponse becomes relevant.
- **Nit (12-attr slog.Info on revoke-all completion)** — Acceptable as-is. The 12 attrs are all individually load-bearing for incident response (revoke-all is a fleet-wide kill switch; on-call needs all of revoked/already_gone/failed/rate_limited/truncated/server_pagination_bug/deadline_exceeded/canceled in one log line to diagnose). Splitting into multiple slog calls would scatter the audit trail.

Filing the four follow-up issues now.

Kevin-layerV:
## Round-2 review response

Pushed fixes for the in-scope items + 2 more follow-up issues for slackdata-pkg items shared with #233.

### Addressed (in this PR)

- **2. Sync verbs could blow 3s ack budget** — `runAdminPolicyMutation`, `handleAdminStatus`, `handleAdminRevoke` now use `adminSyncVerbBudget=1.5s` instead of `asyncWorkTimeout=25s`. Paired with `adminGateBudget=1s` (down from 2s), the full chain fits inside Slack's 3s window: 1s gate + 1.5s body + ~500ms encode. Comment cross-references #358 (countPoliciesForTeam cap) as the keep-status-sync-safe dependency.
- **5. No audit log on non-admin denials** — added `slog.Warn(\"admin command denied: non-admin\", team_id, user_id)` on both `requireAdminSync` and `requireAdminAsync` deny paths. Distinct .Warn level from the success-path .Info so dashboards can filter "denied" without scanning every admin command.
- **7. HelpResponse vs helpMessage drift** — dropped `admin claim` from Block Kit `HelpResponse` so the two surfaces match. Comment in views.go now requires the Block Kit and text variants to be kept in sync.
- **Nit (page-size-vs-byte-cap asymmetry)** — documented inline. At ~50-80 bytes/line, the 3800-byte cap effectively trims at ~50-76 entries, so for typical short IDs the page size IS the cap. Comment makes this explicit so future tweaks to either constant don't silently regress the contract.

### Punted to follow-up issues (slackdata-pkg, shared with #233)

- **1. BindWorkspace silently wipes prior admin set on re-bind** — filed as #359. The PutItem writes `admin_slack_user_ids: SS{[seedAdmin]}` unconditionally on re-bind, dropping every prior admin except seedAdmin. Comment claims preservation; code does the opposite. Code-side fix preferred (UpdateItem ADD instead of PutItem).
- **6. ListPolicies duplicate entries on dual-shape rows** — filed as #360. A row carrying both `attrResourceID` and the same id in `attrAllowedResourceIDs` flattens to two PolicyEntry rows. Dedup-on-flatten is the safer fix.
- **3. countPoliciesForTeam unbounded pagination** — filed earlier as #358.
- **4. TOCTOU in AllowResource/DisallowResource** — filed earlier as #355.

### Nits punted

- **`epochSecondsString` vs `strconv.FormatInt`** — filed as #357.
- **Dead `var _ = fmt.Sprintf` import** — filed as #356.
- **`HTTPBindingsClient.Bearer` required but RedeemBootstrap never plumbs one** — already a `TODO` inline; the failure surface only fires if `s.ExternalIdentityBindings` is wired non-nil before the Auth0-bearer plumbing lands. The cmd/main.go leaves it nil today. Acceptable.

### Test count

88 PASS (unchanged from the previous round; the budget changes are internal timing, not new test surfaces).

CI: green. Mergeable.

Kevin-layerV:
## Round-5 + CI fix shipped — PR is MERGEABLE

Pushed round-5 cr fixes (deadline-mid-walk classify, renderPolicies dual-hint, runAdminPolicyMutation arg cleanup) plus a `go vet` fix on the new `TestHandleAdminRevokeAll_DeadlineMidWalkClassified` test.

### Final state

- **CI**: all checks pass (Lint, Test All Apps, TruffleHog, age-checks, claude-review, etc.).
- **Mergeable**: TRUE.
- **Tests**: 90 PASS (was 58 baseline pre-admin), 66.3% total coverage (above 40% threshold).
- **golangci-lint v2.10.1**: 0 issues.

### Loop summary

5 rounds of cr + 1 CI lint fix + 1 CI vet fix:

| Round | Major fixes | Tests |
|-------|-------------|-------|
| Init | parser/views + slackdata + handler + admin verbs + tests (5 commits) | 81 |
| Lint | hugeParam nolint + G706 nolint + fakeddb unused-helpers retain | 81 |
| 1 | requireAdminSync bounded ctx, AdminClaim ungated, fingerprint render | 88 |
| 2 | sync verb budget (1.5s), non-admin deny audit, help-text drift | 88 |
| 3 | admin over-match boundary, store-nil per-verb, parser regressions | 89 |
| 4 | countPoliciesForTeam degrade audit, drop dead var _ in workspace.go | 89 |
| 5 | runRevokeAllWalk deadline-mid-walk classify, renderPolicies dual hint, runAdminPolicyMutation arg cleanup | 90 |
| CI-vet | t.Cleanup-bound deadline cancel for `go vet` | 90 |

### Follow-up issues filed

Slackdata-package items shared with #233 — file these so whichever PR merges second's rebase picks them up cleanly:

- #355: AllowResource probe-then-update race (UX-only; data outcome correct).
- #356: Dead `var _ = aws.String` / `var _ = fmt.Sprintf` import suppressions.
- #357: `epochSecondsString` reinvents `strconv.FormatInt` (+ math.MinInt64 overflow noted in round 5).
- #358: `countPoliciesForTeam` unbounded pagination cap.
- #359: BindWorkspace silently wipes admin set on re-bind (real bug).
- #360: ListPolicies duplicate entries on dual-shape rows.
- #361: AllowResource doesn't write alias scalar → /qurl admin policies renders empty alias for AllowResource-created rows. Root cause: channel_policies schema gap (one-row-per-channel can't carry multiple aliases). Per Kevin's memory note `project_channel_policies_schema_gap.md`.

### What's still TODO inline (not blocking)

- `HTTPBindingsClient.Bearer` empty-string guard on RedeemBootstrap — currently latent (production leaves `ExternalIdentityBindings` nil).
- `CheckRateLimit` stub returns (true, 0, nil) — design open question (in-mem token bucket vs DDB counter on workspace_state).
- HelpResponse (Block Kit) advertises #233-only verbs; helpMessage (rendered today) sticks to what works in this PR's tree. Documented inline.

### Ready to merge once #228 + qurl-integrations-infra #523 land.

Kevin-layerV:
## Round-6/7/8 + merge — TOCTOU collapsed, follow-ups filed, status reply hardened

Three substantive pushes since the round-5 / MERGEABLE state, plus a merge of `feat/slack-parser-idempotency` (rounds 5-7 of #228).

### Round 6 — TOCTOU collapse (the headline)
- `AllowResource` / `DisallowResource` (`apps/slack/internal/slackdata/policies.go`) collapse the probe-then-update into a single conditional `UpdateItem`. ConditionExpression `attribute_not_exists OR NOT contains(...)` for allow, `attribute_exists AND contains(...)` for disallow; `ConditionalCheckFailedException` → 409/404. Also halves the RCU cost per call. Fence: `TestAllowResource_ConcurrentAllowsExactlyOneWins` (+ Disallow counterpart) — 8 racers, asserts exactly one success and exact-N conflicts. fakeDDB grew `contains(...)` / `NOT contains(...)` evaluators for the new condition shapes.
- Cleanup batch: dead `var _ = aws.String` import-keeper dropped (#356 closed); `epochSecondsString` hand-rolled base-10 loop replaced with `strconv.FormatInt` (#357 closed); stale `[AdminClient.RedeemBootstrap]` doc refs in `views.go` rewritten to `[slackdata.Store.RedeemBootstrap]`; `_ = keyID` and `attrKeyID` dropped (#233 can re-introduce when it consumes); six copies of `requireAdminStoreSync` collapsed to a single guard above the switch.

### Round 7 — empty-alias fallback, qurl_id cap, recordCtxErr clarity
- `renderPolicies` renders `_(alias not recorded — see #361)_` when the alias scalar is empty (AllowResource-created rows pre-schema fix). Empty backticks read as broken UI.
- `qurlIDPattern` caps the suffix at `{1,64}` so a fat-paste fails at parse time rather than shipping a kilobyte URL to qurl-service.
- `recordCtxErr` docstring tightened — explicit contract that callers MUST pass `ctx.Err()`.
- Open follow-ups filed: #399 (RedeemBootstrap rollback semantics — latent until qurl-service ships the binding endpoint) and #400 (CheckRateLimit production strategy).

### Round 8 — dedupe, page-cap, stub-warn
- `ListPolicies` dedupes dual-shape rows (legacy `resource_id` + SS) so `/qurl admin policies` doesn't show the same `(channel, resource)` twice. Test: `TestListPolicies_DedupesDualShapeRow`. (Long-term fix tracked at #360 — schema settle.)
- `countPoliciesForTeam` caps the page walk at 5 pages and returns `(count, capped, err)`. The status handler renders `Channel policies: ≥ N (count truncated; see #358)` when the cap fires so operators don't mistake the cap for an authoritative total.
- `CheckRateLimit` emits a one-shot `slog.Warn` the first time any caller invokes the stub — no-enforcement posture is now visible in CloudWatch instead of silently true. Tracked at #400.
- Doc / readability fixes: `HTTPBindingsClient.Create` body-read drain pattern explained; misleading `newAdminTestHandler` 15-line cleanup-ordering comment trimmed to one accurate line; `ddbToError` 412 fallback explicitly named as defensive-for-future-callers.

### Merge with #228 tip
Rebased the rounds-5/6/7 work from #228 (parser regressions, applyFlag split, empty-quoted-target rejection) into this branch via merge commit. Test `alias with internal double hyphen accepted` re-added after the conflict resolution dropped it.

### CI / test state

- 93 test functions in `apps/slack/internal/` (was 81 on round 5); `go test -race` passes.
- `golangci-lint run --new-from-rev=origin/main ./apps/slack/...`: 0 new findings from this PR's changes.
- All required CI checks green on the latest CI run (Lint, Test All Apps, age-checks, TruffleHog, claude-review).

### Closed / Open

**Closed by this PR**: #355 (TOCTOU), #356 (dead import-keeper), #357 (epochSecondsString).

**Open follow-ups** (linked in PR body): #358 countPoliciesForTeam (now bounded; long-term: maintained counter); #359 BindWorkspace re-bind admin set; #360 ListPolicies dedupe at schema layer; #361 AllowResource alias scalar; #399 RedeemBootstrap rollback; #400 CheckRateLimit strategy.

PR is OPEN and ready for review (added @justin-layerv); annotation in the body notes the stacked-PR constraint (will auto-retarget to `main` after #228 merges).

Kevin-layerV:
Post-rebase claude review verdict: *"Nothing blocking from my side — the items above are one follow-up worth filing (#1), one description fixup (#2), and minor polish."*

Polish items consolidated and filed as #461 so they don't get lost across the merge. Marking this ready for review at `0fb3b44`.

**Note on the commit footer / PR description #361 mismatch (claude's item #2):** the body is correct (#361 is an open follow-up); the squash-merge commit message should drop #361 from the `Closes` list at merge time. Easy hand-edit in the GitHub merge UI — no force-push needed.

Kevin-layerV:
**cr round-14 disposition** ([8501cf0](https://github.com/layervai/qurl-integrations/commit/8501cf0))

| # | Item | Disposition |
|---|------|-------------|
| 1 | `requireAdminAsync` lacks per-call budget | **Fixed.** Wrapped in `context.WithTimeout(ctx, adminGateBudget)` to match `requireAdminSync`. |
| 2 | Commit footer claims `Closes #361`, body says open follow-up | **Reconciled in PR body.** Repo squash-merges with `squash_merge_commit_message = PR_BODY`, so the merged commit message comes from the description (which does not close #361). Stale footer is inert on merge — no force-push history rewrite needed. Body now explicitly calls this out. |
| 3 | `AllowResource` condition vs legacy single-scalar row shape | **Commented.** Added a paragraph on the conditional explaining that legacy rows pass `attribute_not_exists(allowed_resource_ids)`, and the resulting SS-write is intended migration toward post-pivot shape (ResolvePolicy reads both, so dual-write is correctness-safe). |
| 4 | Sequential DELETEs in `runRevokeAllWalk` | No change — reviewer flagged as nit ("Fine for the 50-active-users target"). Bounded-parallel is the right lever if `budget elapsed` ever starts firing on healthy upstreams; filing now would just be speculative. |
| 5 | `renderPolicies` Grow hint 64 → 80 | **Fixed.** 4000B sizes to spec with margin; 3200B forced one re-grow at the 3500B byte-cap. |
| 6 | `{rateLimited, truncated}` co-set renderer test | **Added.** `TestRenderRevokeAllReply_RateLimitedAndTruncatedCoSet` locks the joined-reasons separator and rate-limited-first ordering. |
| 7 | `case AdminClaim` in dispatcher vs `//exhaustive:ignore` | No change — reviewer noted current shape is "fine too". Keeping the explicit case keeps the readable surface honest about which verbs route through the dedicated `handleAdminClaim` path.  |

Tests: `go test ./apps/slack/... ./shared/...` green; `-race` green; `golangci-lint --new-from-rev=origin/main`: 0 issues; `pre-commit`: pass.

Kevin-layerV:
**cr round-15 disposition** ([0778bda](https://github.com/layervai/qurl-integrations/commit/0778bda))

| # | Item | Disposition |
|---|------|-------------|
| 1 | `deletePage` ctx-err reclassify after `c.Delete` | **Fixed.** Symmetric `ctx.Err()` check after the non-nil Delete error so a deadline-mid-walk lands as `r.deadlineExceeded`, not `r.failed++`. |
| 2 | Defensive `cmd.Subcommand != SubcmdAdmin` guard | **Dropped.** Replaced with a comment explaining why the guard isn't needed (entry point only reached from `text=="admin"` / `HasPrefix("admin ")`, parser dispatch produces `SubcmdAdmin` for that class). |
| 3 | `handleAdminRevokeAll` inlines alias-resolve | No change — reviewer marked optional. Response-sink asymmetry (`post` vs `respondSlack`) would force a mirror helper to carry both shapes; not worth the indirection at one call site. |
| 4 | `adminPolicyMutation.call` redundant `*Store` param | **Collapsed.** Closure captures `h.cfg.AdminStore` directly. |
| 5 | `qurlIDPattern` brittleness | **Marked.** Added `TODO(upstream-rebrand)` noting the qurl-service ID-grammar coupling. |
| 6 | Comment density (general observation) | No action — reviewer flagged "not worth a round-trip". |

Tests: `go test ./apps/slack/... ./shared/...` green; `-race` green; `golangci-lint --new-from-rev=origin/main`: 0 issues; `pre-commit`: pass.

Issue #123, opened 2026-04-24
Canonical edge function templates: Cloudflare Worker, Netlify Edge, Vercel Edge


joeollis:
## Problem
Every customer hand-writes the same ~30-line edge/serverless function to wrap `ensureResource` → `mint_link` (see Everhaven's `netlify/functions/create-qurl.js`). There's no canonical template, so customers copy-paste from blog posts or each other's repos and get subtly different results.

## Expected
Canonical templates in `qurl-integrations`, one per major edge platform:
- `templates/cloudflare-worker/`
- `templates/netlify-edge/`
- `templates/vercel-edge/`

Each template:
- Is a self-contained function that mints a QURL for a configured target
- Uses the qurl-typescript SDK's `hideUrl()` helper (layervai/qurl-typescript#54)
- Includes deploy instructions
- Is referenced from the `layerv protect` CLI (#122) as scaffoldable

## Why it matters
New customers get to a working integration in ~5 minutes instead of ~1 hour. `layerv protect` can generate them on demand. AIs can drop them into customer repos without rediscovering the two-step model.

## Related
Epic: layervai/qurl-service#370
Pairs with: #122 (CLI), layervai/qurl-typescript#54 (helper)

PR #233, opened 2026-05-10
feat(slack): qurl get + aliases + admin claim with async response_url

Kevin-layerV:
## Summary

Hand-craft rewrite of PR #233 on top of #228's net/http handler runtime. Implements `/qurl get $alias`, `/qurl aliases`, and `/qurl admin claim` (with modal `view_submission` redeem path), using the DDB-direct `slackdata.Store` for policy + rate-limit + bootstrap-code reads (Justin's 2026-05-12 pivot on qurl-integrations-infra#523).

Rebased onto current `main` 2026-05-18 after #228 + #431 landed.

### Three new slash-command verbs

- **`/qurl get $alias [dm:true] [reason:"…"]`** — resolves alias via `GET /v1/resources/by-alias/:alias`, checks channel policy + rate-limit gates against `slackdata.Store`, mints a qURL via `Client.Create` with a deterministic `Idempotency-Key = sha256(<4-byte big-endian length-prefix> || team || <len-prefix> || channel || <len-prefix> || user || <len-prefix> || trigger_id)`. `dm:true` routes the link through `Config.PostDM` (falls back gracefully to channel ephemeral when not wired); `reason:"…"` flows through to the audit row via `CreateInput.Reason`.
- **`/qurl aliases`** — channel-scoped alias listing via single `GetItem` on `channel_policies` (PK=team, SK=channel). Bounded parallel `GetResourceByAlias` fan-out (8 workers) so a slow customer API can't blow through `asyncWorkTimeout`. Cancellation-aware semaphore acquire — a ctx-cancel during dispatch fills the un-dispatched tail with id-only fallbacks instead of deadlocking the worker pool. Empty `channel_id` is refused (no team-wide listing surface in v1).
- **`/qurl admin claim`** — opens the bootstrap-code modal synchronously via `Config.OpenView` (async-defer can't be used here — Slack rotates `trigger_id` after one use). The dispatcher routes only the bare verb; `admin claim <args>` surfaces a "use the modal" hint without echoing the tail (defense-in-depth against operators logging the bootstrap code by accident). `view_submission` calls `slackdata.Store.RedeemBootstrap` (atomic single-use flip on the `bootstrap_codes` table), then `BindWorkspace` to persist the workspace_mappings row, then DMs success.

### Cross-cutting safety + posture

| Concern | How it's covered |
|---|---|
| Panic recovery in async goroutine | `runAsync` defer-recover with stack (inherited from #228) |
| N+1 sequential resource fetches in /qurl aliases | `fanoutAliasRows` with `aliasesResourceFanoutLimit=8` + cancellation-aware semaphore |
| Internal error text leaks on /qurl aliases | `serviceUnreachableMessage` mapping for all non-success `GetChannelPolicy` paths |
| Bootstrap code in slash-command audit log | `redactSlashCommandText` masks the `admin claim <tail>` form-field before slog; dispatcher refuses to route `admin claim <args>` to the modal handler |
| `IsRedactedSubmissionBlock` on inbound log path | `interactionPayload.LogValue` consults `IsRedactedSubmissionBlock`; verified by `TestInteractionPayload_LogValueRedactsClaimCode` |
| PostDM error silently dropped on claim error | `surfaceClaimError` logs DM failure with structured slog + falls back to field-level modal error |
| 404 on bootstrap_code over-maps to "invalid" | `surfaceClaimError` 404 → ERROR slog (deployment misroute), NOT "invalid code"; verified by `logUnmappedClaimError` test |
| Contradictory DM + modal-open signals | Single-signal contract: DM and close modal, OR keep modal open with field error — never both |
| Bind failure after bootstrap-code burn | `surfaceBindError` distinguishes "code redeemed but bind failed" (support-actionable) from "code invalid" (retry-friendly); fenced by `TestHandleAdminClaimSubmit_BindFailsAfterRedeem` |
| BindWorkspace overwriting admin_slack_user_ids | `attribute_not_exists(slack_team_id)` only — same-owner re-bind surfaces 409 instead of clobbering the existing admin set |
| AllowResource TOCTOU on duplicate add | Probe-then-write replaced with conditional `UpdateItem` (`NOT contains(allowed_resource_ids, :rid)`); two concurrent admins on the same allow now see one 409 |
| `fanoutAliasRows` ctx-cancel vs upstream 404 | Distinct Debug log on `errors.Is(rerr, context.Canceled)`/`DeadlineExceeded` so operators can triage slow aliases vs cancellation |
| 5xx → `serviceUnreachableMessage` on mint | `mapMintError` covers 502/503/504 + wildcard 5xx + bare network err; unmapped non-5xx (401/422/etc.) gets the permanent-class generic copy |
| 429 on alias resolution | `mapAliasResolutionError` mirrors `mapMintError`'s 429 handling so the customer API rate-limit on resolution gets a retry-friendly user signal |
| `epochSecondsString` hand-rolled int → string | Replaced with `strconv.FormatInt` (drops the latent `math.MinInt64` overflow) |
| `ExternalIdentityBindings` foot-gun | Step-2 HTTP call deferred entirely — the field stays nil in `cmd/main.go` until qurl-service #547 ships AND the Auth0 Bearer is plumbed; calling Create with an empty Bearer would burn the bootstrap code and fail at the HTTP layer |
| Dead `var _ = aws.String` re-export | Removed; same package's other files import `aws` directly |
| Stub `CheckRateLimit` invisibility | Once-only `slog.Warn` on first call so the open-gate posture surfaces in CloudWatch instead of only in code review |
| `helpMessage` advertising un-wired verbs | Rendered dynamically against `cfg.PostDM` / `cfg.OpenView` — workspaces without bot-token plumbing don't see verbs that would reply with ":warning: not configured" |
| `mutually_exclusive_fields` fence | `CreateInput.TargetURL` is `omitempty`; alias-only mints emit `{"resource_id":"r_..."}` without stray `target_url:""` |

## Test plan

- [x] `go test -race ./apps/slack/...` — clean
- [x] `golangci-lint run --timeout=5m ./...` — 0 issues
- [x] `gofmt -w` clean; `pre-commit run --files …` passes
- [x] `go build ./...` — clean
- [x] Wire-shape: `target_url` omitted on resource-id mint, `reason` lands when set, `Idempotency-Key` header = `sha256(len-prefixed(team || channel || user || trigger_id))` (4-byte big-endian length tags, not NUL-separators)
- [x] LogValue redaction: bootstrap code never serializes through slog (modal path); `redactSlashCommandText` masks tail (slash-command path)
- [x] Cancellation: `fanoutAliasRows` doesn't deadlock on pre-cancelled ctx
- [x] Single-signal contract: claim error path never DMs AND keeps modal open simultaneously
- [x] Persistence fence: workspace_mappings row carries the redeemer on admin_slack_user_ids after `/qurl admin claim`
- [x] Channel scope: sibling channels' alias bindings do not leak into `/qurl aliases` (fenced by `TestHandleAliases_OtherChannelsDoNotLeak`)
- [x] Race detector clean

## Scope — what's NOT here

- **`/qurl admin allow/disallow/policies/status/revoke[-all]`** — owned by #231 / #234
- **`/qurl list` resources surface + `$r_*` fallback in `/qurl get`** — owned by #234
- **`OpenView` / `PostDM` bot-token wiring in `cmd/main.go`** — landed in the OAuth-install follow-up; today the handlers degrade gracefully (`:warning: not configured`) and `helpMessage` hides the verbs that need the wiring
- **`ExternalIdentityBindings` HTTP call (step 2 of redeem)** — deferred until qurl-service #547 ships; today RedeemBootstrap returns the mapping and BindWorkspace persists the workspace row; the qurl-service binding/key-mint comes in the same PR that flips the field non-nil
- **In-bot rate-limit enforcement** — `CheckRateLimit` is a stub with once-only `slog.Warn` so the open-gate posture is visible in CloudWatch; design TBD per `SLACK_QURL_ROLLOUT.md`

## Dependencies

Stacks on current `main` after #228 (parser/idempotency/views) and #431 (AliasStore wiring) landed.

---


Kevin-layerV:
## Rework needed — qurl-service `/internal/v1/admin/*` endpoints being removed

Justin's review on qurl-integrations-infra#523 (id 4269670962, 2026-05-12 06:28Z) stated:

> *"No API service. Drop that part of my earlier recommendation. The only consumer of these tables is `qurl-bot-slack`."*

This PR's admin-claim modal and aliases-list call qurl-service endpoints that are being removed (qurl-service#489 closing for the same reason).

### Coupling to remove

- `apps/slack/internal/admin_client.go` (439 LOC) — same shape as #231's (minus `GetWorkspaceConfig`)
- `apps/slack/internal/handler_admin_claim.go` — calls `RedeemBootstrap` against `/internal/v1/admin/workspace/redeem`
- `apps/slack/internal/handler_aliases.go:86` — calls `ListPolicies` against `/internal/v1/admin/policy/list`
- `apps/slack/internal/handler_test_helpers_test.go:13-16` — the `pathInternal*` test fixtures

### Replacement

- **Aliases-list**: DDB-direct read against `QURL_CHANNEL_POLICIES_TABLE`
- **Admin-claim modal redeem**: DDB conditional `UpdateItem` against `QURL_BOOTSTRAP_CODES_TABLE` (atomic single-use), then POST to a new integration-agnostic qurl-service endpoint `POST /v1/external-identity-bindings` that binds the Auth0 subject ↔ Slack workspace. **That endpoint does not yet exist** — needs a separate design pass on qurl-service.

### Stated dependencies update

Previously listed as `#487 + #489 + #226`. After pivot:
- #489 is **negative** (do NOT depend on it; it's being closed)
- #487 + #226 still apply

This is a substantial reshape. Flagging so further polish rounds on the current pre-pivot shape can stop. The redeem flow's UX (modal, ephemeral confirm) and the aliases-list output structure carry over; the data layer rebuilds.

Kevin-layerV:
## Rebase deferred 2026-05-14 (subagent J)

Another /pr-ready cascade attempted the rebase of this branch onto the new `#228` tip (`aa60baa refactor(slack): drop admin_client (post-pivot)`). **Conclusion: this is a hand-craft rewrite, not a mechanical rebase** — same finding as the prior subagent. Deferring to coordination round.

### Why not mechanical

`#233`'s tip (`add9163`) carries the post-pivot rewire (`52982dc`) on top of a **pre-#254 Lambda-shaped base** (`9e75de2 feat(slack): parser + idempotency helper + admin client + views`). Concretely:

- `apps/slack/internal/handler.go` on `#233` tip imports `github.com/aws/aws-lambda-go/events` and the dispatch loop is Lambda-shaped (`req.Path`, `req.HTTPMethod`).
- `apps/slack/internal/handler.go` on the new `#228` tip is net/http-shaped (`http.ResponseWriter` / `*http.Request`), with the post-`#254` OAuth surface, async pool, response_url validator, and SIGTERM context propagation.
- `#233`'s tree is missing the entire `apps/slack/internal/oauth/` package (12 files) that `#254` introduced and that `#228` inherited via its own rebase.

The pivot rewire (`slackdata` package, dropping `admin_client.go`) is independently correct, but it was applied to Lambda-shaped handlers. A real rebase has to **port each of `handler_get.go`, `handler_aliases.go`, `handler_admin_claim.go`, `handler_async.go`, `handler_deps.go`** to the net/http shape (including async pool wiring, SIGTERM context, response_url POST plumbing) while preserving the DDB-direct intent. That's a ~1k-LOC rewrite, not a 7-commit replay.

A naive `git rebase --onto origin/feat/slack-parser-idempotency 04536a6 feat/slack-get-aliases-admin-claim` conflicts on the first commit (`aa001cd`) on `handler.go` + `Config` shape — confirmed locally in `/Users/kevin/Projects/layervai/wt-qint-pr3c3`. Resolving that one conflict cleanly would mean rewriting the entire handler.go, then continuing through 6 more commits with cascading conflicts.

### Channel_policies schema gap — still open (DO NOT resolve in code)

Per `project_channel_policies_schema_gap.md` and `SLACK_QURL_ROLLOUT.md`:
- `channel_policies` PK=`slack_team_id` + SK=`slack_channel_id` is unique per channel.
- Pre-pivot UX promised "two distinct aliases per channel."
- Option A: add `aliases` SS attribute to the existing item (single TF change, minimal code).
- Option B: per-alias SK reshape (more invasive).

**Lean A. Justin must decide.** The 2 design-fence skips in the test suite remain in place pending this decision.

### Disposition

- Hold this PR DRAFT.
- Do not force-push until: (a) `#228` `/pr-ready` cycle on `feat/slack-parser-idempotency-pr-ready` settles and the agent decides whether to force-push back to `feat/slack-parser-idempotency` (currently `aa60baa`), and (b) Justin signs off on the channel_policies schema option (A vs B).
- When both are clear: do a hand-craft rewrite on top of the then-current `#228` tip — port the slackdata-direct call sites to the net/http handler shape; keep all the round 1–4 review fixes' intent (panic recovery, parallel alias fan-out, redaction LogValue, host pin case-insensitive, etc.) re-applied as in-line edits rather than separate commits.

### Why no force-push from this subagent

The `/pr-ready` `pr-ready/SKILL.md` final-check requires "would a DE accept this?" — a rebase that silently reverts the Lambda→net/http migration would fail that check immediately. Better to escalate than to push a state that destroys `#254`'s work.

cc @kevinaqliuk → external blocker: channel_policies schema decision.

🤖 Subagent J (rebase + /pr-ready cascade)

Kevin-layerV:
## Hand-craft deferred 2026-05-15 (subagent L, relaunched after 529 timeout)

A third subagent picked up the cascade after the second one (`/pr-ready` on `#228`) hit API 529 mid-attempt. Confirmed the prior finding: **this is a hand-craft rewrite, not a mechanical rebase.**

### What the third subagent attempted

1. Spun an isolated worktree at `/Users/kevin/Projects/layervai/wt-qint-pr3c3`.
2. Reset the branch to `origin/feat/slack-parser-idempotency` (the new `#228` tip, `aa60baa`).
3. Started porting the slackdata package over (it's transport-agnostic — would carry across cleanly).
4. Discovered the **cross-agent overlap**: the concurrent `#231` subagent at `/private/tmp/wt-qurl-integrations-pr231-handcraft-relaunch-…` has already published the slackdata package + parser additions + views additions for admin verbs as 3 commits on top of `#228`'s tip:
   - `f1a0846 feat(slack): parser + views additions for admin verbs`
   - `9365963 feat(shared/client): add GetResourceByAlias + ListByResource`
   - `df0c982 feat(slack): add slackdata package for DDB-direct workspace + policy I/O`

### Why the third subagent stopped

Pushing partial work on `#233` in parallel with `#231` would force a re-rebase once `#231` lands — and `#231` is the cleaner source of truth for the slackdata package, since that PR's scope is "admin verbs + DDB-direct policy state" while `#233`'s scope is "get + aliases + admin claim" (consumer of the slackdata package, not its owner).

### Scope after `#231` lands

`#233`'s rewrite shrinks to:
1. **Net/http-shaped handler files**: `handler_get.go`, `handler_aliases.go`, `handler_admin_claim.go`, and a new `handler_interaction.go` for the modal `view_submission` dispatch.
2. **`shared/client` additions**: `CreateInput.ResourceID` + `CreateInput.Reason` (commit `04536a6` from the old #233 history can be cherry-picked or rewritten — both fields are still needed for the `Create(ctx, {ResourceID: ..., Reason: ...})` calls in `handleGet`).
3. **Test rewrites**: `handler_get_test.go`, `handler_aliases_test.go`, `handler_admin_claim_test.go` ported to the net/http shape with `httptest.NewRecorder()` + `h.ServeHTTP(w, r)`, matching the patterns in `#228`'s `handler_test.go` + `process_test.go`. The fakeddb fixtures (`fakeddb_test.go`, `fakeddb_fixtures_test.go`, `admin_test_helpers_test.go`) carry across mostly unchanged.

The 2 design-fence skips (channel_policies schema gap + bootstrap-redeem 404 pending qurl-service endpoint) remain preserved in the rewrite — they are intentional, not regressions.

### Disposition

**DRAFT**, blocked on:
1. `#231` lands on `origin/feat/slack-parser-idempotency` (slackdata + parser + views + GetResourceByAlias arrive via this PR, not duplicated here).
2. Justin's call on channel_policies schema (`aliases` SS attribute vs per-alias SK reshape) — recorded in memory `project_channel_policies_schema_gap.md`.
3. `#228`'s `/pr-ready` cycle stays green (currently `aa60baa`, all checks pass).

When the three conditions are met, the rewrite is straightforward — `#233`'s post-pivot intent (DDB-direct via `slackdata.Store`, async-defer with host pin, redacted LogValue) is already documented across the prior 4 review rounds and codified in the old `handler_get.go`/`handler_aliases.go`/`handler_admin_claim.go` files at `add9163^^` — they just need re-skinning against `#228`'s net/http surface (`runAsync` + `postResponse` + `validateResponseURL` + `Handler.Go`).

**No work was pushed to this branch** by the relaunched subagent — the original tip `add9163` is intact, the 2 design-fence skips are intact, and the PR body has been updated to reflect the cross-agent dependency.

Kevin-layerV:
## Hand-craft rewrite landed 2026-05-15

Branch reset to `origin/feat/slack-parser-idempotency` (#228 tip `aa60baa`), slackdata package added byte-identical to #231 (`feat/slack-admin-policies` at `2c8f6a5`), shared/client gained `Resource`/`GetResourceByAlias`/`CreateInput.ResourceID`/`CreateInput.Reason`, and three new net/http handlers (`handler_get.go`, `handler_aliases.go`, `handler_admin_claim.go`) plus a view_submission router (`handler_interaction.go`).

### Rounds 1-4 cr-feedback intent — preservation map

See the PR body table for every concern → preservation site mapping. TL;DR:

- Panic recovery, fresh-context poster, case-insensitive host pin: inherited from #228's `runAsync`/`postResponse`/`validateResponseURL`.
- Parallel alias fan-out + cancellation-aware semaphore: `handler_aliases.go` `fanoutAliasRows`, fenced by `TestFanoutAliasRows_RespectsCtxCancellation`.
- `interactionPayload.LogValue` redaction: `handler_interaction.go`, fenced by `TestInteractionPayload_LogValueRedactsClaimCode`.
- `surfaceClaimError` single-signal contract (never both DM + modal-open on the same error): `handler_admin_claim.go`, fenced by `TestHandleAdminClaimSubmit_ContradictorySignalsPreventedOnError`.
- 404 → log+generic on bootstrap_code (not "invalid code"): `surfaceClaimError`/`logUnmappedClaimError`.
- `humanizeRetry` sub-second collapse to "a moment": `handler_get.go`, fenced by `TestHumanizeRetry`.
- `mapMintError` 5xx → `serviceUnreachableMessage`: fenced by `TestMapMintError_Unmapped5xx`.
- `Reason` wire-shape test: `TestCreateInputJSON_Reason`.
- `mutually_exclusive_fields` fence (`TargetURL` omitempty): `shared/client/client.go` + `TestCreateInputJSON_ResourceID`.

### Slackdata zero-diff verification

```
$ git diff origin/feat/slack-admin-policies -- apps/slack/internal/slackdata/
(0 lines)
```

When #231 lands first, this PR's rebase resolves the slackdata files as a no-op merge.

### channel_policies design fence — 2 SKIPs preserved

- `TestHandleAliases_PerRowFailureDegradesToIDOnly`
- `TestHandleAliases_MultipleAliasesInOneChannel`

Both skip with messages referencing `project_channel_policies_schema_gap.md` and pointing at Justin's pending A/B decision. The handler-side rendering code is ready for either resolution (`slackdata.Store.ListPolicies`'s flatten path handles multi-resource rows via the `allowed_resource_ids` SS attribute).

### Test results

- 90 PASS, 2 SKIP, 0 FAIL in `apps/slack/internal/`
- 200+ PASS, 2 SKIP across `apps/slack/...` + `shared/...`
- `go test -race` clean
- `golangci-lint v2.10.1` (CI-matched) — 0 issues

cc @kevinaqliuk @JustinL — ready for review when convenient; channel_policies A/B is the one external blocker for the 2 SKIPs.

Kevin-layerV:
Round-14 cr response — every item addressed. Pushing now; this comment fences the deferral linkage for cr-bot.

- **1 (dead branch in surfaceClaimError)** → dropped the unreachable 409 branch; folded the two ErrCodeBootstrapInvalid / 410-Gone tests into a single guard.
- **2 (string-literal duplication)** → extracted `slackdata.ErrCodeWorkspaceAlreadyBoundToCaller` + `slackdata.ErrCodeWorkspaceAlreadyBound`; both producer and consumer now reference the constant.
- **3 (PR description ↔ idempotency-key code drift)** → PR body updated; description now matches the length-prefixed (4-byte big-endian) scheme in `idempotency.go` instead of the NUL-separated phrasing.
- **4 (BindWorkspace TOCTOU window)** → added a comment in `slackdata/workspace.go` pinning the race scope — the binding itself stays correctly held; the race is only on the 409 message variant.
- **5 (Config fields read on the hot path)** → added a doc-comment on `Config` pinning the contract: fields are captured by value into `Handler.cfg` once and must not be mutated after `NewHandler` returns; runtime swaps go through the `SetX` setter pattern.
- **6 (CheckRateLimit open-gate posture)** → tracked in #436 (priority:medium / pr-polish-followup); flag-flip + token-bucket-vs-DDB design captured there. Out of scope for this PR.
- **7 (admin claim has no pre-modal auth gate)** → added a doc-comment in `handleAdminClaim` explaining the design rationale (claim IS how a workspace transitions to first-admin; bootstrap-code knowledge IS the auth) so a future reviewer doesn't add a deadlocking CheckAdmin gate.
- **8 (bind-failure non-409 test gap)** → added `TestHandleAdminClaimSubmit_BindFailsGenericNon409` fencing the "code redeemed but bind failed — contact support" copy with a negative-control that the workspace-conflict copy MUST NOT leak onto this path. Required a new `SetPutItemErr` helper on `fakeDDB` (mirrors `SetUpdateItemErr`).
- **9 (`asyncWorkTimeout` budget on /qurl aliases)** → added `TestFanoutAliasRows_DeadlineDuringFanoutDoesNotLeak` fencing 80 entries × slow upstream × 150ms ctx deadline. Asserts: every entry produces a line, total wallclock stays under a 1s grace window, un-dispatched majority falls back to id-only.
- **10 (style nit `== false`)** → `handler_admin_claim_test.go:366` rewritten to `if !strings.Contains(...)`.

`go test -race ./apps/slack/...` and `golangci-lint run ./apps/slack/...` both clean.

Kevin-layerV:
Round-15 cr response — every item addressed.

- **Bug (empty channel_id on /qurl get)** → added a symmetric guard in `processGet` mirroring `processAliases`. Refuses the channel-less invocation before the upstream `GetResourceByAlias` fetch, gives the user the actionable "This command must be invoked from a channel." copy instead of the generic mint-failed copy. `apps/slack/internal/handler_get.go:111`.
- **Test coverage (slackdata has no _test.go)** → added `apps/slack/internal/slackdata/store_test.go` with a minimal in-package `stubDDB` and tests for the `*Error` shape (`TestError_FormatPreservesAdminErrorShape`, `TestError_ErrorsAsRoundTrip`), the BindWorkspace 409 distinguish-by-caller branches (`TestBindWorkspace_DistinguishesSameCallerFromDifferentAdmin`), the non-CCFE → 503 path (`TestBindWorkspace_TransportErrorMapsTo503`), RedeemBootstrap's CCFE → 410 mapping (`TestRedeemBootstrap_ConditionalCheckFailedMapsToGone`), and validation guards on both. The `*Error.Code` constants the handler branches on are now fenced inside the package.
- **Perf (countPoliciesForTeam unbounded paginated)** → tracked alongside the rate-limit work in #436 ([comment](https://github.com/layervai/qurl-integrations/issues/436#issuecomment-4483550540)). Out of scope for this PR.
- **Perf (fanoutAliasRows Debug logging)** → tracked in #439 (priority:low / pr-polish-followup). On the alias slice → coordination note included for the new owner. Out of scope for this PR.
- **Security (hashBootstrapCode min length tripwire)** → added `minBootstrapPlaintextLen = 10` and a panic in `hashBootstrapCode` if plaintext is shorter. Production codes are ≥16 chars; the tripwire is set conservatively at 10 so the bot fails fast on the regression class the cr flagged (6-digit OTP) without gating against the issuer's own floor. Fenced by `TestHashBootstrapCode_RejectsShortPlaintext` + `TestHashBootstrapCode_AcceptsAtAndAboveFloor`. `apps/slack/internal/slackdata/bootstrap_codes.go`.
- **Security (BindWorkspace 409 disambiguation race)** → comment already added in round-14; no change.
- **Style (:warning: prefix asymmetry on surfaceBindError)** → added a comment explaining the asymmetry is by design and mirrors surfaceClaimError. DMs render `:warning:` to the emoji; modal-error envelopes don't (block-kit needs an explicit accessory icon), so a `:warning: ` prefix on the modal copy would leak as literal text. `apps/slack/internal/handler_admin_claim.go:233-237`.
- **Style (humanizeRetry d==0)** → expanded the doc-comment to explicitly cover `d == 0` (and negative `d` from `time.Until(past)`) collapsing to `humanFallbackMoment`. Locks the "a moment" copy against future "optimize the zero case" refactors. `apps/slack/internal/handler_get.go:341`.

`go test -race ./apps/slack/...` clean (including new `internal/slackdata` package), `golangci-lint run ./apps/slack/...` clean (no new issues vs baseline).

Kevin-layerV:
Round-16 cr response — blocker fixed, follow-ups filed, polish landed.

**Blocker (item 1)**: short-code panic on user-controlled input → fixed.
- `handleAdminClaimSubmit` now gates `len(code) < slackdata.MinBootstrapPlaintextLen` BEFORE calling `RedeemBootstrap`, surfacing the same `copyCodeInvalidOrExpired` retry-friendly copy the 410 path uses. The slackdata entropy-floor tripwire stays as defense-in-depth for the issuer-side regression case but is now strictly unreachable from user input.
- Exported `slackdata.MinBootstrapPlaintextLen` so the handler can read it.
- Fenced by `TestHandleAdminClaimSubmit_ShortCodeRejectedBeforeHashTripwire`. Uses `SetUpdateItemHook` as a fail-on-call to assert RedeemBootstrap is not reached.
- Extracted `copyCodeInvalidOrExpired` constant since the literal is now used at both the length-gate and the surfaceClaimError 410 branch.

**Follow-up (item 2)**: r.Context() vs h.baseCtx in sync slash-command handlers → filed as #440.
- The current architecture deliberately uses `h.baseCtx` everywhere in slash-command dispatch (handler.go:461-465 documents the choice). Threading `r.Context()` through the sync subset specifically gives client-cancel for handlers like `handleAdminClaim` without breaking the async subset — but the dispatcher restructuring needs to land separately. Tracked with acceptance criteria + the cancelled-r.Context() test.

**Follow-up (item 3)**: handler-side panic recovery → landed scoped to the modal dispatch path.
- Defer/recover in `handleInteraction` under `case callbackIDAdminClaim:` (NOT at the function top — the modal-shaped error envelope would leak onto non-view_submission interactions like buttons / select menus). Logs panic + `debug.Stack()` via slog.Error and writes `respondModalError` as best-effort.

**Polish (item 4)**: humanizeRetry minute-rollover → fixed.
- 59.5s ≤ d < 60s now rolls to "1m" via an explicit `if secs >= 60 return "1m"` guard. Fenced by two new table-driven test cases including `-1s` (negative duration from `time.Until(past)`).

**Polish (item 5)**: BindWorkspace disambiguation budget → fixed.
- Added `bindDisambiguationBudget = 300ms` in slackdata; the post-CCFE disambiguation GetItem now runs with a derived `context.WithTimeout`. Worst-case timeout falls through to the generic `workspace_already_bound` variant — bounded-impact failure mode that matches the race posture already documented at the call site.

**Polish (item 6)**: formatAliasLine target URL mrkdwn safety → filed as #441.
- Alias slice → coordination note included for the new alias owner.

`go test -race ./apps/slack/...` clean. `golangci-lint run ./apps/slack/...` no new issues vs baseline.

Kevin-layerV:
Round-17 cr response — issue #1 (real) fixed, polish items 2/3/4 landed, #5 skipped per cr, #6 verified.

**Bug (item 1)**: tab/whitespace bypass on redactSlashCommandText → fixed.
- `redactSlashCommandText` now tokenizes via `strings.Fields` (catches every Unicode whitespace separator, not just literal space). A hand-crafted POST with `admin claim\\tBOOT-CODE` would previously have leaked into CloudWatch via the dispatcher's "default" branch slog.Info site; now redacted. Fenced by two new `TestRedactSlashCommandText` cases (tab between `claim` and the code; mixed-whitespace between the two prefix tokens).

**Polish (item 2)**: alias resolution before admin-store-nil check → fixed.
- `getWork` now refuses the request when `AdminStore == nil` BEFORE `authenticatedClient` lookup AND `GetResourceByAlias` round-trip. Saves one DDB GetItem + one customer-API call on no-DDB sandbox deploys, and closes the mild info-disclosure (probing alias existence ahead of the not-configured signal). `TestHandleGet_AdminStoreNil` now also asserts the customer alias-resolution endpoint is not hit.

**Polish (item 3)**: BindWorkspace disambig default on transport error → fixed with a 3rd code.
- Added `slackdata.ErrCodeWorkspaceBindUnverified` for the case where the disambiguation GetItem itself fails (timeout / transport blip). The 409 routing in `surfaceBindError` is now a switch with three arms: caller-already-bound, bind-unverified (new), and the default different-admin. `TestBindWorkspace_DistinguishesSameCallerFromDifferentAdmin/disambiguating GetItem fails` updated to expect the new code. The honest "Workspace state could not be confirmed. Please retry…" copy replaces the misleading "ask the different admin" copy on this path.

**Polish (item 4)**: bindDisambiguationBudget doc → fixed.
- Doc-comment now explicitly notes `context.WithTimeout` clamps against the parent ctx, so the effective budget is `min(300ms, parent remaining)`. Pairs with the new BindUnverified surface.

**Polish (item 5)**: `buildAdminStore` re-reads env vars → skipped per cr ("Soft-DRY concern; probably not worth a refactor").

**Polish (item 6)**: `nowOrDefault` fallback verified.
- Grep confirms every `*Store` construction path sets `Now` (`NewStore` in cmd/main.go, `newStoreFromFake` in tests, the in-package `newStore` in store_test.go). The fallback is currently unreachable, but the one-line guard is cheap defense-in-depth for a future caller that grabs a `*Store` directly. Tightened the doc-comment to reflect this rather than removing the branch.

**Style notes (comment-trim, loop-label)**: deferred. Existing comments are load-bearing per cr's own "What's good" section; the `loop:` label is on the alias slice → new alias owner.

`go test -race ./apps/slack/...` clean. `golangci-lint run ./apps/slack/...` no new issues.

Kevin-layerV:
Round-18 cr response — every item addressed.

**Bug (item 1)**: case-sensitive redactSlashCommandText → fixed.
- `redactSlashCommandText` now matches the two prefix tokens with `strings.EqualFold` so `Admin Claim BOOT-SECRET` (mobile autocorrect / habit) also triggers redaction at the unknown-subcommand slog.Info site. Three new `TestRedactSlashCommandText` cases fence the leading-cap, trailing-cap, and all-caps forms.

**Bug (item 2)**: RedeemBootstrap returning empty OwnerID → fixed.
- `RedeemBootstrap` now validates `ownerID != ""` after the conditional UpdateItem succeeds. Surfaces a new `slackdata.ErrCodeBootstrapRowMalformed` (`StatusCode: 500`) with a load-bearing `slog.Error` at the slackdata layer (carries team_id + slack_user_id for storage-context grep). `logUnmappedClaimError` adds a paired Error-level case for the same code at the handler layer (carries user_id + the wrapped error for request-context grep). The double-log is intentional per the existing 404-misroute pattern: different on-call owners + different remediation paths. Bootstrap code is still burned in this path; operators must mint a fresh one after fixing the issuer. Fenced by `TestRedeemBootstrap_MalformedRowReturns500`.

**Style (handler_get_test.go bare alias)** → fixed.
- Dropped `var readAll = io.ReadAll` and inlined `io.ReadAll` at the two call sites. `io` was already imported.

**Style (get-case order)** → fixed.
- `case text == "get" || strings.HasPrefix(text, "get "):` — cheap equality first, matches the surrounding cases.

**Nit (fanoutAliasRows non-determinism comment)** → deferred to the new alias owner (alias slice handoff). Not a bug per cr — just a comment improvement; the new owner will pick up alongside the existing handoff items.

`go test -race ./apps/slack/...` clean. `golangci-lint run ./apps/slack/...` no new issues.

Kevin-layerV:
Round-19 cr response — three substantive items addressed; nits triaged.

**Substantive #1**: strongly consistent disambig read → fixed.
- `BindWorkspace` post-CCFE GetItem now sets `ConsistentRead: aws.Bool(true)`. The CCFE confirms a row exists, but an eventually-consistent read on a stale replica could miss it and send a same-caller re-entry to the "different admin" branch — exactly the failure mode `ErrCodeWorkspaceBindUnverified` was added to handle. 2x RCUs on the rare unhappy path; closes the race directly.

**Substantive #2**: `helpMessage` advertises setalias/unsetalias unconditionally → fixed.
- Added `if h.aliasStore != nil` gate around setalias/unsetalias mirroring the PostDM/OpenView gates above. Sandbox deploys without an aliasStore no longer advertise verbs whose only reply is ":warning: not configured". `TestHelpListsNewVerbs` now uses `newAliasTestHandler` (aliasStore wired); new `TestHelpHidesAliasVerbsWhenAliasStoreNil` fences the unwired path.

**Substantive #3**: `nowOrDefault` called 3x per invocation → fixed.
- `RedeemBootstrap` and `BindWorkspace` now capture `now := s.nowOrDefault().UTC()` once and derive nowEpoch / nowISO / CreatedAt uniformly. Production traces will now show a single instant for the conceptually-one event.

**Nits triaged** (all flagged as awareness, not blockers):
- `mapMintError` 403 Detail echo, `fanoutAliasRows` worst-case latency footer, `adminClaimViewOpenBudget` margin, `copyCodeInvalidOrExpired` collapsing 3 failures, `humanizeRetry` multi-hour, `handleInteraction` partial-header recovery — kept as-is. Each is either documented intent, theoretical against today's API surface, or out-of-scope (alias slice → new owner). `fanoutAliasRows` truncation footer fits naturally with the existing `#439` follow-up on alias logging — flagged there for the new alias owner.

`go test -race ./apps/slack/...` clean. `golangci-lint run ./apps/slack/...` no new issues.

Kevin-layerV:
Round-20 cr response — most actionable item fixed; nits + comments landed.

**Substantive (bug)**: `hashBootstrapCode` panic → error.
- Function now returns `(string, *Error{StatusCode: 410, Code: ErrCodeBootstrapInvalid})` on plaintext shorter than `MinBootstrapPlaintextLen`. `RedeemBootstrap` propagates the error through the same 410 / "invalid or expired" surface the conditional-check failure uses, so a missed handler-side length gate at any future call site degrades to a user-visible reject instead of a per-request panic / log-spam vector. Handler-side length gate stays as the UX fast-path (faster reject, no DDB hit).
- `TestHashBootstrapCode_RejectsShortPlaintext` rewritten from `recover()` to `errors.As(&Error)` + StatusCode + Code assertions — fences strictly more than the old panic-message-substring match.

**Substantive (test gap)**: ConsistentRead=true on the disambig GetItem → fenced.
- New `TestBindWorkspace_DistinguishesSameCallerFromDifferentAdmin/disambig GetItem uses ConsistentRead=true` subtest captures the flag through a stubDDB `getItemFn` and asserts it. A refactor that drops the flag now replays the round-19 eventual-consistency race loudly in unit tests.

**Comments landed:**
- handler.go admin-claim dispatch: added a load-bearing comment explaining why this arm is case-sensitive (mixed-case falls through to the redacted slog.Info site by design). Pointer to `redactSlashCommandText` so a future case-folding refactor doesn't accidentally bypass the redaction.
- slackdata/store.go `ddbToError`: CONTRACT comment on the CCFE→412 branch noting every current caller pre-catches CCFE, and any new op MUST do the same — the handler layer doesn't dispatch on 412.
- slackdata/workspace.go: comment on `ErrCodeWorkspaceAlreadyBound` noting it covers both owner-mismatch and admin-set-mismatch (same user signal).
- handler_get.go: `// 4. Surface` → `// 5. Surface` step-label fix.

**Performance follow-up**: `/qurl aliases` read-amplification → tracked in #436 ([new comment](https://github.com/layervai/qurl-integrations/issues/436#issuecomment-4483868038)). The in-bot rate-limit gate should cover `/qurl aliases` when it lands, not just `/qurl get`.

**Skipped per cr's own framing** ("acceptable" / "comment-only" / "narrow risk"):
- `handleInteraction` wroteHeader sentinel wrapper — cr says "current shape is acceptable."
- early-exit on parent-ctx-canceled in surface*Error — defensive against a narrow window with no concrete trigger today.
- `fanoutAliasRows` explicit ctx-aware wait — cr says "today's test catches the regression"; future SDK refactor would surface in the existing test.

`go test -race ./apps/slack/...` clean. `golangci-lint run ./apps/slack/...` no new issues.

Kevin-layerV:
Round-21 cr response — both must-fix items addressed; suggested test already exists; doc tags landed.

**Before-merge #1**: `reason:"…"` slash-command log → redacted at the slog boundary.
- New `redactReasonFlag` invoked from `redactSlashCommandText` for non-admin-claim paths. `/qurl get $alias reason:"incident 123"` now logs as `get $alias reason:"<redacted>"`. Admin-claim redaction wins precedence (bootstrap code is the more sensitive of the two, so the whole tail still collapses to `admin claim <redacted>` and swallows any reason flag inside it).
- Regex matches BOTH parser-accepted shapes: quoted `reason:"…"` AND unquoted `reason:foo` (`parser.flagPattern` accepts both per parser_test.go's `reason:foo` fixture). A quoted-only redactor would have leaked single-word PII / secrets through the unquoted branch.
- Surface-coupling note in the regex doc: any future free-form flag added to `Command.Flags` must extend this pattern.
- Eight new `TestRedactSlashCommandText` cases fence quoted, empty-quoted, multiple-occurrence, mixed-with-other-flags, unquoted-value, and admin-claim-wins precedence.

**Before-merge #2**: `expires_at` Number-vs-String type → verified + fenced inline.
- `modules/qurl-slack-ddb/main.tf:273-285` already specifies the contract (Number / epoch seconds — required by both DDB TTL and the bot's `expires_at > :now` read-time check). A type mismatch would surface as a ValidationException at the conditional UpdateItem, not a silent let-through. Added a reference comment in `RedeemBootstrap` so future grep finds the schema fence in lockstep with the code.

**Suggested test**: `TestHandleGet_ReasonForwarded` → already exists.
- `TestCreateInputJSON_Reason` at handler_get_test.go:405 invokes `/qurl get $prod-db reason:"incident #123"` end-to-end and captures the outbound `POST /v1/qurls` body. The name is misleading (sounds like a marshal test) but the body is an end-to-end fence covering exactly the suggested invariant. No new test needed.

**Code-quality note**: ddbToError CONTRACT tag → landed on function doc.
- Added "CONTRACT: every caller MUST pre-catch CCFE before invoking ddbToError" to the function-level doc so future call sites see the requirement at the doc-hover, not just in the inline comment.

**Skipped per cr framing** ("low-severity" / "cosmetic" / "out of scope"):
- helpMessage dm:true ordering — cosmetic refactor, cr explicitly says "reads cleaner" but no behavior change.
- userError discoverability comment — type already has the invariant in its name; relocating the doc is style preference.
- formatAliasLine mrkdwn safety — alias slice (already filed as #441).
- /qurl get vs /qurl aliases cross-PR consistency — cr explicitly says "Fix lives in #231/#234."
- aliasesResourceFanoutLimit soft cap on alias_bindings size — alias slice (will fit alongside #441/#439).
- countPoliciesForTeam rate-limit — tracked in #436.
- AWS SDK ctx-canceled wrap inspection — cosmetic per cr.
- runbook line on issuer-vs-bot fall-through entropy — operational doc, not in-repo.

`go test -race ./apps/slack/...` clean. `golangci-lint run ./apps/slack/...` no new issues.

PR #235, opened 2026-05-10
feat(discord): self-destruct timer in /qurl send

justin-layerv:
## Summary

Adds an optional self-destruct timer to the button-driven `/qurl send` flow. The timer is constrained to **7 user-curated presets** — 1/2 second, 1 second, 5 seconds, 30 seconds, 5 minutes, 30 minutes, 1 hour. Forwarded to qurl-s3-connector as `viewer_ttl_seconds` (the field accepted by the connector since [qurl-integrations-infra#477](https://github.com/layervai/qurl-integrations-infra/pull/477)) so the rendered viewer page wipes content after N seconds.

```
/qurl send
  → 📁 Send File / 🗺 Send Location
  → recipient · personal note (optional) · 💥 Self-destruct timer (optional) · expiry
  → Send
```

Customer-facing terminology is "Self-destruct timer" — TTL is jargon. Internal var name (`selfDestructSeconds`) matches the UX label; wire-name to the connector stays `viewer_ttl_seconds`.

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## UX

- New form button alongside the personal-note button — one ActionRow with two buttons (Discord allows up to 5/row), saving a row when target=user pushes the form to the 5-row limit.
- Modal label is "Pick a duration (blank = no timer)"; placeholder is the inline option list — `1/2 second, 1 second, 5 seconds, 30 seconds, 5 minutes, 30 minutes, 1 hour`. Discord modals are TextInput-only (no nested dropdown), and the form already saturates the 5-row max when target=user, so a `StringSelectMenu` row wasn't an option without relocating other controls.
- Parser accepts either the friendly label (`"5 minutes"`) OR the raw seconds value (`"300"`), case- and whitespace-insensitive. Anything outside the preset set renders an inline form warning that lists the allowed options, so the user can correct without losing other selections.
- Empty submit clears the timer.
- Button label reflects current state — "💥 Self-destruct timer (optional)" when unset, "💥 Self-destruct: 5 minutes · edit" when set.

## Why preset-only

The connector's contract is `0.5–3600` inclusive, but exposing the full range as free text would let creators ship values like `0.7s` that hit the 500ms-floor edge case where the page renders just long enough to show a flash of content before blanking. The 7 presets are values we'd recommend out loud — every one produces a useful UX.

## Plumbing

- **`SELF_DESTRUCT_PRESETS`** + **`parseSelfDestructSeconds`** (`utils/time.js`) — frozen preset table is the single source of truth for the modal placeholder, the parser's accept-set, and the formatter's label-rendering. Strict membership matching, no clamping.
- **All four connector upload paths** (`uploadToConnector`, `reUploadBuffer`, `downloadAndUpload`, `uploadJsonToConnector`) accept an optional `viewerTtlSeconds` and route through a new `appendViewerTtl` helper that only adds the form field when a positive finite number is passed — defensive against an upstream caller misrouting NaN/0/string values.
- **`handleSend`** threads `selfDestructSeconds` → `downloadAndUpload` (file) / `uploadJsonToConnector` (location) AND the corresponding `reuploadFn` closure, so a `TOKENS_PER_RESOURCE` refill registers the same timer on each new resource.
- **Persistence**: `qurl_send_configs.self_destruct_seconds` (REAL — sub-second `0.5` survives roundtrip). `ALTER TABLE` migration in `SAFE_ALTERS` for existing tables; `CREATE TABLE` for greenfield. DDB store mirrored.
- **`handleAddRecipients`** reads `sendConfig.self_destruct_seconds` and inherits it into both file and location re-upload paths so additional recipients see the same timer.

## Carve-out

Google-Maps JSON resources hit the connector's render-side carve-out ([qurl-integrations-infra#480](https://github.com/layervai/qurl-integrations-infra/issues/480)) — a TTL on a location upload is silently no-op at view time. The plumbing still forwards the value so behavior matches contract once the carve-out is removed; documented inline.

## Tests

989 tests pass, 100% coverage on `utils/time.js`.

- `parseSelfDestructSeconds` (`tests/time.test.js`): every preset accepted by both label and seconds form, case-/whitespace-insensitivity pinned, off-preset numeric values rejected (`'2'`, `'7'`, `'60'`, `'120'`, `'0.7'`, `'3600.5'`), non-preset friendly phrasings rejected (`'2 minutes'`, `'10 seconds'`, `'forever'`), NaN/Inf/hex/oversized rejected, MIN/MAX/OPTIONS_TEXT exports pinned.
- `formatSelfDestructLabel`: every preset round-trips to its label; off-preset values fall back to compact `"Ns"` rendering (defends against backfilled state).
- **All four connector upload helpers** (`tests/connector-coverage.test.js`): `viewer_ttl_seconds` appears on the multipart form when a positive number is passed; omitted for `null` / `undefined` / `0` / `NaN` / `Infinity` / string / object inputs.
- `saveSendConfig` roundtrip (`tests/database-module.test.js`): 0.5 sub-second persists as REAL, 30 integer persists, omitted defaults to NULL on read.
- `handleSend` self-destruct modal (`tests/qurl-send-state-machine.test.js`): valid preset (`30`) threads through to `uploadJsonToConnector`; friendly label (`"5 minutes"` → 300) threads through; invalid value (`'2'`) re-renders form with inline warning + back-half does not run; empty submit clears the timer.
- `handleSend` file path (`tests/qurl-send-state-machine.test.js`): self-destruct preset threads through to `downloadAndUpload` — symmetric with the location-path positive test so a future refactor that drops the arg from one path but not the other gets caught.
- `handleAddRecipients` inheritance (`tests/qurl-send.test.js`): both file and url/maps re-upload paths inherit `self_destruct_seconds` from the original `sendConfig` — pins the "additional recipients see the same timer" contract on the positive path.

## Security / privacy

- TTL is **not a security boundary** — devtools-strip and screenshots already defeat any client-side timer. Same threat model as PR #477's renderer side.
- The bot already passes `one_time_use: true` on every `mintLinks` call, so the Snapchat-style "no refresh to reset" semantic holds out of the box for the bot flow (direct `/api/upload` callers don't get this — see [qurl-integrations-infra#479](https://github.com/layervai/qurl-integrations-infra/issues/479) for the connector-side coupling follow-up).
- No new logged identifiers — the form-state changes don't add any md5/qurl_link references to log lines.

## Testing

- [x] `npx jest` — 989 tests pass
- [x] `npx eslint src/ tests/` — clean

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -19,7 +19,13 @@ const config = require('./config');
 const db = require('./store');
 const logger = require('./logger');
 const { COLORS, TIMEOUTS, RESOURCE_TYPES, DM_STATUS, MAX_FILE_SIZE, MAX_CONCURRENT_MONITORS, AUDIT_EVENTS } = require('./constants');
-const { expiryToISO, expiryToMs, parseSelfDestructSeconds } = require('./utils/time');
+const {
+  expiryToISO,
+  expiryToMs,
+  parseSelfDestructSeconds,
+  formatSelfDestructLabel,
+  SELF_DESTRUCT_OPTIONS_TEXT,
+} = require('./utils/time');
 const { requireAdmin } = require('./utils/admin');
 const { signQurlOAuthState } = require('./utils/qurl-oauth-state');
 const { deleteLink, getResourceStatus } = require('./qurl');
@@ -55,20 +61,6 @@ let _warnedStateSecretFallback = false;
 // on every process start; tests that need a stable secret should set
 // OAUTH_STATE_SECRET explicitly in their own mocks.
 const _testFallbackSecret = crypto.randomBytes(32).toString('hex');
-// formatSelfDestructLabel renders a seconds value as a compact button
-// label fragment — `0.5s`, `30s`, `1m`, `5m`, `1h`. Avoids `30 seconds`
-// because Discord buttons truncate aggressively and pixel budget on the
-// 2-button row is tight when paired with the personal-note button.
-function formatSelfDestructLabel(seconds) {
-  if (seconds < 60) return `${seconds}s`;
-  if (seconds < 3600) {
-    const m = seconds / 60;
-    return Number.isInteger(m) ? `${m}m` : `${m.toFixed(1)}m`;
-  }
-  const h = seconds / 3600;
-  return Number.isInteger(h) ? `${h}h` : `${h.toFixed(1)}h`;
-}
-
 function stateSecret() {
   // Prefer a dedicated OAUTH_STATE_SECRET so a compromised GITHUB_CLIENT_SECRET
   // can be rotated without also invalidating in-flight OAuth state tokens —
@@ -1267,8 +1259,9 @@ async function handleSend(interaction, apiKey) {
   let recipients = [];
   let personalMessage = null;
   let expiresIn = '24h';
-  // Self-destruct timer (seconds, 0.5–3600). null = no timer (default).
-  // Forwarded to the connector as `viewer_ttl_seconds` at upload time.
+  // Self-destruct timer — one of SELF_DESTRUCT_PRESETS in seconds, or null
+  // for no timer (default). Forwarded to the connector as
+  // `viewer_ttl_seconds` at upload time.
   let selfDestructSeconds = null;
 
   const formId = `qurl_form_${sendNonce}`;
@@ -1563,11 +1556,13 @@ async function handleSend(interaction, apiKey) {
     }
 
     if (compInt.customId === ids.selfDestructBtn) {
-      // Modal collects an optional self-destruct timer (seconds).
-      // Empty submit = no timer (also the way users clear an existing
-      // value). parseSelfDestructSeconds enforces 0.5–3600 mirror of
-      // the connector contract; on validation failure we re-render the
-      // form with an inline warning rather than rejecting the modal.
+      // Modal collects an optional self-destruct timer. Choices come from
+      // SELF_DESTRUCT_PRESETS — Discord modals are TextInput-only (no native
+      // dropdown nested in a modal), so the placeholder lists the 7 options
+      // and parseSelfDestructSeconds enforces preset membership. Empty
+      // submit = no timer (also the way users clear an existing value).
+      // On validation failure we re-render the form with an inline warning
+      // so the user keeps their other selections.
       const destructInputId = 'destruct_value';
       const destructModal = new ModalBuilder()
         .setCustomId(selfDestructModalId)
@@ -1575,12 +1570,12 @@ async function handleSend(interaction, apiKey) {
       destructModal.addComponents(new ActionRowBuilder().addComponents(
         new TextInputBuilder()
           .setCustomId(destructInputId)
-          .setLabel('Seconds before content vanishes (0.5–3600)')
-          .setPlaceholder('e.g. 30 (or leave blank for no timer)')
+          .setLabel('Pick a duration (blank = no timer)')
+          .setPlaceholder(SELF_DESTRUCT_OPTIONS_TEXT)
           .setStyle(TextInputStyle.Short)
           .setMaxLength(32)
           .setRequired(false)
-          .setValue(selfDestructSeconds ? String(selfDestructSeconds) : '')
+          .setValue(selfDestructSeconds ? formatSelfDestructLabel(selfDestructSeconds) : '')
       ));
       await compInt.showModal(destructModal).catch(logIgnoredDiscordErr);
 
@@ -1602,9 +1597,11 @@ async function handleSend(interaction, apiKey) {
       if (error) {
         // Inline warning on the form rather than killing the flow —
         // the user keeps their other selections and can correct the
-        // value with another click.
+        // value with another click. The parser error already carries
+        // the full option list; the modal label already advertises
+        // "blank = no timer", so no extra prompt is needed here.
         await destructSubmit.update({
-          content: formContent({ warning: `Self-destruct: ${error} Try again or leave the field blank to skip.` }),
+          content: formContent({ warning: `Self-destruct: ${error}` }),
           components: formRows(),
         }).catch(logIgnoredDiscordErr);
       } else {
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1307,7 +1307,10 @@ async function handleSend(interaction, apiKey) {
       const preview = personalMessage.length > 80 ? personalMessage.slice(0, 80) + '…' : personalMessage;
       content += `\n\n_Personal message:_ "${preview}"`;
     }
-    if (selfDestructSeconds) {
+    // Same isFinite + > 0 invariant the modal prefill uses — a
+    // corrupted DDB row carrying Infinity is truthy and would otherwise
+    // surface as "_Self-destruct timer:_ (invalid)" in the form preview.
+    if (Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0) {
       content += `\n\n_Self-destruct timer:_ ${formatSelfDestructLabel(selfDestructSeconds)}`;
     }
     return content;
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1307,6 +1307,9 @@ async function handleSend(interaction, apiKey) {
       const preview = personalMessage.length > 80 ? personalMessage.slice(0, 80) + '…' : personalMessage;
       content += `\n\n_Personal message:_ "${preview}"`;
     }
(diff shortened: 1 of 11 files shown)


justin-layerv:
Thanks for the careful pass. Addressed in `f98c25a`:

**🔴 PR description mismatch** — rewritten to reflect the preset-only contract (placeholder = option list, no clamping anywhere, parser strict-matches the 7 friendly labels OR their seconds equivalents). Also added the rationale for why preset-only over the `0.5–3600` range — the 0.7s edge case at the connector floor is precisely what we don't want creators to ship by accident.

**🟡 Orphaned JSDoc in connector.js** — moved `appendViewerTtl` below `connectorAuthHeaders` so each function lives next to its own doc. The connector-side JSDoc is back on its rightful function.

**🟡 Stale "60s→1m" framing** — `formatSelfDestructLabel` now lives in `utils/time.js` and renders by preset table lookup; the "compress 60s→1m" branch is gone. Also dropped the matching stale "0.5–3600" comment on the `SAFE_ALTERS` migration entry.

**🟡 Test gaps on positive paths**:
- `tests/qurl-send-state-machine.test.js`: added a positive-timer file-path test that pins `downloadAndUpload` receives the threaded value (uses `"5 minutes"` to also cover the friendly-label parse on the file branch).
- `tests/qurl-send.test.js`: added two `handleAddRecipients` inheritance tests — one for the file re-upload path (`self_destruct_seconds: 30` → `downloadAndUpload`), one for the url/maps path (`self_destruct_seconds: 300` → `uploadJsonToConnector`). Locks down the "additional recipients see the same timer" claim from the description.

**Nits**:
- `let formCaptured = null` in `connector-coverage.test.js` → just `const appended = []` directly.
- Extracted `SELF_DESTRUCT_INPUT_MAX_LENGTH` shared between the modal's `setMaxLength` and the parser's length cap so a future bump only happens in one place.

Skipped:
- The earlier review's `selfDestructSeconds != null` cosmetic — the parser cannot return 0 (off-preset), so the truthy check is exact today.
- The earlier review's "empty-clears starts from null state" — the parser test already pins `'' → { seconds: null, error: null }` symmetrically across both empty-after-set and never-set inputs.

989 tests pass, 100% coverage on `utils/time.js`.

justin-layerv:
Both findings addressed in `9dc64a5`:

**1. Hex / scientific notation bypass** — confirmed real:
```
Number("0x1")  → 1   (matches "1 second")
Number("0x1e") → 30  (matches "30 seconds")
Number("0x12c")→ 300 (matches "5 minutes")
Number("5e-1") → 0.5 (matches "1/2 second")
```
Tightened the parser with `/^[+-]?\d+(\.\d+)?$/` before `Number()` so only decimal notation can reach `findPresetBySeconds`. Updated the test from the 3-case hex-prefix variant to a 14-case set covering all four bypass classes (plain hex `0x1/0x1e/0x12c`, hex float `0x1p3`, signed hex `+0x1/-0x1`, and scientific notation `5e-1/3e2/1e0/0.5e0`). The misleading comment is gone — the new comment explains exactly why the regex gate matters.

**2. DDB roundtrip parity** — added a `saveSendConfig` test in `ddb-store.test.js` that mirrors the sqlite roundtrip across the same 0.5 / 30 / omitted cases. A future refactor that drops `self_destruct_seconds` from the Item now fires on both stores.

Side effect: the regex tightening makes `findPresetBySeconds`'s `Number.isFinite` guard unreachable from the parser, but it's still load-bearing for `formatSelfDestructLabel(NaN)` if a corrupted DB row makes it through. Added a formatter test for that path so the defensive guard stays pinned at 100% coverage.

Skipped the optional `Map<canonicalLabel, preset>` precompute — `N=7` is too small to justify the indirection.

991 tests pass, 100% on `utils/time.js`.

justin-layerv:
All round-3 nits addressed in `33687fd`:

**1. Doubled colon** — parser errors now read as verb phrases (`must be one of: …`, `is too long (max 32 characters).`) so the handler renders `Self-destruct timer ${error}` with one colon and natural reading order.

**2. Length-cap message includes the limit** — `is too long (max ${SELF_DESTRUCT_INPUT_MAX_LENGTH} characters).` so the rare upstream-bypass path doesn't leave the user guessing.

**3. `inheritedDestruct` hoisted** — moved above the `if (hasFile) / if (hasLocation)` branches in `handleAddRecipients`. Both branches can fire for a sendConfig with both kinds, so the single-source-of-truth prevents drift.

**4. Button-label state pinning** — addressed via UX upgrade rather than just a test. `formContent` now echoes the chosen timer (`_Self-destruct timer:_ 30 seconds`) on the form preview line, parity with the personal-message preview. Doubles as the visible state signal that was missing — the new state-machine test pins the preview line so a future refactor that decouples the button label from `formatSelfDestructLabel` fires here too.

**5. `(invalid)` fallback** — `formatSelfDestructLabel(NaN)` and `formatSelfDestructLabel(Infinity)` now return `(invalid)` instead of `NaNs`/`Infinitys`. Defends the button label + modal prefill against a corrupted DB row surfacing garbage.

**Skipped (per cr's explicit flag)**:
- Empty-submit-clears confirmation — cr noted "Not changing anything; flagging for future UX consideration", and parity with the personal-note flow is the right call.

993 tests pass, 100% on `utils/time.js`.

justin-layerv:
All round-4 items addressed in `aa26a86` — including the (2) clear-after-set test:

**(2) Clear-after-set test** — rewrote the empty-submit test to open the destruct modal twice: first `"30"`, then `""`. Asserts (a) the form preview shows `_Self-destruct timer:_ 30 seconds` after the SET submit, (b) the preview line is gone after the CLEAR submit, (c) the eventual `uploadJsonToConnector` call carries `null`. The previous test could pass even if the handler ignored empty submits entirely; the new one pins the actual clear semantic.

**(1) Lost-input doc-comment** — added inline rationale for why we don't stash the rejected raw for the next modal open. Trade-off documented: clean retype (with the option-list warning visible) over let-them-edit-the-typo, since the right value is by definition one of the listed presets.

**(3) `+0.5` regex tightening** — `DECIMAL_SECONDS_RE` is now `/^\d+(\.\d+)?$/`. Tests extended with `+0.5`, `+1`, `+30`, `+300`. The accept-set matches the placeholder exactly.

**(4) Modal prefill `isFinite + > 0`** — swapped truthy check for `Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0`. A corrupted row carrying `Infinity` is truthy and would otherwise prefill the modal with the literal `"(invalid)"` formatter fallback. Same invariant as `appendViewerTtl` on the connector boundary now.

**(5) `OPTIONS_TEXT.length <= 100` test** — added so a future preset addition that pushes the joined string past Discord's `setPlaceholder` cap surfaces in CI, not as a runtime API error.

994 tests pass, 100% on `utils/time.js`.

justin-layerv:
All round-5 cleanups landed in `9f626a1`:

**(1) `appendViewerTtl` typeof guard removed** — `Number.isFinite('30')` is `false`, so `typeof === 'number'` was redundant. Same invariant as the modal prefill now.

**(2) `findPresetBySeconds` isFinite early-return removed** — `NaN !== anything` and `Infinity !== <finite>` both naturally fail the equality loop. `formatSelfDestructLabel` keeps its own non-finite check for the `(invalid)` fallback so the user-facing rendering contract is unaffected.

**(3) `formContent` guard now mirrors modal prefill** — `Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0`. All three boundaries (wire append, modal prefill, form preview) share the same positive-finite check, so a corrupted Infinity row can't surface as `_Self-destruct timer:_ (invalid)` in the preview line.

**(4) Format → parse round-trip test** — added in `tests/time.test.js`. Pins that every preset's `formatSelfDestructLabel(seconds)` parses back to the same `seconds`. This is the contract that holds when a user re-opens the modal to edit and submits without changing — without it, a future formatter/parser change could silently mutate a stored value on resubmit.

Skipped (cr explicitly flagged "not worth optimizing"): the `Map<canonicalLabel, preset>` precompute. N=7 doesn't justify the indirection.

995 tests pass, 100% on `utils/time.js`.

justin-layerv:
Both round-6 nits landed in `d0bdf24`:

**Button-label guard alignment** — `formRows`' destruct-button-label predicate now matches the modal-prefill / formContent invariant (`Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0`). Unreachable today but the alignment defends against a future caller piping a corrupted stored value.

**`findPresetByLabel` precompute** — preset labels are canonicalized once at module load into a parallel array; lookup is now an `indexOf` instead of a re-canonicalize-each-iteration loop. Trivial perf, slightly cleaner read.

995 tests pass, 100% on `utils/time.js`.

Per the cr ("Nothing blocking. Ship it.") this should be the final lap.

PR #236, opened 2026-05-10
feat(discord): self-destruct timer is a true dropdown, not a modal

justin-layerv:
## Summary

Replaces the modal-with-text-input UX from #235 with a real `StringSelectMenu` in the form. When the modal first shipped the label `Pick a duration (blank = no timer)` implied a dropdown promise that the typed-input field couldn't deliver — users would read it, see a text field, and either retype the placeholder text or wonder where the dropdown was.

Discord modals can't host StringSelectMenus, so the select has to live in the form itself. The form was already saturating the 5-row max when `target=user`, so this PR shuffles the bottom row to fit:

| | before (#235) | after (this PR) |
|---|---|---|
| Row 1 | `targetSelect` | `targetSelect` |
| Row 2 | `userSelect` (when target=user) | `userSelect` (when target=user) |
| Row 3 | `messageBtn` + `selfDestructBtn` (modal-opener) | `selfDestructSelect` (true dropdown) |
| Row 4 | `expirySelect` | `expirySelect` |
| Row 5 | `send` + `cancel` | `messageBtn` + `send` + `cancel` |

The personal-note button moves into the bottom row (Discord allows up to 5 buttons per ActionRow, 3 fits comfortably). Its label shortens from `Add a personal note (optional)` to `Add a note` to keep the row visually balanced. The optional-edit button sits leftmost so left-to-right reading order puts it before the commit action.

## UX changes

- **Dropdown header always reflects current state.** When unset, the `No self-destruct timer` option is `default: true` and the header shows `No self-destruct timer` — same convention `expirySelect` uses (which always shows the current expiry as the default-true option). When a preset is picked, the header shows `💥 5 minutes` (etc.).
- **Eight options:** `No self-destruct timer` (default), then the seven curated durations in ascending order with a 💥 prefix — `💥 1/2 second`, `💥 1 second`, `💥 5 seconds`, `💥 30 seconds`, `💥 5 minutes`, `💥 30 minutes`, `💥 1 hour`. The bomb emoji rides only on active-timer states; pinning it on `No timer` would read as a contradictory icon.
- **Clearing is one click**, not a modal-open-then-submit-empty dance. The user just picks `No self-destruct timer`.
- **No typing, no validation errors.** The option set is the source of truth; users can only pick a preset.
- **Form preview line** (`_Self-destruct timer:_ 5 minutes`) and the existing `personalMessage` preview keep their parity.

## Type of Change

- [ ] Bug fix
- [x] New feature (UX rework of a freshly-merged feature)
- [ ] Breaking change

## Plumbing changes

- **`src/utils/time.js`** — Drops `parseSelfDestructSeconds` and all the text-parsing infra (`DECIMAL_SECONDS_RE`, `canonicalize`, `findPresetByLabel`, `SELF_DESTRUCT_OPTIONS_TEXT`, `SELF_DESTRUCT_INPUT_MAX_LENGTH`, the precomputed canonical-label index). Adds a small `selfDestructSelectValueToSeconds` — strict string equality against `String(preset.seconds)`, so a forged `'0x1'` can't slip past as `1` (same hex/scientific-notation foot-gun the previous parser defended against).
- **`src/commands.js`** — Drops the ~70-line `selfDestructBtn` modal handler block (modal builder, awaitModalSubmit, parser invocation, error-warning rendering, lost-input doc-comment, the whole apparatus). Replaces with a one-liner select handler that calls `selfDestructSelectValueToSeconds` and re-renders.
- **No connector changes.** `viewer_ttl_seconds` wire-field plumbing is unchanged — `appendViewerTtl` and the four upload paths still thread the same `selfDestructSeconds`.
- **No DB changes.** `self_destruct_seconds` column still REAL on both stores.
- **No `handleAddRecipients` changes.** Inheritance still pulls from `sendConfig.self_destruct_seconds`.

## Tests

989 pass, 100% on `utils/time.js`.

- `time.test.js`: 12 parser tests removed (parser is gone), 3 `selfDestructSelectValueToSeconds` tests added (no-timer sentinel, every preset value, forged-value defense), plus a format ↔ value round-trip pin so the dropdown's `default` flag stays honest.
- `qurl-send-state-machine.test.js`: 5 modal-flow tests become 6 select-flow tests:
  - picking a preset persists into the upload call
  - picking a preset re-renders with the visible preview line
  - **every preset value (all 7) threads through to the upload call** — locks down the whole option set, not just one example
  - `no-timer` option AFTER a preset actually clears (active-clear, not trivially-already-null)
  - empty-values array falls back to no-timer (defense)
  - forged select value falls back to no-timer (defense)

## Why ship this on top of #235 instead of folding it in

#235 was already merged when the UX issue surfaced. Keeping the rework as a follow-up PR keeps the rollback story simple — revert this PR, the bot is back to the modal flow without losing the connector-side plumbing the modal flow already exercises end-to-end in sandbox.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1360,18 +1360,18 @@ async function handleSend(interaction, apiKey) {
     // isFinite + > 0 invariant the form preview uses — a corrupted DDB
     // row carrying Infinity is truthy and would otherwise mark a wrong
     // option default.
-    // hasTimer is the `default` predicate for both the no-timer option
-    // (negated) and each preset (with an equality check). An off-preset
-    // finite value (only reachable today via a hypothetical backfilled
-    // sendConfig.self_destruct_seconds outside the preset set) leaves
-    // every option un-defaulted; Discord then renders the first option
-    // ('No self-destruct timer') in the collapsed header while the
-    // formContent preview line still echoes the off-preset value via
-    // formatSelfDestructLabel's `${seconds}s` fallback. The mismatch is
-    // visible but harmless (the wire still carries the value); a more
-    // defensive choice would be `default: !hasTimer || !findPresetBySeconds(...)`,
-    // skipped here because the path is exotic and the asymmetry honest.
+    // hasTimer = current state has a positive-finite seconds value.
+    // hasMatchingPreset = that value matches one of the 7 presets.
+    // The two diverge only for off-preset finite values (only reachable
+    // today via a hypothetical backfilled sendConfig.self_destruct_seconds
+    // outside the preset set). Defaulting the no-timer option on
+    // `!hasMatchingPreset` keeps the dropdown header sensible — without
+    // it, every option is un-defaulted and Discord falls back to
+    // rendering the first option's label in the collapsed header,
+    // which would conflict with the formContent preview line that
+    // echoes the off-preset value via formatSelfDestructLabel.
     const hasTimer = Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0;
+    const hasMatchingPreset = hasTimer && SELF_DESTRUCT_PRESETS.some((p) => p.seconds === selfDestructSeconds);
     // No setPlaceholder — the No-timer option ships default-true when
     // !hasTimer, so the dropdown header always reflects the current
     // state and a placeholder would never render. Same default-true
@@ -1385,11 +1385,11 @@ async function handleSend(interaction, apiKey) {
         .setMinValues(1)
         .setMaxValues(1)
         .addOptions(
-          { label: 'No self-destruct timer', value: SELF_DESTRUCT_NO_TIMER_VALUE, default: !hasTimer },
+          { label: 'No self-destruct timer', value: SELF_DESTRUCT_NO_TIMER_VALUE, default: !hasMatchingPreset },
           ...SELF_DESTRUCT_PRESETS.map((p) => ({
             label: `\u{1F4A5} ${p.label}`,
             value: String(p.seconds),
-            default: hasTimer && selfDestructSeconds === p.seconds,
+            default: hasMatchingPreset && selfDestructSeconds === p.seconds,
           }))
         )
     ));
@@ -1413,7 +1413,7 @@ async function handleSend(interaction, apiKey) {
     rows.push(new ActionRowBuilder().addComponents(
       new ButtonBuilder()
         .setCustomId(ids.messageBtn)
-        .setLabel(personalMessage ? '✏\u{FE0F} Edit note' : '✏\u{FE0F} Add a note')
+        .setLabel(personalMessage ? '✏\u{FE0F} Edit note' : '✏\u{FE0F} Add a note (optional)')
         .setStyle(ButtonStyle.Primary),
       new ButtonBuilder()
         .setCustomId(ids.sendBtn)
@@ -1603,9 +1603,10 @@ async function handleSend(interaction, apiKey) {
       // selfDestructSelectValueToSeconds owns the conversion and falls
       // back to null (no timer) for any unexpected value (forged
       // interaction / option-list drift), which is the safe default.
-      // Optional chaining defends a malformed interaction that omitted
-      // the `values` array entirely — discord.js normalizes empty to
-      // [] but a forged payload could skip the field.
+      // Optional chaining is defense-in-depth — discord.js normalizes
+      // the gateway payload to `values: []` for String selects, so
+      // missing `values` is only reachable via a constructed-by-hand
+      // compInt (test path or forged event). The `?.` is still cheap.
       selfDestructSeconds = selfDestructSelectValueToSeconds(compInt.values?.[0]);
       await safeCompUpdate(compInt, { content: formContent(), components: formRows() });
       continue;
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1368,6 +1368,11 @@ async function handleSend(interaction, apiKey) {
     rows.push(new ActionRowBuilder().addComponents(
       new StringSelectMenuBuilder()
         .setCustomId(ids.selfDestructSelect)
+        // Explicit min/max=1 (Discord's default, matches userSelect's
+        // explicit settings on this form). The empty-values defense in
+        // the handler stays as belt-and-suspenders for forged payloads.
+        .setMinValues(1)
+        .setMaxValues(1)
         .addOptions(
           { label: 'No self-destruct timer', value: SELF_DESTRUCT_NO_TIMER_VALUE, default: !hasTimer },
           ...SELF_DESTRUCT_PRESETS.map((p) => ({
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1360,18 +1360,18 @@ async function handleSend(interaction, apiKey) {
     // row carrying Infinity is truthy and would otherwise mark a wrong
     // option default.
     const hasTimer = Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0;
+    // No setPlaceholder — the No-timer option ships default-true when
+    // !hasTimer, so the dropdown header always reflects the current
+    // state and the placeholder would never render. Same convention
+    // expirySelect uses (no placeholder; the current expiry is always
+    // the default-true option).
     rows.push(new ActionRowBuilder().addComponents(
       new StringSelectMenuBuilder()
         .setCustomId(ids.selfDestructSelect)
-        // Placeholder is a defensive fallback only — the No-timer option
-        // ships defaulted, so the dropdown header always reflects the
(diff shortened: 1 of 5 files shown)


justin-layerv:
All 3 round-1 items addressed:

**🔴 PR-description-vs-code mismatch** — fixed the description (commit-free, just `gh pr edit`). Code intent is unchanged: the `No timer` option ships `default: true` on initial render so the dropdown header always reflects current state, matching how `expirySelect` always shows the current expiry rather than its placeholder. The `default: !hasTimer` mechanism is intentional, not a bug — the original description was just wrong about which side of the toggle wins.

**🟡 Emoji style** — `💥` literal → `\u{1F4A5}` escape in the placeholder (commit `44351e5`). Matches the `\u{1F4A5}`/`\u{1F4E4}`/`\u{FE0F}` convention the rest of the file uses.

**🟡 Description on No-timer only** — dropped. The 7 preset labels are self-describing ("5 minutes" defines itself), so adding descriptions to all 7 was redundant; pinning one only on No-timer made the open-list visually uneven. Cleaner to drop.

Skipping (cr explicitly said "non-blocking"): N/A — addressed everything.

988 tests still pass.

justin-layerv:
3 of 4 round-2 items addressed in `c31a269`:

**Bottom-row reorder** — `[Add note, Send, Cancel]` now. The optional-edit affordance sits leftmost so left-to-right scanning hits it before Send.

**Test gap on full preset set** — added `30` and `1800` to the table. All 7 presets now pinned end-to-end through to the upload call.

**`values: []` defense test** — added a one-liner. The fall-through to `null` was already exercised by the forged-value test, but the empty-array path is now pinned explicitly so a refactor can't quietly regress it.

**Skipped (cr explicitly flagged "take it or leave it"):** the `selfDestructSelectValueToSeconds` ↔ `findPresetBySeconds` abstraction. At N=7 a 4-line loop is more legible than a transformed-key helper, and the two have different return shapes (preset object vs seconds value) which would make the unified helper signature awkward.

989 tests pass, 100% on `utils/time.js`.

justin-layerv:
3 of 5 round-3 items addressed in `2d04d0e`:

**(1) Optional chaining** — `compInt.values?.[0]` now defends a forged interaction that omits the `values` field entirely (the empty-array path was already covered).

**(3) Bottom-row rationale trimmed** — reframed as "left-to-right reading order puts the optional affordance before the commit action". The previous "reduces accidental Send" framing oversold it.

**(5) Placeholder dropped + UX tightened** — the placeholder was dead code given `default: !hasTimer` on the No-timer option. Removed.

Side-effect of dropping the placeholder: the dropdown header now literally shows the option's label, so I made two compensating changes:
- Renamed `No timer` → `No self-destruct timer` so the header is self-explanatory without the placeholder context (`No timer ▾` alone could read as "no timer for what?").
- Added 💥 emoji prefix to the 7 preset labels (active-timer states), so the header reads `💥 30 seconds` / `💥 5 minutes`. Visual cue ties the dropdown to the self-destruct concept. The No-timer option deliberately doesn't carry the bomb emoji — putting it next to "no" reads as contradictory.

**Skipped:**
- **(2) Explicit `setMinValues`/`setMaxValues`** — `expirySelect` doesn't set them either; staying consistent with prior art on this file.
- **(4) Direct test of the `default` flag on re-render** — testing the default flag specifically would be inconsistent with how the other two selects on this form are tested. The format↔value round-trip pin already covers the contract that drives the flag.

989 tests pass, 100% on `utils/time.js`.

justin-layerv:
All 3 round-4 items addressed in `076ccb0`:

**Stale modal-prefill cross-reference** — `formContent`'s comment now points at `hasTimer` in `formRows` (the only other site applying the same predicate now). The invariant itself is unchanged.

**`expirySelect` "no placeholder" claim** — wrong, it does set `Link expiry`. Corrected the parenthetical to focus on the actual convention being followed: current-value is always default-true.

**Option label drift in PR description** — updated description to reflect `No self-destruct timer` (not `No timer`) and the bomb-emoji prefix on the 7 active-state options.

989 tests still pass.

justin-layerv:
Round-5 doc-rot addressed in `94685c9`:

`tests/connector-coverage.test.js:197` referenced `parseSelfDestructSeconds` (gone in this PR's main commit). Reframed the comment to point at the dropdown as the contract instead. Test logic unchanged — it still pins `appendViewerTtl`'s defense against upstream-caller misuse, which is independent of how the bot collects the value.

`git grep parseSelfDestructSeconds` is now clean across `src/` and `tests/`.

The other observations in round 5 were either explicitly non-blocking or "fine to leave as-is":
- `hasTimer && ...` redundancy on the preset default flag — defensive, makes the predicate read symmetrically with the no-timer option.
- `setMinValues(1) / setMaxValues(1)` — explicitly matches `expirySelect` house style.
- Off-preset DDB row → dropdown header shows wrong default — already documented as "unreachable through the dropdown today".

989 tests still pass.

justin-layerv:
2 of 4 round-6 items addressed in `7632df6`:

**(1) Dead MIN/MAX constants dropped** — `SELF_DESTRUCT_MIN_SECONDS` / `MAX_SECONDS` were the parser's contract boundary; with the parser gone they had no consumers (grep across `src/` + `tests/` showed only self-references). Per CLAUDE.md "no half-finished implementations", dropped both + the `MIN/MAX track the first and last preset` test.

**(3) Explicit `setMinValues(1)` / `setMaxValues(1)` on `selfDestructSelect`** — Discord's defaults are `1`/`1` so behavior is unchanged, but this matches `userSelect`'s explicit settings elsewhere on the form and makes the empty-values defense test's claim "Discord's `setMinValues=1` means an empty values array shouldn't happen" literally true (it was relying on the default).

**Skipped (cr explicitly noted both as feel-free-to-ignore):**
- **(2) Precompute `Map<value, seconds>`** — perf microopt for a 7-element loop on a click event; not worth the indirection.
- **(4) Muscle-memory note** about the Send-button position shift — release-note territory, not code.

988 tests still pass (one dropped with the MIN/MAX test).

justin-layerv:
Round-7 was approving — addressed both consider-level notes in `adff022`:

**Edge-case doc-comment near `hasTimer`** — documents the off-preset scenario (hypothetical backfilled `sendConfig.self_destruct_seconds` outside the preset set). All options un-defaulted → Discord shows the first option in the header → mismatch with `formContent`'s `${seconds}s` fallback. The path is exotic and the mismatch harmless (wire still carries the value); the comment makes the rationale legible to a future reader who might otherwise assume every `hasTimer=true` branch marks exactly one default.

**Missing-`values`-field branch** — converted the empty-array test to an `it.each` over `[[], undefined]`. The optional chaining defends both cases; the missing-field path was technically uncovered until now.

Skipped per cr's explicit guidance:
- Sentinel paranoid prefix (cr: "Current value is fine")
- `safeCompUpdate` vs `.catch` unification on the neighboring `messageBtn` handler (cr: "Not introduced by this PR")

989 tests pass.

justin-layerv:
Took the off-preset fix in `0d7d1e9`:

**(2) `hasMatchingPreset` gate** — applied. Now the dropdown's `default` flags are consistent with what Discord actually renders: an off-preset finite value (e.g. backfilled `60`) leaves the No-timer option default-true and the form preview shows the value as informational text. Header and preview no longer disagree.

**(1) Note button "(optional)" restored** — agreed it's load-bearing for first-time users. The bottom row's pixel budget can absorb 4 more chars; not splitting hairs.

**(3) Optional-chain comment** — reframed to "defense-in-depth, cheap" since discord.js normalizes the gateway payload to `values: []`. The previous wording over-promised what discord.js's behavior necessitates.

989 tests pass.

justin-layerv:
Round 9 was approving. Took the one cr-flagged copy-consistency nit in `01041f2`:

**`Personal message:` → `Note:`** in the form preview. The button label says "Add a note (optional)" / "Edit note" — same field, two names. Picked "note" since it's the affordance label users actually click. Test assertion updated to match.

Skipped the other 3 explicitly-non-blocking nits:
- Per-option label length guard — cr called this "not worth adding a test for" (longest label 13 chars vs 100-char cap)
- Add-note ↔ Send adjacency visual hierarchy — sandbox-eyeball concern, can't fix from code
- Default-flag-on-re-render belt-and-suspenders test — would test discord.js's builder behavior, not ours

989 tests pass.

justin-layerv:
Round 10's real finding addressed in `4935b9c`:

**Dropped the misnamed "round-trip" test** — accurate critique. The test asserted two independent properties (value→seconds AND seconds→label) without composing them, so the "round-trip" framing was a holdover from the parser-era test where format(seconds) → user-typed-unchanged → parse(label) actually did round-trip. With the parser gone there's no compose-chain left to pin, and both halves are already covered by dedicated tests.

Skipped (cr explicit):
- `Map<string, number>` precompute — "Optional — the for-loop is fine"
- Corrupted-DDB visual inconsistency — already documented in the `hasMatchingPreset` comment; "Not a blocker"

988 tests pass.

PR #240, opened 2026-05-11
docs(discord): zero-downtime upgrade design + cross-process RESUME spike

justin-layerv:
## Why

Every deploy of `bot_gateway` today causes a 30–90s outage that breaks three independent user-visible surfaces:

1. **Slash command initiation** — users see "interaction failed" while the new task boots.
2. **In-flight `/qurl send` flows** — anyone mid-flow (waiting to drop a file, click a button, submit a modal) gets silently dropped. This is a bug *today*, deploy or not — any process crash mid-flow loses the user.
3. **Bot presence** — the green dot disappears for the same window.

Targeting an enterprise availability bar means all three have to survive a deploy invisibly. Tracking issue: [qurl-integrations-infra#122](https://github.com/layervai/qurl-integrations-infra/issues/122).

## What's in this PR

Two artifacts that unblock the rest of the work without changing any production code path:

### 1. `apps/discord/docs/zero-downtime-design.md` — the design doc

Captures:

- **Why current code can't get there** — Discord's one-active-WS-per-token rule plus `MESSAGE_CREATE` being Gateway-only (rules out HTTPS Interactions as an alternative).
- **Event-shipper architecture** — thin `bot_gateway` forwards Discord dispatches to SQS; stateless worker tier (extends `bot_http`) consumes the queue and runs all business logic. Worker deploys become standard rolling deploys with 0-downtime trivially because workers are stateless.
- **Three implementation pillars** — persistent flow state in DDB (replaces every `await*` in `commands.js`), cross-process Gateway RESUME via `@discordjs/ws`'s public `retrieveSessionInfo` / `updateSessionInfo` callbacks, and hot-standby with direct push-handoff over a control channel (sub-second WS gap, not seconds-of-polling).
- **SLI / SLO definitions** for slash command availability, flow continuity, presence, and resume success rate.
- **Per-phase rollback paths** — every phase has a documented flip-back, including app-side feature flags `ENABLE_EVENT_SHIPPER` and `ENABLE_GATEWAY_RESUME`.
- **Deliberately deferred items** — sharding inflection, worker autoscaling tuning, cross-region failover.

Reviewed against a DE lens before posting; corrections incorporated:

- MESSAGE_CREATE cache is **positive-only** — caching "user has no flow" would silently drop the user's single-shot file upload.
- Standby discovery uses a DDB heartbeat row, not ECS Service Connect (which resolves to the active itself and would need a self-exclusion filter).
- Documented degradation path when no peer is reachable on SIGTERM (handoff falls back to cold-start; doesn't fail outright).

### 2. `apps/discord/scripts/gateway-resume-spike.js` + tests — runnable validation

Two-phase Node script that proves Pillar 2's mechanism against a real Discord bot token:

- **phase1** opens a fresh Gateway connection, captures `session_id` / `resume_url` / `sequence` via `updateSessionInfo`, persists to a local file, closes clean (code 1000).
- **phase2** (run within Discord's ~60s resume buffer) loads the file, feeds it back via `retrieveSessionInfo`, observes whether Discord ACKs with `RESUMED` (success — Pillar 2 mechanism validated) or with `READY` + a `updateSessionInfo(null)` (resume rejected, graceful IDENTIFY fallback).

The persistence helpers (atomic write, malformed-file fallback) plus a contract test pinning the `@discordjs/ws` option names against the installed library version are unit-tested. Manual validation against the sandbox bot is the remaining step — see the script's header runbook.

## What's NOT in this PR

This PR is design + de-risking only. No production code path changes. The follow-up PRs (PRs 1–15) are pre-staged and will land in phases per the design doc.

## Test plan

- [x] `npm test` passes locally (1006/1006).
- [x] `npm run lint` passes locally with `--max-warnings 0`.
- [x] Spike-helper unit tests cover atomic-write, malformed-file fallback, contract pinning on `@discordjs/ws` option names (15 tests).
- [ ] **Manual:** run the spike's phase1 + phase2 against the sandbox bot token, confirm `RESUMED` dispatch in phase2's log (see `apps/discord/scripts/gateway-resume-spike.js` header for runbook).

## Open questions for reviewers

- Anything in the design that should be expanded vs. moved to a follow-up?
- Anyone want a fourth alternative architecture evaluated in §"Why not other approaches" before we close on event-shipper?
- The SLO numbers (slash command 99.95% / flow continuity 99.99% / presence 99.9%) are starting placeholders — open to tightening or loosening based on what's reasonable to engineer for at our scale.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,5 +32,6 @@ e2e/node_modules/
 apps/discord/.claude/
 apps/discord/scripts/*
 !apps/discord/scripts/loadtest-standalone.js
+!apps/discord/scripts/gateway-resume-spike.js
 apps/discord/.github/
 apps/discord/railway.json
--- a/apps/discord/docs/zero-downtime-design.md
+++ b/apps/discord/docs/zero-downtime-design.md
@@ -1,0 +1,342 @@
+# Zero-downtime upgrades for the qURL Discord bot
+
+**Status:** design — ratified, implementation in progress
+**Tracking:** `qurl-integrations-infra#122` (deploy outage), `qurl-integrations#TBD` (this PR)
+**Owners:** posey + reviewers
+
+## Summary
+
+Every deploy of `bot_gateway` today causes a 30–90 s outage. Three independent
+user-visible surfaces break: slash command initiation, in-flight `/qurl send`
+flows (button clicks, file drops, modals), and the bot's presence indicator. To
+meet an enterprise availability bar, all three must survive a deploy with no
+user-observable impact.
+
+The redesign collapses to one architectural decision (event-shipper split) plus
+three implementation pillars. After this work, deploying business logic is a
+standard rolling deploy with zero user impact, and the rarely-deployed gateway
+tier uses Discord's own session-resume primitive plus a hot-standby with direct
+push-handoff to keep its own outage sub-second.
+
+## Why the current design can't get there
+
+Three independent layers break during a deploy of the current `bot_gateway`
+service (singleton, `desired_count=1`, `deployment_minimum_healthy_percent=0`):
+
+1. **Slash command initiation.** Discord enforces a 3 s ACK deadline on every
+   interaction. While the gateway task is being replaced, slash commands
+   return "interaction failed" in the user's Discord client.
+2. **In-flight `/qurl send` flows.** The send flow uses `awaitMessageComponent`,
+   `awaitMessages` (file drop in DM), and `awaitModalSubmit` in nine distinct
+   sites across `apps/discord/src/commands.js`. Each is a Promise resolved by
+   an event the Gateway delivers to *this process*. SIGTERM = those Promises
+   never resolve = the user's flow silently dies. This is a bug today even
+   outside of deploys (any process crash mid-flow drops the user).
+3. **Bot presence.** The green dot in Discord member lists flips off for the
+   same window.
+
+Two Discord constraints make these hard:
+
+- **One active Gateway WebSocket per bot token.** A second `IDENTIFY` on the
+  same token invalidates the first session. Rules out the naive
+  `desired_count=2` rolling deploy. Already enforced in
+  `qurl-integrations-infra/qurl-bot-discord/terraform/variables.tf` via a
+  validation block rejecting `gateway_desired_count != 1`.
+- **`MESSAGE_CREATE` is Gateway-only.** Discord's HTTPS Interactions endpoint
+  (an alternative interaction delivery surface) carries interaction events but
+  *not* regular DM messages. Users dropping a file as a DM message is the
+  load-bearing step of `/qurl send`; switching to HTTPS Interactions would
+  break that step. We considered and rejected this approach.
+
+## The redesign at a glance
+
+Two tiers:
+
+```
+                     Discord Gateway WS
+                            │
+                            ▼
+      ┌─────────────────────────────────────────┐
+      │  GATEWAY TIER (bot_gateway)             │
+      │  - Maintains Discord Gateway WebSocket  │
+      │  - Forwards every dispatch to SQS       │
+      │  - Zero business logic                  │
+      │  - Hot-standby + leader-election lock   │
+      │  - Cross-process RESUME via DDB         │
+      │  - Deployed rarely (Node/discord.js     │
+      │    version bumps, sharding changes)     │
+      └────────────────┬────────────────────────┘
+                       │
+              SQS FIFO Queue (per-guild
+              ordering, DLQ, autoscaling
+              signal for workers)
+                       │
+                       ▼
+      ┌─────────────────────────────────────────┐
+      │  WORKER TIER (extends today's bot_http) │
+      │  - Stateless queue consumer             │
+      │  - All /qurl command logic              │
+      │  - All flow-state transitions in DDB    │
+      │  - Discord REST for outbound responses  │
+      │  - Standard rolling deploys (min=100,   │
+      │    max=200, ALB drain) — 0-downtime     │
+      │    trivially because stateless          │
+      │  - Auto-scales on queue depth           │
+      └─────────────────────────────────────────┘
+```
+
+The user-visible deploy story becomes:
+
+- **Worker tier deploy (the common case — every PR):** rolling deploy, standard
+  ECS `deployment_minimum_healthy_percent = 100`, `maximum = 200`. New tasks
+  reach steady state before old tasks stop. ALB target-group `deregistration_delay = 90`
+  matches `idle_timeout` so in-flight HTTP requests drain. Workers are stateless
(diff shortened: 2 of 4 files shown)


justin-layerv:
## Spike validation result — RESUME-OK ✅

Ran the spike end-to-end against the sandbox bot token on 2026-05-10. Result: **cross-process Discord Gateway RESUME mechanism is validated.**

### Setup

- Sandbox `bot_gateway` ECS service scaled to `desired_count=0` for the duration of the test to eliminate token-contention (a contending IDENTIFY on the same token invalidates the session and causes RESUME to fail for unrelated reasons).
- Sandbox restored to `desired_count=1` immediately after; verified `running=1` healthy (task `8df66f0b…`).
- Token retrieved from SSM (`/qurl-bot-discord/DISCORD_TOKEN`) via the `layerv-integrations` AWS profile, held in a `chmod 600` tmpfs file, deleted after the run.

### Result

```
[phase2] retrieveSessionInfo(0) -> stored session (will RESUME)
[phase2] updateSessionInfo(0) seq=2 session=2a3ff10f…
[phase2] updateSessionInfo(0) seq=3 session=2a3ff10f…
[phase2] updateSessionInfo(0) seq=4 session=2a3ff10f…
[phase2] RESUMED dispatch received — cross-process resume SUCCEEDED
[phase2] RESULT: RESUME-OK (Pillar 2 mechanism validated)
```

Same `session_id` (`2a3ff10f…`) survives across processes. Discord replays events with incrementing `seq` to the resumed session. This is the load-bearing proof for Pillar 2.

### Two design findings from the run (commits in this PR)

**1. `retrieveSessionInfo` must respect the null clear signal from `updateSessionInfo`.** The first run produced an infinite RESUME-reject loop because my callback was returning the loaded-at-startup session unconditionally — Discord rejected, library cleared, my callback re-returned the same dead session, ~200ms cycle. Production DDB-backed code maintains a mirror that `updateSessionInfo` writes into; `retrieveSessionInfo` reads from the mirror, not from the original load. Fixed in commit `3e4952f`.

**2. `manager.destroy()` invalidates the session — production must NOT use it for cross-process handoff.** Reading `@discordjs/ws`'s `destroy()` implementation: it unconditionally calls `updateSessionInfo(null)` AND sends a close-1000 frame (which Discord treats as terminal). The `recover: Resume` flavor uses close-4200 (resumable on Discord's side) but ALSO triggers `internalConnect()` to resume in the same process — not useful for cross-process. The right shape: persist final state, then `process.exit()` without closing the WS. TCP drops, Discord sees a network disconnect, session stays resumable for ~60s. Validated in commit `a2a8086`.

This matches the production SIGTERM handler sequence in the design doc's Pillar 3: persist → push to standby → exit (no clean close).

### What's still untested by this spike

- The push-handoff timing (Pillar 3) — separate end-to-end test once PR 13 ships.
- SQS forwarding of dispatches (Pillar 1) — covered by PRs 10/11.
- The leader-election lock primitive (PR 13).
- Behaviour under sustained traffic (the spike only ran for ~10s phase1 + ~5s phase2 — no high-volume event replay tested).

### Test plan checklist

- [x] `npm test` passes locally (1006/1006).
- [x] `npm run lint` passes with `--max-warnings 0`.
- [x] Unit tests for spike persistence helpers (15/15).
- [x] **Manual spike validation against sandbox bot — RESUMED dispatch confirmed.**

Ready for review.

vikramlayerv:
A few clarifying questions before I start in on the follow-up PRs — flagging while the doc is fresh:

**1. SQS FIFO vs Standard.** FIFO caps at 300 TPS (3000 in high-throughput mode) across the whole queue. If every `MESSAGE_CREATE` the bot receives in any DM (flow or not) ships through, FIFO becomes the throughput ceiling well before sharding inflection. Standard + DDB-OCC at the flow-transition layer would enforce per-flow ordering and scale further. Is FIFO chosen because it gives a per-guild guarantee DDB-OCC can't, or for simplicity?

**2. SIGTERM step ordering — release vs destroy.** Step 7 (`releaseLock`) runs *before* step 8 (`WebSocketManager.destroy`). If the standby's `acquireLock` in `POST /control/yours` (step 2) races slow, it could miss and fall back to the heartbeat-cadence retry. Should release move *after* destroy, or gate on a standby-side `connect()` ACK (push-pull handshake)?

**3. Cold-start fallback latency floor.** §Pillar 3 says no-peer SIGTERM degrades to "a few seconds." Worst case is bounded by the standby's lock-acquisition cadence — heartbeat 5s, TTL 15s — so the no-peer floor could be 15s + RESUME RTT before WS is back. Is "a few seconds" the optimistic case, or do we want the heartbeat/TTL tightened for the no-peer path?

**4. Positive-cache effectiveness across N workers.** The non-flow-DM hot path benefits if the same worker handles both the flow-create event and the subsequent `MESSAGE_CREATE`. With SQS fanout there's no natural affinity — so workers that didn't see the create still pay a DDB `GetItem` per non-flow DM that lands on them. Is there a routing/affinity story (e.g., consistent-hash on flow-id), or do we accept the read-through cost across all workers?

**5. `ENABLE_EVENT_SHIPPER=false` rollback window.** Once command handlers are refactored out of the gateway role (Pillar 1), can the gateway role still serve in-process if the flag flips off, or does rollback only work *before* that PR lands? Worth pinning the rollback-cliff in the doc — `git revert` of the command-handler-removal commit gets us back, but the doc reads as if the flag alone is sufficient indefinitely.

Happy to start implementing on any of these as you confirm the answers — none are blockers for Pillar 1 / flow-state work, but #1 and #5 shape PR ordering.

justin-layerv:
Followups to your review + Vikram's questions landed in #241. TL;DR:

* FIFO → Standard (per Vikram #1); doc has a new "Queue" section explaining the application-layer idempotency contract.
* Pillar 3 SIGTERM handoff rewritten around `transferLock` CAS to close the lock race Vikram #2 flagged.
* Heartbeat/TTL tightened 5s/15s → 2s/6s — no-peer cold-fallback floor drops from ~15s to ~7s (Vikram #3).
* Positive cache removed (Vikram #4) — Standard queue has no worker affinity, cache would be stale-no-flow on N-1 workers; accept the bounded DDB read-through cost instead.
* `ENABLE_EVENT_SHIPPER` cliff documented (Vikram #5) — flag-rollback only valid while PR 10's dual-path is gated.
* Spike-script bugs: `MAX_IDENTIFY_ATTEMPTS` off-by-one, exit-code precedence so budget-exhausted is reachable, `mode: 0o600` on persist, SIGINT handler, `connect()` timeout race.
* `@discordjs/ws` and `@discordjs/rest` lifted from transitive → direct deps so a discord.js minor bump can't silently swap them.
* `npm run lint` now covers `scripts/` (was a hole — the spike script wasn't actually linted).
* Tests: `jest.spyOn` over monkey-patch, contract test destroys its manager, new `mode 0o600` assertion.

PR #241, opened 2026-05-11
fix(discord): address PR #240 review — design + spike followups

justin-layerv:
## Why

Follow-up to merged PR #240 (zero-downtime design + RESUME spike). Vikram left 5 clarifying questions on the design and the automated review surfaced several correctness items in the spike script and tests. This PR addresses everything in one pass — none of these touch production code, but the design changes shape PR B (SQS queue, in flight on qurl-integrations-infra) so it needs to land first.

### Design doc (`apps/discord/docs/zero-downtime-design.md`)

- **Queue: FIFO → Standard** (Vikram #1). FIFO's 300/3000 TPS cap is a real cliff at sharding scale; the ordering it would enforce isn't load-bearing since DDB is the source of truth. Standard works because we already have application-layer idempotency: Discord's `/callback` rejects dup ACKs, OCC version checks reject dup transitions, audit events tolerate dups. New "Queue" section documents the contract.
- **MESSAGE_CREATE positive-cache removed** (Vikram #4). With Standard queue there's no natural worker affinity, so a cache would be stale-no-flow on every worker except the one that handled flow-create. Coherent fanout would need a separate broadcast channel (Redis pubsub / DDB Streams) — more infra than the bounded DDB GetItem cost warrants. Doc explicitly accepts read-through.
- **Pillar 3 SIGTERM handoff rewritten around `transferLock` CAS** (Vikram #2). The earlier sequence had a real race: standby's `acquireLock` waited on a lock the active hadn't released, while active waited for an ACK the standby couldn't send → deadlock or always-cold-fallback. Atomic `transferLock(active → standby)` in the standby's POST handler closes the window. Active no longer calls `releaseLock()` (transferLock did it) and no longer calls `manager.destroy()` (Pillar 2 contract says destroy invalidates the session).
- **Heartbeat 5s/TTL 15s → 2s/6s** (Vikram #3). Drops the no-peer cold-fallback floor from ~15s+RESUME to ~7s. Cost: 2.5× DDB writes/sec (0.2/s → 0.5/s) ≈ $1.62/shard/month on on-demand — trivial. (Math corrected from initial draft per review.)
- **Control-channel auth: shared HMAC via SSM**. Committed off the "mTLS or HMAC" undecided wording. Narrow surface (one in-VPC endpoint, fixed payload); HMAC complexity matches the risk envelope; same trust class as `DISCORD_TOKEN`.
- **Standby-discovery freshness filter on read**. DDB TTL deletion is eventual (up to ~48h worst-case AWS-documented), so peer picking filters by `updated_at > now - 6s` at read rather than relying on TTL.
- **Clock-skew assumption documented**. ±2s of skew is absorbed by the 2s/6s ratio; beyond that `version` provides causal ordering. Documented but not engineered for — chrony failure on Fargate is a bigger incident.
- **ECS deployment-strategy invariant**. `bot_gateway` deploy config must serialize replacements (`minimumHealthyPercent=50` / `maximumPercent=100` at `desired_count=2`). Tfvars-level, not a deploy-time guess.
- **`ENABLE_EVENT_SHIPPER` rollback cliff documented** (Vikram #5). Flag rollback is valid only while PR 10's dual-path scaffolding is in place. After PR 10 removes the in-process fallback (post-soak), rollback is `git revert`, not a flag flip. PR 10's description must note the soak recommendation (1 week of clean prod traffic).

### Spike script (`apps/discord/scripts/gateway-resume-spike.js`)

- `MAX_IDENTIFY_ATTEMPTS` off-by-one fixed (comment said "1 IDENTIFY", guard let through 2).
- phase2 result-classification order: `budgetExhausted` check moved before `identified-without-budget`. The earlier order made exit-code 3 unreachable.
- `persistSession` writes with `mode: 0o600`. Session_id + resume_url is RESUME-usable for ~60s; world-readable in `/tmp` is sloppy even on a dev machine.
- phase1 SIGINT handler persists partial state. Ctrl+C between connect and the idle persist call previously left phase2 with no session and a confusing "no persisted session" error.
- phase1 `connect()` raced against 30s timeout. phase2 already had this; phase1 was missing it.

### Package + tooling

- `@discordjs/ws` (`~1.2.3`) and `@discordjs/rest` (`~2.6.0`) lifted from transitive (via `discord.js@14.25.1`) to explicit direct dependencies. Locks the surface area against `discord.js` minor bumps that could swap the @discordjs/ws major out from under us.
- `npm run lint` now covers `scripts/` in addition to `src/` and `tests/`. `scripts/**/*.js` has a `no-console` eslint override since CLI scripts legitimately log to stdout.

### Tests (`apps/discord/tests/gateway-resume-spike.test.js`)

- `jest.spyOn(fs, 'readFileSync')` + `afterEach(jest.restoreAllMocks())` replaces the direct monkey-patch in the EACCES propagation test (safer if assertions throw mid-body).
- Contract test now `await mgr.destroy({code:1000})` in a `try / finally` so a future @discordjs/ws version that opens a constructor-time timer doesn't leak a handle.
- New: `persistSession writes file with mode 0o600` test, gated on `process.platform !== 'win32'`.

## Test plan

- [x] `npm run lint` passes with `--max-warnings 0` (now covers `scripts/`).
- [x] `npm test` passes locally (1007/1007, up from 1006 with the new mode test).
- [x] Local validate: the 0o600 mode test fires and asserts the actual file mode bits.
- [ ] **Manual re-validation of the spike against sandbox is NOT required for this PR** — no Discord-API-facing behavior changed. Persistence helpers + result classifier are the only logic touched; covered by unit tests.

## Open questions intentionally left for follow-ups

These are flagged in the doc but deferred:

- **Sharding inflection (~2,500 guilds)**: schemas are already shard-aware, but per-shard leader election and queue consumer dispatch will need a generalization PR. Revisit when guild count crosses 1,500.
- **Worker-tier autoscaling thresholds**: initial alarms in PR B are conservative; we'll tune in the first month of real traffic.

### DIFF
--- a/apps/discord/.eslintrc.json
+++ b/apps/discord/.eslintrc.json
@@ -31,6 +31,12 @@
       "rules": {
         "no-console": "off"
       }
+    },
+    {
+      "files": ["scripts/**/*.js"],
+      "rules": {
+        "no-console": "off"
+      }
     }
   ]
 }
--- a/apps/discord/docs/zero-downtime-design.md
+++ b/apps/discord/docs/zero-downtime-design.md
@@ -67,9 +67,10 @@ Two tiers:
       │    version bumps, sharding changes)     │
       └────────────────┬────────────────────────┘
                        │
-              SQS FIFO Queue (per-guild
-              ordering, DLQ, autoscaling
-              signal for workers)
+              SQS Standard Queue (DLQ,
+              autoscaling signal for
+              workers; ordering enforced
+              at the DDB layer)
                        │
                        ▼
       ┌─────────────────────────────────────────┐
@@ -139,21 +140,53 @@ handler:
 
 1. Reads message metadata: `(channel_id, author_id, attachments)`.
 2. Constructs `flow_id` candidate keys from the metadata.
-3. Looks up the flow row in DDB. With ~thousands of guilds and ~tens of DMs/s
-   peak, this is a cheap lookup. To keep DDB reads off the per-message hot
-   path for non-flow DMs (the common case), we cache **positive** flow
-   memberships per worker: when worker A creates a flow for user X, A writes
-   to DDB and *also* publishes "user X is in flow Y" to a short-TTL in-memory
-   table on every worker (via the existing SQS dispatch — workers see flow-
-   create events too). Misses always re-read DDB. We **never** cache absence,
-   because a user dropping a file is a single irreversible action — a stale
-   "no flow" cache entry on the worker that gets the message would silently
-   drop the user's upload. Read-through on miss costs one DDB GetItem per
-   non-flow DM, which we accept; the optimization is for DMs *during* an
-   active flow, not against them.
+3. Looks up the flow row in DDB. **Every** `MESSAGE_CREATE` does a single
+   `GetItem` against the flow table by composite key. No in-process cache —
+   the earlier draft had a per-worker positive cache that would only have
+   been populated on whichever worker handled the matching flow-create
+   event, leaving every other worker with stale-no-flow assumptions. With
+   SQS Standard (at-most-once-per-consumer fanout) there's no natural
+   affinity between flow-create and the user's subsequent file-drop; a
+   coherent cache would require a separate broadcast channel (Redis pubsub
+   or DDB Streams), which is more infrastructure for marginal cost savings.
+   The volume math says the DDB cost is bounded: at ~thousands of guilds
+   and the bot's invite-only-beta DM patterns, GetItem-on-every-non-flow-
+   DM is tens-to-hundreds of reads per second, fully within DDB's
+   on-demand pricing without provisioned capacity tuning. Revisit only if
+   the volume crosses an order of magnitude.
 4. If matched, advances the flow via OCC. If two events race (e.g., user
    uploads twice quickly), one OCC retry path wins, the other gets a clear
    `"flow already advanced"` error.
+
+### Queue: SQS Standard + application-layer idempotency
+
+The queue between the gateway tier and the worker tier is SQS Standard, not
+FIFO. The rationale is that the ordering FIFO would enforce isn't load-
+bearing — the source of truth for "did this flow advance" is the DDB flow
+row, not the order in which events were processed — and FIFO's per-queue
+throughput ceiling (300 TPS, 3000 with high-throughput mode) is a real
+cliff at sharding scale that Standard doesn't have.
+
+Standard's at-least-once delivery means duplicate events. Application-layer
+idempotency covers it:
+
+- **`INTERACTION_CREATE` duplicates** are naturally idempotent at Discord:
+  `/callback` returns "Unknown interaction" on the second call because the
+  interaction token has already been ACK'd. The worker's only real cost on
+  a duplicate is a wasted REST call and a logged 4xx.
+- **`MESSAGE_CREATE` duplicates** are blocked by OCC at `transitionFlow`:
+  the second arrival reads the same `version`, attempts the same transition,
+  and the conditional `UpdateItem` fails with `ConditionalCheckFailedException`.
+  The handler treats this as "another worker already advanced this flow"
+  and exits without acting.
+- **`GUILD_CREATE` / `GUILD_DELETE` duplicates** are audit-only; duplicate
+  log lines are tolerable.
+
+Producer-side dedup hint: the gateway sets `MessageAttributes.event_id` from
+Discord's per-dispatch `s` (sequence) field as a defense-in-depth — workers
+log when they see a `event_id` they processed within the last 5 minutes
+(in-process LRU, ~10k entries) so a real-world dup rate can be quantified.
+The LRU is best-effort; OCC is the correctness primitive.
 
 ### Pillar 2 — Cross-process Gateway session resume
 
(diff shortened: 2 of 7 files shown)


justin-layerv:
Addressed all 8 items from the review in commit ef0bc18. Summary:

**Merge-blockers**

* **1 (math fix)**: 5× → 2.5× on heartbeat-cost math. Doc now states the correct rate (~$1.62/shard/month total on DDB on-demand) and stays internally consistent.
* **2 (mode-only-on-create)**: added `fs.chmodSync(tmp, 0o600)` after `fs.writeFileSync` in `persistSession`. New test pre-seeds a 0o644 stale `.tmp` and asserts the post-rename file lands at 0o600.
* **3 (HMAC replay)**: signed body now `{active_instance_id, expected_version, nonce, ts}`. Standby rejects `|ts - now| > 5_000ms` and dedups against an in-memory LRU of seen nonces. Dual-secret-accept window for SSM rotation also documented (SSM parameter holds `{current, previous}` JSON; 24h rotation window; active always signs with `current`).

**Design gaps**

* **4 (transferLock → connect failure)**: documented a connection-watchdog loop on the standby. Every 1s: if `heldLock && !connected`, attempt connect with exponential backoff; after 5 attempts release the lock and `process.exit(1)` so ECS replaces. Covers both the "transferLock succeeded but connect failed" path AND the cold-start no-peer path. PR 13 owns the implementation; PR 15 (chaos) exercises both.
* **5 (clock-skew margin)**: tightened the engineering target from ±2s to ±1s, with the explicit note that the `version`-check on the conditional write is the actual safety primitive, not the timing math. The 2.5× heartbeat-to-TTL ratio gives 1 missed-renewal margin at ±1s; beyond that `version` absorbs.

**Smaller items**

* **7 (classifyResult coverage)**: extracted the result-classification block into a pure function exported from the spike module. 6 new unit tests pin the dispatch order, including the specific case where `resumed=false / budgetExhausted=true / identified=true / postResumeSessionCleared=true` must return exit-code 3 (the reorder bug we caught against a real Discord run).
* **8 (SIGINT race)**: single `exiting` boolean guard. Second SIGINT exits immediately rather than racing the in-flight `persistSession`.
* **6 (patch-range deps)**: informational, no change — confirmed intentional; lockfile pins the actual versions.

**Test state**

  Before: 16/16. After: 23/23 (+6 classifier tests, +1 mode-overwrite test). Full suite: 1014/1014 (up from 1007). `npm run lint` clean.

The watchdog in #4 is documented in the design but **not implemented in the spike** — the spike's scope is Pillar 2 mechanism validation; the watchdog is Pillar 3 infra that lands in PR 13.

Ready for another pass.

justin-layerv:
Round-two feedback addressed in f360755.

| # | Reviewer ask | Fix |
|---|---|---|
| 1 | SIGINT-only handler claims to mirror production SIGTERM | Added `process.on('SIGTERM', ...)` to the same handler; distinct exit codes (130 / 143) so a parent script can distinguish cancellation vs. external termination. |
| 2 | Race-guard rationale overstated | Tightened the comment — `persistSession` is sync and Node serializes signal callbacks via the event loop, so concurrent interleave isn't actually possible. The `exiting` guard's real job is debouncing a second signal that arrives between handler-return and `process.exit`. |
| 3 | Connect-timeout setTimeout not `.unref()`'d | Added `.unref()` on the 30 s connect-timeout in phase1. Benign in practice (we `process.exit` anyway) but future-proof. |
| 4 | `restoreAllMocks` only covers one describe | Tried `restoreMocks: true` in jest.config first — **broke 65 tests** in `qurl-send-state-machine.test.js`, which set up `jest.spyOn` once per describe and assert across accumulated call counts. Reverted to a file-scope `afterEach` at the head of the spike test file (covers every describe in the file without leaking to other files). Added a comment in `jest.config.js` explaining why `restoreMocks` is deliberately NOT global. |
| 5 | HMAC payload doesn't bind recipient | Signed body now `{active_instance_id, peer_instance_id, expected_version, nonce, ts}` — at sharding inflection (N>1 standbys), each standby rejects requests not addressed to itself. Closes the cross-shard replay window before it can become a sharding-PR surprise. |
| 6 | `identifyAttempts` overruns MAX after `budgetExhausted` flips | Added an early short-circuit at the top of `retrieveSessionInfo` that re-throws cleanly without incrementing the counter. Cosmetic but the misleading "attempt 3/1" log was real. |
| Test+ | One more `classifyResult` case | Added `budgetExhausted=true, identified=false, postResumeSessionCleared=false` — the "RESUME hung, never READY, retrieveSessionInfo threw" path. |

**Test state**: 24/24 spike file (was 23), 1015/1015 full suite (was 1014). Lint clean.

The reviewer's framing was "LGTM with the cosmetic items above" — all items now resolved. Ready for merge.

justin-layerv:
Round-three feedback addressed in 2ba0061.

| # | Reviewer ask | Fix |
|---|---|---|
| 1 | Stale `restoreMocks` comment in tests | Deleted. The file-scope `afterEach` at the top of the file is what does the work; the per-describe comment was left over from the reverted approach in round 2. |
| 2 | Comment said `>= MAX` but code is `> MAX` | Comment fixed. Behavior was always correct (MAX=1 with `>` lets exactly one IDENTIFY through); just the comment lied. |
| 3 | PR description math still says 5× | **PR body updated** — now says "2.5× DDB writes/sec (0.2/s → 0.5/s) ≈ $1.62/shard/month". |
| 4 | HMAC rotation runtime path unspecified | Committed to **rolling redeploy** (matches every other SSM-backed secret on the bot). Doc now states: bot does NOT hot-reload SSM, current/previous are captured at boot, rotation is write-SSM → redeploy → wait 24 h → scrub → redeploy. Compromise rotation skips the wait. Considered hot-reload-with-TTL and rejected — would need new IAM scope, refresh loop, mid-handoff failure story. |
| 5 | `resumed + budgetExhausted` test comment overstated | Re-phrased to "defensive only, not reachable today" with the actual reasoning (RESUMED implies no IDENTIFY was attempted). |
| 6 | Over-commented spike (SIGINT block, `.unref()` block) | Trimmed ~30 lines of narration to ~10 lines covering only the non-obvious WHY (matched exit codes, debounce purpose, event-loop pin concern). Per CLAUDE.md "default to no comments." |
| 7 | Watchdog exit path needs chaos-suite tie-in | Added explicit one-liner in the design doc: PR 15's chaos suite kills Discord-endpoint resolution on the standby to exercise the retries-exhausted branch. Otherwise the failure ladder was "documented but never executed." |

**Test state**: 24/24 spike, 1015/1015 full suite (unchanged from round 2). Lint clean.

justin-layerv:
Round 4 feedback addressed in 774c7b7.

**Substantive**

| # | Reviewer ask | Fix |
|---|---|---|
| HMAC freshness vs ±5 s clock-skew tolerance | Pick a story | Tiered the skew envelope explicitly: ≤ ±1 s zero impact, ±1–2 s degraded (HMAC near-edge rejects, active retries POST; version check is lock-side safety), > ±2 s treated as chrony-failure incident paged on a `clock_skew_seconds` canary metric. Makes the auth section and lock section internally consistent and gives ops a real escalation ladder. |
| Verify `@discordjs/ws` / `@discordjs/rest` match discord.js transitive | `npm ls` check | <details><summary>Dedup verified — all three roots resolve to the same version</summary><pre>qurl-discord-bot@2.1.0\n├── @discordjs/rest@2.6.0\n├─┬ @discordjs/ws@1.2.3\n│ └── @discordjs/rest@2.6.0 deduped\n└─┬ discord.js@14.25.1\n  ├── @discordjs/rest@2.6.0 deduped\n  └── @discordjs/ws@1.2.3 deduped</pre>Single version in the tree; no risk of running a different minor than discord.js was tested against.</details> |

**Minor**

| # | Reviewer ask | Fix |
|---|---|---|
| `"attempt 2/1"` log awkwardness | Format or guard | Moved the over-budget log to its own branch, distinct message: `"BUDGET EXHAUSTED after 1 attempt(s), aborting"`. The in-budget path keeps the `attempt N/MAX` format. |
| `persistAndExit` signal handler can throw | Wrap in try/catch | Wrapped. `persistSession` can throw on disk-full / chmodSync EACCES / ENOENT-on-rename — the earlier code set `exiting=true` before the throw, stranding phase1 (subsequent signals debounced, `process.exit` never reached). try/catch logs and exits 1 cleanly. |
| Nonce-LRU sizing tied to freshness window | Document the coupling | Added: at 5 s freshness, 1k entries absorbs ~200 handoffs/sec before still-fresh nonces evict (allowing replay). Real rate is one-per-deploy; ~3 orders of magnitude over-provisioned. Future contributor knows the size must move with the freshness window. |
| Cold-fallback floor math | ~7 s is best-case, not bounded | Updated: best case ~7 s (standby's heartbeat fires immediately after active dies), worst case ~9 s (standby just renewed before active died, dead-active discovery one heartbeat cycle later). Both still well below the original ~15 s+ ceiling. |
| TOCTOU window in mode-enforcement | Effectively unreachable | Noted — local-fs scope, no remote attacker surface. Comment in `persistSession` already explains the mechanism; not adding more code for an unreachable concern. |
| Mode-bit test helper extraction | Not blocking | Skipped per reviewer's note. |
| SIGINT/SIGTERM handler unit test | Small follow-up | Skipped per reviewer's note. Worth filing as a discrete follow-up issue. |

**Test state**: 24/24 spike file, 1015/1015 full suite. Lint clean.

This round was the smallest yet — design-doc consistency tightening + log-format polish. The reviewer's "approve-with-nits" stands; this should close it out.

justin-layerv:
Round 5 feedback addressed in 0ca57fe.

| # | Severity | Reviewer ask | Fix |
|---|---|---|---|
| 1 | must-fix | `persistSession` race window between `writeFileSync` and `chmodSync` when `.tmp` is pre-existing | Best-effort `fs.unlinkSync(tmp)` BEFORE `writeFileSync`. Forces the create-path so `mode: 0o600` is honored; the chmodSync stays as belt-and-suspenders against the (vanishingly unlikely) race where another process recreates `.tmp` between unlink and write. Picked the unlink approach over `openSync('wx')` so a previous crashed run is recovered gracefully — the spike's a dev tool, hard-failure on stale `.tmp` would be more friction than safety. |
| 2 | nice-to-have | `connect()` timeout leaks the manager handle | Wrapped `Promise.race` in try/catch; on timeout, `manager.destroy({code: 1000})` fires before re-throw. No more in-flight WS leaking into `process.exit`. |
| 3 | nice-to-have | Discord's `s` is per-shard; `event_id` will collide across shards | Doc one-liner: at the sharding PR the producer-side `event_id` migrates to `${shard_id}:${s}`. Worker-side LRU shape doesn't change. Won't be a future surprise. |
| 4 | nit | `afterEach` ordering — inner cleanup runs before `restoreAllMocks` | Comment in tests noting the dependency: today no test spies on `fs.unlinkSync`, but if a future one does, move the unlink cleanup later or restore the specific spy explicitly. |
| 5 | nit | "at-most-once-per-consumer fanout" wording | Reworded to "each message is delivered to one worker; there's no natural affinity..." — says the same thing about the cache-coherence problem without misleading vocab. |
| 6 | nit | Watchdog backoff `min(..., 5s)` cap is dead code at MAX_ATTEMPTS=5 | Doc comment notes the cap is unreachable at the current attempt cap, kept in case a future contributor bumps `MAX_ATTEMPTS` and forgets to revisit. |

**Test state**: 24/24 spike, 1015/1015 full suite. Lint clean. No new tests this round — the changes are either secret-handling hardening (covered by the existing mode-overwrite test) or doc / comment edits.

Five review rounds; each one smaller than the last. This should close it out.

justin-layerv:
Round 6 addressed in 732e52a.

| # | Severity | Reviewer ask | Fix |
|---|---|---|---|
| 1 | must-fix | phase2's `setTimeout` missing `.unref()` | Both phase2 timers (30 s connect race + 3 s settling wait) now `.unref()` for parity with phase1. |
| 2 | optional | `/tmp` symlink-follow risk between unlink and writeFileSync | Closed at the directory layer. Module load creates `$TMPDIR/qurl-spike-$UID/` with `mkdirSync({recursive: true, mode: 0o700})` + a follow-up `chmodSync(0o700)` for the pre-existing-directory case; the session file lives inside that owner-only dir. `SPIKE_SESSION_FILE` env still overrides for tests. |
| 3 | optional | `classifyResult` JSDoc doesn't cover exit 1 / 130 / 143 | Hoisted the full exit-code table to the module-level header doc (0/1/2/3/130/143 with what produces each). |
| 4 | style nit | All-flags-set test could be `resumed === true short-circuits` | Left as the quadruple enumeration. The unreachable-state-as-guardrail framing is what the test is explicitly about per the round-3 comment update. |
| Doc | clarity | Mid-rotation window: new active signs `<new>` while old standby only knows `<old>`? | Stated explicitly: ECS's serial-replacement invariant (PR 14 pins `minimumHealthyPercent=50 / maximumPercent=100` at `desired_count=2`) means the standby is ALWAYS replaced first — new standby boots with both `<new>` and `<old>` before ECS moves on to the active. No bad window. |
| Doc | clarity | Watchdog backoff cap reachable at `MAX_ATTEMPTS≥6` | Existing comment already covers this; confirming the framing is intentional future-proof code. |

**Test state**: 24/24 spike, 1015/1015 full suite. Lint clean. Commit GPG-signed.

The new `qurl-spike-$UID/` directory creation runs at module load, which means the persistence-helper tests exercise it implicitly — if `mkdirSync` threw, the test `require` would fail. It doesn't. Tests pass clean.

PR #245, opened 2026-05-11
feat(discord): reserve flow_state audit events for event-shipper Pillar 1

justin-layerv:
## Why

This is **Phase 1.0** of the event-shipper redesign (`apps/discord/docs/zero-downtime-design.md`, merged in #240). Reserves the three audit-event names that PR 4's state-machine harness will emit, and documents the CloudWatch-side SLI computation contract those events feed.

**Why ship Phase 1.0 first**: every downstream event-shipper PR (PR 4 = state-machine harness, PRs 5/6/7-9 = per-command `await*` conversions, PR 10 = gateway producer, PR 11 = worker consumer) needs a stable observability surface to emit against. Locking the event names + meta schema in a single up-front PR prevents two classes of churn: AUDIT_EVENTS edits in every downstream PR, and CloudWatch metric-filter churn in `qurl-integrations-infra`.

## What's in this PR

Constants + docs + a contract test only. **No emissions are wired** (those land in PR 4); **no CloudWatch metric filters are created** (those land in `qurl-integrations-infra` once PR 4 is producing events in sandbox).

Three-stage rollout — **names reserved → emissions wired → filters lit** — keeps each layer reviewable in isolation. This is the same pattern the existing `AUDIT_EVENTS` entries follow: `GUILD_INSTALL` and `monitoring.tf`'s filter live together as a paired unit.

## Events reserved

| Constant | Literal | When emitted | Purpose |
|---|---|---|---|
| `FLOW_CREATED` | `flow_created` | `createFlow()` | SLI denominator (`total_flows`) |
| `FLOW_TRANSITION` | `flow_transition` | `transitionFlow()` | The canonical metric — `qurl_bot_flow_transition_total{stage_from,stage_to,result,terminal}` (design doc § Pillar 1) |
| `FLOW_DELETED` | `flow_deleted` | explicit `deleteFlow()` | SLI numerator (`completed_flows`). TTL-reaped flows deliberately do NOT emit |

Detailed comments in `constants.js` for each entry cover: when it fires, the meta schema (LOW-cardinality-safe-to-dimension vs HIGH-cardinality-forensic-only field distinction), and which SLI it feeds. Mirrors the existing comment discipline on `GUILD_INSTALL` / `INTERACTION_HANDLED` / `DEPENDENCY_AUTH_FAILURE`.

## SLI computation contract (new section in design doc)

The \"`silently_dropped_flows / total_flows`\" SLI from the design doc's existing table was previously underspecified — the doc named the numerator without saying how it's computed. Closed in this PR:

```
total_flows            = count(FLOW_CREATED)
completed_flows        = count(FLOW_DELETED)
silently_dropped_flows = total_flows - completed_flows
```

The \"computed as a difference\" framing intentionally sidesteps the async-TTL-reap problem (DDB TTL deletion is asynchronous, so a synchronous \"reaped\" signal would require a separate sweeper). It also captures every unclean-drop class — process crash mid-flow, worker hang, TTL reap of abandoned flow — under one metric, which is what the SLI is designed to measure.

**Asymmetry the SLI math relies on**: if a future change adds a \"delete-on-TTL-reap\" sweeper, it MUST emit a distinct event (e.g. `FLOW_REAPED`), not `FLOW_DELETED`. Pinned in both the inline `constants.js` comment AND the design-doc section so whoever writes that PR sees it.

## Tests

New `tests/constants-flow-state-events.test.js` (5 tests) locks:
- The three literal string values (CloudWatch metric filters match literals; a typo or rename silently breaks the filters with no compile-time signal).
- The `Object.freeze` guard remains in place.
- The lower_snake_case naming convention every new entry must follow.

Full suite still passes: **1020 / 1020 tests across 42 suites**.

## Explicitly out of scope

- **Queue / worker-dispatch / resume / clock-skew constants** — those land with PRs 10 / 11 / 13 respectively, paired with their own emissions. Reserving them now without consumers creates \"what is this?\" surface area without solving any actual churn problem.
- **Helper wrapper module** (e.g. `emitFlowTransition()`) — the existing pattern is direct `logger.audit(AUDIT_EVENTS.X, {...})` calls. Helper modules exist only where there's real auxiliary logic (`gateway-metrics.js` has `setInterval` timers, not just wrappers).
- **Terraform CloudWatch metric filters** — those live in `qurl-integrations-infra` and pair with PR 4's emissions, not this PR.

## Verification

- [x] `npx jest tests/constants-flow-state-events.test.js` — 5/5 pass
- [x] `npx jest` (full suite) — 1020/1020 pass across 42 suites
- [x] `npx eslint` clean on changed files
- [ ] CI on this PR

### DIFF
--- a/apps/discord/docs/zero-downtime-design.md
+++ b/apps/discord/docs/zero-downtime-design.md
@@ -125,7 +125,10 @@ sharding lands later).
 The state-machine harness `apps/discord/src/flow-state.js` (PR 4) provides
 `createFlow`, `loadFlow`, `transitionFlow` (DDB conditional `UpdateItem` with
 `version = :expected` for OCC), and `deleteFlow`. Every transition emits a
-metric `qurl_bot_flow_transition_total{stage_from,stage_to,result}` (PR 3).
+metric `qurl_bot_flow_transition_total{stage_from,stage_to,result,terminal}`
+(materialized by the event-shipper observability Phase 1.0 PR's audit-event
+reservation plus the matching CloudWatch metric filter in
+`qurl-integrations-infra`).
 
 **Application-level concurrency rule:** a user with an existing non-expired
 flow row in a given `(channel_id)` cannot start a second flow there. The second
@@ -656,11 +659,12 @@ by design, because DDB TTL deletion is asynchronous and a synchronous
 captures every unclean drop: process crash mid-flow, worker hang, TTL reap
 of an abandoned flow.
 
-PR 3 (event-shipper observability, Phase 1.0) reserves the
+The event-shipper observability Phase 1.0 PR reserves the
 `FLOW_CREATED` / `FLOW_TRANSITION` / `FLOW_DELETED` audit-event names in
-`apps/discord/src/constants.js`. PR 4 wires the emissions in the
-state-machine harness. The paired CloudWatch metric filters land in
-`qurl-integrations-infra` once PR 4 is producing events in sandbox.
+`apps/discord/src/constants.js`. The state-machine harness PR wires the
+emissions. The paired CloudWatch metric filters land in
+`qurl-integrations-infra` once the harness is producing events in
+sandbox.
 
 If a future change adds a "delete-on-TTL-reap" sweeper, it MUST emit a
 distinct event (e.g. `FLOW_REAPED`) — not `FLOW_DELETED` — to preserve
--- a/apps/discord/docs/zero-downtime-design.md
+++ b/apps/discord/docs/zero-downtime-design.md
@@ -637,6 +637,35 @@ deploy story. It's not SLO'd because we don't control it (Discord does), but
 we alert if it drops below 99% over a 24 h window — that signals a Discord-side
 change or our session-state staleness, both of which need investigation.
 
+### Flow continuity — computation contract
+
+The `silently_dropped_flows` numerator is computed as a difference, not
+emitted as a direct event:
+
+```
+total_flows            = count(FLOW_CREATED)
+completed_flows        = count(FLOW_DELETED)
+silently_dropped_flows = total_flows - completed_flows
+```
+
+Every `createFlow()` emits `FLOW_CREATED`; every explicit `deleteFlow()`
+(terminal stage, abort, admin cleanup) emits `FLOW_DELETED`. TTL-reaped
+flows — the silent-drop case the SLI is designed to catch — emit no event
+by design, because DDB TTL deletion is asynchronous and a synchronous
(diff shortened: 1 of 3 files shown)


justin-layerv:
**Round 1 review addressed (`32aa872`).**

The must-fix item and both polish items pulled in-PR; one PR-4-concern captured in the rollout-plan task notes.

| # | Item | Status |
|---|---|---|
| 1 | **(must-fix) Dimension-list inconsistency** — inline FLOW_TRANSITION comment named the materialized metric as `{stage_from, stage_to, result}` but the schema/PR body include `terminal` | ✅ Updated the materialized metric reference to `{stage_from,stage_to,result,terminal}`. Added the explicit "AND included in the materialized metric's dimension list above" note on the `terminal` payload-field documentation so future readers can't miss the intentionality. Synced the design doc's § Pillar 1 reference line for consistency. |
| 2 | **"PR N" refs are roadmap ordinals, not GitHub PR numbers** | ✅ Replaced with descriptive phrases (`"the state-machine harness PR"` / `"the event-shipper observability Phase 1.0 PR"`). Survives renumbering AND reads self-explanatory without cross-referencing the design doc. Applied to both the constants block AND my new design-doc computation contract section. |
| 3 | **Snake_case test scope ↔ description mismatch** — loop only iterates the 3 new keys but the description claims a global-convention check | ✅ Widened to iterate `Object.entries(AUDIT_EVENTS)` — now a real regression guard for every current and future event name. Added two more sub-checks: (a) literal length ≤ 80 (CloudWatch metric filter pattern length is bounded), and (b) the JS key is UPPER_SNAKE_CASE so `flowCreated: 'flow_created'` (literal passes, key doesn't) gets caught. |
| 7 | **PII posture for `flow_id`** — composite encoding contains raw Discord user IDs | 📌 Captured for the state-machine harness PR (task #21). The HIGH-cardinality / forensic-only tagging prevents metric-cost explosion, but ANY log line emitting flow_id carries user IDs into CloudWatch Logs. That PR's review must verify the user-ID exposure shape against the bot's PII posture — either keep raw (current plan, forensic-only) or hash/truncate the user_id segment before emitting. |
| 4 | **`test.each` DRY** for the three literal-string tests | 📌 Deferred per reviewer's "Optional, not blocking." Current shape is 11 readable lines; compressing saves ~5 lines at the cost of indirection. |
| 5, 6 | **Design observations** (asymmetric-reap is right, HIGH/LOW cardinality tagging is good) | 💬 Acknowledged. No asks. |

Local verification: 1020/1020 tests pass across 42 suites.

PR #246, opened 2026-05-11
feat(discord): state-machine harness for flow_state DDB table

justin-layerv:
## Summary

Wire the four lifecycle methods the event-shipper architecture needs to drive multi-step flows across gateway handoffs and SQS-worker redeliveries — `createFlow` / `loadFlow` / `transitionFlow` / `deleteFlow` — backed by the `flow_state` DDB table.

This is **PR 4 of the zero-downtime upgrade rollout**. PRs 5–9 will convert `/qurl revoke`, `/qurl setup`, and `/qurl send` from in-memory `await*` interactions to flow_state, so a deploy mid-flow no longer drops the user.

## Why now

Today every multi-step Discord interaction ("click this button" → "fill out this modal" → "confirm send") is held in the gateway process's memory via `client.awaitMessageComponent`/`awaitModalSubmit`. A rolling deploy that swaps the WebSocket tier mid-flow drops that state, the bot stops responding, and the user is left staring at a button that no longer works.

The harness here gives us a durable substrate so the next deploy can stop, restart, and resume without losing in-flight interactions.

## Correctness primitives

| Primitive | Mechanism |
|-----------|-----------|
| Exactly-once advancement under SQS at-least-once | OCC: `version + 1` write gated by `ConditionExpression: #v = :expected` |
| Block transitions on logically-expired rows (SLI-honesty guard) | `ConditionExpression: #e >= :now` on Update — CCFE maps to `result: 'not_found'` via the recheck logic |
| No silent TTL orphans | Writer asserts `expires_at` is a finite-integer Number, strictly in the future; DDB TTL only reaps Number-typed attributes |
| Honest "is this flow alive?" | Reader returns `null` for `now > expires_at` even before async reap (~48h lag) |
| Fail-safe on corrupted rows | Reader returns `null` + warns for missing/non-numeric `expires_at` (vs. silently returning an immortal row). Recheck path has symmetric warn |
| SLI math integrity | `silently_dropped = count(FLOW_CREATED) - count(FLOW_DELETED)` — DELETE is conditional on `attribute_exists`, so redelivery doesn't double-emit FLOW_DELETED and inflate the numerator |
| Mandatory payload encryption | `encryptStrict(JSON.stringify(payload))` — fails closed if `KEY_ENCRYPTION_KEY` is unset (payload transitively carries `attachment_url` + recipient lists) |
| No flow_id drift | Canonical `parseFlowId`/`buildFlowId` in `src/flow-id.js` locks the mixed-separator format (`:` allowed in `shard_id`, `#` forbidden everywhere); `createFlow` validates structure at entry |
| Audit-shape honesty | `terminal` forced to `false` on non-success transitions; `extended` true iff the new `expires_at` is strictly greater than the prior (genuine extension, not shortening) |

## What does NOT change

- No callers wired yet — flows still go through `await*` in this PR. PRs 5–9 do the conversions one flow at a time.
- No new audit event names — `FLOW_CREATED` / `FLOW_TRANSITION` (with `terminal`, `extended`, `version`) / `FLOW_DELETED` (with `reason`) landed in #245.
- No new infra — `flow_state` table is provisioned by qurl-integrations-infra #504.
- No new IAM — the bot's task role already covers data-plane DDB verbs for every table in the module.

## Deferred / explicit follow-up

- **DDB client setup is duplicated** from `src/store/ddb-store.js` rather than extracted. Marked `TODO(PR 12)`: consolidate once a third consumer (the gateway-tier RESUME table) arrives.
- **2 DDB calls per transition on the happy path** (pre-read for `stage_from` + Update). Tracked as #253 — collapse to a single call via `ReturnValues: 'ALL_OLD'` + `ReturnValuesOnConditionCheckFailure: 'ALL_OLD'` (also eliminates the recheck-Get).
- **No integration test** against a real DDB table — that lands with the first caller conversion (PR 5) where we can exercise the actual flow end-to-end.

## Test plan

- [x] `flow-id.test.js`: 28 tests covering roundtrip, every rejection surface (empty / non-string / contains `#`), and `:` preservation inside `shard_id`
- [x] `flow-state.test.js`: 65 tests covering: happy path, idempotent recreate (OCC conflict on Put), OCC conflict-vs-not-found discrimination on transition (via post-failure recheck), TTL writer guard (string / float / NaN / Infinity / past / equal-to-now rejected), TTL reader contract (logical expiry + grace), fail-safe on missing/non-numeric `expires_at` on both read and recheck paths, payload encryption shape (including null/undefined asymmetry between create and transition), at-most-once `FLOW_DELETED` emission on redelivery, payload corruption resilience in `loadFlow`, decrypt-throws propagation (fail-loud on KEK misconfig), `terminal` forced to false on non-success result paths, `extended` true only when `new > prior` (shortening, equal, and missing-prior all emit false), and entry-point `parseFlowId` rejection on createFlow
- [x] Full Jest suite: **1113 passing** (~89 new tests across the two suites; 24 net-new beyond the original `flow-state` plan after 9 rounds of code review)
- [ ] First caller wiring (PR 5) will exercise this against a real sandbox DDB table

### DIFF
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -268,7 +268,20 @@ const AUDIT_EVENTS = {
   //                   varies per flow type — revoke is shorter than
   //                   send). LOW-cardinality (boolean) — safe to
   //                   dimension on AND included in the materialized
-  //                   metric's dimension list above.
+  //                   metric's dimension list above. FORCED TO false
+  //                   on non-success results (not_found, conflict,
+  //                   error) — the transition didn't advance, so
+  //                   nothing terminal happened, so the audit is
+  //                   honest by construction. Consumers can still
+  //                   slice `count_by(terminal=true)` safely.
+  //   - `extended`:   bool. True if the transition extended/reset
+  //                   the row's expires_at (via the `set_expires_at`
+  //                   option). LOW-cardinality (boolean) — safe to
+  //                   dimension on. Forensic-only today; lets an
+  //                   incident query answer "did this transition
+  //                   bump the deadline?" without re-reading the
+  //                   row. Not currently used as a metric filter
+  //                   dimension.
   //   - `flow_id`:    HIGH-cardinality. Forensic-query ONLY — same
   //                   posture as FLOW_CREATED.
   FLOW_TRANSITION: 'flow_transition',
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -274,14 +274,21 @@ const AUDIT_EVENTS = {
   //                   nothing terminal happened, so the audit is
   //                   honest by construction. Consumers can still
   //                   slice `count_by(terminal=true)` safely.
-  //   - `extended`:   bool. True if the transition extended/reset
-  //                   the row's expires_at (via the `set_expires_at`
-  //                   option). LOW-cardinality (boolean) — safe to
-  //                   dimension on. Forensic-only today; lets an
-  //                   incident query answer "did this transition
-  //                   bump the deadline?" without re-reading the
-  //                   row. Not currently used as a metric filter
-  //                   dimension.
+  //   - `extended`:   bool. True iff the transition GENUINELY
+  //                   extended the row's expires_at (the new value
+  //                   is strictly greater than the prior). A
+  //                   set_expires_at that shortens, equals, or
+  //                   leaves the value untouched emits false —
+  //                   so `count_by(extended=true)` is a faithful
+  //                   "this transition bumped the deadline forward"
+  //                   count and not a "set_expires_at was passed"
+  //                   count. Also false on non-success transitions
+  //                   (nothing extended) and on rows whose prior
+  //                   expires_at was missing/corrupted (no honest
+  //                   baseline to extend FROM). LOW-cardinality
+  //                   (boolean) — safe to dimension on. Forensic-
+  //                   only today; not currently used as a metric
+  //                   filter dimension.
   //   - `version`:    integer. On success: the row's NEW version
   //                   after the OCC bump. On non-success (conflict,
   //                   not_found, error): the version the caller
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -282,6 +282,18 @@ const AUDIT_EVENTS = {
   //                   bump the deadline?" without re-reading the
   //                   row. Not currently used as a metric filter
   //                   dimension.
+  //   - `version`:    integer. On success: the row's NEW version
+  //                   after the OCC bump. On non-success (conflict,
+  //                   not_found, error): the version the caller
+  //                   expected (i.e., `expectedVersion`). Lets a
+  //                   forensic query correlate retries by attempt
+  //                   identity ("which attempt won version 5?")
+  //                   without needing the live row — important
+  //                   because already-deleted flows can't be
+  //                   JOINed against the live table. LOW-cardinality
+  //                   per-flow (bounded by transitions-per-flow,
+  //                   ~10 in practice); NOT a metric dimension —
+  //                   forensic-only field.
   //   - `flow_id`:    HIGH-cardinality. Forensic-query ONLY — same
   //                   posture as FLOW_CREATED.
   FLOW_TRANSITION: 'flow_transition',
--- a/apps/discord/src/flow-id.js
+++ b/apps/discord/src/flow-id.js
@@ -1,0 +1,74 @@
+// flow-id — canonical parse/build for the shard-aware composite key
+// used by `flow_state.flow_id` (see qurl-integrations-infra
+// modules/qurl-bot-ddb flow_state table).
+//
+// Format: `<shard_id>#<guild_id>#<channel_id>#<user_id>`
+//
+// Asymmetric separator rules — `shard_id` may contain `:` (the
+// canonical shard encoding is `k:n`, e.g. `0:1` for single-shard,
+// `3:8` for shard 3 of 8); `:` is therefore NOT used as the
+// flow_id-level separator. The flow-level separator is `#`, which
+// is FORBIDDEN inside every component (including shard_id). This
+// keeps split-on-first-`#` parseable AND lets shard_id retain its
+// natural `k:n` shape.
+//
+// Why centralize this in one module instead of inlining the join /
+// split at every callsite (handler, flow-state, future leader-
+// election lookup, audit-log forensic queries):
+//
+//   - A handler computing `flow_id = `${shard}#${guild}#${chan}#${user}``
+//     and a worker computing the same on its side will drift the
+//     first time someone changes the separator convention. The
+//     OCC-on-flow_id contract relies on byte-identical keys across
+//     both sides; the canonical build/parse pair removes the drift
+//     surface.
+//   - Validation lives in one place. A `guild_id` containing `#`
+//     (impossible per Discord's snowflake encoding today, but cheap
+//     to defend against if Discord ever ships a non-numeric guild
+//     identifier) would silently produce a malformed flow_id and a
+//     parseFlowId() at the other end would yield wrong fields. Reject
+//     at build time, not at deserialization time.
+
+const SEP = '#';
+
+// Snowflake-ish components MUST NOT contain `#`. We intentionally do
+// NOT check the colon — `:` is legal inside `shard_id` and irrelevant
+// for the other components.
+function assertNoSep(value, fieldName) {
+  if (typeof value !== 'string' || value.length === 0) {
+    throw new TypeError(`flow-id: ${fieldName} must be a non-empty string (got ${typeof value === 'string' ? 'empty string' : typeof value})`);
+  }
+  if (value.includes(SEP)) {
+    throw new TypeError(`flow-id: ${fieldName} must not contain '${SEP}' (got ${JSON.stringify(value)}); the '#' character is reserved as the flow_id component separator`);
+  }
+}
+
+// Build a flow_id from its four components. Order is fixed:
+// shard → guild → channel → user. Callers should treat the returned
+// string as opaque; do not parse it inline — use parseFlowId().
+function buildFlowId({ shard_id, guild_id, channel_id, user_id }) {
+  assertNoSep(shard_id, 'shard_id');
+  assertNoSep(guild_id, 'guild_id');
+  assertNoSep(channel_id, 'channel_id');
+  assertNoSep(user_id, 'user_id');
+  return `${shard_id}${SEP}${guild_id}${SEP}${channel_id}${SEP}${user_id}`;
+}
+
+// Inverse of buildFlowId. Returns `null` for any malformed input
+// (wrong type, wrong shape, empty component) so callers can decide
+// whether to log + skip or throw. Throwing on a malformed flow_id
+// pulled from DDB would break the audit/forensic path — a
+// best-effort parse + null sentinel preserves that path.
+//
+// Split is exact: exactly 3 `#` separators produces 4 components.
+// More or fewer is rejected (returns null).
+function parseFlowId(flow_id) {
+  if (typeof flow_id !== 'string' || flow_id.length === 0) return null;
+  const parts = flow_id.split(SEP);
+  if (parts.length !== 4) return null;
+  const [shard_id, guild_id, channel_id, user_id] = parts;
+  if (!shard_id || !guild_id || !channel_id || !user_id) return null;
+  return { shard_id, guild_id, channel_id, user_id };
+}
+
+module.exports = { buildFlowId, parseFlowId };
--- a/apps/discord/src/flow-state.js
+++ b/apps/discord/src/flow-state.js
@@ -101,6 +101,12 @@ if (!AWS_REGION) {
 
 const TABLE_NAME = `${TABLE_PREFIX}flow-state`;
 
+// Test escape hatch: `DDB_TEST_ENDPOINT=http://localhost:8000`
+// points the client at DynamoDB Local (or any HTTP-compatible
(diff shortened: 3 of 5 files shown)


justin-layerv:
Addressed the review feedback in a02ef7d:

### Fixed now (7 items)

1. **Stage-OCC clarification** — `transitionFlow` docstring now spells out that OCC gates on `version` only, not `stage`; callers wanting strict stage transitions must assert `stage_from` themselves.
2. **Reader/writer symmetry on `expires_at`** — `loadFlow` now matches `assertExpiresAt`'s float rejection (`Math.floor(x) !== x`). A regression writer's `1234567890.5` row no longer round-trips undetected. New test covers it.
3. **Test-authoring note** — comment near the module-load env throws points future test authors at `tests/flow-state.test.js`'s pattern so they don't hit a confusing module-load stack trace.
4. **`undefined` vs `null` payload asymmetry** — documented in `createFlow`'s docstring (both persist as null on create; on transition `undefined` preserves and `null` clears).
5. **Recheck failure no longer silent** — `transitionFlow` now `logger.warn`s when the post-CCFE recheck-Get itself throws (e.g. DDB availability blip). Conservative `result: 'conflict'` fallback still applies but the signal is no longer hidden. New test covers it.
6. **Negative `grace_seconds`** — documented as accepted (skews "expired" earlier; permissive but explicit).
7. **`getCalls` closure scope** — moved inside the TTL-race test so a future copy-paste can't bleed state between tests.

Also documented the **audit-emit-on-rethrow contract** in `transitionFlow`'s docstring — N retries of a transient error legitimately produce N `result: 'error'` events. The SLI math (`silently_dropped = count(FLOW_CREATED) - count(FLOW_DELETED)`) doesn't read the error bucket, so this is by design.

### Deferred (filed as follow-up)

- **Collapse `transitionFlow` into a single round trip** via `ReturnValuesOnConditionCheckFailure: 'ALL_OLD'` → #253. Reviewer framed this as "worth a follow-up at minimum"; a performance-only refactor is the wrong scope to bundle with the harness's first ship. The current 2-call shape is correct (just chattier than necessary) and the test surface is identical to what the optimized version will need.

### Test counts

- `flow-id.test.js`: 28 (unchanged)
- `flow-state.test.js`: 43 (was 41 — +2 for float-rejection symmetry and recheck-failure warn)
- Full suite: **1091 passing**

justin-layerv:
Round-3 review: 3 separate claude-review runs surfaced 10 substantive items. Addressed 9 inline in ee6e5e7; 1 remains deferred to #253.

### Closed (foot-guns / correctness)

1. **`assertExpiresAt` accepts `0` and negative** — tightened to `expires_at > nowEpochSeconds()`. The forgotten-`+ ttl_seconds` foot-gun (writing `expires_at = nowEpochSeconds()`) now throws at the call site instead of silently producing a born-expired row that inflates `silently_dropped_flows`. The bare `> 0` floor wouldn't have caught this — `nowEpochSeconds()` evaluates to ~1.7e9.
2. **`grace_seconds` not type-checked** — added `Number.isFinite` guard. `grace_seconds: 'forever'` would have produced `expires + 'forever'` → NaN → expired-treated-as-alive.
3. **`flow-state` didn't actually use `flow-id`** — `createFlow` now calls `parseFlowId(flow_id)` and rejects malformed keys at the entry point. Other methods trust the caller (a malformed key hits `not_found`). Module header documents the contract.

### Closed (audit-shape honesty)

4. **`terminal: true` on non-success audits** — `transitionFlow` now forces `terminal: false` on `not_found`/`conflict`/`error` paths. If the transition didn't advance, nothing terminal happened. 3 new regression tests pass `terminal: true` on each non-success path and assert the audit emits `false`.
5. **`extended: bool` missing from FLOW_TRANSITION audit** — added. `constants.js` FLOW_TRANSITION payload-fields doc updated with the new field + the forced-terminal-on-non-success semantic.

### Closed (API clarity)

6. **`extend_expires_at` name implied monotonicity not enforced** — renamed to `set_expires_at` (no existing callers). Docstring explicitly notes "set, not extend" and points monotonic-only callers at `loadFlow().expires_at` first.
7. **`createFlow` redelivery with mutated payload** — strengthened comment to note payload mutation is silently discarded (the contract is "create-once for this flow_id"; payload evolution is `transitionFlow`'s job).

### Closed (cleanup)

8. **`createFlow` called `nowEpochSeconds()` twice** — compute once, reuse for `created_at` + `updated_at`. Tested.
9. **Redundant null guards in `encryptPayload`/`decryptPayload`** — kept `encryptPayload`'s (load-bearing for DDB attribute shape consistency; annotated why) and dropped `decryptPayload`'s redundant first guard.
10. **`loadFlow` always `ConsistentRead: true`** — documented the choice in the docstring with the canonical use case + an "add a parameter when an orthogonal-read caller arrives" note.

### Deferred (#253)

- **Collapse `transitionFlow` into a single round trip** via `ReturnValuesOnConditionCheckFailure: 'ALL_OLD'`. Reviewer in round 1 framed this as "worth a follow-up at minimum"; performance-only refactor remains the wrong scope for the harness's first ship.

### Test counts

- `flow-id.test.js`: 28 (unchanged)
- `flow-state.test.js`: 53 (+10 from round 2)
- Full suite: **1101 passing**

All static `expires_at: 1000` fixtures replaced with a `futureExpiry()` helper. All FLOW_TRANSITION audit assertions updated to include the new `extended: bool` field.

justin-layerv:
Round-4 polish addressed in 64ff53e:

1. **`assertExpiresAt` called `nowEpochSeconds()` twice** → cached the value. Eliminates the tiny race where the comparison and the error-message `now` could straddle a second boundary.
2. **`decryptPayload` propagates decrypt throws (intentional)** → pinned the intent in a comment so a future contributor doesn't "helpfully" add a try/catch and silently degrade a KEK misconfig into `payload: null`. Symmetric with `encryptStrict`'s fail-closed posture.
3. **Redelivery debug breadcrumb** → `createFlow` now emits `logger.debug('flow-state.createFlow: idempotent redelivery (row already exists)', { flow_id })` on the `{ created: false }` branch. Debug-level (not warn) keeps legitimate-redelivery traffic out of normal logs but surfaces under `LOG_LEVEL=debug` for triage.
4. **Inline TODO(#253) on `transitionFlow`** → calls out the deferred round-trip collapse with the explicit rationale (performance-only refactor, wrong scope for the harness's first ship). Closes the "either fix or point to deferral" loop.
5. **Decrypt-throws test gap** → new test `loadFlow propagates decrypt-side throws (fail-loud on KEK misconfig)` locks in the behavior so a future try/catch in `decryptPayload` gets caught at PR time.

### Test counts

- `flow-id.test.js`: 28
- `flow-state.test.js`: 55 (+2 from round 3)
- Full suite: **1103 passing**

justin-layerv:
Round-5 review addressed in 6b1eae4:

### Correctness
1. **Post-CCFE recheck conflated "row exists" with "row alive"** → recheck now projects `expires_at` and applies the same logical-expiry filter as `loadFlow`. A row that slipped past `expires_at` between pre-read and Update will report `not_found` (caller picks the right retry strategy), not `conflict`.

### Audit enrichment
4. **`version` missing from FLOW_TRANSITION audit** → added. On `success` it's the new version (post-OCC-bump); on `not_found`/`conflict`/`error` it's `expectedVersion`. Lets forensic queries correlate retries by attempt identity without JOINing the live table — critical because already-deleted flows can't be JOINed. Bound is per-flow (~10 max transitions), so LOW-cardinality enough to live in the audit. `constants.js` docs updated.

### Tests
2. **`payload: null` on transitionFlow** → new test asserts that null *clears* (the documented asymmetry with createFlow), via UpdateExpression containing `#p = :payload` and `:payload` being `null`.
5. **Terminal-on-success symmetry** → new negative-control test asserts that `terminal: false` on a successful transition stays `false`, companion to the existing forced-to-false tests on non-success paths.

### Documentation polish
3. **`set_expires_at` TOCTOU** → comment noting the "strictly in the future" check fires at call time, not DDB-write time. Non-fatal with sane TTLs; pinning the caveat anyway.
7. **`DDB_TEST_ENDPOINT`** → one-line note pointing at `localhost:8000` for DynamoDB Local. Was previously undocumented outside the code.

### Acknowledged, no change
6. Encrypt mock is pass-through. The `.startsWith('enc:v1:')` assertion is the right proxy for "encryption happened".
8. `__TABLE_NAME` export convention. Codebase has no unified pattern (logger.js uses `__testExports`, ddb-store.js exports nothing test-only). Left as-is.

### Test counts
- `flow-id.test.js`: 28
- `flow-state.test.js`: 58 (+3 from round 4)
- Full suite: **1106 passing**

justin-layerv:
Round-6 polish in abd3743:

1. **`ProjectionExpression: 'flow_id, expires_at'` redundant** → trimmed to just `'expires_at'`. Partition key is always returned.
2. **Local `futureExpiry` shadowed the outer helper** → renamed to `expiresAt`.
3. **`createFlow` redelivery returns no version** → JSDoc now explicit: a caller needing the existing row's version after `{ created: false }` must follow up with `loadFlow()`. Pre-empts a sharp edge for PR 5+ consumers.
5. **`created_at` invariance test gap** → new regression guard. `transitionFlow`'s UpdateExpression never writes `created_at`; a future refactor adding `#c = :created_at` will fail at PR time.

### Acknowledged, no change
4. **`res?.Attributes?.version ?? expectedVersion + 1` defensive fallback** — reviewer noted unreachable in practice with `ReturnValues: 'UPDATED_NEW'`; "fine as-is, just noting it isn't covered by a test." Skipped.

### Test counts
- `flow-id.test.js`: 28
- `flow-state.test.js`: 59 (+1 from round 5)
- Full suite: **1107 passing**

justin-layerv:
Round-7 caught a real correctness bug. Addressed in a58d4d7.

### 🐛 Bug fix (the must-fix from the review)

`transitionFlow`'s `ConditionExpression` was missing the `#e >= :now` guard. A logically-expired row (DDB TTL reap is async, ~48h lag) could pass `attribute_exists` + version-match and have its `stage` rewritten by `UpdateCommand` — even though `loadFlow` would then return null for it. The next `FLOW_DELETED` would never fire and `silently_dropped = count(CREATED) - count(DELETED)` would inflate on every transition-of-expired-row.

The asymmetric guard the post-CCFE recheck and `loadFlow` both already had — now extended to the Update itself.

- Added `#e >= :now` to `ConditionExpression`. CCFE on expired-row now fires at write-time; recheck logic maps to `result: 'not_found'`.
- Used `>=` (not `>`) to stay consistent with `loadFlow`'s "alive at boundary" semantic.
- Cached `now` once via `updateNow` so the condition and `:updated_at` can't straddle a second boundary.
- Symmetric observability: recheck now warns on corrupted-`expires_at` rows, matching `loadFlow`'s warn (operator greps the same string across both paths).

### ⚠️ Tightened `grace_seconds >= 0`

Reviewer-flagged inconsistency: rest of the module is type-check tight; permissive negatives let a `-3600` typo silently shorten lifetimes by an hour. Docstring + error message + tests updated.

### 📋 Test gap fills

- `undefined payload persists as null (symmetric with explicit null)` — locks the documented semantic.
- `rejects negative grace_seconds (typo guard)`.
- `post-CCFE recheck warns when row has corrupted expires_at and reports not_found` — regression guard for the symmetric warn.

### ✏️ Doc cleanup

- Dropped the admin-CLI claim from `__TABLE_NAME` comment. The `__` prefix should signal test-only; a real third-caller export gets designed when that caller actually arrives.

### Test counts
- `flow-id.test.js`: 28
- `flow-state.test.js`: 62 (+3 from round 6)
- Full suite: **1110 passing**

justin-layerv:
Round-8 fix-up in 4e95b6a, leading with the merge-blocker:

### 🔧 Merge-blocker fix: `extended` reflects genuine deadline extension

`extended: true` was emitted whenever `set_expires_at !== undefined`, even when the new value SHORTENED the lifetime. The field name reads as "the deadline was extended", so a forensic `count_by(extended=true)` was over-counting shortenings as extensions.

- Widened pre-read `ProjectionExpression` to include `expires_at`.
- Now computes: `extended = set_expires_at !== undefined && priorExpires is finite-integer && set_expires_at > priorExpires`.
- Strict `>` — no-op rewrites and shortenings emit `false`.
- Missing/corrupted prior `expires_at` → `false` (no honest baseline; corruption signal already surfaces via the recheck warn).
- `constants.js` FLOW_TRANSITION docs rewritten to spell out the new semantic + when `false` is correct.
- 3 new tests: shortening → false, equal → false, missing-prior → false. Plus the existing happy-path test updated to use prior+new values.

### 📝 Doc polish (intentional-asymmetry guards)

Both are one-liner notes that a future reader might try to "fix" but shouldn't:
- `assertExpiresAt`: writer rejects `expires_at == now` while readers (loadFlow + recheck) treat `==` as alive. Writer is intentionally stricter — relaxing would race the next-second boundary into born-expired territory.
- `transitionFlow` `set_expires_at` docstring: spell out that `#e >= :now` blocks reviving an expired row even with a fresh-future `set_expires_at`. Resurrecting expired rows would inflate the silently_dropped math. Callers needing fresh lifecycle after expiry should `createFlow` again.

### Skipped per reviewer's framing
- **Payload size pre-check** — depends on caller-side confidence, no current flow types known to balloon.
- **Audit-emit atomicity** — reviewer noted "probably not worth restructuring".
- **Test env pollution** — Jest worker isolation makes it non-blocking.

### Open question on #253
Reviewer's note "would `ReturnValuesOnConditionCheckFailure` also eliminate the recheck-Get?" — yes, and #253's description already covers this: *"CCFE path → 1 round trip (failure response distinguishes conflict vs not_found by err.Item presence)"*. Same follow-up.

### Test counts
- `flow-id.test.js`: 28
- `flow-state.test.js`: 65 (+3 from round 7)
- Full suite: **1113 passing**

PR #256, opened 2026-05-12
feat(discord): convert /qurl revoke to flow_state harness

justin-layerv:
## Summary

First command converted off in-memory `awaitMessageComponent` and onto the DDB-backed `flow_state` harness from #246. This is **Pillar 1 of the zero-downtime upgrade rollout** (see `docs/zero-downtime-design.md`) and the smallest at-risk surface — `/qurl revoke` only has a single 60-second click window, so converting it first keeps the blast radius minimal while we shake out the dispatcher pattern that PR 6 (`/qurl setup`) and PR 7–9 (`/qurl send`) will reuse.

## Business context

Today, a rolling deploy mid-revoke drops the user's interaction silently — the `awaitMessageComponent` Promise on the old process can't be resumed on the new one. We have live users in prod (per `project_live_customers` memory) and even though we're not charging yet, "click a thing in Discord and have it not work" is a reputation-damaging UX. The 15-PR rollout chain closes this for every `await*` site; this PR closes the first one.

## What changed

- **`src/flow-dispatch.js`** (new): routes `MessageComponent` / `ModalSubmit` interactions by `customId` prefix. The flow_id is derived from trusted interaction context (`user.id` + `channelId` + `SHARD_ID`) — **never from `customId`**, which is replayable client input. Stage-mismatch and missing-row cases both reply with a recoverable "superseded — run the command again" message. Universal try/catch around the handler call so a throw post-side-effect surfaces as a user-visible error rather than an unhandledRejection.

- **`src/commands.js` `handleRevoke`**: writes a flow row (`stage='awaiting_revoke_select'`, payload=null, TTL=60s) and renders the select menu. Returns immediately — no inline `awaitMessageComponent`. Concurrent revokes are **superseded** (existing flow row gets `admin_cleanup`'d, new one created). This is a deliberate exception to the design-doc "one flow per channel" rule: the revoke menu is a stateless listing, blocking the second invocation would just confuse users who don't remember whether they cancelled the prior menu. The `/qurl send` PRs will hold the block-second-flow rule.

- **`src/commands.js` `handleRevokeSelect`**: dispatcher-side handler. Runs `getGuildApiKey` + `deleteFlow` in parallel (independent reads). `deleteFlow` is the at-most-once dedup primitive — only the worker whose conditional delete returns `{ deleted: true }` proceeds to the qURL API call.

- **`src/index.js`** listener splits dispatch by interaction kind: chat-input / autocomplete → `handleCommand`; component / modal → `handleFlowInteraction`.

- **`src/config.js`**: adds `SHARD_ID = process.env.SHARD_ID || '0:1'` as the single source of truth for the flow_id's shard component.

- **`tests/setup-env.js`** (new): sets `DDB_TABLE_PREFIX` + `AWS_REGION` globally so source modules' fail-fast module-load guards don't throw across the test suite.

## Trust model

`customId` routes the event class; **identity** (user, channel, shard) comes from interaction context. A hostile client editing `customId` to spoof "user-VICTIM" cannot reach the victim's flow row. Pinned by `flow-dispatch.test.js`.

## Test plan

- [x] `tests/flow-dispatch.test.js` — routing happy path, stage-mismatch, missing-row, unknown customId, DM context, handler-throw safety net, customId-spoofing rejection (13 cases)
- [x] `tests/commands-comprehensive.test.js` — slash-command path (menu rendered, supersede behavior, error when both create attempts race) + dispatcher-side `handleRevokeSelect` path (revoke executes, dedup via deleteFlow=false, apiKey-missing, partial revoke)
- [x] `revoke-filter.test.js` — unchanged, still passes (hits real SQLite layer)
- [x] `qurl-send-back-half.test.js` — `revokeAllLinks` direct unit tests unchanged
- [x] All 1133 tests pass, lint clean, coverage 82.23/75.02/81.99/83.83 (clear of 78/68/78/78 floor)

## Deferred / follow-up

- `INTERACTION_HANDLED` audit event is not emitted from the dispatcher path. The event's documented contract (`src/constants.js:152`) keys on `command_name`, which component events don't have. A separate flow-handler observability event (or extending the existing audit event's payload contract) belongs in PR 6 once the second dispatcher consumer lands.
- `nowEpochSeconds` (in `flow-state.js`) and the apiKey-resolution pattern (in `handleCommand`) are both candidates for extraction into shared utilities, but extracting them in this PR would expand the diff into modules unrelated to the revoke conversion. Worth doing as a separate cleanup PR once PR 6 introduces a third callsite.
- Supersede path issues 2 DDB calls (`deleteFlow` + `createFlow`). Could be 1 if `flow-state.createFlow` grew a `force: true` option, but that's a flow-state contract change beyond the revoke conversion's scope.

### DIFF
--- a/apps/discord/docs/zero-downtime-design.md
+++ b/apps/discord/docs/zero-downtime-design.md
@@ -136,6 +136,22 @@ flow row in a given `(channel_id)` cannot start a second flow there. The second
 cancel it first."` This avoids needing a GSI on the flow table and keeps the
 state model simple.
 
+**Per-command override matrix.** The block-second-flow rule above is the
+default. Two commands carry intentional exceptions because their UX shape
+makes blocking the wrong call:
+
+| Command | Behavior on existing flow | Rationale |
+|---|---|---|
+| `/qurl revoke` | **Supersede** (admin_cleanup + recreate) | The menu is a stateless listing of recent sends; an admin who can't remember whether they cancelled the prior dropdown shouldn't be told "finish your existing dropdown first." Re-running shows a fresh menu and the orphan menu's selection lands on `loadFlow → null → "superseded"`. |
+| `/qurl setup` | Block (default) | Multi-step OAuth / API-key paste flow with in-progress state; the user should finish or cancel the active step. |
+| `/qurl send` | Block (default) | Multi-step send flow (recipient picker, attachment capture, confirm); abandoning mid-flow because a duplicate `/qurl send` slipped through would lose user input. |
+
+The supersede semantics are **enforced at the harness level** by `deleteFlow`'s
+conditional `stage = :expected` gate (see `apps/discord/src/flow-state.js`).
+A revoke command can only supersede a revoke flow; it cannot accidentally
+admin_cleanup a sibling setup or send flow that happens to share the same
+`flow_id` keying.
+
 **MESSAGE_CREATE handler.** Today the file-drop step uses
 `captureChannel.awaitMessages({...})` on a DM channel. Under the redesign,
 **workers** subscribe to a `MESSAGE_CREATE` event class from the queue. The
--- a/apps/discord/jest.config.js
+++ b/apps/discord/jest.config.js
@@ -1,6 +1,11 @@
 module.exports = {
   testEnvironment: 'node',
   testMatch: ['**/tests/**/*.test.js'],
+  // Global env setup — sets DDB_TABLE_PREFIX + AWS_REGION before any
+  // test file loads, so source modules with fail-fast module-load
+  // guards (flow-state, ddb-store) can be required without throwing.
+  // See tests/setup-env.js for the rationale.
+  setupFiles: ['<rootDir>/tests/setup-env.js'],
   // Note: `restoreMocks` is NOT set globally — qurl-send-state-machine.test.js
   // and others rely on jest.spyOn results persisting across tests within a
   // describe (the spies are set up in module-level scope and tests assert
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -31,6 +31,8 @@ const { requireAdmin } = require('./utils/admin');
 const { signQurlOAuthState } = require('./utils/qurl-oauth-state');
 const { deleteLink, getResourceStatus } = require('./qurl');
 const { downloadAndUpload, reUploadBuffer, mintLinks, uploadJsonToConnector, isAllowedSourceUrl } = require('./connector');
+const { createFlow, deleteFlow } = require('./flow-state');
+const { flowIdForInteraction, registerFlow } = require('./flow-dispatch');
 
 // Max tokens the QURL API allows per resource. When exceeded, a new
 // resource must be created (re-upload) to get a fresh token pool.
@@ -2593,6 +2595,25 @@ function formatRevokeDescription(s) {
   return truncate(base);
 }
 
+// Single non-terminal stage between createFlow and deleteFlow.
+// Stage name is the source of truth for the dispatcher routing
+// table (registerFlow at the bottom of the file) — changing it is
+// a single edit.
+const REVOKE_STAGE_AWAITING_SELECT = 'awaiting_revoke_select';
+
+// customId for the select menu. PREFIX-ONLY — no nonce, no flow_id
+// encoded. Concurrent revokes deliberately drop: the second call
+// supersedes the first via deleteFlow + createFlow. See
+// flow-dispatch's trust-model comment on why customId is unsafe as
+// an identity signal.
+const REVOKE_SELECT_CUSTOM_ID = 'qurl_revoke_select';
+
+// TTL for the awaiting-select flow row, in seconds. Matches the
+// pre-existing 60-second click window for the menu. DDB TTL reap is
+// asynchronous (~48 h) but loadFlow filters expired rows
+// synchronously, so the UX cuts off at 60 s on the dot.
+const REVOKE_FLOW_TTL_SECONDS = 60;
+
 async function handleRevoke(interaction, apiKey) {
   if (!apiKey) {
     return interaction.reply({ content: 'qURL API key is not configured.', ephemeral: true });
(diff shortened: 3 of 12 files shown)


vikramlayerv (changes requested):
## Review — request changes before merge

This PR is the precedent for 4 more (`/qurl setup`, `/qurl send` PRs 7–9), so issues here propagate. Pattern, trust model, and dispatcher shape are sound; ship after the items below land.

---

### Why these have to land in THIS PR

The dispatcher pattern, the `flow_id` shape, the supersede semantics, and the `registerFlow` API will all be copied by PR 6 / 7 / 8 / 9. If the contract is wrong here, four PRs ship the same defect and each becomes harder to retrofit because it's now "the convention." The cheapest place to fix is now — most of these are 5-line changes that prevent multi-PR rework.

---

### Critical

**1. `deleteFlow` doesn't check stage → cross-flow contamination when PR 6 lands.**
`flow-state.js:708` deletes purely on `attribute_exists(flow_id)`. `handleRevoke` (`commands.js:79–110`) supersedes on `created=false` by calling `deleteFlow` without a stage guard. As soon as PR 6 (`/qurl setup`) reuses the `flow_id` shape (which it must — same `user.id + channelId + SHARD_ID`), a user who runs `/qurl revoke` while their `setup` modal is open will silently `admin_cleanup` the setup flow.

The trust-model invariant "one flow per (user, channel)" then degrades to "most-recently-started flow wins" with no stage-mismatch reply because the cleanup happens command-side, not dispatcher-side.

**Fix at the harness level** (preferable): gate `deleteFlow` on an `expectedStage` parameter; conditional-delete adds `stage = :expected`. Alternative: `loadFlow` first in the supersede path and only delete if `stage === REVOKE_STAGE_AWAITING_SELECT`. Harness fix is cleaner since the contract is reused 4×.

---

### High

**2. Retry `createFlow` in supersede path has no try/catch.** `commands.js:97–110`. If `deleteFlow` succeeds but the retry `createFlow` throws (DDB throttle / IAM blip), the exception propagates to the universal handler → generic error reply. Worse: the prior flow row is gone. Wrap the retry `createFlow` for symmetry and surface a recoverable message.

**3. Document the "parallelization safety" rule on `Promise.all([getGuildApiKey, deleteFlow])`.** `commands.js:152–158`. Current code is fine — `getGuildApiKey` is a cheap DDB read and the dedup loser short-circuits before the qURL API call. But the pattern propagates to PRs 7–9 where the parallel call could be a paid Google Places lookup or a recipient resolution — at-least-once redelivery would cost a paid call per duplicate. Add a comment in `flow-dispatch.js` (or alongside the handler) pinning: **"OK for idempotent / cheap reads only; do NOT parallelize the dedup primitive with paid side-effect calls."**

---

### Architecture (fold in — they're the precedent)

**4. Supersede deviation must live in the design doc, not just code comments.**
PR description correctly calls out "revoke = supersede" as a deliberate exception to the design-doc "one flow per channel" rule, but the deviation isn't recorded in `apps/discord/docs/zero-downtime-design.md`. PR 6/7–9 engineers will either (a) copy "supersede" for `/qurl send` without realizing send shouldn't, or (b) copy "block" for revoke without realizing revoke does supersede. Add a paragraph under Pillar 1 listing the per-command override:

| Command | Behavior on existing flow | Rationale |
|---|---|---|
| `/qurl revoke` | Supersede | Stateless listing; blocking confuses admins |
| `/qurl setup` | TBD — block | Stateful multi-step |
| `/qurl send` | Block | Stateful multi-step |

One source of truth — same pattern as the CLAUDE.md cross-repo invariants.

**5. `registerFlow` exact-match vs. `prefix` naming.** `flow-dispatch.js:13`. Handler lookup does `handlers.get(customId)` not `startsWith`. The parameter is named `prefix`, which implies prefix semantics. PR 6 modal IDs will likely want real prefix matching (e.g. `setup_modal:<nonce>`). Decide **before PR 6 commits the wrong assumption:**

- (a) Keep exact-match → rename param to `customId` / `routeKey`.
- (b) Switch to prefix-match → name stays right, lookup logic changes.

Either is fine; the cost of leaving it ambiguous is PR 6 picking the other interpretation and tripping a registration collision.

---

### Track as follow-ups (not blocking)

- Listener `index.js:280–285` silent-drops unknown future Discord interaction types — add `logger.debug('interaction: unrouted', { type })` fallthrough.
- `flow_id` in debug logs exposes `user.id + channelId` — switch to a `flow_id_hash` in dispatcher debug paths.
- Document "deleteFlow-first ordering" as a hard requirement for handlers (60s TTL race could otherwise commit irreversible work after row vanishes).
- `nowEpochSeconds` + apiKey-resolution extraction — file an issue now (per the team's "file follow-ups immediately" policy) so PR 7's reviewer doesn't rediscover the duplication.
- `INTERACTION_HANDLED` audit-event contract gap is the right deferral; decide once PR 6's modal contract is in hand.

---

Trust model, dispatcher abstraction, OCC + `deleteFlow` dedup primitive, handler safety-net, and the `tests/setup-env.js` decision are all correct. The fixes above are surgical and don't touch the design — they just prevent four downstream PRs from inheriting the rough edges.

🤖 Reviewed via 2-agent fanout

justin-layerv:
## Review resolution — round 1

Thanks for the thorough review. Six items addressed in 906e8c3, four tracked as follow-ups.

### Addressed

| # | Item | Fix |
|---|------|-----|
| Critical 1 | `deleteFlow` lacks stage gate → cross-flow contamination | Harness-level fix: `ConditionExpression` now `attribute_exists(flow_id) AND #s = :stage`. Post-CCFE recheck logs the stage-mismatch vs row-absent discriminator. Test added pinning that a revoke supersede cannot admin_cleanup a sibling `awaiting_setup_modal` row at the same flow_id. |
| High 2 | Retry `createFlow` no try/catch | Wrapped in try/catch; throw on retry surfaces the recoverable "Could not start a revoke session" reply instead of falling through to the universal error envelope. |
| High 3 | Document parallelization safety | Comment block alongside `Promise.all([getGuildApiKey, deleteFlow])` in `handleRevokeSelect` pinning the rule for PR 6/7/8/9: idempotent / free reads only; never parallelize the dedup primitive with paid side-effect calls. |
| Arch 4 | Supersede deviation in design doc | Added per-command override matrix to `docs/zero-downtime-design.md` Pillar 1 (revoke=supersede with rationale; setup/send=block). Notes that the supersede semantics are now harness-enforced via the stage gate. |
| Arch 5 | `prefix` vs `customId` naming | Renamed param `prefix` → `customId`. Comment in module header documents how to extend the lookup if a future flow needs nonce'd suffix matching. Stayed with exact-match — keeping the contract narrow until a real prefix-match use case appears. |
| Follow-up | Listener unrouted fallthrough | Added `logger.debug('interactionCreate: unrouted interaction type', { type })` in `index.js`. |

### Deferred / tracked

| # | Item | Where |
|---|------|-------|
| FU a | `flow_id` PII hashing in debug logs | Invite-only-beta scope means debug logs aren't user-exposed. Will revisit when log retention or audience changes. |
| FU b | "deleteFlow-first ordering" as a hard requirement | Implicitly covered by the design-doc per-command override section above; promotable to a named harness invariant when PR 7's `handleSendSelect` lands as a second concrete example. |
| FU c | `nowEpochSeconds` + apiKey resolution extraction | Filed as [#261](https://github.com/layervai/qurl-integrations/issues/261) per your "file follow-ups immediately" callout. Lists the existing callsites + scope guidance for picking up between PR 5 and PR 6. |
| FU d | `INTERACTION_HANDLED` contract gap | Deferred until PR 6's modal contract is in hand, per your suggestion. |

Test count up to 1134 (one new stage-mismatch case in `flow-state.test.js`). Lint clean. Ready for re-review.

PR #260, opened 2026-05-12
fix(discord): opt into connector per-recipient vid path (qurl-integrations-infra#525)

vikramlayerv:
## Closes refresh-bypass on self-destruct (bot half)

Companion to **qurl-integrations-infra#526**, which adds the server-side enforcement primitive. This PR makes the Discord bot opt into the new per-recipient vid path so the connector actually generates per-mint vids on \`/qurl send\`.

Without this PR, the connector's enforcement code is dormant — the legacy mint_link path (1 qURL + N tokens, shared target_url) keeps shipping and refresh-bypass stays open.

## What changes

\`apps/discord/src/connector.js\` — \`mintLinks\` gains an optional 5th argument \`expiresIn\` (duration string like \`1h\`, \`168h\`). When set, included in the connector's mint_link request body, which is the connector's signal to take the per-vid path (N × \`CreateOneTimeQURL\`). Backward-compat: 4-arg legacy callers continue to work unchanged.

\`apps/discord/src/commands.js\` — \`mintLinksInBatches\` accepts \`expiresIn\` in opts and forwards it to \`mintLinks\`. Also reads \`link.resource_id\` from the connector response (per-vid path returns per-mint resource_id) with a fallback to the batch's \`currentResourceId\` for legacy responses + cold-start pre-#525 connectors. Four call sites updated:

- \`handleSend\` file send
- \`handleSend\` location send (no-op today per the #480 carve-out; provisioned for when it's removed)
- \`handleAddRecipients\` file
- \`handleAddRecipients\` location (same carve-out note)

All four pass the user-chosen send TTL (\`expiresIn\` from \`sendConfig.expires_in\`) so the per-recipient qURLs honor the sender's choice.

## Backward compat

| Connector | Bot side | Result |
|---|---|---|
| Pre-#526 | Pre-this-PR | Legacy mint_link, no enforcement. Status quo. |
| **Post-#526 (merged)** | Pre-this-PR | Legacy mint_link path on connector (bot omits expires_in) — enforcement dormant. Safe to merge connector first. |
| Pre-#526 | This PR (deployed early) | Bot sends \`expires_in\`; old connector ignores it; legacy mint_link. No regression. |
| Post-#526 | This PR | **Per-recipient vid enforcement live.** Refresh after self-destruct = 410. |

Deploy order: connector first, then bot. Both PRs are independently safe.

## Tests

- \`tests/qurl-send.test.js\` — 7 \`mockMintLinks.toHaveBeenCalledWith\` assertions updated to expect the 5th \`expires_in\` arg.
- All 1113 bot tests green.

## Test plan

- [x] \`npx jest\` — 1113 passed, 0 failed
- [x] Lint clean (\`npm run lint\` if configured; pre-commit hooks)
- [x] Diff confined to mintLinks signature + 4 call sites + 7 test assertions
- [ ] After connector PR #526 lands in sandbox: deploy bot here, set self-destruct=30s, view from one recipient, wait 30s, refresh → expect 410 \"Content expired\"
- [ ] Multi-recipient: with 3 recipients sharing a batch, recipient B opening at T=45s (past A's 30s window) should STILL see the content fresh for their own 30s window (per-recipient vid means B has their own clock)

## Pairs with

- **qurl-integrations-infra#526** — connector server-side enforcement (must ship first for this PR's effect to be observable; safe to merge in either order)
- Tracking issue: qurl-integrations-infra#525

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -745,16 +745,31 @@ async function fetchGuildMembers(guild) {
  * and handleAddRecipients (file/location). Centralizing this means a fix to
  * the re-upload / batching / quota logic lands in one place.
  *
+ * `expiresIn` (optional duration string like "1h", "168h") opts into the
+ * connector's per-recipient vid path (qurl-integrations-infra#525). Each
+ * minted link becomes its own one-time-use qURL with a unique vid in
+ * target_url, so refresh after the viewer self-destruct fires returns 410
+ * for that specific recipient. Omitting it keeps the legacy shape (1 qURL
+ * + N tokens, shared target_url, no server-side refresh enforcement).
+ *
+ * Per-recipient resource_id: when the connector takes the per-vid path,
+ * each link comes back with its own fresh resource_id; we honor that
+ * (link.resource_id) for downstream tracking. The legacy path returns the
+ * shared resource_id; fallback to the currentResourceId we drove the
+ * batch with preserves the existing contract.
+ *
  * @param {object} opts
  * @param {string} opts.initialResourceId — resource_id from the first upload
  * @param {() => Promise<{resource_id: string}>} opts.reuploadFn — called when
  *   the current resource's token pool is drained. Must return a fresh resource.
  * @param {string} opts.expiresAt — ISO string; forwarded to mintLinks.
+ * @param {string} [opts.expiresIn] — duration string ("1h", "168h"); when
+ *   set, opts into per-recipient vid enforcement on the connector side.
  * @param {number} opts.recipientCount — number of tokens to mint in total.
  * @param {string} opts.apiKey — QURL API key.
  * @returns {Array<{qurl_link: string, resourceId: string}>}
  */
-async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, recipientCount, apiKey }) {
+async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, expiresIn, recipientCount, apiKey }) {
   const allLinks = [];
   let currentResourceId = initialResourceId;
   let tokensUsed = 0;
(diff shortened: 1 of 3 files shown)


vikramlayerv:

3-agent fan-out (correctness + simplify + efficiency). The fold-in items were addressed in commit \`7f54b45\`.

|---|---|---|---|---|
| Minor (folded) | Test assertion uses \`expect.any(String)\` for the 5th \`expires_in\` arg — too loose. | Fixed — pinned literal \`'6h'\` on the file-send test. A bug swapping ISO into the expires_in slot now fails CI. | done | n/a |
| Minor (folded) | No test exercised the per-vid response branch (\`link.resource_id || currentResourceId\`). | Fixed — new \`file send: per-vid response shape\` test mocks the connector returning per-link resource_id; asserts each recipient stores its per-vid id, not the batch's currentResourceId. | done | n/a |
| Minor | \`7d\` in \`EXPIRY_LABELS\` — confirm qurl-service \`ExpiresIn\` accepts \`d\` units (not just \`m\` / \`h\`). | Already shipped: \`qurl.js:206\` round-trips the same string format pre-#260; #260 doesn't widen the surface. Connector forwards verbatim, qurl-service validates server-side. | none — already in production usage | low |
| Minor | Drift risk between \`expiresIn\` (handleSend local) and \`sendConfig.expires_in\` (handleAddRecipients). | Both flow through the same \`EXPIRY_LABELS\` dropdown values; handleAddRecipients reads what handleSend wrote to DB. Verified: no re-prompt for expiry on Add Recipients. | none — guaranteed-equal by data flow | none |
| Minor | mintLinks positional 5th arg vs opts-object. | Rest of the codebase uses positional args (\`createOneTimeLink\`, \`uploadJsonToConnector\`, \`downloadAndUpload\`). Opts-object refactor would churn unrelated call sites. | none — convention | none |

CI green: 1114 / 1114 tests pass.

— via /pr-polish

vikramlayerv:
Closing — pair of qurl-integrations-infra#526 (closed). The per-recipient vid mechanism is being discarded.

This bot-side PR has no value standalone (it's only useful when the connector is taking the per-vid mint path, which won't ship). Branch retained for reference.

PR #277, opened 2026-05-13
refactor(discord): extract executeSendPipeline from handleSend

justin-layerv:
## Summary

Refactor — extract the back-half of `handleSend` into a standalone `executeSendPipeline(interaction, params)` function so PR 7b's `handleSendFormSend` can be the second caller. **1231 tests pass; lint clean.** (1194 pre-refactor + 37 new across multiple rounds for the five strict-input entry gates.)

The closure-captured state from `handleSend`'s Step-3 form-fill becomes explicit destructured params: `apiKey`, `resourceType`, `attachment`, `locationUrl`, `locationName`, `recipients`, `target`, `isVoiceContext`, `expiresIn`, `selfDestructSeconds`, `personalMessage`, `sendNonce`.

`handleSend` retains its front-half (file/location chooser, recipient/expiry/note form-loop). Step 4 — the line range already demarcated by the comment `// --- Step 4: Process and send ---` — is now a single `return await executeSendPipeline(...)` call.

## Scope honesty: behaviorally equivalent on every existing call path

Original framing said "byte-identical to the pre-refactor." It isn't — the additions are:

1. **Five strict-input entry gates** (none reachable by today's caller; the gates are forcing functions for PR 7b's flow_state-deserialized caller):
   - `isVoiceContext` must be a strict boolean.
   - `target` must be `'user' | 'channel'`.
   - `attachment.url` re-validated against `isAllowedSourceUrl` (defense-in-depth SSRF; `RESOURCE_TYPES.FILE`-only).
   - `expiresIn` must be a key of `EXPIRY_LABELS`.
   - `personalMessage` must be `null` or `string`.
   Each gate shares the same `cancelEdit()` + `clearCooldown()` + `throw` pattern. The cancel-edit replaces the caller's stale "Preparing send..." ephemeral with a specific cancel; the outer `handleCommand` catch still appends a generic followUp.
2. **Removed the duplicate `releaseSlot()` from the catch block.** The `finally` always runs after a return-from-catch and `releaseSlot` is idempotent via `fileSendSlotClaimed` — the catch-branch call was dead code that obscured the single-release-path invariant.

## Why now

PR 7b will replace `/qurl send`'s bouncy file-via-DM UX with new `/qurl file` and `/qurl map` slash commands, both flow_state-backed and fully in-channel. The terminal handler in that PR (`handleSendFormSend`) needs to invoke the same back-half pipeline that `handleSend` uses today. Extracting it as a focused refactor first means PR 7b's review can focus on the new front-half shape — not on whether the back-half extraction is faithful.

## What's NOT in this PR

- **Direct happy-path unit tests for `executeSendPipeline`** beyond the gate-rejection cases. The existing suite (`qurl-send.test.js`, `qurl-send-state-machine.test.js`, `qurl-send-back-half.test.js`, `commands-comprehensive.test.js`) pins end-to-end behavior via `handleSend`. Direct param-object tests will land in PR 7b alongside the second caller — tracked in #278.
- **`/qurl file` / `/qurl map`** — that's PR 7b.
- **Defensive guards on `recipients` shape/cap** — #280.
- **`recipients` mutation cleanup (clone-on-mutate)** — #281.
- **`QURL` → `qURL` prose-message brand sweep** (pre-existing drift in two moved strings) — #279.
- **`AUDIT_EVENTS.PERSIST_FAILED` on `recordQURLSendBatch` failure** — #284.
- **Split `executeSendPipeline` into mint / persist / dispatch phases** — #285.
- **Gate-error aggregation + `_test` export-bag split** — two new follow-ups filed from round 12.

## Diff shape

`apps/discord/src/commands.js` + new tests in `qurl-send-back-half.test.js`. The bulk of the commands.js diff is the function being moved (insertion at line ~803, deletion from former lines 1650–2241). Net adds include the new function header + destructure + docstring + 5 strict-input gates + 1 cleanup + the call-site object.

## Test plan

- [x] **1231 unit tests pass** (1194 pre-existing + 37 new across 5 new `describe` blocks: `isVoiceContext` gate, `target` gate, SSRF gate, `expiresIn` gate, `personalMessage` gate). Test count delta breakdown across review rounds: round 5 +3 (isVoiceContext base), round 6 +4 (numeric/wrapper/cooldown), round 10 +13 (target + SSRF), round 11 +17 (expiresIn + personalMessage).
- [x] `npm run lint` clean (`--max-warnings 0`)
- [x] Function structure inspection: `executeSendPipeline` at line ~803, `handleSend` at line ~1414, all other top-level functions unchanged

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -840,8 +840,13 @@ async function executeSendPipeline(interaction, {
   attachment,
   locationUrl,
   locationName,
+  // MUTATES: the Add Recipients post-send branch dedupe-pushes into
+  // this array (see docstring's "transferred ownership" note).
   recipients,
   target,
+  // Defaults to text-channel wording on a missing/undefined value;
+  // PR 7b should validate at the flow_state-payload schema layer to
+  // catch a voice-context send that lost this signal in serialization.
   isVoiceContext = false,
   expiresIn,
   selfDestructSeconds,
@@ -978,7 +983,10 @@ async function executeSendPipeline(interaction, {
   } catch (error) {
     logger.error('Failed to prepare QURL links', { error: error.message, apiCode: error.apiCode });
     clearCooldown(interaction.user.id); // allow retry on failure
-    releaseSlot();
+    // Slot release lives in the `finally` block below — it ALWAYS runs
+    // after a return-from-catch, and `releaseSlot` is idempotent via
+    // the `fileSendSlotClaimed` flag. Dropping the duplicate call here
+    // keeps the single-release-path invariant visible at a glance.
     // Surface a specific message for known upstream failure codes so the
     // user knows what to do (re-upload to refresh the per-resource quota)
     // instead of seeing a generic "try again" that won't help.
@@ -2313,9 +2321,12 @@ async function handleSend(interaction, apiKey) {
 
   // --- Step 4: Process and send (back-half — extracted to executeSendPipeline). ---
   // `return await` (not bare `await`) so the docstring's "Resolved value"
-  // contract matches the call shape — and the tail call propagates a
-  // throw with one less microtask hop than a bare-await + implicit
-  // undefined return.
+  // contract matches the call shape. `return await` keeps the async
+  // frame on the stack until the awaited promise settles, which
+  // preserves the executeSendPipeline frame in rejection traces —
+  // `return promise` (no await) would detach it, costing forensic
+  // signal. There's a ~1-microtask perf cost vs the no-await form;
+  // negligible at this scale, worth the better traces.
   return await executeSendPipeline(interaction, {
     apiKey,
     resourceType,
@@ -4576,6 +4587,12 @@ module.exports = {
       REVOKE_TRUNC_LIMIT,
       mintLinksInBatches,
       activeMonitors,
+      // The top-level back-half driver. Exported here so PR 7b's
+      // tests (and the follow-up direct unit spec in #278) can pin
+      // its contract against a constructed param object — without
+      // re-driving handleSend's Step-1-through-Step-3 wizard or the
+      // future flow_state-backed form-fill.
+      executeSendPipeline,
       // Test-only file-concurrency hooks. The slot counter is module-
       // private (live state) and exposing a setter lets the cap branch
       // be tested without a parallel-send harness.
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -865,7 +865,17 @@ async function executeSendPipeline(interaction, {
   // a wrong value here surfaces only as a subtle copy mismatch in
   // a non-ephemeral channel post — exactly the silent-regression
   // shape the round-4 cr review flagged.
+  //
+  // Clear cooldown BEFORE the throw so the caller's setCooldown-
+  // in-flight doesn't lock the user out for the cooldown window
+  // (~30s) with no user-visible feedback. clearCooldown is a no-op
+  // when no cooldown is set, so this is safe for callers that
+  // haven't reached the cooldown branch yet — and it matches
+  // handleSend's existing convention of clearing on every error
+  // path. Today's caller never reaches the throw; the cleanup
+  // exists for PR 7b's handleSendFormSend.
   if (typeof isVoiceContext !== 'boolean') {
+    clearCooldown(interaction.user.id);
     throw new TypeError(`executeSendPipeline: isVoiceContext must be a boolean (got ${typeof isVoiceContext}: ${JSON.stringify(isVoiceContext)})`);
   }
   await interaction.editReply({ content: `Preparing links for ${recipients.length} recipient(s)...`, components: [] }).catch(logIgnoredDiscordErr);
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -782,13 +782,24 @@ async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, re
 // notes below capture only non-obvious contract guarantees that a
 // reader couldn't infer from the call sites alone.
 //
+// Entry gates (fire-and-forget cancel-edit + throw):
+//   - `isVoiceContext` must be a strict boolean.
+//   - `target` must be `'user'` or `'channel'`.
+//   - `attachment.url` re-validated against `isAllowedSourceUrl`
+//     when `resourceType === RESOURCE_TYPES.FILE`.
+// Each gate clears the caller's stale ephemeral with a cancel-
+// edit then throws — the user still sees the outer catch's
+// generic followUp, but the gate's specific cancel replaces
+// the stale "Preparing send..." rather than co-existing with it.
+//
 // Caller contract (non-obvious bits — what the signature can't say):
 //
-//   - `attachment.url` is assumed PRE-VALIDATED against
-//     `isAllowedSourceUrl`. The pipeline does NOT re-check; an
-//     attacker-controlled URL reaching `downloadAndUpload` here
-//     would bypass the SSRF defense. Callers reconstructing this
-//     from a persisted source MUST re-validate before invoking.
+//   - `attachment.url` re-validated at entry against
+//     `isAllowedSourceUrl` — defense-in-depth, but callers
+//     SHOULD still validate at their boundary (the entry gate
+//     is the second line, not the first). A tampered persisted
+//     payload reaching the gate produces a generic "Internal
+//     error" cancel without leaking the SSRF detail.
 //
 //   - `recipients` is MUTATED in-place on the Add Recipients post-
 //     send branch (deduped push of new IDs). Treat as transferred
@@ -869,20 +880,22 @@ async function executeSendPipeline(interaction, {
   // handleSend's existing convention of clearing on every error
   // path. Today's caller never reaches the throw; the cleanup
   // exists for PR 7b's handleSendFormSend.
+  // Shared cancel-edit shape for all three entry gates. Fire-and-
+  // forget so a failed edit doesn't become the observable outcome —
+  // the throw is the load-bearing signal (test pins + logger.error
+  // in handleCommand's outer catch). The outer catch will still
+  // append a generic "There was an error" followUp; this edit
+  // replaces the caller's stale ephemeral so the user sees a
+  // specific cancel alongside the generic followUp, not a stale
+  // "Preparing send..." alongside it.
+  const cancelEdit = () => interaction.editReply({
+    content: '❌ Internal error — send cancelled. Please rerun the command.',
+    components: [],
+  }).catch(logIgnoredDiscordErr);
+
   if (typeof isVoiceContext !== 'boolean') {
     clearCooldown(interaction.user.id);
-    // Clear whatever stale ephemeral the caller's button handler
-    // left on screen ("Preparing send…" is the canonical case)
-    // BEFORE throwing. Fire-and-forget — the throw is the
-    // load-bearing signal (test pins + logger.error in
-    // handleCommand's outer catch), so a failed editReply here
-    // must not become the observable outcome. Without this, the
-    // user sees a persistent stale message alongside the generic
-    // "There was an error" followUp the outer catch will add.
-    interaction.editReply({
-      content: '❌ Internal error — send cancelled. Please rerun the command.',
-      components: [],
-    }).catch(logIgnoredDiscordErr);
+    cancelEdit();
     // Render `typeof=` + `value=` separately so a prod-log reader
     // doesn't have to disambiguate `(got object: null)` (where
     // "object" describes the typeof and "null" is the value — a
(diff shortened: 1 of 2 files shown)


justin-layerv:
Addressed cr feedback in 4bd8314:

- **Item 1 (`recipients` mutation as a side-effect)** — added an explicit "transferred ownership" note in the contract docstring. PR 7b's `handleSendFormSend` will reconstruct the array from the flow_state payload and now knows not to treat it as a snapshot.
- **Item 2 (partial docstring)** — completed the param enumeration. All 12 params now have explicit prose, not just 5.
- **Item 4 (return-value drop)** — documented in a new "Resolved value" paragraph.

Deferred to PR 7b (explicitly flagged by the reviewer):
- **Item 3 (`_test` export)** — once PR 7b lands a second caller, adding `executeSendPipeline` to `_test` lets the pipeline contract be pinned directly. Out of scope here.

Deferred to a future PR (explicitly flagged as future):
- **Item 5 (size)** — the collector-setup block (the post-DM management window) is the natural cleavage if `executeSendPipeline` ever gets split. Not for this PR.

`/simplify` returned clean: code-reuse review found nothing, quality review confirmed the 12-param destructure is justified at this scale, efficiency review confirmed no regression. 1194 tests still pass, lint clean.

justin-layerv:
Round 2 addressed in f94166c:

- **Item 1 (doc-vs-code mismatch on `return`)** — fixed by adding `return` at the call site, per reviewer's preference. Docstring now matches.
- **Item 2 (`target` value space)** — fixed. Enum now `'user' | 'channel'` with explicit note that voice/text is on `isVoiceContext`, not `target`.
- **Item 3 (`isVoiceContext` no default)** — fixed. Destructure now `isVoiceContext = false`, with a doc note explaining why (PR 7b safety).

Deferred (issues filed):

- **Item 4 (no dedicated unit tests for `executeSendPipeline`)** — issue #278. Natural to land in PR 7b alongside the second caller; reviewer explicitly flagged as not blocking.
- **Pre-existing `'QURL'` → `'qURL'` brand-spelling drift** in two log messages — issue #279. Pre-existing in the back-half; not introduced by this refactor. Folding into a pure-refactor PR would break the "every test passes unchanged" safety property.

Acknowledged not actioned:

- **Item 5 (doc block could shed weight)** — reviewer's exact framing was "Not blocking; calling out for the next refactor on this surface." The per-param prose is load-bearing for PR 7b's second caller, so I'm keeping it for now. Happy to revisit on the next pass.

`/simplify` reran clean post-fix. 1194 tests pass, lint clean.

justin-layerv:
Round 3 addressed in b2e48b4 — all 5 nits applied:

- **Item 1 (missing `_test` export)** — added. Matches the existing pattern (`monitorLinkStatus`, `revokeAllLinks`, etc.). #278's direct unit spec can now dock onto it cleanly.
- **Item 2 (redundant `releaseSlot()` in catch)** — dropped. The `finally` always runs after the return-from-catch, and the call is idempotent via the flag.
- **Item 3 (inverted microtask claim)** — corrected. The right rationale is **stack-trace preservation on rejection** (async frame stays on stack until promise settles), not microtask reduction. `return await` actually adds a microtask.
- **Item 4 (`isVoiceContext` default is load-bearing)** — left as `= false` (legacy text-channel fallback) and added an inline note: PR 7b should validate at its flow_state-payload schema layer. Per reviewer's "Not blocking for this PR" framing.
- **Item 5 (`recipients` mutation marker on the param line)** — added `// MUTATES` inline comment in the destructure.

`/simplify` reran clean. 1194 tests still pass, lint clean.

justin-layerv:
Round 4 addressed in 5d48834:

- **Concern 2 (`isVoiceContext` default has silent-wrong-branch failure)** — fixed in code. Dropped the `= false` default and added a `TypeError` at the entry of `executeSendPipeline` when `typeof isVoiceContext !== 'boolean'`. Matches the strict-input convention of the flow-state primitives. `handleSend` always threads the flag so no path regresses; 1194 tests still pass.

Deferred (issues filed):

- **Concern 1 (defensive guards on `recipients` invariants)** — #280. Reviewer's strongest concern; "could land as a tiny follow-up before 7b's second caller appears." Keeping PR #277 pure-refactor; the tiny follow-up will land the guards.
- **Concern 3 (`recipients` mutation code smell)** — #281. Two-line defensive-clone fix; eliminates the "transferred ownership" caveat from the docstring.

Confirmed:

- **Concern 4 (no direct unit test for `executeSendPipeline`)** — #278 (filed in round 2). Natural to land in PR 7b alongside the second caller.

Also noted in the round-4 review:

- **Diff stat discrepancy** (`+1402 / -1350` vs the actual `+1462 / -1355`) — pre-amendment numbers from my PR body, now stale after rounds 1–4 added comment lines. Not editing for that — the actual diff is what reviewers see in the Files tab.

`/simplify` reran clean. Lint clean.

justin-layerv:
Round 5 addressed in 0519486 + PR description edit:

- **Item 1 (PR description's "byte-identical" overstates)** — edited the body to call out the two non-byte-identical additions explicitly: the new `isVoiceContext` strict gate (new failure mode, defensible for PR 7b safety) and the `releaseSlot()` cleanup (dead code removed). Audit trail now matches the diff.
- **Item 2 (TypeError gate has zero test coverage)** — added 3 cases in `qurl-send-back-half.test.js` pinning the gate: `undefined`, `'true'` (string), `null` (the `typeof null === 'object'` foot-gun). Each asserts the rejection fires before the "Preparing links..." editReply, so any soft-warn regression trips the test. 1194 → 1197 tests.

Confirmed (already filed):

- **Item 3 (`recipients` mutation contract surprise)** — #281.
- **Item 4 (brand-spelling sweep for `'QURL'` → `'qURL'`)** — #279.

Acknowledged not actioned:

- **Item 5 (docstring shape — JSDoc `@param` vs prose)** — pure nit. The prose form matches the rest of `commands.js` per the reviewer's own note. Leaving as-is.

Lint clean. `/simplify` reran clean.

justin-layerv:
Round 6 addressed in d9bd4c4:

- **Item 1 (cooldown side-effect at the gate)** — fixed. `clearCooldown(interaction.user.id)` now fires immediately before the `TypeError` throw. Matches handleSend's existing convention of clearing on every error path. Pinned by a new test case.
- **Item 4(a) (extra gate test cases)** — added. `0`, `1` (numeric miscoding), and `new Boolean(true)` (wrapper-object foot-gun) all rejected. Plus the cooldown-cleanup test from Item 1. 1197 → 1201 tests.

Forward-looking concerns recorded for PR 7b:

- **Item 2 (`recipients` mutation on retry)** — already #281. Concern that PR 7b's `handleSendFormSend` retrying with a previously-mutated `recipients` reference would silently double-add was added to PR 7b's task description (#24) so the implementation handles it.
- **Item 3 (`activeFileSends` watchdog double-claim under retry)** — recorded in PR 7b's task. Mitigated either by landing #280's input guards before PR 7b lands, or by PR 7b tracking slot claims at the flow_state row level.
- **Item 4(b) (direct happy-path tests timing recommendation)** — recorded. Will land alongside PR 7b's `handleSendFormSend`, not after. Already noted in #278.

`/simplify` reran clean. Lint clean.

justin-layerv:
Round 7 addressed in 87cdcb4:

- **Main item (caller-contract gaps — interaction shape)** — fixed. Docstring now enumerates `user.id`, `channelId`, `channel`, `member?.displayName`, `guild?.members?.fetch`, `editReply`, and `channel.send` with the call site each touches. PR 7b's `handleSendFormSend` has a discoverable contract instead of a debug-by-null trail.
- **Minor 1 (`JSON.stringify(undefined)` → "got undefined: undefined")** — fixed. Switched to `String(...)` for the value rendering. Doc-comment notes the rationale + the BigInt-doesn't-serialize edge case.
- **Observation (`undefined`-case test double-invokes pipeline)** — collapsed. Rejection captured once into a promise, reasserted twice. Adds a comment explaining the rationale for future readers.

Deferred (issues filed):

- **Minor 2 (`recordQURLSendBatch` failure path has no audit event)** — #284. Orphan rate is currently logs-only; the audit event makes it dashboardable. Land as its own focused PR before PR 7b.
- **Minor 4 (function size / mint-persist-dispatch split)** — #285. Three-way split would cut PR 7b's review surface. Could bundle with #280's input guards (same entry point).

Confirmed cross-refs:

- **Minor 3 (brand spelling on moved log strings)** — already #279.

1201 tests pass. Lint clean. `/simplify` reran clean.

justin-layerv:
Round 8 addressed in 11ba98d:

- **🟡 String(null) rendering ambiguity** — fixed. Throw message now reads `(got typeof=object, value=null)` instead of `(got object: null)`. Removes the `typeof null === 'object'` foot-gun for a prod-log reader. Comment explains the rationale.
- **🟡 SSRF re-validation requirement** — added to the `attachment` bullet of the caller contract. Today handleSend's Step-2 validates `attachment.url` against `isAllowedSourceUrl` before reaching this pipeline; PR 7b's `handleSendFormSend` reconstructing from a flow_state row MUST re-validate. The doc note makes this discoverable instead of hidden in handleSend.
- **🟡 Cooldown test cleanup hygiene** — added an `afterEach` to the gate describe that explicitly clears the cooldown for the unique test user id. Today the gate's own `clearCooldown` does the cleanup so the afterEach is a no-op (defense-in-depth); a future regression that moves the throw before `clearCooldown` would leak sendCooldowns state without it. Required exposing `clearCooldown` via `_test` (matches the existing `isOnCooldown` / `setCooldown` exposure).

Confirmed (already filed / cross-ref):

- **🟡 Brand drift in moved log strings** — #279.
- **🟡 Direct unit tests deferred to PR 7b** — #278.

Acknowledged not actioned (forward-looking / pre-existing):

- **🟢 Forward-looking dead-code marker on the gate throw** — Reviewer: "If PR 7b lands a different shape and never trips the gate, consider whether to keep it or yank." Decision for PR 7b's review, not for this PR.
- **🟡 `expiresAt` clock measured from re-invocation** — pre-existing behavior, reviewer re-acknowledged. The existing comment already covers it; no change for this PR.

1201 tests pass. Lint clean. `/simplify` reran clean.

justin-layerv:
Round 9 addressed in fad52ad:

- **Issue 1 (gate strands user UI)** — fixed. The TypeError now does a fire-and-forget `editReply({ content: '❌ Internal error — send cancelled. Please rerun the command.', components: [] })` BEFORE propagating. The throw stays the load-bearing signal (test pins + outer-catch logger.error), but the caller's stale ephemeral gets cleared with a specific message rather than persisting alongside the generic followUp. Test pin updated: the two existing gate tests now assert the cancel-message shape instead of "editReply not called."
- **Issue 2 (docstring violates CLAUDE.md comment density)** — trimmed. Dropped per-param WHAT descriptions (the destructure IS the signature), the verbose "Resolved value" paragraph, and the expanded interaction-shape table. Kept the load-bearing WHYs: SSRF re-validation contract, `recipients`-mutates ownership, `target` vs `isVoiceContext` discrimination, gate forcing-function rationale, and a compressed terse interaction-surface list with parenthetical annotations.
- **Issue 3 (`during PR 7a` reference rots)** — subsumed by issue 2. All PR / caller refs ("PR 7a", "PR 7b's handleSendFormSend", "during PR 7a") are gone — replaced with generic, durable framing.

Diff: −110 / +85 (mostly the docstring trim).

Other notes from round 9:
- **🟢 Brand spelling on moved log strings** — already #279. Reviewer confirmed "leaving them untouched here matches the don't-widen-scope-during-refactor convention."
- **🟡 Direct unit tests for executeSendPipeline** — already #278. Reviewer agrees with the deferral.

1201 tests pass. Lint clean.

justin-layerv:
Round 10 addressed in dbc4abc:

- **Item 1 (dual-message UX)** — docstring updated to be honest about it. The gate's cancel-edit replaces the stale "Preparing send…" but the top-level catch still appends its generic followUp; the prose no longer claims the gate eliminates the followUp.
- **Item 2 (`attachment.url` SSRF gate — reviewer's strongest concern)** — added. The entry-point gate runs `isAllowedSourceUrl(attachment.url)` on `resourceType === RESOURCE_TYPES.FILE` paths. Defense-in-depth — handleSend's Step-2 still validates before calling; this catches future callers that forget. Location sends bypass the gate (locationUrl is sanitizer-built from user text, no attacker-controlled origin).
- **Item 3 (`target` allowed-set gate)** — added. Symmetric shape to the isVoiceContext gate. Accept-set is `'user' | 'channel'`; 'voice' / unknown / non-string all throw with the cancel-edit.

All three entry gates share a `cancelEdit()` helper so the cancel-ephemeral shape stays consistent across the three failure modes.

Tests: 1201 → 1214 (+13 new gate cases). Coverage:
- `target` invalid set: `'voice'` (silent-suppress shape), empty string, unknown values, non-string
- `target` accept-set: `'user'`, `'channel'`
- SSRF rejects: null attachment, missing url field, non-string url, `http://localhost/`, `http://127.0.0.1:8080/`, `http://169.254.169.254/` (AWS metadata)
- SSRF gate bypassed for `resourceType !== file`

Confirmed (already filed):
- **Item 4 (direct happy-path tests)** — #278.
- **Brand spelling** — #279.

Lint clean. `/simplify` reran clean.

justin-layerv:
Round 11 addressed in c5013c2:

- **Item 2 (stranded preamble — comment / code adjacency)** — fixed. Reordered so each comment block sits next to its gate. No more cancelEdit-rationale block intervening between unrelated comments and the isVoiceContext gate.
- **Item 3 (`personalMessage` lacks entry gate)** — added. Symmetric shape: null or string, otherwise throw. Closes the `[object Object]` silent-render concern in the DM embed.
- **Item 4 (`expiresIn` shape unchecked)** — added. Allowed set is `Object.keys(EXPIRY_LABELS)` — single source of truth, can't drift.

Five entry gates now: `isVoiceContext`, `target`, `attachment.url` (SSRF), `expiresIn`, `personalMessage`. All share the `cancelEdit` + `clearCooldown` + throw pattern.

Tests: 1214 → 1231 (+17 new gate cases):
- `expiresIn` invalid: `'25h'`, `'never'`, empty, undefined, number
- `expiresIn` accept-set: `'30m'`, `'1h'`, `'6h'`, `'24h'`, `'7d'`
- `personalMessage` invalid: object, array, number, boolean
- `personalMessage` accept-set: null, empty string, short string

Forward-looking items not actioned:

- **Item 1 (docstring length)** — reviewer flagged as "Not a blocker; the surrounding file style supports heavy comments." Already trimmed substantially in round 9.
- **Item 5 (brand spelling)** — #279.

`/simplify` reran clean. 1231 tests pass. Lint clean.

justin-layerv:
Round 12 addressed:

**Description fix (item 1)** — PR body updated. Test count now accurately shows 1194 → 1231 (+37 new) with the round-by-round breakdown. The stale "+3 added in round 5" / "1197 final count" framing is gone.

**Already pinned (item 3)** — `expect(interaction.editReply).toHaveBeenCalledTimes(1)` in the gate test does pin the synchronous editReply-before-throw ordering. Reviewer said "disregard if so." Disregarded.

**Confirmed deferred / already filed:**

- **Item 2 (weak accept-path tests)** — folded into #278 scope. Direct happy-path tests against `executeSendPipeline` are the right surface to land stronger gate-passed assertions.
- **Item 4 (docstring length)** — reviewer's own framing: "verbosity is defensible." No action.
- **Item 5 (`recipients` mutation enforcement)** — already #281. Reviewer's "interim debug assertion" suggestion is an interim shape; the real fix in #281 (clone-on-mutate) supersedes.

**Newly filed:**

- **Item 6 (gate-error aggregation)** — #286. Multi-fault reporting; cheaper debugging when PR 7b's deserializer corrupts multiple fields together.
- **Item 7 (`_test` export-bag split)** — #287. Per-area sub-bags before the bag hits ~50 symbols.

`/simplify` reran clean. 1231 tests pass, lint clean.

PR #288, opened 2026-05-13
refactor(discord): defensive recipients gates on executeSendPipeline

### LINKED ISSUE #280: discord: defensive guards on executeSendPipeline invariants (recipients shape + cap)
Tracking issue raised by PR #277 cr review (round 4, concern 1 — flagged as the strongest of the four forward-looking concerns).

## Context

PR #277 extracted the back-half of \`/qurl send\` into \`executeSendPipeline(interaction, params)\`. The docstring promises two invariants on \`recipients\`:

1. Non-empty (the pipeline assumes \`recipients.length >= 1\` for the mintLinks/DM-fanout/post-send-collector paths).
2. ≤ \`config.QURL_SEND_MAX_RECIPIENTS\` (the per-send cap).

Today \`handleSend\`'s front-half enforces both right before the call site. Once PR 7b adds \`handleSendFormSend\` as a second caller — reconstructing params from a decrypted \`flow_state\` payload — the invariants depend on the caller doing the same check. A serialization bug, a manual DDB row edit, or a future payload-schema change could violate either without anyone noticing until the failure surfaces deep inside \`mintLinksInBatches\` (zero links → "Failed to create any links") or in DM fan-out past the per-send cap.

## Scope

Add two cheap guards at the top of \`executeSendPipeline\` (in \`apps/discord/src/commands.js\`):

\`\`\`js
if (!Array.isArray(recipients) || recipients.length === 0) {
  throw new TypeError(\`executeSendPipeline: recipients must be a non-empty array (got length \${recipients?.length})\`);
}
if (recipients.length > config.QURL_SEND_MAX_RECIPIENTS) {
  throw new RangeError(\`executeSendPipeline: recipients.length (\${recipients.length}) exceeds QURL_SEND_MAX_RECIPIENTS (\${config.QURL_SEND_MAX_RECIPIENTS})\`);
}
\`\`\`

Matches the strict-input convention already in place for the harness primitives (flow-state.js's createFlow / transitionFlow / deleteFlow validate every input).

## Acceptance criteria

- Both guards land at the entry of \`executeSendPipeline\`.
- The pre-call checks in \`handleSend\` (lines ~2272–2276, post-PR-#277) can stay as belt-and-suspenders OR be dropped — either is fine.
- Tests pin both guard branches (constructed param object with empty array → \`TypeError\`; oversized array → \`RangeError\`).
- The existing 1194 tests still pass — the guards can't trip from any current path.

## Why deferred from PR #277

PR #277 was framed as a pure refactor: every line of the back-half byte-identical to pre-extraction. Adding validation guards is genuinely new behavior — defensible, but a deviation from that framing. The reviewer's exact framing: "could be a follow-up in #278 [or this issue]" — i.e., bundling with the unit-tests work in #278 is reasonable, or a one-shot tiny PR ahead of 7b.

## Surfaced by

PR #277 cr review (round 4, concern 1). Reviewer: "Strongest of the four is (1) defensive guards — could land as a tiny follow-up before 7b's second caller appears."

justin-layerv:
## Summary

Closes #280. Adds the two `recipients`-invariant gates that the cr review on #277 flagged as the strongest non-blocking concern — landing them as a focused follow-up before PR 7b's `handleSendFormSend` becomes the second caller.

## What's in this PR

Two entry gates at `executeSendPipeline`, sharing the existing strict-input pattern (`cancelEdit` + `clearCooldown` + `throw`):

```js
if (!Array.isArray(recipients) || recipients.length === 0) {
  throw new TypeError(`executeSendPipeline: recipients must be a non-empty array (got ${...})`);
}
if (recipients.length > config.QURL_SEND_MAX_RECIPIENTS) {
  throw new RangeError(`executeSendPipeline: recipients.length (${recipients.length}) exceeds QURL_SEND_MAX_RECIPIENTS (${config.QURL_SEND_MAX_RECIPIENTS})`);
}
```

Seven entry gates now (was five): `isVoiceContext`, `target`, `attachment.url` SSRF, `expiresIn`, `personalMessage`, `recipients` non-empty, `recipients` ≤ cap.

## Why now (not in #277)

#277 was scoped as a focused refactor. Each gate added there was defensible inline, but bundling a sixth + seventh would have started pushing the "pure refactor" framing. Landing as its own tiny PR before PR 7b means:

- PR 7b's `handleSendFormSend` reconstructing from a decrypted `flow_state` payload inherits the guards by default
- A future cr cycle on PR 7b doesn't re-litigate the gate shape (already pinned here)
- The `recipients` non-empty gate also fences the `recipients.length` read on the very next line — a non-array reaching that line would crash with a less actionable "Cannot read properties of null" message

## What's NOT in this PR

- Drop of the matching pre-call checks in `handleSend` — they're idiomatic UX (polite editReply, no throw) and stay as belt-and-suspenders. The gates here are defense-in-depth for callers that skip those checks.

## Test plan

- [x] 1231 → 1241 (+10 new gate cases): non-array-or-empty (empty array, null, undefined, plain object, string, number), oversized (cap+1 trips the RangeError), cooldown cleanup on the rejection path, accept-set (one + several recipients).
- [x] All existing 1231 tests pass unchanged — today's callers thread valid recipients.
- [x] `npm run lint` clean (`--max-warnings 0`).

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -971,7 +971,10 @@ async function executeSendPipeline(interaction, {
   // `personalMessage` shape gate. The DM-embed renderer expects
   // null OR string. A non-string object would silently stringify
   // to `[object Object]` in the recipient's DM — exact silent-
-  // regression shape the other gates fence.
+  // regression shape the other gates fence. Intentionally renders
+  // only `typeof`, NOT the value (no truncForLog), since a string
+  // here would be user-authored DM prose; we don't want even a
+  // truncated 64-char prefix of that landing in prod logs.
   if (personalMessage !== null && typeof personalMessage !== 'string') {
     clearCooldown(interaction.user.id);
     cancelEdit();
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -850,19 +850,11 @@ async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, re
 // early-exit `return interaction.editReply(...)` paths discard
 // observable result; the function exists for its user-visible
 // side-effects (editReply + DM fan-out + channel-announce).
-// Shared throw-message renderer for the entry-gate family. Bounds the
-// stringified value at 64 code points and appends `…` when truncation
-// happened, so a prod-log reader can tell "the caller passed exactly
-// these 64 chars" from "the value was longer and we cut it." The
-// String() coercion may itself throw on a pathological value with a
-// throwing Symbol.toPrimitive / toString — fall back to a sentinel so
-// the gate's intended throw isn't itself replaced by an opaque error
-// from the renderer.
-//
-// Slice on code points (via the string iterator), NOT code units, so
-// the 64th character can't land on a high surrogate and produce a
-// malformed UTF-16 pair before the `…`. Cosmetic for log readability
-// but matches "do the boring thing right" for the gate family.
+// Bounded value-renderer for entry-gate throw messages. `String()` is
+// wrapped because a hostile @@toPrimitive / toString would otherwise
+// replace the gate's intended TypeError with the renderer's opaque
+// throw. Code-point (not code-unit) slicing avoids splitting a
+// surrogate pair across the `…` boundary.
 function truncForLog(v) {
   let s;
   try {
@@ -870,14 +862,11 @@ function truncForLog(v) {
   } catch {
     return '<unrepresentable>';
   }
-  // Bound the spread cost. A 1MB input would otherwise materialize a
-  // 1M-entry code-point array just to take the first 64. 128 code
-  // units is a safe over-budget that covers any 64-code-point prefix
-  // even if every code point uses a surrogate pair (each pair = 2
-  // code units). `truncated` is the authoritative truncation signal
-  // because a string of exactly 65 surrogate pairs (130 code units)
-  // has head-cps.length === 64 — relying on cps.length alone would
-  // silently drop the `…` marker on that shape.
+  // Pre-slice bounds the [...head] spread cost on pathological inputs
+  // (1MB string → 128-element array, not 1M). `truncated` is the
+  // authoritative truncation signal because 65 surrogate pairs (130
+  // code units) yields head-cps.length === 64 — relying on cps.length
+  // alone would silently drop the `…` on that shape.
   const truncated = s.length > 128;
   const head = truncated ? s.slice(0, 128) : s;
   const cps = [...head];
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -851,13 +851,18 @@ async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, re
 // observable result; the function exists for its user-visible
 // side-effects (editReply + DM fan-out + channel-announce).
 // Shared throw-message renderer for the entry-gate family. Bounds the
-// stringified value at 64 chars and appends `…` when truncation
+// stringified value at 64 code points and appends `…` when truncation
 // happened, so a prod-log reader can tell "the caller passed exactly
 // these 64 chars" from "the value was longer and we cut it." The
 // String() coercion may itself throw on a pathological value with a
 // throwing Symbol.toPrimitive / toString — fall back to a sentinel so
 // the gate's intended throw isn't itself replaced by an opaque error
 // from the renderer.
+//
+// Slice on code points (via the string iterator), NOT code units, so
+// the 64th character can't land on a high surrogate and produce a
+// malformed UTF-16 pair before the `…`. Cosmetic for log readability
+// but matches "do the boring thing right" for the gate family.
 function truncForLog(v) {
   let s;
   try {
@@ -865,7 +870,8 @@ function truncForLog(v) {
   } catch {
     return '<unrepresentable>';
   }
-  return s.length > 64 ? `${s.slice(0, 64)}…` : s;
+  const cps = [...s];
+  return cps.length > 64 ? `${cps.slice(0, 64).join('')}…` : s;
 }
 
 async function executeSendPipeline(interaction, {
@@ -928,7 +934,7 @@ async function executeSendPipeline(interaction, {
   if (target !== 'user' && target !== 'channel') {
     clearCooldown(interaction.user.id);
     cancelEdit();
-    throw new TypeError(`executeSendPipeline: target must be 'user' or 'channel' (got ${String(target)})`);
+    throw new TypeError(`executeSendPipeline: target must be 'user' or 'channel' (got ${truncForLog(target)})`);
   }
 
   // Defense-in-depth SSRF re-check. handleSend's Step-2 validates
@@ -959,7 +965,7 @@ async function executeSendPipeline(interaction, {
   if (!Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, expiresIn)) {
     clearCooldown(interaction.user.id);
     cancelEdit();
-    throw new TypeError(`executeSendPipeline: expiresIn must be one of ${Object.keys(EXPIRY_LABELS).join('|')} (got ${String(expiresIn)})`);
+    throw new TypeError(`executeSendPipeline: expiresIn must be one of ${Object.keys(EXPIRY_LABELS).join('|')} (got ${truncForLog(expiresIn)})`);
   }
 
   // `personalMessage` shape gate. The DM-embed renderer expects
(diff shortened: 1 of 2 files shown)


justin-layerv:
Round 1 addressed in 38e7a02:

- **Nit 1 (stale line 948 reference)** — fixed. Replaced with the durable "the editReply below." Line numbers in comments rot too quickly.
- **Nit 2 (vacuous-pass in accept-path)** — aligned to the existing file convention. Initially diverged with `expect.assertions(2)` + fake assertions; /simplify also flagged the divergence. The broader file-wide tightening across all four accept-path sites is filed as #291.
- **Nit 3 (dead `typeof clearCooldown === 'function'` guard)** — removed.

Optional follow-up filed: **#290** — element-level validation of `recipients` array (today's gate validates shape + cap but not the elements; `[null]` or `[{ no_id: 'oops' }]` would clear both gates and crash downstream).

`/simplify` reran clean post-fix. 1241 tests pass, lint clean.

justin-layerv:
Round 2 addressed in 3ecf5fe:

- **Nit 1 (rendering inconsistency for null vs `{}`)** — fixed. Non-array branch now uses `typeof=X, value=Y` pattern matching the isVoiceContext gate. Prod-log readers can distinguish `null` / `undefined` / `{}` / strings.
- **Nit 2 (cooldown cleanup only covers empty-array path)** — added a parallel test for the oversized (RangeError) branch.
- **Nit 4 (docstring element-shape sharpening)** — done. `recipients` entry now mentions `{ id, username }` element shape AND explicitly states the gate only validates outer shape + length.
- **Nit 3 (vacuous-pass cross-ref)** — already #291.

Plus four new tests pinning the rendering-shape contract on the throw message itself (so a future refactor that drops the typeof/value split is caught by message-shape tests, not just by message-substring matches).

1241 → 1246 (+5). Lint clean. `/simplify` clean.

justin-layerv:
Round 3 addressed in e3bd4b3 — all 4 nits applied:

- **Nit 1 (boundary at length === cap)** — added an exact-cap accept-path test. Pins that the gate uses `>`, not `>=`. Sourced from `config.QURL_SEND_MAX_RECIPIENTS` so a cap bump doesn't drift.
- **Nit 2 (`String(recipients)` unbounded)** — `.slice(0, 64)` applied to BOTH the recipients gate and the sibling `isVoiceContext` gate. Symmetric — both have the same pathological-input risk for a future caller.
- **Nit 3 (plain-object label misleading)** — replaced with two distinct test cases: `'plain object'` (`{}`) and `'array-like object'` (`{ 0: 'u1', length: 1 }`). The array-like case explicitly documents the `Array.isArray`-strict behavior.
- **Nit 4 (vacuous-pass cross-ref missing)** — added `tracked in #291` to the accept-path comment.

1246 → 1248 (+2 new tests). Lint clean. `/simplify` clean.

justin-layerv:
Round 4 addressed in 698f774:

- **Nit 1 (truncation marker)** — fixed. Extracted shared `truncForLog(value)` helper, applied consistently to both the `isVoiceContext` gate and the new `recipients` gate. 64-char cap + `…` on truncation. A prod-log reader can now distinguish exact-fit from cut-off.
- **Nit 2 (`String(obj)` can throw)** — also addressed via the same helper. `truncForLog` wraps `String()` in try/catch with `<unrepresentable>` fallback. Reviewer said "probably not worth the noise" but since the renderer is now consolidated, the hardening is cheap.
- **Nit 3 (vacuous-pass)** — already #291.
- **Nit 4 (comment density)** — recipients gate's rendering comment now references the `isVoiceContext` gate for shared rationale ("Same rendering rationale as the isVoiceContext gate above") instead of duplicating.

Tests: 1248 → 1250 (+2 new: truncation-marker rendering pin + hostile-toString fallback pin).

Lint clean. `/simplify` clean.

Verdict: I believe this is ready to merge — but waiting on the cr re-run to confirm round-5 has no new findings.

justin-layerv:
## Round 5 — All three nits applied (1f9eb46)

**1. `truncForLog` symmetry across all value-rendering gates** ✅
- Applied to `target` rejection message (commands.js:937)
- Applied to `expiresIn` rejection message (commands.js:968)
- All four value-printing gates (isVoiceContext, target, expiresIn, recipients) now share identical 64-char bounding behavior.

**2. UTF-16 code-point safety in `truncForLog`** ✅
- Switched from `String#slice(0, 64)` (UTF-16 code units) to `[...s].slice(0, 64).join('')` (code points).
- A surrogate-pair emoji at the 64-boundary would otherwise leak a lone high surrogate before the `…`.
- Test added: 64 emoji renders as 64 emoji + `)` (no `…`); 65 emoji renders as 64 emoji + `…`.

**3. Comment compression in recipients gate** ✅
- Collapsed duplicative `typeof=` / `value=` rationale into one block referencing the isVoiceContext gate above.

Test count: 119 in the back-half suite (was 116), 1253 total. Lint clean. Awaiting next cr round.

justin-layerv:
## Round 6 — Two nits applied, one deferred (4a341d5)

**1. `personalMessage` value-omission rationale** ✅
- Added 4-line note explaining the intentional `typeof`-only rendering: a string here is user-authored DM prose, so we don't want any prefix of it (truncated or otherwise) landing in prod logs.

**2. Test config hoist** ✅
- Replaced three inline `require('../src/config')` calls with a single `const RECIPIENT_CAP` at the top of the recipients describe block.

**3. `failGate(interaction, ErrorCtor, msg)` helper across 7 gates** — deferred to #292
- Reviewer explicitly framed as "not blocking — the explicit form does keep each gate locally readable."
- Falls outside this PR's "defensive recipients gates" framing; cap-by-cap churn would have doubled the diff in late cr rounds.
- Filed as `pr-polish-followup` + `priority: low`.

Test count unchanged at 119 (back-half) / 1253 (total). Lint clean.

justin-layerv:
## Round 7 — All four nits applied (2173632)

**1. `truncForLog` worst-case allocation bounded** ✅
- Pre-slice `s` to 128 code units before the `[...]` spread. A 1MB input no longer materializes a 1M-entry code-point array on the error path. Use a `truncated` flag (not just `cps.length > 64`) as the authoritative signal — needed for the 65-surrogate-pair case where head-cps.length === 64.

**2. `detail` rendering unified** ✅
- Empty-array now renders `typeof=object, value=<empty array>` to match the canonical shape. One grep pattern alerts on the gate now.

**3. Comment accuracy** ✅
- Fixed to describe the actual `recipients.length` chain (cap check → editReply), not just the editReply.

**4. Second catch branch of `truncForLog` pinned** ✅
- Added `Object.create(null)` to the `<unrepresentable>`-fallback test.each. Realistic deserialized-payload shape: no prototype → no `toString` → `String(v)` throws.

Test count: 120 in the back-half suite (was 119), 1254 total. Lint clean.

justin-layerv:
## Post-round-8 `/simplify` pass (2f17a9d)

Round 8 was "ship it" with one non-actionable micro-observation. Running `/simplify` per the pr-ready loop surfaced two CLAUDE.md compliance fixes:

**1. Strip "Round-N cr nit" prefixes from added comments** ✅
- Five test-file comments + the RECIPIENT_CAP hoist note. Per CLAUDE.md: "Don't reference the current task, fix, or callers... those belong in the PR description and rot as the codebase evolves." Kept the WHYs intact; dropped only the round numbering.

**2. Compress `truncForLog` header** ✅
- 13 lines → 5. The original restated parts of what the code already says. Kept the two load-bearing WHYs: hostile `@@toPrimitive` justification and surrogate-pair-safety justification.

**3. Factory consolidation — deferred to #293**
- `/simplify` Agents 1 + 2 both flagged 7 duplicate `makePipelineParams` factories in `qurl-send-back-half.test.js`. But 6 of 7 predate #288 — the consolidation is a file-wide refactor unrelated to "defensive recipients gates." Filed as `pr-polish-followup` + `priority: low`.

Test count unchanged at 120 (back-half) / 1254 (total). Lint clean.

This should be the convergence round.

justin-layerv:
## Round 9 — both items applied (d5bfe67)

**1. Missed round-reference at line 1921** ✅
- Re-phrased: "not just the two introduced in round 4" → "not just the recipients gate that introduces it." Caught by Round-9 cr — same anti-pattern as the prefix strip, just phrased differently. Apologies for missing it in the /simplify sweep.

**2. `isVoiceContext` truncation pin** ✅
- Added the missing one-liner. The describe whose name says "all value-rendering gates" now actually covers all three (`isVoiceContext`, `target`, `expiresIn`). Closes the asymmetry the reviewer flagged.

Test count: 121 in the back-half suite (was 120), 1255 total. Lint clean.

Issue #293, opened 2026-05-13
discord: consolidate 7 duplicate makePipelineParams factories in qurl-send-back-half.test.js


justin-layerv:
## Context
Surfaced by `/simplify` during PR #288 convergence. `tests/qurl-send-back-half.test.js` has 7 separate `makePipelineParams` / `makeParams` factories across the per-gate describe blocks (lines 1394, 1514, 1573, 1638, 1684, 1741, 1926 at HEAD of the PR). Each returns the same default payload shape, differing only in the param being tested.

## Drift hazard
Adding a new pipeline param (e.g. when PR 7b lands `/qurl file`) would require updating all 7 factories. The PR #288 round-7 block at line 1926 already uses the strictly-more-useful `(overrides = {})` spread shape — that signature is the right one for all of them.

## Scope
- Define one file-scope `makePipelineParams(overrides = {})` (probably near `makeInteraction()` at `tests/qurl-send-back-half.test.js:222`).
- Delete the 7 per-describe variants.
- Pure test refactor — production code untouched.

## Why deferred from #288
- The duplicate factories predate #288 by 6 of 7 instances; #288 added the pattern but didn't introduce it.
- The consolidation is a file-wide refactor unrelated to "defensive recipients gates."
- Bundling it would have expanded #288's diff by ~85 lines for a stylistic cleanup.

## Acceptance criteria
- One file-scope `makePipelineParams(overrides = {})`; zero per-describe duplicates.
- 1254-test suite stays green; lint clean.
- Net deletion of ~60-85 lines.

Issue #291, opened 2026-05-13
discord: tighten vacuous-pass on accept-path tests in qurl-send-back-half.test.js


justin-layerv:
Tracking issue raised by PR #288 cr review.

## Context

\`apps/discord/tests/qurl-send-back-half.test.js\` has four "accepts the allowed value" tests, one per entry gate (\`target\`, \`expiresIn\`, \`personalMessage\`, \`recipients\`):

\`\`\`js
test.each([...])('accepts the allowed shape: %s', async (_label, value) => {
  const interaction = makeInteraction();
  try {
    await executeSendPipeline(interaction, makePipelineParams(value));
  } catch (err) {
    expect(err.message).not.toMatch(/the gate's rejection regex/);
    return;
  }
});
\`\`\`

If the call resolves cleanly (mocks line up), the \`try\` returns with **zero assertions fired** and the test passes vacuously. Today the mocks DON'T line up — downstream calls (\`downloadAndUpload\`, \`mintLinksInBatches\`, etc.) throw something the \`expect.not.toMatch\` then validates. But a future test-setup tightening that lets the call resolve happily would silently pass these.

## Sites

- \`tests/qurl-send-back-half.test.js:1554\` — target accept-path
- \`tests/qurl-send-back-half.test.js:1672\` — expiresIn accept-path
- \`tests/qurl-send-back-half.test.js:1716\` — personalMessage accept-path
- \`tests/qurl-send-back-half.test.js:1805\` — recipients accept-path

## Scope

Add \`expect.hasAssertions()\` at the top of each test (or wrap the four tests in a single helper). The pattern:

\`\`\`js
test.each([...])('accepts the allowed shape: %s', async (_label, value) => {
  expect.hasAssertions();
  const interaction = makeInteraction();
  try {
    await executeSendPipeline(interaction, makePipelineParams(value));
    // If we get here, the call resolved — no gate-rejection
    // message to assert against; record one satisfied assertion
    // for the hasAssertions() contract.
    expect(true).toBe(true);
  } catch (err) {
    expect(err.message).not.toMatch(/the gate's rejection regex/);
  }
});
\`\`\`

OR (cleaner alternative): wrap as a helper that records the satisfied-or-failed-with-non-gate-message assertion centrally.

## Why deferred from PR #288

[...]

PR #296, opened 2026-05-13
refactor(discord): factor entry-gate bookkeeping into failGate helper (#292)

### LINKED ISSUE #292: discord: factor 7-gate clearCooldown+cancelEdit+throw triplet into a failGate helper
## Context
Surfaced by Round-6 cr feedback on #288. The `executeSendPipeline` entry gates (7 of them as of this PR) all follow the same triplet:

```js
clearCooldown(interaction.user.id);
cancelEdit();
throw new TypeError(`executeSendPipeline: ... `);
```

A tiny helper would deduplicate the bookkeeping:

```js
function failGate(interaction, ErrorCtor, msg) {
  clearCooldown(interaction.user.id);
  cancelEdit();
  throw new ErrorCtor(msg);
}
```

## Why this was deferred from #288
- The cr reviewer explicitly framed this as "not blocking — the explicit form does keep each gate locally readable."
- #288's framing is "defensive recipients gates," not "refactor the gate bookkeeping pattern."
- Cap-by-cap edits during the cr cycle would have churned every gate twice (the helper + each gate's call-site).

## Scope
- Add `failGate(interaction, ErrorCtor, msg)` adjacent to `truncForLog` in `apps/discord/src/commands.js`.
- Convert all 7 `executeSendPipeline` entry gates to use it.
- Update tests if any inspect the call ordering directly (current tests assert behavior, so should pass unchanged).

## Acceptance criteria
- All 7 gates use `failGate(...)`.
- `cancelEdit` is no longer captured as a closure in `executeSendPipeline` if the helper takes the interaction; if it stays a closure, the helper accepts it as a 4th arg or `cancelEdit` is hoisted to be parameterless.
- 1253-test suite remains green; `npm run lint` clean.
- The change reduces line count in `executeSendPipeline`'s gate region by ≥10 lines.

## Out of scope
- Behavior changes (e.g. unifying error class, swallowing the throw). The helper is purely a bookkeeping consolidation.

justin-layerv:
## Summary
Closes #292. Factors the repeated `clearCooldown + cancelEdit + throw` triplet across all 7 entry gates in `executeSendPipeline` into a single nested `failGate(ErrorCtor, msg)` helper.

## Why
The triplet was duplicated at all 7 gate sites. Each gate body had ~3 lines of identical bookkeeping noise wrapping the gate's actual signal (the `if`-condition + the gate-specific error message). After #288 added two new gates (recipients shape + cap), the cr reviewer flagged this in round 7 as worth a follow-up — "bookkeeping consolidation, distinct from the recipients-gates framing."

## How
- New `failGate(ErrorCtor, msg)` nested inside `executeSendPipeline` (closes over `interaction` + `cancelEdit` — no parameter sprawl).
- Always-throws: linters + readers see one explicit-throw function, not 7 "clear + cancel + throw" sequences.
- The 7 gate bodies now call `failGate(...)` instead of repeating the bookkeeping.

## Notable change: SSRF gate `logger.warn` reordering
The SSRF gate previously did `clearCooldown → cancelEdit → logger.warn → throw`. After refactor, `logger.warn → failGate(...)`. This is:
- **Semantically equivalent** — both run synchronously before the throw; cancelEdit is fire-and-forget so its scheduling order is unobservable.
- **A latent-hazard fix** — if cancelEdit's `editReply` ever threw synchronously in some future Discord.js version, the SSRF breadcrumb would have been lost. Now logged first.

## Diff
- `apps/discord/src/commands.js`: **+27 / -29** (net -2 lines, but the structural win is each gate going from 3-line bookkeeping to a single `failGate(...)` call).
- No test changes — existing tests pin the throw + cancel-edit + clearCooldown behavior, which is unchanged.

## Test plan
- [x] `npm run lint` — clean
- [x] `npx jest` — 1255 tests pass (no regression)
- [x] Manual review: walked through each gate's call site, confirmed equivalent behavior

## Out of scope
- Test refactor (`makePipelineParams` consolidation) — tracked in #293
- Accept-path test tightening — tracked in #291

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -906,12 +906,14 @@ async function executeSendPipeline(interaction, {
     components: [],
   }).catch(logIgnoredDiscordErr);
 
-  // Shared bookkeeping for every entry-gate failure. Nested so it
-  // closes over `interaction` + `cancelEdit` without parameter
-  // sprawl at the call sites. Always throws — a future control-
-  // flow change here (e.g. removing the throw) would surface as
-  // fall-through past every gate and trip the downstream
-  // invariants the gates exist to guard.
+  // Nested so it closes over `interaction` + `cancelEdit` without
+  // parameter sprawl at the call sites. The `@returns {never}`
+  // annotation lets static analysis treat the call as terminating.
+  /**
+   * @param {ErrorConstructor} ErrorCtor
+   * @param {string} msg
+   * @returns {never}
+   */
   function failGate(ErrorCtor, msg) {
     clearCooldown(interaction.user.id);
     cancelEdit();
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -908,11 +908,10 @@ async function executeSendPipeline(interaction, {
 
   // Shared bookkeeping for every entry-gate failure. Nested so it
   // closes over `interaction` + `cancelEdit` without parameter
-  // sprawl at the call sites. Always throws — the function-return
-  // type is `never` in practice, so a future control-flow change
-  // here (e.g. removing the throw) would surface as fall-through
-  // past every gate and trip the downstream invariants the gates
-  // exist to guard.
+  // sprawl at the call sites. Always throws — a future control-
+  // flow change here (e.g. removing the throw) would surface as
+  // fall-through past every gate and trip the downstream
+  // invariants the gates exist to guard.
   function failGate(ErrorCtor, msg) {
     clearCooldown(interaction.user.id);
     cancelEdit();
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -910,7 +910,7 @@ async function executeSendPipeline(interaction, {
   // parameter sprawl at the call sites. The `@returns {never}`
   // annotation lets static analysis treat the call as terminating.
   /**
-   * @param {ErrorConstructor} ErrorCtor
+   * @param {new (msg: string) => Error} ErrorCtor
    * @param {string} msg
    * @returns {never}
    */
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -906,6 +906,19 @@ async function executeSendPipeline(interaction, {
     components: [],
   }).catch(logIgnoredDiscordErr);
 
+  // Shared bookkeeping for every entry-gate failure. Nested so it
+  // closes over `interaction` + `cancelEdit` without parameter
+  // sprawl at the call sites. Always throws — the function-return
+  // type is `never` in practice, so a future control-flow change
+  // here (e.g. removing the throw) would surface as fall-through
+  // past every gate and trip the downstream invariants the gates
+  // exist to guard.
+  function failGate(ErrorCtor, msg) {
+    clearCooldown(interaction.user.id);
+    cancelEdit();
+    throw new ErrorCtor(msg);
+  }
+
   // `isVoiceContext` strict-boolean gate. Unique among the params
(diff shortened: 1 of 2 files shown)


justin-layerv:
## Round 1 — phrasing nit applied (b642348)

Dropped the "function-return type is `never` in practice" framing from the failGate comment per cr nit. That phrasing was TypeScript-only and slightly overstated the JS enforcement — the always-throws contract is genuinely just a comment + the existing clearCooldown-branch tests anticipating exactly this kind of consolidation (test at line 1879: "future refactor that drops one of the two clearCooldown calls is caught by exactly one test failing").

All other items were "LGTM" / non-actionable. Other than this comment tweak, no other changes.

justin-layerv:
## Round 2 — all three nits applied (c84a300)

**1. `@returns {never}` JSDoc** ✅
- Lets static analysis treat `failGate(...)` calls as terminating. Replaces the prior comment-only "always throws" framing with a machine-readable signal.

**2. `@param {ErrorConstructor} ErrorCtor`** ✅
- Documents the type contract for the 7 native-Error-subclass call sites without adding runtime overhead.

**3. Comment volume trimmed** ✅
- 6-line prose comment → 3 lines. The fall-through rationale lives in the original PR description and commit history; keeping it permanently inline was excess for a 3-line function body.

1255 tests pass, lint clean.

justin-layerv:
## Round 3 — two items applied, one skipped (3602724)

**1. JSDoc type tightened** ✅
- `@param {ErrorConstructor}` → `@param {new (msg: string) => Error}`. The former technically refers to the Error constructor only — TypeErrorConstructor / RangeErrorConstructor are distinct types. The structural type matches what all 7 call sites actually pass.

**2. SSRF `logger.warn` ordering pinned with a test** ✅
- New test uses `mock.invocationCallOrder` to assert `logger.warn` fires before `interaction.editReply` (cancelEdit's underlying call). Under the old order (`clearCooldown → cancelEdit → logger.warn → throw`), editReply's invocation order would have come first; under the new order it comes after. A future refactor moving the warn back after cancelEdit now fails loudly rather than silently regressing.

**3. TDZ comment skipped** — the reviewer self-resolved this one ("the existing comment block above failGate is already clear enough").

1256 tests pass (+1), lint clean.

justin-layerv:
## Round 4 — concrete suggestion applied (2092822)

**Test pin tightened** ✅
- Added `expect(interaction.editReply).toHaveBeenCalledTimes(1)` before the invocationCallOrder check. Anchors the `[0]` index to a single observable call so a future change adding more editReply churn on the SSRF path fails this test rather than silently passing against whichever call happens to be `[0]`.

The other two round-4 items were informational only ("for awareness" on the cancelEdit-vs-failGate declaration-style mix, and a note on SSRF sync-throw subtlety in the PR description). Neither requested a code change.

1256 tests still pass, lint clean.

justin-layerv:
## Round 5 — two nits applied, one out-of-scope (637d22b)

**1. Symmetric `toHaveBeenCalledTimes(1)` on logger.warn** ✅
- Round 4 added the anchor to editReply but missed the symmetric one on logger.warn. Applied. Belt-and-suspenders shape matches.

**2. Stale line-number ref dropped** ✅
- Replaced "the 'Preparing links…' edit on line 1014" with a grep-able quote of the literal content. Future insertions above won't rot the reference.

**3. SSRF gate's `Error` vs other gates' `TypeError`/`RangeError`** — out of scope. The reviewer explicitly framed this as pre-existing and "not blocking" (the SSRF gate isn't a type error in spirit). Leaving for a future cleanup if a project-specific error class lands.

1256 tests pass, lint clean.

PR #297, opened 2026-05-13
refactor(discord): consolidate 7 makePipelineParams factories into one helper (#293)

### LINKED ISSUE #293: discord: consolidate 7 duplicate makePipelineParams factories in qurl-send-back-half.test.js

justin-layerv:
## Summary
Closes #293. Hoists 7 duplicate `makePipelineParams` factories in `qurl-send-back-half.test.js` into one file-scope helper.

## Why
Each of the 7 executeSendPipeline-gate describe blocks had its own near-identical `makePipelineParams` factory — same default payload shape, differing only in which field the test varied. Adding a new pipeline param (e.g. when PR 7b lands `/qurl file`) would have required updating all 7. The PR #288 round-7 block already used the strictly-more-useful `(overrides = {})` spread shape; that's now the canonical signature.

## How
- New file-scope `makePipelineParams(overrides = {})` next to `makeInteraction()` at the top of the file.
- Deleted the 7 per-describe variants.
- Call sites went from positional `makePipelineParams(value)` to spread-overrides `makePipelineParams({ field: value })` across ~30 invocations.
- The location-skip SSRF test also dropped its inline params literal and uses `makePipelineParams({ ...overrides })` to inherit the canonical defaults.

## Diff
- `apps/discord/tests/qurl-send-back-half.test.js`: **+59 / -163** (net **-104** lines)
- The issue's acceptance criterion was 60-85 lines; exceeded.

## Test plan
- [x] `npm run lint` — clean
- [x] `npx jest` — **1256 tests pass** (unchanged from baseline)
- [x] Visual diff review: every call site converted to spread-overrides shape; no per-test factory left.

## Out of scope
- Accept-path test tightening (vacuous-pass concern) — tracked in #291.

### DIFF
--- a/apps/discord/tests/qurl-send-back-half.test.js
+++ b/apps/discord/tests/qurl-send-back-half.test.js
@@ -242,7 +242,9 @@ function makeInteraction(overrides = {}) {
 // the baseline explicitly: `{ attachment: { ...DEFAULT_ATTACHMENT,
 // url: 'x' } }`.
 // Frozen so a test forgetting to spread (`DEFAULT_ATTACHMENT.url = ...`)
-// fails loudly instead of silently corrupting the constant for later tests.
+// can't silently corrupt the constant for later tests. (In strict
+// mode this throws; in this file's CommonJS non-strict context the
+// assignment is a silent no-op — either way, the constant stays clean.)
 const DEFAULT_ATTACHMENT = Object.freeze({ url: 'https://cdn.discordapp.com/x', name: 'x.png', contentType: 'image/png' });
 function makePipelineParams(overrides = {}) {
   return {
--- a/apps/discord/tests/qurl-send-back-half.test.js
+++ b/apps/discord/tests/qurl-send-back-half.test.js
@@ -241,7 +241,9 @@ function makeInteraction(overrides = {}) {
 // and `name`). If a test wants to vary just one nested field, spread
 // the baseline explicitly: `{ attachment: { ...DEFAULT_ATTACHMENT,
 // url: 'x' } }`.
-const DEFAULT_ATTACHMENT = { url: 'https://cdn.discordapp.com/x', name: 'x.png', contentType: 'image/png' };
+// Frozen so a test forgetting to spread (`DEFAULT_ATTACHMENT.url = ...`)
+// fails loudly instead of silently corrupting the constant for later tests.
+const DEFAULT_ATTACHMENT = Object.freeze({ url: 'https://cdn.discordapp.com/x', name: 'x.png', contentType: 'image/png' });
 function makePipelineParams(overrides = {}) {
   return {
     apiKey: 'apikey',
--- a/apps/discord/tests/qurl-send-back-half.test.js
+++ b/apps/discord/tests/qurl-send-back-half.test.js
@@ -242,9 +242,7 @@ function makeInteraction(overrides = {}) {
 // the baseline explicitly: `{ attachment: { ...DEFAULT_ATTACHMENT,
 // url: 'x' } }`.
 // Frozen so a test forgetting to spread (`DEFAULT_ATTACHMENT.url = ...`)
-// can't silently corrupt the constant for later tests. (In strict
-// mode this throws; in this file's CommonJS non-strict context the
-// assignment is a silent no-op — either way, the constant stays clean.)
+// can't corrupt the constant for later tests.
 const DEFAULT_ATTACHMENT = Object.freeze({ url: 'https://cdn.discordapp.com/x', name: 'x.png', contentType: 'image/png' });
 function makePipelineParams(overrides = {}) {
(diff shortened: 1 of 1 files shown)


justin-layerv:
## Round 1 — two nits applied (c6a8932)

**1. Dropped `PR 7b` caller-ref from helper docstring** ✅
- Per CLAUDE.md. The general framing ("adding a new pipeline param is a one-line edit here") carries the same value without the rot risk.

**2. Trimmed redundant half of describe intro comment** ✅
- Kept the load-bearing "gate at function entry — mocks don't need to be configured" part; dropped the call-site-redundant "pass `{ isVoiceContext }` to vary that field" half.

The pre-existing `PR 7b` markers at lines 1394 + 1417 are left as-is — reviewer flagged them as pre-existing and "worth a future cleanup pass when 7b actually lands."

The accept-path vacuous-pass concern (#291) stays tracked separately as agreed.

Net diff now: -106 lines (was -104). 1256 tests pass, lint clean.

justin-layerv:
## Round 2 — concrete fix + docstring caveat applied (04e9f34)

**1. `DEFAULT_ATTACHMENT` extracted** ✅
- The spread-overrides idiom was silently dropping `contentType: 'image/png'` from the baseline whenever a test overrode `attachment`. Now: helper spreads `{ ...DEFAULT_ATTACHMENT }` for its default; the two tests that vary the URL spread the baseline: `{ ...DEFAULT_ATTACHMENT, url: '...' }`.

**2. CAVEAT block added to helper docstring** ✅
- Warns future authors that overrides REPLACE nested objects wholesale, with the spread idiom as the prescribed remedy. Closes the structural gap the reviewer identified ("the spread-overrides idiom invites whenever an override field is itself an object").

**3. try/catch-without-fallback-expect** — out of scope, kept tracked in #291 per reviewer's framing.

1256 tests pass, lint clean.

justin-layerv:
## Round 3 — freeze applied (29d354f)

**`Object.freeze(DEFAULT_ATTACHMENT)`** ✅
- A future test that writes `DEFAULT_ATTACHMENT.url = 'x'` (forgetting the spread) now throws in strict mode instead of silently corrupting the constant for every later test.

**Comment trim** skipped — reviewer self-resolved as "I'd keep it" because the CAVEAT half is load-bearing.

1256 tests pass, lint clean.

justin-layerv:
## Round 4 — doc fix applied (161d1fe)

**`Object.freeze` comment corrected** ✅
- Round 3's "fails loudly" framing was strict-mode-only. This file's eslint config is CommonJS non-strict (`sourceType: "script"`, no `'use strict'`), so the assignment is a silent no-op. Comment now captures both modes — the freeze still protects the constant in either, just throws only under strict mode.

The informational note about `contentType` showing up in two converted sites is intentional — that's the direct consequence of round 2's spread-baseline fix. Reverting would undo the cleanup.

1256 tests pass, lint clean.

justin-layerv:
## Round 5 — freeze comment trimmed (480c59a)

**Comment density** ✅
- Dropped the strict-mode-vs-CommonJS dual-mode aside (added in round 4). That was HOW, not WHY — the load-bearing rationale is the missing-spread protection. Keeps round 4's correctness fix by removing the inaccurate "fails loudly" framing entirely, rather than qualifying it.

The other round-5 observations were explicitly non-actionable: comment-block layout flagged "for the next time someone touches this block," and the `recipients` shared-by-reference concern self-resolved as "this is fine — every call gets a fresh array."

1256 tests pass, lint clean.

PR #299, opened 2026-05-13
feat(discord): add recipient-parser module for /qurl file + /qurl map (7b.1)

justin-layerv:
## Summary
**PR 7b.1 of 3** in the rollout to replace `/qurl send`'s bouncy-DM front-half with `/qurl file` + `/qurl map`. Pure-additive piece — lands the mention-parsing module that 7b.2 will wire into the new slash commands' `recipients:` option. **No existing code paths touched**; the module is unused until 7b.2.

## Business context
User feedback on `/qurl send`: the current UX bounces users into DMs to choose recipients after picking a file — confusing and adds friction. The 7b rollout splits this into two purpose-specific commands (`/qurl file`, `/qurl map`) with the recipient picker living in-channel, backed by `flow_state` (the harness landed in PR #275).

For power users, the new commands expose a `recipients:` slash-command option that takes free-form mentions: `/qurl file recipients:<@alice> <@bob> <@&team>`. **That's the option this PR parses.**

## What this module does
- Parses Discord mention syntax: `<@123>` / `<@!123>` users, `<@&id>` roles (expanded via guild member cache).
- Filters: dedupe via `seen` Set, sender exclusion, best-effort bot filter.
- Two-set cap design (`seen` + `ids`): `cappedCount = seen.size - ids.size` is accurate whether overflow source is direct mentions or role expansion.
- Returns `{ ids, invalidTokens, cappedCount }` — never throws.
- **Mass-mention safety:** `@everyone` / `@here` (lowercase only, matching Discord's case-sensitive tokenizer) are pre-escaped with U+200B via regex — handles `@everyone!`, `@here:`, embedded `here@everyone`. Belt-and-suspenders re-escape after per-token truncation. Naive caller interpolation can't fan-out-ping.
- **Per-token cap:** `invalidTokens` entries capped at 256 chars + ellipsis, protecting Discord's 4096-char embed budget.
- **ReDoS guard:** input length-capped at 4000 chars before regex scan.
- **Logger escalation:** at `MASSIVE_OVERSHOOT_MULTIPLIER × cap`, log level promotes from `debug` to `warn` for oncall pattern visibility.

## Why a STRING option (not Discord's `mentionable` type)
Discord's `mentionable` slash-command option type is **single-value**. The power-user path needs many mentions / role mentions / a mix in one shot. A free-form string is the only Discord shape that carries that.

## Test plan
- [x] `npm run lint` — clean
- [x] `npx jest tests/recipient-parser.test.js` — **43 / 43 pass**
- [x] `npx jest` — **1299 / 1299 pass** (was 1256; +43 new)
- [x] Coverage: 5 describe blocks — basic shape, filtering (incl. malformed-interaction precondition, empty-object tolerance), role mentions (incl. cross-role overflow, explicit direct+role dedupe, repeated-invalid-role dedupe), invalid tokens (incl. `@everyone`/`@here` escape with punctuation residue, `@Everyone` case-sensitivity, animated emoji, residue dedupe), cap + ReDoS safety (incl. cappedCount accurate across direct + role overflow, dedupe-then-cap ordering, log threshold escalation at exactly-2× boundary, MAX_INPUT_LENGTH boundary, truncation-inside-token guard, per-token cap, escape-before-truncate ordering)

## Final diff
- `apps/discord/src/recipient-parser.js`: **new file, 354 lines** (heavy WHY comments — option semantics, regex tradeoffs, two-set cap design, mass-mention escape case-sensitivity, markdown-injection caller-responsibility, partial-cache caveat, shared-regex safety, TODO(7b.2) coordination markers)
- `apps/discord/tests/recipient-parser.test.js`: **new file, 680 lines** (43 tests)

## Rollout sequence
- **7b.1 (this PR):** parser module + tests, no wiring
- 7b.2: add `/qurl file` + `/qurl map` slash commands (flow_state-backed in-channel form, wires this parser into the `recipients:` option; narrows the option's `max_length` ≤ MAX_INPUT_LENGTH; surfaces `role.memberCount` vs `role.members.size` divergence; user-visible cappedCount copy; wraps `invalidTokens` in code block when rendering)
- 7b.3: hard-remove `/qurl send` (no grace period, no compat shim)

### DIFF
--- a/apps/discord/src/recipient-parser.js
+++ b/apps/discord/src/recipient-parser.js
@@ -95,11 +95,12 @@ const ROLE_MENTION_RE = /<@&(\d+)>/g;
 // before the API ceiling. 7b.2 should narrow the option's
 // max_length to ≤4000 so this cap becomes defense-in-depth rather
 // than the user-visible truncation point.
-// KEEP IN SYNC with the slash-option `max_length` in 7b.2's
-// builder. If 7b.2 narrows the option's max_length to ≤ 4000,
-// this cap becomes defense-in-depth; if it doesn't, this is the
-// user-visible truncation point (and the partial-mention trim
-// below becomes lossy on legitimate over-cap content).
+// TODO(7b.2): KEEP IN SYNC with the slash-option `max_length` in
+// 7b.2's builder. If 7b.2 narrows the option's max_length to
+// ≤ 4000, this cap becomes defense-in-depth; if it doesn't, this
+// is the user-visible truncation point (and the partial-mention
+// trim below becomes lossy on legitimate over-cap content).
+// Grep `TODO(7b.2)` to find the 7b.1 → 7b.2 coordination markers.
 const MAX_INPUT_LENGTH = 4000;
 
 // Per-token cap on entries in `invalidTokens`. Discord's embed total
@@ -339,7 +340,7 @@ function parseRecipientMentions(raw, interaction) {
       ? logger.warn
       : logger.debug;
     logFn('recipient-parser: capping recipient list at QURL_SEND_MAX_RECIPIENTS', {
-      unique_count: seen.size, cap, capped_count: cappedCount,
+      uniqueCount: seen.size, cap, cappedCount,
     });
   }
 
--- a/apps/discord/src/recipient-parser.js
+++ b/apps/discord/src/recipient-parser.js
@@ -13,9 +13,11 @@
 //   - User mentions:  <@123456> or <@!123456>   → user ID 123456
 //   - Role mentions:  <@&987654>                → expand to current
 //                                                 guild members of that role
-//   - Whitespace, commas, semicolons, pipes, and slashes as
-//     separators (lenient — covers free-form paste from contact
-//     lists / CSV-ish formats)
+// Mention extraction is regex-based — separators don't matter for
+// the mentions themselves (`<@111><@222>` matches both). The
+// separator class only affects the residue/invalid-token strip
+// pass: whitespace, commas, semicolons, pipes, and slashes (lenient
+// — covers free-form paste from contact lists / CSV-ish formats).
 //
 // What we reject (returned in `invalidTokens`):
 //   - Channel mentions <#...>
@@ -46,12 +48,18 @@
 // same per-send cap the in-channel form enforces), and excludes the
 // invoking user + bots (matches the form's self-send + bot rejection).
 //
-// Role expansion uses interaction.guild.members.cache. If the cache is
-// cold (DM context, bot just restarted) the role expansion silently
-// resolves to no members for that role — the role mention lands in
-// `invalidTokens` so the caller can surface "couldn't expand @team-blue,
-// pick users from the menu instead." This matches the existing form's
-// posture (fall through to the picker rather than 404 a slash command).
+// Role expansion uses interaction.guild.members.cache via discord.js's
+// `Role.members` getter, which filters the guild's member cache for
+// the role. PARTIAL-CACHE BLIND SPOT: in large guilds running without
+// `GUILD_MEMBERS` intent (or before chunking), `role.members.size`
+// reflects only the currently-cached subset, not the role's true
+// population. We cannot distinguish "small role" from "large role,
+// partial cache." Cold-cache / DM-context degrades safely (token
+// lands in `invalidTokens`), but a partial-cache silent under-resolve
+// is harder to surface — 7b.2 should consider logging `role.memberCount`
+// vs `role.members.size` when they diverge, and the user-facing copy
+// should say "expanded N members" so users notice if a role appears
+// to expand to far fewer recipients than they expect.
 
 const config = require('./config');
 const logger = require('./logger');
@@ -65,9 +73,13 @@ const USER_MENTION_RE = /<@!?(\d+)>/g;
 const ROLE_MENTION_RE = /<@&(\d+)>/g;
 
 // Hard cap on input length so an adversarial regex-blowup attack
-// (10MB of `<@1>` repetitions) doesn't tie up the worker. The form's
-// 200-char descriptions and Discord's natural 2000-char message cap
-// make 4000 chars more than enough headroom for legitimate use.
+// (10MB of `<@1>` repetitions) doesn't tie up the worker. Discord's
+// slash-command STRING option max is 6000 chars (when not narrowed
+// via setMaxLength); 4000 is a conservative budget that comfortably
+// fits any legitimate recipient-list paste while leaving headroom
+// before the API ceiling. 7b.2 should narrow the option's
+// max_length to ≤4000 so this cap becomes defense-in-depth rather
+// than the user-visible truncation point.
 const MAX_INPUT_LENGTH = 4000;
 
 // Resolve a list of mention tokens (raw "<@..>" / "<@&..>" / etc.) to
@@ -122,14 +134,13 @@ function parseRecipientMentions(raw, interaction) {
   const guild = interaction.guild;
   const cap = config.QURL_SEND_MAX_RECIPIENTS;
 
-  // Mark an ID as considered. Returns true if it's net-new to the
-  // candidate set (caller uses this to drive "role contributed
-  // anything" detection).
+  // Mark an ID as considered: dedupe via `seen`, add to `ids` only
+  // while under cap. Role-contribution detection lives in the role
+  // loop's `usable` counter — see comment there.
   function consider(id) {
-    if (seen.has(id)) return false;
+    if (seen.has(id)) return;
     seen.add(id);
     if (ids.size < cap) ids.add(id);
-    return true;
   }
 
   for (const m of input.matchAll(USER_MENTION_RE)) {
--- a/apps/discord/src/recipient-parser.js
+++ b/apps/discord/src/recipient-parser.js
@@ -285,6 +285,7 @@ function parseRecipientMentions(raw, interaction) {
     .replace(USER_MENTION_RE, ' ')
     .replace(ROLE_MENTION_RE, ' ');
   const leftover = stripped.split(/[\s,;|/]+/).filter(Boolean);
+  const residueSeen = new Set();
   for (const tok of leftover) {
     // Pre-escape `@everyone` / `@here` so a caller interpolating
     // `invalidTokens` into a user-visible message can't accidentally
@@ -307,6 +308,12 @@ function parseRecipientMentions(raw, interaction) {
     if (escaped.length > MAX_INVALID_TOKEN_LENGTH) {
       escaped = `${escaped.slice(0, MAX_INVALID_TOKEN_LENGTH)}\u2026`;
     }
+    // Dedupe residue tokens by the post-escape rendered form so
+    // `<#456> <#456>` or `alice alice` produces ONE entry, not two.
+    // Symmetric with the role-error dedup path (pushRoleErrorIfNew
+    // above) \u2014 a caller's "couldn't parse: X, X" embed is hostile.
+    if (residueSeen.has(escaped)) continue;
+    residueSeen.add(escaped);
     invalidTokens.push(escaped);
   }
 
--- a/apps/discord/src/recipient-parser.js
+++ b/apps/discord/src/recipient-parser.js
@@ -72,12 +72,15 @@ function parseRecipientMentions(raw, interaction) {
   }
   // Length-cap BEFORE regex matching to keep the global-flag iteration
   // bounded under adversarial input. The /g flag scans linearly but
-  // pathological repetitions still allocate per-match strings. If the
-  // cut lands inside a `<...>` token, drop the trailing partial so the
-  // strip-pass doesn't surface a manufactured "invalid token" the user
-  // didn't actually type wrong.
-  let input = raw.length > MAX_INPUT_LENGTH ? raw.slice(0, MAX_INPUT_LENGTH) : raw;
-  if (input.length === MAX_INPUT_LENGTH) {
+  // pathological repetitions still allocate per-match strings.
+  const truncated = raw.length > MAX_INPUT_LENGTH;
+  let input = truncated ? raw.slice(0, MAX_INPUT_LENGTH) : raw;
+  // If the cut lands inside a `<...>` token, drop the trailing partial
+  // so the strip-pass doesn't surface a manufactured "invalid token"
+  // the user didn't actually mistype. Gate strictly on `truncated` —
+  // a naturally-MAX-length input with a stray `<` mid-string is
+  // legitimate content that the strip-pass should surface as-is.
+  if (truncated) {
     const lastOpen = input.lastIndexOf('<');
     const lastClose = input.lastIndexOf('>');
     if (lastOpen > lastClose) {
@@ -158,7 +161,12 @@ function parseRecipientMentions(raw, interaction) {
   const cap = config.QURL_SEND_MAX_RECIPIENTS;
   let finalIds = [...ids];
   if (finalIds.length > cap) {
-    logger.debug('recipient-parser: capping recipient list at QURL_SEND_MAX_RECIPIENTS', {
+    // Massive over-cap (>2× the limit) signals user confusion —
+    // probably pasted a list they didn't trim. Surface at warn so
+    // oncall sees the pattern. Modest overshoot is normal-ish
+    // (typed too many) and stays at debug.
+    const logFn = finalIds.length > cap * 2 ? logger.warn : logger.debug;
+    logFn('recipient-parser: capping recipient list at QURL_SEND_MAX_RECIPIENTS', {
       raw_count: finalIds.length, cap,
     });
     finalIds = finalIds.slice(0, cap);
--- a/apps/discord/src/recipient-parser.js
+++ b/apps/discord/src/recipient-parser.js
@@ -146,10 +146,13 @@ function parseRecipientMentions(raw, interaction) {
   for (const m of input.matchAll(USER_MENTION_RE)) {
     const id = m[1];
     if (id === senderId) continue;
-    // Bot filter — best-effort via cache. A cache miss leaves the bot
-    // in the result; downstream send-pipeline has its own bot check
-    // (see existing form's `Cannot send to a bot` warning path) so
-    // this layer being lossy is acceptable.
+    // Bot filter — best-effort via cache. A cache miss (cold cache,
+    // partial cache without GUILD_MEMBERS intent, or DM-context with
+    // `guild === undefined`) leaves the bot in the result; downstream
+    // send-pipeline has its own bot check (see existing form's
+    // `Cannot send to a bot` warning path) so this layer being lossy
+    // is acceptable. 7b.2 wiring should NOT assume the parser
+    // filtered bots.
     const member = guild?.members?.cache?.get(id);
     if (member?.user?.bot) continue;
     consider(id);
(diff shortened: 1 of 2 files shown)


justin-layerv:
## Round 1 — six concrete nits applied (c19fed9)

**Source changes:**
1. **Mid-mention truncation manufactures invalid tokens** ✅ — trim back to last `<` before scanning if the slice lands inside an open token.
2. **`invalid_tokens` → `invalidTokens`** ✅ — public API now consistent with the camelCase function name. Locking in before 7b.2 wires the call sites.
3. **`__MAX_INPUT_LENGTH` → `MAX_INPUT_LENGTH`** ✅ — dropped the test-private prefix; the constant is exportable as-is.
4. **Over-defensive cap check** ✅ — dropped the `typeof === 'number' && Number.isFinite(cap) && cap > 0` paranoia; trusting `config.js:279`'s `intEnv minPositive: true` invariant.
5. **"Pass 1" / "Pass 2" labels** ✅ — dropped per CLAUDE.md (those were WHAT, not WHY).

**Test changes:**
6. **jest.mock comment** ✅ — noted the mock value (25) is for test convenience; prod default is 50.
7. **Boundary test at exactly `MAX_INPUT_LENGTH`** ✅ — pins the `>` vs `>=` inequality direction.
8. **Partial-mention-after-truncation test** ✅ — locks in fix #1 by asserting `<@…` residue never lands in `invalidTokens`.
9. **Repeated role mention test** ✅ — pins that `<@&id>` appearing multiple times expands once via Set dedup, no double-counted invalid_tokens.

**Deferred to 7b.2 (per reviewer's framing):**
- Dedupe `invalidTokens` at the user-visible seam (back-half embed should dedupe `alice alice`-style repeats when rendering).
- Bot-filter cache-miss surfacing — confirm the back-half's bot-rejection has a user-visible signal.

24 tests in the recipient-parser suite (was 21), **1280 total**. Lint clean.

justin-layerv:
## Round 2 — bug fix + 3 test additions + logger promotion (ac4030b)

**REAL BUG (item 1)** ✅
- Trim-trailing-partial was gated on `input.length === MAX`, which matched BOTH truncated AND naturally-MAX-length input. Fixed to gate strictly on the `truncated` flag. Now a legitimate 4000-char input with a stray `<` mid-string isn't silently truncated.

**Logger promotion (item 7)** ✅
- `logger.debug` → `logger.warn` when `raw_count > cap * 2`. Strong signal of user confusion (probably pasted an untrimmed list) — oncall benefits from seeing it.

**Test additions:**
- **`@everyone` / `@here` land in invalidTokens** (item 3) ✅ — pins they aren't accidentally treated as mentions, and the test comment documents the caller-side escape requirement.
- **Newline separator** (item 4) ✅ — pins that `/[\s,]+/` includes `\n`.
- **Mixed cap + invalidTokens** (item 5) ✅ — 30 valid + 1 channel-mention typo produces 25 ids AND keeps the typo. A future bug that truncated invalidTokens alongside ids is now caught.

**Not changed (item 2 — please reconcile):**
The reviewer flagged that direct user mentions could be silently displaced by role expansion when cap is exhausted. But the current code processes `USER_MENTION_RE` (Pass 1) **before** `ROLE_MENTION_RE` (Pass 2). The cap slice operates on Set insertion order, so direct user mentions are always preserved first.

For the reviewer's example `<@&team-blue> <@alice>`: alice enters via Pass 1, team-blue's members via Pass 2 — alice survives the cap slice regardless of input order. The only way alice could be displaced is if there were >cap direct user mentions ahead of her, which is the expected behavior.

If I'm missing a scenario, happy to add a test pinning the user-mentions-win-over-role-expansion ordering — let me know.

**Deferred to 7b.2:** bot-filter cache reliability for large guilds + caller-side "can't resolve @team-blue" UX copy (item 6).

27 tests in recipient-parser suite (was 24), **1283 total**. Lint clean.

justin-layerv:
## Round 3 — 7 of 8 nits applied (f58dc53)

**Source changes:**
1. **Mass-mention safety** (item 1) ✅ — `@everyone` / `@here` in `invalidTokens` are now pre-escaped with U+200B after the `@`. Pushes safety into the boundary; naive caller interpolation can't fan-out-ping.
2. **`cappedCount` in return shape** (item 2) ✅ — return is now `{ ids, invalidTokens, cappedCount }`. Lets 7b.2 surface "kept 25 of 30" UX without re-parsing. Updated 23 existing `toEqual` fixtures to include `cappedCount: 0`.
3. **Widened separator regex** (item 3) ✅ — `/[\s,]+/` → `/[\s,;|/]+/`. Common CSV-ish paste patterns no longer surface `;` / `|` as invalid tokens.
4. **Short-circuit role expansion** (item 4) ✅ — `if (ids.size >= cap) break;` in the role loop. Skips iterating large roles whose members would be cap-trimmed anyway.
5. **`added` comment** (item 5) ✅ — clarifies it counts iterations past the sender/bot filter, not net-new additions to `ids`.
7. **ReDoS-guard ordering comment** (item 7) ✅ — pins that the length cap MUST precede `matchAll`.

**Test additions:**
8. **Explicit direct+role dedupe** (test gap noted at end of review) ✅ — `<@101> <@&7000>` where role 7000 contains 101 → `ids: ['101']`. Makes the invariant non-incidental rather than implicit in the existing merge test.

**Description/cosmetic:**
6. **PR body line counts** ✅ — updated to reflect current file sizes (224 / 359).

28 tests in recipient-parser suite (was 27), **1284 total**. Lint clean.

justin-layerv:
## Round 4 — both real bugs + both nits applied (86492e4)

**Bugs:**
1. **Inner-loop cap short-circuit** (item 1) ✅ — `if (ids.size >= cap) break;` now sits inside the per-member iteration, not just between roles. A single fat role no longer runs 50k Set.add calls before the slice trims them back.

2. **Regex-wide mass-mention escape** (item 2) ✅ — switched from exact-match to `tok.replace(/@(everyone|here)/g, '@​$1')`. The escape now applies anywhere in any token (`@everyone!`, `@everyone.fix`, `@here:`, embedded `here@everyone`) and replaces all occurrences via `/g`.

**Drift/minor:**
3. **Header separator comment** (item 3) ✅ — updated from "Whitespace + commas" to "Whitespace, commas, semicolons, pipes, and slashes" to match the impl.
4. **dedupe-then-cap test** (item 4) ✅ — new fixture pins that `cappedCount` reflects post-dedupe count: 56 mentions → 28 unique → `cappedCount: 3` (not 31).

**Plus extra tests for item 2:** `@everyone!`, `@everyone.fix`, `@here:`, `here@everyone` (embedded). These were the exact bypass cases the old equality check would have allowed.

30 tests in recipient-parser suite (was 28), **1286 total**. Lint clean.

(Round-3 cr previously failed with a transient Anthropic API 500; re-running cr surfaced this round's substantive feedback — glad it did, items 1 and 2 were real safety holes.)

justin-layerv:
## Round 5 — real bug fixed via redesign (c2d101c)

**Item 1 — accurate cappedCount across role expansion** ✅

You were right, this was a real UX hole. The round-4 inner-loop break left `ids` plateaued at cap, so `cappedCount` always reported 0 when role expansion was the overflow source — a caller couldn't surface "kept 25 of 70" to the user.

**Redesign:** introduced a `seen` Set that tracks every post-filter unique candidate (unbounded by cap), with `ids` as the cap-bounded subset. `cappedCount = seen.size - ids.size` — accurate regardless of overflow source. Trades the perf optimization (inner-loop break) for correctness; the per-iteration cost past cap is just filter checks + a `seen.has` lookup.

**Bonus fix:** the dedupe-then-cap test was actually failing in my intermediate state because the old `overCap` counter incremented per ID *appearance*, not per *unique ID*. `seen` deduplicates correctly.

**New regression test:** `<@&role1> <@&role2>` with 50+20 members, cap=25 → `cappedCount: 45` (was 0 before fix). Pins the exact scenario the cr surfaced.

**Items 2-6 deferred to 7b.2 per your framing:**
- (2) `MAX_INPUT_LENGTH` ≤ slash-option `max_length` pin
- (3) `GUILD_MEMBERS` intent + partial-cache surfacing
- (4) `<@everyone>` is harmless (Discord doesn't expand the angle-bracketed form)
- (5) Preamble trim — keeping for now; the WHY-heavy comments earn their keep on the non-obvious bits
- (6) Test param naming `users:` → `members:` — keeping for now to avoid churn

31 tests in recipient-parser suite (was 30), **1287 total**. Lint clean.

justin-layerv:
## Round 6 — 5 doc/test fixes + PR description (9dfb940)

**Doc/clarity:**
1. **`consider()` dead return removed** ✅ — comment claimed the return drove role-contribution detection, but both call sites discarded it; `usable++` is what actually drives that. Tightened to a 4-line side-effecter.
2. **Separator doc tightened** ✅ — clarified that the separator class only affects residue/invalid-token splitting, not mention extraction.
3. **`MAX_INPUT_LENGTH` rationale fixed** ✅ — old comment cited the 2000-char MESSAGE cap; slash-option STRING max is 6000. New rationale points at the slash-option ceiling and flags that 7b.2 should narrow option `max_length` so this cap becomes defense-in-depth.
4. **`role.members` large-guild caveat expanded** ✅ — now covers the partial-cache scenario (large guild without `GUILD_MEMBERS` intent), with a note that 7b.2 should log `role.memberCount` vs `role.members.size` divergence and surface expansion counts to the user.

**Tests:**
5. **Malformed `interaction.user` precondition test** ✅ — pins that the parser doesn't throw on `{ guild }` (no user); sender-exclusion no-ops, caller bug surfaces at the back-half's `interaction.user.id` read instead.

**PR description:** updated diff section to accurate line counts (246 / 446) and test counts (32, was 28 / was 31). Description text count was stale across the cr cycle as items landed.

**Skipped:** items 6 (informational), 7 (confirmation), 8 (test count), 9 (snowflake-shape fixtures — reviewer explicitly flagged "Low priority").

32 tests in recipient-parser suite, **1288 total**. Lint clean.

justin-layerv:
## Round 7 — bug fix + 3 test pins + 2 doc improvements + 1 doc fix (bdf0ed8)

**Real bug (item 1):** Repeated invalid role mentions duplicated in `invalidTokens` ✅ — added `invalidRoleIds` Set inside the role loop. `<@&999> <@&999> <@&999>` now yields one `invalidTokens` entry, not three.

**Doc improvements:**
- (2) Bot-filter doc now flags DM-context + partial-cache bypass; 7b.2 wiring should NOT assume parser filtered bots.
- (6) Role-expansion doc notes O(guild_size) per-call cost of `Role.members`.

**Test additions:**
- (1-pin) Invalid-role dedupe test.
- (3) Direct-mentions-win-cap ordering test — pins that pass 1 runs before pass 2; direct mentions survive the cap over role expansion. (Caught a subtle issue: had to use numeric IDs since the regex requires `\d+`.)
- (5) Logger.warn vs debug threshold test — 26 unique (cap+1) → debug; 51 unique (2×cap+1) → warn. Pins the `>2×` boundary.

**Doc fix (item 4):** Stale test comment said `/[\s,]+/` but actual regex is `/[\s,;|/]+/`. Updated.

35 tests in recipient-parser suite (was 32), **1291 total**. Lint clean.

justin-layerv:
## Round 8 — 4 polish fixes applied (8c3c84b)

Reviewer framed this as "ship-ready, polish-passes rather than gate items." Applied the 4 cheap ones:

1. **`interaction == null` defensive guard** ✅ — mirrors the `raw` guard. Caller bugs return empty result instead of TypeError.
2. **ZWSP escape consistency** ✅ — normalized to `​` form throughout the tests (was a mix of literal/escape). Future reviewer scanning the file sees the escape, not invisible whitespace.
3. **`raw_count` → `unique_count`** ✅ — `seen.size` is post-dedupe/filter, not raw input count.
4. **Hoisted `pushIfNew` closure** ✅ — now a standalone `pushRoleErrorIfNew(roleId, rawToken)` helper.

**Deferred to 7b.2 (item 5):** `partialRoleExpansions` in return shape — reviewer explicitly framed as "Could equally be done in 7b.2 once the UX is decided."

35 tests, **1291 total**. Lint clean.

justin-layerv:
## Round 9 — 4 nits + test gap applied (24dc8b6)

Reviewer was explicit "Approve." Applied:

1. **Docstring drift** ✅ — line 87 now says `{ ids, invalidTokens, cappedCount }`.
2. **`capped_count` in log payload** ✅ — oncall doesn't have to re-derive from `unique_count` and `cap`.
3. **`MASSIVE_OVERSHOOT_MULTIPLIER = 2` constant** ✅ — replaces the magic `2` in the threshold check; the name documents the oncall-signal intent.
4. **"Same treatment" clarity nit** ✅ — empty-role branch now spells out the consequence inline (lands in `invalidTokens`).
5. **Test gap: `cappedCount: 0` when no mentions** ✅ — pins that `parseRecipientMentions('alice bob', int)` doesn't false-positive the "user pasted too many" inference.

**Deferred to 7b.2:** strip-pass separator `/` splits URLs (informational), role-expansion O(N×guild_size) optimization, partial-token trim depends on `max_length ≤ 4000` narrowing.

36 tests in recipient-parser suite (was 35), **1292 total**. Lint clean.

justin-layerv:
## Round 10 — 4 cheap nits + PR description refresh (d02ab37)

1. **Shared `/g` regex safety note** ✅ — inline comment next to the regex declarations: matchAll clones, replace doesn't mutate lastIndex.
2. **Case-sensitivity invariant documented + tested** ✅ — added "INTENTIONALLY case-sensitive" comment block AND a test pinning `@Everyone <@111>` → `invalidTokens: ['@Everyone']` un-escaped. A future "defensive hardening" PR can't widen to `/i` without flipping this test.
5. **Truncation-guard heuristic one-sidedness explicit** ✅ — now spelled out: false-positive-resistant on residue, lossy by design on legitimate content past MAX. With a note that 7b.2's max_length narrowing makes this defense-in-depth.

**PR description** ✅ — refreshed metrics (297/551 lines, 37 tests; was 246/446/32 in the body).

**Skipped item 3** — boundary test at exactly `cap * 2`. Reviewer explicitly flagged as "probably fine" and not load-bearing unless the threshold is oncall-critical.

**Deferred to 7b.2:** partial-cache telemetry + role-expansion large-guild benchmark.

37 tests, **1293 total**. Lint clean.

justin-layerv:
## Round 11 — all 7 nits applied (2184695)

**Doc:**
1. **Markdown-safety caveat** ✅ — module header now flags that `invalidTokens` is NOT escaped against Discord markdown (`[link]()`, `||spoiler||`, backticks). 7b.2's error renderer MUST wrap in a code block or escape `[`, `]`, `` ` ``, `|`, `*`, `_`. Mass-mention escape is special-cased because it's the only shape with off-channel side effects.
2. **3-state conflation note** ✅ — role-expansion comment now spells out that missing/empty/all-filtered roles all go to `invalidTokens`, with a note that 7b.2 may want to distinguish them.
7. **Variable shadowing** ✅ — `roleMember` rename in the inner role loop.

**Tests:**
3. **`usable++` invariant comment** ✅ — strengthened the "user-in-both" test's comment to make the ordering invariant explicit (the test already pinned the behavior).
4. **`guild: null` test** ✅ — alongside `guild: undefined`. discord.js returns null for DMs.
5. **Exactly-cap boundary** ✅ — new test pinning that `cappedCount: 0` AND no log fires when input is exactly at cap.
6. **Logger payload shape** ✅ — strengthened the overshoot test to assert `{ unique_count, cap, capped_count }` on both debug and warn branches. Future field rename surfaces here, not in prod logs.

39 tests in recipient-parser suite (was 37), **1295 total**. Lint clean.

justin-layerv:
## Round 12 — 2 source notes + boundary test sub-case (66b5876)

1. **`MAX_INPUT_LENGTH` keep-in-sync marker** ✅ — comment next to the constant flagging coupling with 7b.2's slash-option max_length.
2. **`cap` config-invariant note** ✅ — one-line pointer at `intEnv minPositive: true` so the boot-time guarantee is explicit without a runtime guard.
3. **2× cap boundary test** ✅ — threaded into the existing overshoot test as the middle threshold-walking case (cap+1 → 2×cap → 2×cap+1). Pins the strict `>` direction.

**Skipped:** `interaction == null` vs `{}` asymmetry — reviewer explicitly flagged as "fine either way"; the asymmetric behavior is intentional and already documented inline.

**Deferred to 7b.2:** markdown-injection wrapper on the renderer, `role.memberCount` divergence logging, `roleExpansionMs` debug counter.

39 tests, **1295 total**. Lint clean.

justin-layerv:
## Round 13 — near-must-fix + 2 polish (7ce4ae2)

**Real UX fix (item 1):** `invalidTokens` per-token cap at `MAX_INVALID_TOKEN_LENGTH = 256` chars ✅ — a single ~4000-char garbage token would otherwise blow Discord's 4096-char embed budget. Residue tokens now truncate with `…`; role-error tokens (bounded by Discord ID width) are exempt. Test pins both branches (500 → trim; 100 → no-op).

**Polish:**
- **`SECURITY:` prefix on header boundary blocks** (item 5) ✅ — eyes catch the prefix faster than plain prose.
- **`interaction === {}` test** (item 6 gap) ✅ — pins optional-chain contracts against a future refactor that flips `.guild?.x` to `.guild.x?`.

**Skipped:**
- Item 2 (sender-exclusion silent-no-op) — reviewer flagged as "defensible," not a change request.
- Item 3 (`consider` allocation) — reviewer explicit "style note, not a real ask."

**Deferred to 7b.2:** Item 4 (note role-expansion O(guild_size) coupling in 7b.2's commit message).

41 tests in recipient-parser suite (was 39), **1297 total**. Lint clean.

justin-layerv:
## Round 14 — escape-order test + TOKEN cap export (7b2df1b)

1. **Escape-before-truncate ordering test** ✅ — pins that `@everyone` near the start of an oversized token gets ZWS-escaped before truncation. A future refactor flipping to truncate-then-escape would fail this test.
2. **`MAX_INVALID_TOKEN_LENGTH` exported** ✅ — alongside `MAX_INPUT_LENGTH` so 7b.2's renderer can budget per-token size without duplicating the constant.

**Skipped (per reviewer's framing):**
- Surrogate-pair split at boundary — "Probably not worth fixing."
- Snowflake regex looseness — informational only.
- snake_case vs camelCase in log meta — "Minor; up to you"; test pins the field names so divergence is intentional.

42 tests in recipient-parser suite (was 41), **1298 total**. Lint clean.

justin-layerv:
## Round 15 — residue dedupe symmetry (cc83ced)

**Real inconsistency fixed (item 1):** residue tokens (`<#456> <#456>`, `alice alice`) now dedupe via a `residueSeen` Set, parallel to the existing role-error `invalidRoleIds` dedupe. The old comment claimed natural dedupe — only true within an un-separated run. New test pins both `<#456>` and bare-name dedupe cases.

**Deferred / skipped:**
- (Item 2) `cap` read-once snapshot — reviewer "matters approximately never."
- (Item 3) 7b.2 should `require('./recipient-parser').MAX_INPUT_LENGTH` rather than hardcode — already exported; will be confirmed in 7b.2 review.
- (Item 4) `role.members` O(guild_size) — bounded by input cap; 7b.2 will add divergence logging.

43 tests in recipient-parser suite (was 42), **1299 total**. Lint clean.

(For context: this is round 15 of cr. The cr cycle has been thorough but is finding ever-smaller items round-over-round. Happy to keep iterating, but I think we're past the substantive-concerns phase.)

justin-layerv:
## Round 16 — 4 polish fixes applied (d7c41e4)

1. **`usable` comment clarity** ✅ — now explicit that `usable++` runs BEFORE the seen-check inside `consider`, with explicit pointer to the test that catches the ordering.
2. **Belt-and-suspenders re-escape** ✅ — second `.replace(/@(everyone|here)/g, ...)` after truncation. Today's order is safe; this removes a reasoning gotcha if the escape pattern ever changes.
4. **Stale `config.js:279` line refs** ✅ — replaced with named references in both source and test.
6. **Test constant imports** ✅ — `MAX_INVALID_TOKEN_LENGTH` imported, hardcoded `257` replaced with `MAX_INVALID_TOKEN_LENGTH + 1`.

**Skipped:**
- Item 3 (role-expansion ceiling) — informational for 7b.2.
- Item 5 (env-var literal in log) — reviewer "fine for ops-readability."
- Test gap (literal `<text>` mid-truncation) — reviewer "Consider for 7b.2" not blocking.

43 tests, **1299 total**. Lint clean.

justin-layerv:
## Round 17 — convergence applied (499fe22)

1. **camelCase log fields** ✅ — `unique_count` → `uniqueCount`, `capped_count` → `cappedCount`. Aligns with `apps/discord/src/`'s log-meta convention.
2. **PR description line counts synced** ✅ — 354/680, 43 tests, 1299 total.
3. **`TODO(7b.2):` marker on MAX_INPUT_LENGTH** ✅ — `git grep TODO(7b.2)` finds the coordination markers.
4. **Animated emoji test (`<a:name:id>`)** ✅ — alongside the static `<:name:id>` case.

The cr cycle has been 17 rounds — exceptional thoroughness. The last several rounds have been progressively smaller polish items (camelCase consistency, TODO markers, line-count syncs). Substantive concerns landed in earlier rounds (cap accounting, mass-mention escape, per-token truncation, residue dedupe).

**Final state:** 43 tests, 1299 total, lint clean. Module is pure-additive and unused-until-7b.2. Ready to merge.

Issue #283, opened 2026-05-13
Prod E2E smoke does not exercise viewer_ttl_seconds < 300s — session_duration regressions land green


vikramlayerv:
## Incident (2026-05-13)

\`qurl-integrations-infra#540\` introduced a connector-side session_duration forwarding bug that broke every \`/qurl send\` with a Snapchat-TTL value (e.g. \`viewer_ttl_seconds=30\`). The prod e2e smoke ran *immediately* after the breaking deploy and **passed** — because every smoke test uploads with no \`viewer_ttl_seconds\`, taking the connector's \"use server default\" branch and never exercising the session_duration → qurl-service contract.

The breaking PR shipped to prod with a green smoke signal. User-visible failure surfaced only when real traffic hit the new code path.

## Coverage gap

Grep the prod \`*.smoke.test.ts\` files in \`qurl-integrations/e2e/tests/\`: **zero** references to \`viewer_ttl_seconds\` or \`session_duration\`. The Snapchat-TTL feature was added long ago but never picked up smoke coverage.

## What to add

A new smoke case in \`qurl-send-revoke.smoke.test.ts\` (or similar) that:

1. Sends a \`/qurl send\` with \`viewer_ttl_seconds=30\` (or any value < 300s).
2. Asserts the bot's response contains a working qurl_link (not \"Failed to create links\").
3. Recipient-side: clicks the link → file renders → recipient sees auto-destruct countdown.

Add the same case to the OAuth-setup smoke if onboarding flow also exercises this path.

## Acceptance

- Smoke case exists, runs in CI, and would have failed against \`qurl-integrations-infra@56fedcf\` (the broken state we just rolled back from).
- Verifies both directions: bot accepts the request AND recipient-side gets the file.

Tag: pr-polish-followup, observability, incident-postmortem

Issue #276, opened 2026-05-13
Bot "Failed to create links" path doesn't log or increment any error metric


vikramlayerv:
## Symptom

During today's sandbox /qurl send outage, two users reported \"Failed to create links. Please try again.\" Bot HTTP-service CloudWatch log group \`/ecs/qurl-bot-discord-sandbox-http\` had **zero events** in the window — the bot wasn't logging the error path that produces that message.

## What to fix

Wherever the bot wraps a downstream failure (connector 4xx, qurl-service 5xx, network timeout) as \`\"Failed to create links. Please try again.\"\` for Discord display, before returning that message:

1. \`logger.error\` with: the downstream URL hit, response status, response body (truncated to 1KB), and request id. Without this, every future regression of this shape is invisible from the outside.
2. Emit a counter metric \`qurl_send_create_link_failure_total{reason=\"connector_400\" | \"qurl_service_5xx\" | \"timeout\" | ...}\`.

## Where to alarm

\`#alerts-sandbox\` (sandbox) / \`#all-layerv\` (prod) when the counter rate exceeds (e.g.) 5 events / 5min for any single \`reason\` label. Tracking the reason label rather than a single counter so a real qurl-service-side bug doesn't mask itself as \"transient noise.\"

## Acceptance

- Bot code change in \`qurl-integrations/apps/discord/\` adds the structured log + the prom metric.
- The metric is scraped by the bot's existing Prometheus exporter (the bot already exposes \`/metrics\` per \`project_discord_connector_migration_done.md\`).
- \`alerts-infra\` adds the corresponding alarm (separate PR).

## Repro for this incident

Today (2026-05-12 ~22:14 EDT), sandbox bot received slash commands but the bot didn't log a single event during the user's failure window (10 min). The user-facing error came from a wrapper that swallowed the downstream details. There's no log line to grep for \"why did this 8:14 PM /qurl send fail\" — the answer had to be inferred from cross-PR contract analysis.

Tag: pr-polish-followup, observability

PR #300, opened 2026-05-13
fix(discord): close two post-incident observability gaps (#276 + #283)

vikramlayerv:
## What

Two follow-ups from the 2026-05-13 \`/qurl send\` outage post-mortem:

### #276 — bot \"Failed to create links\" path doesn't log or alarm

The catch block at \`apps/discord/src/commands.js::handleSend\` logs at \`logger.error\` level but doesn't emit an audit event. The CloudWatch log-metric filters that drive alarms only key off the \`audit.event\` field in our log envelope, so the existing log line is invisible to the alarm layer.

**Fix:** add a new \`QURL_SEND_CREATE_LINK_FAILURE\` audit event, emitted once per failed send. Categorized via a new \`classifyMintFailure\` helper into \`timeout\` / \`upstream_4xx\` / \`upstream_5xx\` / \`unknown\` so the dashboard can split a real qurl-service outage from a per-user 4xx (validation, contract drift, etc.).

The audit event also carries \`api_code\`, \`status_code\`, and \`kind\` (file/location) for forensic-query convenience. Names frozen in \`AUDIT_EVENTS\` so a runtime mutation can't silently break the metric.

**Paired terraform** (alarm + filter) lives in qurl-integrations-infra and will land in a follow-up PR — kept separate so this PR doesn't span repos.

### #283 — e2e smoke doesn't exercise \`viewer_ttl_seconds < 300s\`

PR qurl-integrations-infra#540 added connector-side \`session_duration\` forwarding. Prod qurl-service had a 5-minute floor at deploy time. The existing smoke uploads with no \`viewer_ttl_seconds\`, taking the connector's \"use server default\" branch — bypassing the session_duration → qurl-service contract entirely. The contract regression was invisible to smoke; users hit it for ~4 hours before someone noticed.

**Fix:** add \`viewerTtlSeconds\` parameter to the \`uploadFile\` helper, plus a new \`qurl-send-viewer-ttl.smoke.test.ts\` that pins both paths:
- TTL'd path (\`viewer_ttl_seconds=30\`) — the regression-class catcher
- No-TTL path — control, catches refactors that break the \"use server default\" branch

## Files

| File | Change |
|---|---|
| \`apps/discord/src/constants.js\` | + \`QURL_SEND_CREATE_LINK_FAILURE\` audit event with comment |
| \`apps/discord/src/commands.js\` | + \`classifyMintFailure\` helper, audit emission in the catch block, enriched logger.error fields (status, sendId) |
| \`e2e/helpers/qurl-api.ts\` | + optional \`viewerTtlSeconds\` on \`uploadFile\` |
| \`e2e/tests/qurl-send-viewer-ttl.smoke.test.ts\` | new — TTL'd + no-TTL smoke cases |

## Tests

- \`apps/discord\` Jest: 45 suites, 1256 tests, all pass
- \`e2e\` TypeScript compile: clean (only pre-existing tsconfig deprecation warning)

## Closes

- ✅ qurl-integrations#276 (bot send-failure observability)
- ✅ qurl-integrations#283 (TTL smoke coverage)

## Follow-up (separate PR)

- qurl-integrations-infra terraform: log-metric filter on \`audit.event = qurl_send_create_link_failure\` + alarm wired to the existing \`alerts_connector\` SNS topic. Will reference this PR.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -99,9 +99,10 @@ function emitMintFailureAudit(error, { sendId, kind }) {
   logger.audit(AUDIT_EVENTS.QURL_SEND_CREATE_LINK_FAILURE, {
     send_id: sendId,
     reason: classifyMintFailure(error),
-    // `?? null` (not `|| null`) preserves a legitimate `0` status (e.g.
-    // fetch-on-DNS-failure surfaces 0) or empty-string `apiCode` —
-    // collapsing those to null would lose a dimension silently.
+    // `?? null` (not `|| null`) preserves an empty-string `apiCode` —
+    // `|| null` would collapse it to null and lose a forensic dimension.
+    // (Status 0 doesn't apply: native fetch throws on transport errors
+    // rather than returning a 0-status response.)
     api_code: error?.apiCode ?? null,
     status_code: error?.status ?? null,
     kind,
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -52,15 +52,12 @@ const TOKENS_PER_RESOURCE = 10;
 // 429 handling).
 function classifyMintFailure(error) {
   if (!error) return 'unknown';
-  // AbortError disambiguation runs BEFORE the message-regex branch
-  // because undici/fetch raises AbortError on deadline-fired aborts
-  // AND on user-cancellations. Bucket as `timeout` only when
-  // error.cause corroborates a timeout signal; otherwise fall through
-  // to `unknown` so a future cancel-button adoption doesn't silently
-  // mis-bucket cancellations as timeouts. Without this ordering, an
-  // AbortError whose message happens to contain "timeout" would
-  // short-circuit through the regex below and bypass the cause check.
-  // See PR #300 review (Justin).
+  // AbortError is ambiguous: undici/fetch raises it on deadline-fired
+  // aborts AND on user-cancellations. error.cause is the only reliable
+  // disambiguator — bucket as `timeout` only when cause corroborates a
+  // timeout signal; otherwise fall through to `unknown` so a future
+  // cancel-button adoption doesn't silently mis-bucket cancellations
+  // as timeouts. See PR #300 review (Justin).
   if (error.name === 'AbortError') {
     const causeStr = String((error.cause && (error.cause.message || error.cause)) || '');
     if (/timeout/i.test(causeStr)) return 'timeout';
@@ -1212,10 +1209,13 @@ async function executeSendPipeline(interaction, {
     // un-instrumented and tonight's outage shape (4+ hours of users
     // hitting this before anyone noticed) repeats. qurl-integrations#276.
     //
-    emitMintFailureAudit(error, {
-      sendId,
-      kind: resourceType === RESOURCE_TYPES.FILE ? 'file' : 'location',
-    });
+    // Explicit map (RESOURCE_TYPES.MAPS → 'location' for parity with the
+    // UPLOAD_SUCCESS / addRecipients vocabulary) instead of `=== FILE ?
+    // 'file' : 'location'` so a future third RESOURCE_TYPES value
+    // surfaces as `null` rather than silently bucketing as 'location'.
+    // quota_exceeded skip lives inside emitMintFailureAudit.
+    const kindMap = { [RESOURCE_TYPES.FILE]: 'file', [RESOURCE_TYPES.MAPS]: 'location' };
+    emitMintFailureAudit(error, { sendId, kind: kindMap[resourceType] ?? null });
     logger.error('Failed to prepare QURL links', {
       error: error.message,
       apiCode: error.apiCode,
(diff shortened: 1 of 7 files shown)


vikramlayerv:
## pr-polish: intentionally unresolved items (updated after Justin + Claude round-14)

Loop closed at SHA `18bfbf6` after **14 polish rounds**. Latest Claude verdict: *"Approving in spirit — a few minor observations below, none blocking."* Justin's `AbortError false-positive risk` ask is addressed in-code; the broader loop went through 9 additional iterative rounds of doc-rot cleanup + correctness fixes that Claude reviews surfaced. 1273/1273 jest tests pass; mergeable: MERGEABLE.

### Justin's ask (resolved)
- **AbortError false-positive risk** — `classifyMintFailure` now gates `AbortError → timeout` on word-anchored `error.cause` corroboration only. Bare AbortError + any non-timeout cause buckets as `unknown`. Pinned by 3 new tests. Filed [#303](https://github.com/layervai/qurl-integrations/issues/303) earlier for the further `error.cause = 'user-cancel'` distinguisher when the cancel button lands.

### Real bugs surfaced + fixed during the loop (would have shipped otherwise)
1. `handleAddRecipients`'s outer catch was un-instrumented for the same incident shape (#276) — would have under-counted the alarm. Round 6.
2. `kind: hasFile ? 'file' : 'location'` mis-labeled mixed sends because the file branch returns early on its own failure. Replaced with `activeKind` per-branch tracking. Round 7. Pinned by the `mixed file+location, file succeeds, location fails → kind=location` test.
3. AbortError ordering bug: `/timeout/i` message-regex ran before the AbortError branch, so an AbortError with timeout-shaped message bypassed cause-corroboration. Round 9. Regex eventually removed entirely (round 11) — name/code paths cover real shapes; substring match was a false-positive class.
4. `result.qurl_link` assertion in e2e smoke was unreachable because connector `/upload` doesn't return that field (mint is separate). Would have failed in real CI even on healthy contract. Round 11.

### Intentionally unresolved (non-blocking per Claude)
| Severity | Finding | Why accepted | Future plan |
|---|---|---|---|
| Minor | `/\btimed?\s*out\b/i` on AbortError cause still matches "not due to timeout" — `timeout` is a complete word | Word-anchoring drops "rtimeout"/concatenation oddities; full negation defense would require shape-checking (cause.name === 'TimeoutError'), which breaks existing positive tests. Real undici deadlines raise TimeoutError separately | None — acceptable; inline comment honestly documents the limit |
| Minor | 429 buckets as `upstream_4xx`; future `rate_limit` reason class foreshadowed in docstring | Premature to add now; cost is dashboard-side split, not data integrity | Sequenced with the actual upstream 429-handling work |
| Minor | `logger.audit` throw-safety: if audit throws, `logger.error` + `clearCooldown` are skipped | Vanishingly unlikely with current logger; would harden via try/catch | None — accepted, no concrete trigger |
| Minor | smoke `result.hash` regex pins MD5 (`^[0-9a-f]{32}$`) — sha256 migration would fail noisily | Discoverable not silent; intentional contract pin | None — pin is the contract |
| Minor | `addRecipients` CDN-expiry path (signed Discord URLs expire ~24h) now contributes `unknown`/`kind=file` audits | Operator-knowable baseline once terraform-side filter lands | Terraform PR (paired with this one) — filter `unknown+isExpired` subset out of alarm |
| Cross-repo | Audit emission is dead weight until paired terraform log-metric filter + alarm lands | Tracked in PR body; separate qurl-integrations-infra PR | Sibling PR (cross-repo gate) |

### Loop-overhead note
13+ rounds is a lot. Several rounds were doc-rot cleanup from earlier rounds — Claude even called this out explicitly in round 13 ("Worth a quick post-mortem on whether the multi-round pattern is amplifying overhead"). End state is solid, but a more careful initial pass could have saved ~4-5 rounds. Future PRs should treat the first 3 rounds as the substantive loop and stop unless something material surfaces.

### Filed during polish
- [#303](https://github.com/layervai/qurl-integrations/issues/303) — AbortError `error.cause = 'user-cancel'` distinguisher + location-side audit-emission test

— via /pr-polish

justin-layerv:
Closing — fully superseded. This PR's two bundled fixes were split, rebased onto current main, polished through review, and merged as three focused PRs:

- **#283 (e2e smoke):** #626 — merged, and verified live-green against the sandbox connector.
- **#276 (send-failure observability):** #627 (the discord audit event) + qurl-integrations-infra#928 (the CloudWatch metric-filters + spike/sustained alarms) — both merged. The cross-repo contract was confirmed end-to-end by a live synthetic-match in sandbox (metric incremented +1, alarms healthy).

The rebase also fixed drift this PR couldn't have anticipated (the `newLinks` rename, the deleted test files) and several real bugs the review surfaced (the alarm would have excluded its own incident's 4xx; the CDN-expiry skip was silently suppressing real connector 403s). Thanks @vikramlayerv for the original work.

Remaining for #276: promote the alarm to prod (it's sandbox-only so far). Tracked follow-ups: #631, #634.

PR #301, opened 2026-05-13
feat(discord): add /qurl file + /qurl map slash commands (7b.2)

justin-layerv:
## Why

`/qurl send`'s button-driven wizard works but forces every user through a multi-step flow even when they could type the send in one shot. PR 7b.2 ships two slash-command-driven entry points so power users can stay on the keyboard, and lays the foundation for `/qurl send`'s retirement in PR 7b.3 (the legacy form's in-memory `await*` primitives don't survive deploys mid-flow — the new commands' flow_state-backed confirm card does).

This is the second of three PRs converting `/qurl send` to a flow_state-backed shape:
- **7b.1 (#299, merged):** Recipient-mention parser.
- **7b.2 (this PR):** `/qurl file` + `/qurl map` subcommands + confirm card + flow_state wiring.
- **7b.3 (next):** Hard-remove `/qurl send`'s front-half (no grace period).

## Conflict with #294

PR #294 (different author) adds `/qurl send_file` + `/qurl send_location` as **stubbed** shortcuts that delegate to `/qurl send`'s Step 3 form. The "next commit on that branch" plans to extract the legacy form into a shared helper.

This PR takes the architecturally opposite direction agreed in the PR 7-9 rollout plan: **retire** the form (PR 7b.3) rather than refactor it. Building flow_state-backed handlers now — instead of reusing the in-memory form — is what makes the per-shard rollout in PRs 10-15 actually safe under deploy churn (the legacy form drops every in-flight send when an ECS task cycles).

After this PR lands, #294 can be closed (its work is superseded by /qurl file + /qurl map).

### New subcommands
- **`/qurl file`** — required `attachment:`, optional `recipients:` (text mention list), `expires-in:`, `self-destruct:`, `personal-message:`.
- **`/qurl map`** — required `location:` (Google Maps URL or place text), optional `recipients:`, `location-name:` (override), plus the same shared options.

### Dual-input recipients
- **With `recipients:` text** → `parseRecipientMentions` (PR 7b.1) → guild `members.fetch` in parallel via `batchSettled` → bot/self filter → confirm card.
- **Without `recipients:` text** → confirm card renders a `UserSelectMenu`; picker interactions transition the same flow row's payload (stage unchanged, version + TTL refreshed via `transitionFlow`) until Send.

### Confirm card (flow_state-backed)
- `supersedeOrCreate({ stage: 'awaiting_send_confirm', ttl_seconds: 180 })` at command invocation.
[...]

PR #308, opened 2026-05-14
fix(discord): restore button-driven selection on /qurl file + /qurl map confirm card

justin-layerv:
## Summary

PR 7b.2 (#301) shipped \`/qurl file\` and \`/qurl map\` with slash-option-only entry for expiry, self-destruct, and personal message — the confirm card had no inline controls. After picking recipients, users hit a Send/Cancel-only card with no way to change those values short of canceling and re-running the slash command. Reported in-channel by a user testing the live bot.

This PR restores parity with the legacy \`/qurl send\` UX:

- **Expiry StringSelectMenu** (30m / 1h / 6h / 24h / 7d), current value marked \`default: true\`
- **Self-destruct StringSelectMenu** (No timer + presets), distinct value-space from the slash option (\`no-timer\` sentinel vs \`'none'\`)
- **"Add a note" / "Edit note" button** → ModalBuilder with the current raw user input pre-filled (NOT the sanitized form — see "Round-trip correctness" below)

Slash-option defaults still feed the initial card state — this is additive UX, not a behavior change.

## Architecture

Each menu handler reuses the same flow_state transition + OCC fencing pattern as the recipient picker:
- Same \`expectedStage\` (\`SEND_STAGE_AWAITING_CONFIRM\`), \`expectedVersion\` from the row — race-fenced against picker and sibling menus
- All four entry paths (picker, expiry, self-destruct, note modal) defer-ack first then re-render via \`editReply\` through the shared \`rerenderConfirmCard\` helper
- Note button does NOT touch flow state — opens the modal directly, so repeated open/close doesn't bump the version or fence siblings
- Modal submit \`deferUpdate\`s first (3s ack budget against slow DDB writes), then \`editReply\` on success / \`followUp\` on throw
- Picker is **always attached** on every render (slash entry + every menu interaction) so the row layout is stable from frame 0 — matches \`handleSendUserSelect\`'s post-pick contract

## Round-trip correctness

Storage splits the personal message into two fields:
- \`personalMessageRaw\` — trimmed user input, capped via \`safeCodepointSlice\` at \`PERSONAL_MESSAGE_INPUT_MAX = 280\` (the Discord-enforced input bound on both ingress paths). **For modal pre-fill ONLY.** Never read it from any rendering path or downstream pipeline — only \`personalMessage\` (the sanitized form) is safe for those.
- \`personalMessage\` — \`sanitizeMessage(raw)\`, the render-ready sanitized form.

A static contract test pins that \`executeSendPipeline\`'s function body never references \`personalMessageRaw\` — defense against future bit-rot where a refactor might accidentally read raw instead of sanitized.

Both forms derive from the same trimmed-and-capped raw at every ingress (slash entry + modal submit), so an Edit-and-resubmit round-trip is idempotent. Without this split, pre-filling from the sanitized form would re-sanitize on no-op resubmit and double-escape the backslashes on every cycle.

## Cache-miss recipient fallback

\`rerenderConfirmCard\` resolves picked users through three layers in priority order:
1. \`interaction.guild.members.cache\` — freshest data when populated
2. \`payload.recipientAliases\` (new field) — persisted at slash-entry / picker-rePick time; survives cache eviction between pick and menu-click
3. \`user-\${id}\` fabricated fallback — last resort

The recipient warningsBlock is also persisted in payload so menu interactions render with the same warning surface as the initial card.

## Defense-in-depth

Both menu handlers validate the picked value against their closed legitimate set **before** \`deferUpdate\` so the forgery branch uses the cheaper single-call \`reply\` ack:
- **Expiry**: rejects values not in \`EXPIRY_LABELS\`, surfaces \"Unrecognized expiry value\" via ephemeral \`reply\`, logs \`warn\` for forensics, NO transitionFlow.
- **Self-destruct**: rejects values not in \`{no-timer sentinel, ...preset seconds}\`, same shape as expiry — \`reply\` + \`warn\` + no save. Discord enforces the option set server-side, so an off-set value here is forgery; silently mutating state (a previous design) would clear a user's prior timer on every probe.

Modal submit \`sanitizeMessage\`s the input: NFKC + bidi strip + @-mention defuse + markdown escape + 500-codepoint cap.

## Tests

Coverage matrix across 6 describe blocks:

- \`handleSendConfirmExpirySelect\` — happy path, forged value rejection via reply (asserts no deferUpdate), conflict / not_found / throw, payload field preservation
- \`handleSendConfirmSelfDestructSelect\` — happy path, \`no-timer\` sentinel mapping, forged value rejection via reply + warn-log (asserts no deferUpdate, no transitionFlow), conflict / not_found, legitimate-value-no-false-positive
- \`handleSendConfirmNoteButton\` — modal opens without mutating flow state, modal pre-filled from \`personalMessageRaw\` (NOT the sanitized form), legacy-row fallback to empty, idempotent after picker
- \`handleSendConfirmNoteModal\` — deferUpdate gates the ack, happy path with trim + sanitize, round-trip idempotency (no double-escape on no-op resubmit), empty / whitespace → \`null\`, conflict / not_found via \`editReply\`, throw → \`followUp\` with negative-assertion that \`reply\`/\`update\` aren't called post-defer
- \`rerenderConfirmCard\` cache-miss fallback — renders persisted alias when member-cache is empty, warningsBlock carries across menu interactions
- \`renderConfirmCardRows\` — 4 rows always (picker stable), button construction order, Add/Edit label flip

Plus \`/qurl map\` payload-parity test pinning that both entry points share the same payload-construction path (\`recipientAliases\`, \`warningsBlock\` populated identically).

Plus the static CONTRACT test pinning \`executeSendPipeline\` never reads \`personalMessageRaw\`.

Happy-path tests for all three menu handlers pin the user-facing rendering — \`expect(lastEdit.content).toMatch(/7 days/)\` etc. — so a refactor that lands the right value in DDB but renders the old label would fail.

\`/simplify\` ran clean.

## Follow-up

#309 — Modal typed-note-lost-on-conflict UX paper cut (filed as enhancement; future work to surface typed text via ephemeral followUp on supersede).

## Test plan

- [ ] CI green (1485 tests pass locally, lint clean)
- [ ] \`cr\` workflow clean
- [ ] **Manual verification in beta bot guild after merge** — UI not yet exercised in a live Discord client; the test suite verifies wire-protocol + render content but not visual layout / row order / select default-selection rendering

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3873,36 +3873,27 @@ function formatPersonalMessagePreview(message) {
   return `> ${safeCodepointSlice(oneLine, 80)}…`;
 }
 
-// Build the ActionRow set for the confirm card. When `attachPicker`,
-// row 0 is a UserSelectMenu (recipients absent OR a prior pick is in
-// the payload and the user might want to re-pick); otherwise no
-// picker row. Self-destruct + Expiry StringSelectMenus follow (each
-// requires its own row — Discord forbids combining selects with
-// buttons in the same ActionRow). The last row packs the optional
-// note button + Send + Cancel.
+// Build the ActionRow set for the confirm card. Always 4 rows:
+// UserSelectMenu (always attached so re-picks stay available and
+// the layout is stable from frame 0), Self-destruct StringSelectMenu,
+// Expiry StringSelectMenu (each select needs its own row — Discord
+// forbids combining selects with buttons in the same ActionRow),
+// and the bottom row packing Note button + Send + Cancel.
 //
 // Row math (Discord 5-row max):
-//   attachPicker=true:  UserSelect + SelfDestruct + Expiry + [Note,Send,Cancel] = 4 rows
-//   attachPicker=false: SelfDestruct + Expiry + [Note,Send,Cancel] = 3 rows
-//
-// Renamed from `needsPicker` so the content/rows divergence at the
-// post-pick call site (content rendered WITHOUT picker, rows rendered
-// WITH picker — to let the user re-pick) is impossible to silently
-// merge in a future refactor.
+//   UserSelect + SelfDestruct + Expiry + [Note, Send, Cancel] = 4 rows
 function renderConfirmCardRows({
-  attachPicker, sendDisabled, expiresIn, selfDestructSeconds, personalMessage,
+  sendDisabled, expiresIn, selfDestructSeconds, personalMessage,
 }) {
   const rows = [];
-  if (attachPicker) {
-    const maxValues = Math.min(USER_SELECT_PER_PICK_CAP, config.QURL_SEND_MAX_RECIPIENTS);
-    rows.push(new ActionRowBuilder().addComponents(
-      new UserSelectMenuBuilder()
-        .setCustomId(SEND_USER_SELECT_CUSTOM_ID)
-        .setPlaceholder(`Pick recipients (1–${maxValues})`)
-        .setMinValues(1)
-        .setMaxValues(maxValues)
-    ));
-  }
+  const maxValues = Math.min(USER_SELECT_PER_PICK_CAP, config.QURL_SEND_MAX_RECIPIENTS);
+  rows.push(new ActionRowBuilder().addComponents(
+    new UserSelectMenuBuilder()
+      .setCustomId(SEND_USER_SELECT_CUSTOM_ID)
+      .setPlaceholder(`Pick recipients (1–${maxValues})`)
+      .setMinValues(1)
+      .setMaxValues(maxValues)
+  ));
   // Self-destruct StringSelectMenu. Mirrors the OLD /qurl send form's
   // shape at commands.js:~2334: "No self-destruct timer" + 7 presets.
   // The `default: true` flag on the matching option keeps the
@@ -3913,6 +3904,16 @@ function renderConfirmCardRows({
   // render the first option's label — wrong UX).
   const hasTimer = Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0;
   const hasMatchingPreset = hasTimer && SELF_DESTRUCT_PRESETS.some((p) => p.seconds === selfDestructSeconds);
+  if (hasTimer && !hasMatchingPreset) {
+    // Symmetric with the expiry off-set warn below: a finite positive
+    // selfDestructSeconds that misses every preset means a corrupted
+    // row. The "No timer" fallback keeps the header sensible but the
+    // content line still renders "self-destructs in N seconds" — log
+    // so the asymmetric state surfaces in forensics.
+    logger.warn('renderConfirmCardRows: off-preset selfDestructSeconds in payload — falling back to No timer default', {
+      selfDestructSeconds: truncForLog(selfDestructSeconds),
+    });
+  }
   rows.push(new ActionRowBuilder().addComponents(
     new StringSelectMenuBuilder()
       .setCustomId(SEND_CONFIRM_SELF_DESTRUCT_SELECT_CUSTOM_ID)
@@ -4216,14 +4217,6 @@ async function handleQurlSlashSend(interaction, params) {
       interaction,
     });
     const rows = renderConfirmCardRows({
-      // Always attach the picker so the row layout is stable from
-      // frame 0 — without this, a slash entry that supplied
-      // `recipients:` would render 3 rows initially and then jump to
-      // 4 the first time the user touched any menu (rerenderConfirmCard
-      // hard-codes attachPicker: true). Always-attached also matches
-      // handleSendUserSelect's post-pick contract: the picker stays
-      // available for re-picks, even after a successful pick.
-      attachPicker: true,
       sendDisabled: needsPicker,  // Send stays disabled until UserSelectMenu fires
       expiresIn,
       selfDestructSeconds,
@@ -4581,7 +4574,6 @@ async function handleSendUserSelect(interaction, { flow_id, row }) {
       interaction,
     }),
     components: renderConfirmCardRows({
-      attachPicker: true,
       sendDisabled: true,
       expiresIn: payload.expiresIn,
       selfDestructSeconds: payload.selfDestructSeconds,
@@ -4661,13 +4653,11 @@ async function handleSendUserSelect(interaction, { flow_id, row }) {
     }).catch(logIgnoredDiscordErr);
   }
 
-  // Intentional flag divergence: the CONTENT renders the resolved
-  // recipient summary ("**To:** N user(s)") — needsPicker:false —
-  // while the ROWS keep the UserSelectMenu attached so the user can
-  // re-pick. Send is enabled because we now have a valid recipient
-  // set. The two flags are deliberately fed opposite values; the
-  // rename from `needsPicker` to `attachPicker` on the rows side
-  // makes this asymmetry impossible to silently consolidate.
+  // The CONTENT renders the resolved recipient summary
+  // ("**To:** N user(s)") with needsPicker:false; the ROWS always
+  // keep the UserSelectMenu attached (renderConfirmCardRows always
+  // produces 4 rows) so the user can re-pick. Send is enabled now
+  // that we have a valid recipient set.
   const content = renderConfirmCardContent({
     resourceType: payload.resourceType,
     resourceLabel: payload.resourceLabel,
@@ -4682,7 +4672,6 @@ async function handleSendUserSelect(interaction, { flow_id, row }) {
   return interaction.editReply({
     content,
     components: renderConfirmCardRows({
-      attachPicker: true,
       sendDisabled: false,
       expiresIn: payload.expiresIn,
       selfDestructSeconds: payload.selfDestructSeconds,
@@ -4695,10 +4684,8 @@ async function handleSendUserSelect(interaction, { flow_id, row }) {
 // payload (expiry / self-destruct / note). Each menu handler builds
 // `newPayload` and calls this with the post-transition `interaction` +
 // payload. Keeping the re-render shape in one place avoids the three
-// handlers drifting on `attachPicker` / `sendDisabled` / content flags.
+// handlers drifting on `sendDisabled` / content flags.
 //
-// `attachPicker: true` matches `handleSendUserSelect`'s post-pick
-// behavior (the picker stays attached so the user can re-pick).
 // `sendDisabled` is derived from recipientIds — without it, a user
 // who opened /qurl file without `recipients:` could change expiry
 // before picking and see an enabled Send button against an empty
@@ -4747,7 +4734,6 @@ async function rerenderConfirmCard(interaction, newPayload) {
   return interaction.editReply({
     content,
     components: renderConfirmCardRows({
-      attachPicker: true,
       sendDisabled: needsPicker,
       expiresIn: newPayload.expiresIn,
       selfDestructSeconds: newPayload.selfDestructSeconds,
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -4616,6 +4616,13 @@ async function handleSendUserSelect(interaction, { flow_id, row }) {
   const newRecipientAliases = Object.fromEntries(
     valid.map((u) => [u.id, resolveRecipientAlias(u, interaction)])
   );
+  // Asymmetry vs. the menu/modal handlers' no-op short-circuits:
+  // the picker DOES NOT skip transitionFlow on a same-recipient-set
+  // re-pick. Re-running here recomputes warningsBlock (droppedBots /
+  // droppedSelf can flip if the picker selection now includes the
+  // sender or a bot) and refreshes recipientAliases (display names
+  // may have changed since the prior pick). Both are user-visible
+  // and worth the version bump.
   const newPayload = {
     ...payload,
     recipientIds: valid.map((u) => u.id),
@@ -4763,11 +4770,11 @@ async function rerenderConfirmCard(interaction, newPayload) {
 // option gate — Discord enforces the choice set server-side, but a
 // forged interaction could land an off-set value here.
 async function handleSendConfirmExpirySelect(interaction, { flow_id, row }) {
-  // `interaction.values && ...` is paranoia, not load-bearing —
-  // Discord guarantees `values` is an array for StringSelectMenu
-  // interactions. The guard catches a forged interaction missing
-  // the field entirely; legitimate UI paths always provide it.
-  const picked = interaction.values && interaction.values[0];
+  // `?.[0]` is paranoia, not load-bearing — Discord guarantees
+  // `values` is an array for StringSelectMenu interactions. The
+  // guard catches a forged interaction missing the field entirely;
+  // legitimate UI paths always provide it.
+  const picked = interaction.values?.[0];
   // Validate before deferring so the forgery branch can use `reply`
   // (the cheaper, single-call ack) instead of `followUp` after a
   // wasted deferUpdate. Symmetric with the self-destruct handler.
@@ -4830,9 +4837,9 @@ async function handleSendConfirmExpirySelect(interaction, { flow_id, row }) {
 // selfDestructOptionToSeconds. The helper falls back to null for any
 // unexpected value (forged interaction), which is the safe default.
 async function handleSendConfirmSelfDestructSelect(interaction, { flow_id, row }) {
-  // `interaction.values && ...` is paranoia (see expiry handler) —
-  // legitimate StringSelectMenu interactions always carry the array.
-  const pickedValue = interaction.values && interaction.values[0];
+  // `?.[0]` paranoia (see expiry handler) — legitimate
+  // StringSelectMenu interactions always carry the values array.
+  const pickedValue = interaction.values?.[0];
   // Validate against the closed legitimate set BEFORE acknowledging.
   // Discord enforces the choice list server-side; an off-set value
   // here means a forged interaction. Silently mapping forged values
@@ -4932,11 +4939,14 @@ async function handleSendConfirmNoteButton(interaction, { flow_id, row }) {
 }
 
 // Note modal-submit → sanitize input, update payload.personalMessage.
-// Modal-submit interactions get a fresh 3s ack window; using
-// `editReply` after `update()` mirrors handleSetupModal's shape.
-// `update()` ACKs against the ORIGINAL message (the confirm card)
-// because flow-dispatch wires modal submits to the same message-bound
-// interaction context.
+// Modal-submit interactions get a fresh 3s ack window. We defer
+// FIRST (deferUpdate) for budget headroom under slow DDB writes,
+// then editReply on the deferred interaction. The deferred ack
+// targets the ORIGINAL message (the confirm card) because flow-
+// dispatch wires modal submits to the same message-bound
+// interaction context. Post-defer error paths use followUp
+// (ephemeral) — reply / update would 409 against the acked
+// interaction.
 async function handleSendConfirmNoteModal(interaction, { flow_id, row }) {
   // Defer the modal-submit ack BEFORE the DDB write — without this,
   // a slow conditional write (throttled partition) can push past
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -4713,6 +4713,18 @@ async function rerenderConfirmCard(interaction, newPayload) {
   const memberCache = interaction.guild?.members?.cache;
   const persistedAliases = newPayload.recipientAliases || {};
   const validRecipients = recipientIds.map((id) => {
+    // Asymmetric inputs through the same render path:
+    //   - cache-hit returns a Discord.js User object (UNSANITIZED
+    //     fields — username/displayName fresh from the gateway).
+    //   - cache-miss returns { displayName: persistedAlias },
+    //     where the alias was ALREADY sanitized at pick time by
+    //     resolveRecipientAlias.
+    // Both then re-flow through renderConfirmCardContent's
+    // `resolveRecipientAlias` pass. The cache-hit branch sanitizes
+    // fresh; the cache-miss branch is a no-op IF sanitize stays
+    // idempotent (pinned by tests/sanitize.test.js's idempotence
+    // block). If anyone splits these into divergent post-processing,
+    // the cache-miss branch loses its protection.
     const cached = memberCache?.get?.(id);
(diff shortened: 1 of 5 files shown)


justin-layerv:
Round-2 fixes pushed in 940f0d8. cr items addressed:

**🐛 Double-escape bug — fixed.** Added `personalMessageRaw` field to the flow payload. Pre-fill reads raw; sanitize derives at submit time. Both ingress paths (slash entry + modal submit) trim → `safeCodepointSlice(_, 280)` → store both forms. New tests pin the round-trip is idempotent and that legacy flow rows missing the field fall back to empty pre-fill.

**🟡 Self-destruct forgery — logs warn now.** `selfDestructSelectValueToSeconds` still silently falls back to null (to avoid punishing a legitimate UI bug), but unknown values that aren't the no-timer sentinel or a known preset get `logger.warn`ed. New negative test confirms the no-timer sentinel doesn't false-positive.

**🟡 `handleSendConfirmNoteModal` duplication — eliminated.** `rerenderConfirmCard` now accepts an `ack` callback (default: `editReply`). Modal handler passes `interaction.update`. ~25 lines deleted; one render path shared by picker + 3 menus + modal.

**🟢 setMaxLength vs 500 cap nit — resolved.** Introduced `PERSONAL_MESSAGE_INPUT_MAX = 280` constant, used by both the modal's `setMaxLength` and both raw-storage caps. No more skew between input bound and storage cap. Swapped `substring(0, 500)` for `safeCodepointSlice(_, PERSONAL_MESSAGE_INPUT_MAX)` — codepoint-aware, surrogate-safe.

**🟡 Expiry vs self-destruct handler near-duplication — left as-is.** cr called this optional; the forgery-warn divergence between the two handlers (expiry rejects via followUp + log, self-destruct silently falls back + log) makes extraction premature. Will revisit if a third menu lands in this shape.

1481 tests pass (was 1478, added 3). Lint clean. UI still pending a live-bot verification — will exercise in the beta guild after merge.

justin-layerv:
Round-3 fixes pushed in 4935759. cr round-2 items addressed:

**1. ✅ `handleSendConfirmNoteModal` 3s ack risk — fixed.** Added early `await interaction.deferUpdate()`. Success/conflict/not_found now use `editReply`; throw path uses `followUp({ ephemeral: true })` since `reply()` would 409 post-defer. Test asserts the negative invariant (`reply` and `update` NOT called on the throw path).

**2. ✅ `rerenderConfirmCard` dropped warnings + cache-miss fallback — fixed.** Two new payload fields:
   - `recipientAliases: { [id]: aliasString }` — resolved at slash-entry + picker-rePick. Cache-miss fallback chain is now: cache → aliases → `user-${id}`. The raw snowflake is no longer the user-facing fallback.
   - `warningsBlock: string` — slash-entry records the warning surface; menu interactions carry it forward. Picker re-pick OVERWRITES with fresh partition warnings (recipient set may have changed).
   - Worst-case ~800 bytes per flow row, bounded by 3-min TTL.

**3. ✅ `PERSONAL_MESSAGE_INPUT_MAX` cap drift — fixed.** Both slash-option `setMaxLength` sites now reference the constant (was only the modal in the previous round). Single source of truth for 280.

**4. 🟡 Forgery-asymmetry note for PR description — adding now.** Will update body separately.

**5. ✅ Negative-assertion on modal-throw test — added** (`reply`/`update` NOT called post-deferUpdate).

**6. ✅ `/qurl map` payload-parity test — added.** Pins that `handleQurlMap` produces the same `recipientAliases`/`warningsBlock` shape as `handleQurlFile`. Was a blind spot in the round-2 test set.

**🟡 Expiry/self-destruct handler near-duplication — still left as-is.** Round-1 cr called this optional; the forgery-handling divergence (expiry rejects + warn, self-destruct falls back + warn) plus differing post-validation logic makes extraction premature. Will revisit when a third menu of this shape lands.

1484 tests pass (was 1481), lint clean.

justin-layerv:
Round-4 fixes pushed in 53c5c54. cr round-3 items addressed:

**1. ✅ Self-destruct symmetric reject — fixed.** Now rejects forged values via `reply` + warn-log, no save. Previous silent-fallback would have erased a user's prior timer on every forgery probe. New behavior is symmetric with the expiry handler.

**2. ✅ Validate-before-defer — fixed (both menu handlers).** Forgery branches now use `reply` (single-call ack) instead of `followUp` after a wasted `deferUpdate`. Saves one Discord API call per probe and reads more naturally.

**3. ✅ `personalMessageRaw` defense — explicit CONTRACT comment** at the field declaration. Skipped the cr-suggested rename in favor of comment + the structural defense already provided by `executeSendPipeline`'s explicit destructuring (no `...payload` spread). Future-bit-rot signal is the comment.

**4. ⏭️ Modal typed-note-lost-on-conflict UX paper cut → filed as #309** ([link](https://github.com/layervai/qurl-integrations/issues/309)). Labeled `enhancement / priority: low / javascript`. Includes pseudocode + acceptance criteria for the future ephemeral-followUp echo of typed text.

**5. ⏭️ `renderConfirmCardRows` 5-param signature — skipped per cr (\"not blocking\").** Touches 3 call sites for cosmetic gain; the params are flat scalars from the same payload — a `selectionState` sub-object would add indirection without changing what's passed.

**6. ℹ️ `warningsBlock` size note — informational, no action.** cr called it "Acceptable" (bounded by `QURL_SEND_MAX_RECIPIENTS` * a few words, far below DDB's 400KB ceiling).

1484 tests pass, lint clean.

justin-layerv:
Round-5 fixes pushed in 8ab6fdb. cr round-4 items addressed:

**1. ✅ Dead `ack` parameter — dropped.** After round-3 added \`deferUpdate\` to the modal handler, all four entry paths use \`editReply\` through \`rerenderConfirmCard\`. Cleaned up to a single straight-line render path.

**2. ✅ `attachPicker` stabilized at \`true\` on every render.** Slash entry (with OR without recipients) and every menu interaction now produce 4 rows. No more mid-flow row-count shift. Matches \`handleSendUserSelect\`'s post-pick contract (picker stays available for re-picks). Test updated to assert the new always-4-rows contract.

**3. ✅ PR description updated.** Fixed two stale claims:
- Forgery handling section now reflects symmetric reject (round-3 flipped self-destruct).
- Architecture section dropped the stale "ack callback" claim.

**4. ✅ Contract test made real.** Round-3 added the CONTRACT comment but cr noticed I'd misled it about a "contract test" — there wasn't one. Now there is: static check extracts the \`executeSendPipeline\` function body via brace-depth tracking and asserts the substring \`personalMessageRaw\` doesn't appear inside it.

**5. ⏭️ \`personalMessageRaw\` rename — skipped (cr non-blocking).** "Neither blocking; the current state is fine as long as the contract test stays green." Contract test is now green.

**6. ℹ️ Payload-size note + legacy-row UX during deploy — informational, no action.**

**7. 🟡 Handler near-duplication + SEND_NOTE_MODAL_INPUT_ID comment nit — skipped per cr ("Leaving as-is is fine").**

1485 tests pass, lint clean.

justin-layerv:
Round-6 fixes pushed in 9c37f71. cr round-5 items addressed:

**1. ✅ `truncForLog` on forgery warns — applied.** Both menu handlers now truncate the forged value before logging. Bounds the log payload on a 4000-char garbage probe. Mirrors the existing precedent at \`executeSendPipeline\` entry-validation gates.

**2. ✅ Sibling-race merge test added.** Lock in the modal-submit merge semantics when row.payload carries a sibling mutation (e.g. user types in modal while another desktop window changes expiry). The dispatcher reloads the row at submit time → spread preserves the sibling's change → modal submit's note lands alongside it. Test pins this so a future refactor that explicitly carries forward specific fields can't silently undo a sibling's edit.

**Confirmed informational (no action per cr):**
- Brief client-side UI discrepancy on forgery rejection (corrects on next legitimate interaction).
- Sliding TTL via \`set_expires_at\` is intentional (active engagement extends the 3-min window).
- Brand spelling — checked clean.

1486 tests pass (was 1485, added 1), lint clean.

justin-layerv:
Round-7 pushed in 8f9f689 ("Approve once #1 is addressed" → addressed).

**1. ✅ showModal failure surfaces ephemeral reply.** \`handleSendConfirmNoteButton\`'s catch no longer ends with a bare warn-log — now also fires \`interaction.reply({ ephemeral: true, content: '❌ Could not open the note editor — try again.' })\` so the user gets a remediation hint instead of Discord's generic "interaction failed" toast. Test pins the new behavior + asserts flow_state stays untouched (showModal failure is a pure UX surface, not a state mutation).

**2. ⏭️ Parallel \`recipientIds\` / \`recipientAliases\` maps → filed #310** ([link](https://github.com/layervai/qurl-integrations/issues/310)). cr explicitly said "Not blocking" and "when convenient" — labeled \`enhancement / priority: low / javascript\` with full scope + acceptance criteria for the future unification to \`recipients: [{id, alias}]\`.

**3, 5. ℹ️ Flow row byte count + DDB redaction surface — informational, no action.** Bounded, no regression vs pre-PR. CLAUDE.md-level note deferred until/unless a sensitive-field allowlist lands.

**4. ✅ ButtonBuilder count tripwire — keeping per cr ("That's the desired behavior — keep it").**

**Nits — observations, no action per cr.**

1487 tests pass (was 1486, added 1), lint clean.

justin-layerv:
Round-8 pushed in 9e7b6ff. cr round-7 actionable items (1, 2, 4) addressed:

**1. ✅ Contract test failure hint.** \`If this test fails\` paragraph now points at the field-declaration docblock in handleQurlSlashSend. Future engineers tripping the coarse substring red get a clear pointer to where to document the invariant (at the field decl), not inside executeSendPipeline.

**2. ✅ \`personalMessageRaw\` preservation asserted in expiry handler's "preserves all other payload fields" test.** Locks the round-trip invariant — a regression that destructured \`personalMessage\` (without \`Raw\`) into newPayload would silently break the next Edit-note pre-fill.

**4. ✅ rerenderConfirmCard fallback shape normalized.** Both branches now set \`displayName\` (NOT \`username\`) on the fabricated user object. Removes the asymmetric shape footgun if anyone touches resolveRecipientAlias's precedence later.

**3. ⏭️ Modal-typed-note-lost in-card hint → added to #309 scope.** Comment posted on the issue suggesting the in-card supersede copy point at the recoverable-text followUp.

**5. ℹ️ Forgery rejection rate-limit — cr "probably acceptable" since only authenticated users can fire the interaction. No action.**

**6. ℹ️ PERSONAL_MESSAGE_INPUT_MAX vs sanitizeMessage 500 cap — cr "Not actionable here." No action.**

1487 tests pass, lint clean.

justin-layerv:
Round-9 pushed in 45dfa6f. cr round-8 items addressed:

**1. ✅ Defensive modal field read.** \`getTextInputValue\` now wrapped in try/catch with empty-string fallback. If the customId allowlist ever drifts, the modal-submit interaction still resolves (routes to clear-note branch) instead of stalling past \`deferUpdate\`. Also renamed \`SEND_NOTE_MODAL_INPUT_ID\` → \`SEND_NOTE_MODAL_FIELD_ID\` per cr style nit.

**2. ✅ Expiry corrupted-row defense.** Mirrors the existing self-destruct defense. \`hasExpiryMatch\` check; if no match, defaults the '24h' option (the codebase default). Test pins via self-destruct handler (which doesn't overwrite expiresIn during rerender).

**3. ✅ Idempotence assumption documented** in \`rerenderConfirmCard\` — the persisted aliases are re-passed through \`resolveRecipientAlias\` on render; no-op today because NFKC + bidi-strip have fixed points, but flagged so a future sanitize-semantics change is forced to confront it.

**4. ✅ Brace-walker limitation noted in CONTRACT test.** Future engineer adding a string/template/regex literal containing braces to \`executeSendPipeline\`'s body now has a clear hint.

**5. ⏭️ Menu conflict tombstoning → filed #311** ([link](https://github.com/layervai/qurl-integrations/issues/311)). cr framed as "follow-up, not blocker." Issue has full re-read + re-render pseudocode + acceptance criteria.

**6, 7. ℹ️ Forged-value log + handler duplication — no action, cr framed informational / optional.**

1488 tests pass (was 1487, added 1 corrupted-row test), lint clean.

justin-layerv:
Round-10 pushed in 30955f5. cr round-9 actionable items (1, 2, 4, 5) addressed:

**1. ✅ \`personalMessageRaw\` null'd on sanitize-empty.** ZWSP-only / bidi-only input passes \`.trim()\` but is stripped by \`sanitizeMessage\`. Previously the raw was retained while sanitized was empty → button labeled "Add a note" but pre-fill carried invisible chars. Now both fields go to null in lockstep. Applied at both slash-entry and modal-submit sites.

**2. ✅ Idempotence test for \`sanitizeDisplayNamePlain\`.** 5 cases: ASCII / post-RLO / post-NFKC / capped-input / surrogate-pair emoji. Pins the load-bearing assumption rerenderConfirmCard's cache-miss path relies on — a future sanitize-semantics change that breaks idempotence flips the cache-miss render into a double-sanitize bug and trips these tests.

**4. ✅ Optional chaining applied** in rerenderConfirmCard's memberCache resolution. Matches the existing pattern at \`resolveRecipientAlias\`.

**5. ✅ Explicit registration test for the 4 new customIds.** Asserts duplicate-register throws "already registered" — catches the registration-omission failure shape directly. The existing keyed-by-stage siblingMessage test covers the message inheritance; this new test covers the registration-presence invariant.

**3, 6. ℹ️ Brace-walker fragility / showModal dual-failure — informational, no action per cr.**

1494 tests pass (was 1488), lint clean.

justin-layerv:
Round-11 pushed in 1f12bfc. cr round-10 was effectively LGTM — addressed one of the two pure-nit optionals:

**1. ✅ Extracted \`isLegitimateSelfDestructSelectValue\` helper** to \`utils/time.js\`. Same semantics as the inline check, single source of truth. 4 new unit tests pin no-timer sentinel / all presets / forged values / numeric-not-stringified.

**7. ⏭️ \`SEND_NOTE_MODAL_FIELD_ID\` rename — deferred.** cr called it "Pure nit" (forensic log clarity). Changing the literal would break any modal currently open at deploy — old modal submits the old field id, new code reads the new id, defensive try/catch routes to clear-note → user input dropped. Deploy-window UX cost outweighs marginal forensic gain.

**2, 3, 4, 5, 6. ℹ️ All other observations confirmed informational** ("fine today", "good as-is" per cr).

1498 tests pass (was 1494, +4 isLegitimate predicate tests), lint clean.

justin-layerv:
Round-12 pushed in 57dccf7. cr round-11 (4 paper-cut items) all addressed:

**1. ✅ \`getTextInputValue\` throw → surfaces ephemeral followUp.** "Could not read your note input — try again. Your existing note, if any, is unchanged." Returns early, no transitionFlow, payload preserved. Closes the silent-clear failure mode.

**2. ✅ Off-EXPIRY_LABELS payload now logs a warn.** Surfaces corrupted DDB rows in forensics rather than papering over with the 24h fallback.

**3. ✅ Runtime CONTRACT test added** via Proxy-wrapped \`row.payload\` that throws on \`get('personalMessageRaw')\`. Drives \`handleSendConfirmClick\` through to executeSendPipeline. Survives future syntax changes (string/regex literals containing braces) that the static brace-walker can't.

**4. ✅ showModal-failure dual-rejection negative-assertion test.** Pins that if both showModal AND the fallback reply reject, no unhandled rejection escapes — installs \`unhandledRejection\` listener, microtask-flushes, asserts the listener never fires.

1501 tests pass (was 1498, +3 contract/error-surface tests), lint clean.

justin-layerv:
Round-13 pushed in fe930c1. cr round-12 single actionable nit applied:

**3. ✅ \`interaction.values\` guard clarified** as paranoia (not load-bearing). Comment now notes Discord guarantees the array; the guard catches only forged/missing-field interactions, not legitimate UI shapes.

**1, 2, 4, 5. ℹ️ All other round-12 observations explicitly informational/non-blocking per cr.**

1501 tests pass, lint clean. PR is at convergence — all rounds of cr feedback addressed including nits, builds passing, /simplify clean (no findings on the comment-only diff).

justin-layerv:
Round-14 pushed in 2abe7a4. cr round-13 actionable items (1, 2) addressed:

**1. ✅ \`attachPicker\` parameter dropped.** All call sites passed \`true\` (round-9 stabilized this); the param + conditional branch + header-comment 3-row math described a state no caller reached. Renderer now produces 4 rows unconditionally. Surrounding comments cleaned up.

**2. ✅ Symmetric self-destruct corruption warn added.** Off-preset finite \`selfDestructSeconds\` now logs alongside the "No timer" fallback. Mirrors the off-EXPIRY_LABELS warn — closes the "header reads 'No timer' but content line says 'self-destructs in N seconds'" silent-asymmetry gap.

**3. ℹ️ Menu DDB write throttling — cr "Probably acceptable", no action.**

1501 tests pass (no test changes needed — attachPicker drop doesn't change behavioral output; tests already asserted 4-rows always), lint clean.

justin-layerv:
Round-15 pushed in ac00b9e. cr round-14 items 1, 2, 5 addressed — item 5 caught a real bug:

**1. ✅ getValue catch simplified.** Dropped the \`getValueThrew\` flag; return from catch directly. Same behavior, fewer moving parts.

**2. ✅ Brace-walker LIMITATION comment expanded** to call out destructured default object literals (like \`personalMessage = { fallback: null }\`) alongside string/regex literals. Runtime Proxy CONTRACT test below covers these; static check is belt-and-suspenders.

**5. 🐛 ZWSP-only test caught a real bug.** Round-9's "null out raw if sanitize stripped to empty" fix was incomplete: \`sanitizeMessage\` returns \`''\` (not \`null\`) on full strip. So \`personalMessage\` ended up as \`''\` in DDB — falsy in practice but inconsistent with the documented field contract ("null if no note"). Fix: \`|| null\` after sanitize. Both ingress paths updated. Field shape now matches the docstring.

**3, 4, 6. ℹ️ Conflict/not_found dedup / payload size / ephemeral→MessageFlags — all explicitly non-blocking per cr ("Optional", "informational", "Flagging only as a future cross-cutting cleanup").**

1502 tests pass (was 1501, +1 ZWSP regression test that flushed out the bug), lint clean.

justin-layerv:
Round-16 pushed in b48ca21. cr round-15 actionable items (1, 2, gap) addressed:

**1. ✅ Asymmetric-input comment** at rerenderConfirmCard's construction site. Flags that cache-hit returns unsanitized User; cache-miss returns pre-sanitized alias — both re-flow through resolveRecipientAlias relying on idempotence. Future refactor that splits the post-processing now has visible warning at the construction site.

**2. ✅ Pipeline-direct runtime CONTRACT test added.** Wraps the params object that executeSendPipeline receives in a Proxy that throws on \`get('personalMessageRaw')\`. Complements the handler-level Proxy by catching a future destructure that adds personalMessageRaw to executeSendPipeline's signature.

**Test gap: ✅ corrupted-SD warn assertion added** to the expiry-handler's "preserves all other payload fields" test. Pins the forensic surface fires from the expiry-handler entry path, not just the rerenderConfirmCard cache-miss block.

**3, 4. ℹ️ Sentry alert / payload size — non-blocking per cr (infra / informational).**

1503 tests pass (was 1502), lint clean.

justin-layerv:
Round-17 pushed in 5611e47. cr round-16 actionable items addressed:

**1. ✅ No-op re-pick short-circuit** on expiry + self-destruct handlers. Same-value click skips transitionFlow → no version bump → no sibling fence. Visible feedback (rerender) still fires. Tests pin both no-op paths. Modal handler intentionally NOT short-circuited (raw-vs-sanitized equality is more nuanced; modal submits are user-deliberate).

**2. ✅ Stale \`attachPicker\` test title fixed.** Renamed to describe behavior, not the round-13-removed param.

**4. ⏭️ Corruption-warn dedup per flow_id → filed #312** ([link](https://github.com/layervai/qurl-integrations/issues/312)). cr framing: "Likely acceptable... not free." Issue has LRU-bounded design + acceptance criteria.

**3, 5. ℹ️ "Unrecognized" copy nit / payload size growth — non-blocking per cr.**

1505 tests pass (was 1503, +2 no-op tests), lint clean.

justin-layerv:
Round-18 pushed in be77f97. cr round-17 item 1 (in this PR) + trivial items 3, 5 addressed; items 2, 4 are "follow-up candidates" per cr framing.

**1. ✅ Modal no-op skip applied.** Symmetric with expiry/self-destruct (round-17). Compares BOTH personalMessage AND personalMessageRaw (round-trip idempotence is what makes the comparison safe). The existing round-trip-no-double-escape test now also pins the no-op skip. Three pre-existing clear-note tests (empty/whitespace/ZWSP) updated to use a payload with a previous note so the clear write fires through.

**3. ✅ Inline comment near \`setValue(payload.personalMessageRaw || '')\`** notes the bidi-survives-prefill is intentional (ephemeral-to-self surface).

**5. ✅ Send-disabled test added.** Asserts \`ButtonBuilder\` for the Send button got \`setDisabled(true)\` on no-recipients initial render, while expiry/self-destruct/note remain interactable.

**2, 4. ⏭️ Follow-up candidates per cr framing** ("Items 2–5 are follow-up candidates if you'd rather keep this PR focused"):
- Item 2: \`derivePersonalMessageForms(raw)\` helper extraction (slash-entry + modal-submit duplication)
- Item 4: \`isLegitimateSelfDestructSelectValue\` for...of → .some() style consistency

1506 tests pass (was 1505), lint clean.

justin-layerv:
Round-19 pushed in d8458b8. cr round-18 polish items addressed:

**1. ✅ Doc-vs-code drift fixed.** handleSendConfirmNoteModal header now describes the actual defer → editReply pattern (was referencing an \`update()\` call that doesn't exist).

**2. ✅ Optional chaining style applied.** \`interaction.values?.[0]\` at both menu handlers, matches the existing \`memberCache?.get?.(id)\` pattern.

**3. ✅ Picker no-op asymmetry documented as intentional.** Comment at handleSendUserSelect's newPayload site explains the picker doesn't skip on same-recipient-set re-pick because the re-run recomputes warningsBlock (droppedBots/Self can flip) and refreshes recipientAliases (display names may have changed) — both user-visible, worth the version bump.

**4. ✅ Picker preservation test added.** Pins that re-pick preserves personalMessageRaw + personalMessage via the spread. A regression that destructured \`personalMessage\` without \`Raw\` into newPayload would silently drop Edit-pre-fill.

**5, 6. ⏭️ Deferred per cr "not blocking":**
- getter-wrap for personalMessageRaw at type/module level (larger refactor)
- resolveRecipientAlias precedence unit test (sanitize/utils concern, separate PR scope)

1507 tests pass (was 1506), lint clean.

PR #322, opened 2026-05-14
feat(discord): allow self-send on /qurl file + /qurl map

justin-layerv:
## Summary

Self-send is now supported on `/qurl file` and `/qurl map`. The sender can include themselves as a recipient — useful for:

- Test the link before forwarding it
- Cross-device handoff (send from desktop, receive on phone)
- Self-archival with the self-destruct timer / expiry controls

### UX

**Before:** mentioning `@me` (or picking yourself in the UserSelectMenu) produced a warning — *"You cannot send a qURL to yourself — skipped."* — and the sender was filtered out of the recipient list.

**After:** the sender enters the recipient list normally. The confirm card surfaces a NEUTRAL notice instead:

```
📁 Sending file: report.pdf
ℹ️ Send includes you.
**To:** 3 users (alice, bob, you)
**Expires:** 24h
```

Implicit opt-in: there's no new slash option or subcommand. Self-send is triggered the same way any other recipient is — typing `@me` in the `recipients:` text or picking yourself in the UserSelectMenu. Both are deliberate actions, so there's no accidental-self-send footgun.

### Code changes

- **`recipient-parser.js`**: drop the `id === senderId` filter in both the user-mention loop and the role-expansion loop. The bot filter stays. Drop the unused `senderId` capture. Updated docstring/comments.
- **`commands.js partitionRecipients`**: return `{ valid, droppedBots, selfIncluded }`. Sender stays in `valid[]`; `selfIncluded` is a boolean flag set when the sender appears in `users[]`.
- **`commands.js renderRecipientWarnings`**: drop the `droppedSelf` parameter + the corresponding warning line.
- **`commands.js renderConfirmCardContent`**: accept a `selfIncluded` prop, render `ℹ Send includes you.` above the recipient preview.
- **`commands.js` initial payload + UserSelect transitionFlow payload**: persist `selfIncluded` in `flow_state.payload` so menu re-renders (expiry / self-destruct / note modal) keep the notice.
- **`commands.js` empty-valid branches**: simplified — self-only is now legitimate, so the bot-only path is the only remaining empty trigger.
- **Tests**: full suite update across `recipient-parser.test.js` and `qurl-file-map.test.js`. +2 new tests pinning the `selfIncluded` notice rendering. 1417 tests pass.

### Intentionally NOT changed

- The bot filter — bots still can't receive qURL links.
- SEND_FLOW_TTL_SECONDS, customId wire literals, DDB schema — untouched.
- The DM dispatch path — already DMs whoever's in the recipient list, no change needed.

## Test plan

- [x] `pnpm test` — 1417 tests pass
- [x] `pnpm lint` — clean (`eslint --max-warnings 0`)
- [x] `grep -rn 'droppedSelf' apps/discord` returns zero results
- [ ] CI green
- [ ] Claude review clean
- [ ] Live smoke: `/qurl file` with `recipients:@me` shows the notice + delivers a DM to the sender

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2698,6 +2698,12 @@ async function resolveRecipientUsers(interaction, ids) {
 // drops them all — small absolute cost but burns per-guild rate-limit
 // budget that legitimate ops elsewhere could use. The durable fix is
 // the v2 resolve-then-cap refactor tracked in #304.
+//
+// Self-mention cap behavior: since self-send is supported, `<@me>` now
+// CONSUMES a cap slot like any other mention. So `<@me> <@u1>..<@u25>`
+// yields self + 24 others (was: 25 others, pre self-send). This is
+// symmetric with the bot cap-skew above and aligns with the supported-
+// recipient model — sender is just another recipient.
 function partitionRecipients(users, senderId) {
   // Dedup contract is OWNED upstream: parseRecipientMentions dedupes
   // via a Set (recipient-parser.js:197-198) and the UserSelectMenu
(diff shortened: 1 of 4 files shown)


justin-layerv:
Thanks both! r1 cr addressed in 815d2e5:

**From cr #1 (4453663356):**
- ✅ #1 (all-invalid empty-reasons coherence) — added `'No usable recipients in pick'` fallback so the message stays coherent if a future filter is removed and `valid.length === 0` is hit with `droppedBots === 0`.
- ✅ #2 (cap-skew docstring) — added a paragraph noting self-mention now consumes a cap slot (symmetric with bot cap-skew above).
- ✅ #3 (positive dispatch assertion) — added "Preparing send" editReply + terminal deleteFlow assertions to the sender-only Send-click test.
- ✅ #4 (role-via-membership note) — added a paragraph in the role-expansion comment block.
- Acknowledged #5 (self-DM live smoke) — will verify on rollout, not blocking.
- Nit re `=== true` style — left as-is; defensive coercion is consistent with how `rerenderConfirmCard` already treats payload booleans.
- Nit re lint rule for codemod-resistance — out of scope for this PR; useful idea for a follow-up.

**From cr #2 (4453684777):**
- ✅ #1 (stale `interaction.user?.id` comment in `recipient-parser.js:158`) — updated to `interaction.guild` deref.
- Acknowledged #2 (picker hint copy) — out of scope; would be a small follow-up if we ever revisit picker UX.
- ✅ #3 (re-render persistence test gap) — added a focused test in `handleConfirmExpirySelect` suite pinning that `selfIncluded` survives an expiry-select transition through `rerenderConfirmCard`. Catches a future refactor that drops the prop.

PR #323, opened 2026-05-14
feat(discord): allow @everyone in /qurl file + /qurl map (gated on MENTION_EVERYONE)

justin-layerv:
## Summary

Allow `@everyone` in the `recipients:` text on `/qurl file` and `/qurl map`. Gated behind Discord's MENTION_EVERYONE permission so any guild member can't blast a /qurl send to all 25 cached members.

This is **Part 1 of 2** of the feature requested ("send to @everyone or to a role"). Part 2 — a `MentionableSelectMenu` picker for selecting roles from a dropdown — is tracked in #324. Role mentions via text (`@TeamName` → `<@&roleId>`) already work today; that path is unchanged.

### UX

- **User types `@everyone alice` and has Mention Everyone permission** → all non-bot guild members expand into the recipient list; the standard cap (25) + bot filter apply.
- **User types `@everyone alice` and does NOT have permission** → confirm card shows a permission-specific warning:
  > ⚠ Some recipients were dropped:
  > • `@everyone` requires the **Mention Everyone** permission — skipped.

  …and Alice expands as usual.
- **DM context**: the `@everyone` permission warning is suppressed (Discord doesn't expand `@everyone` in DMs, so the copy would be misleading). Other warnings still surface.
- **`@here` is intentionally NOT implemented.** It would need GUILD_PRESENCES intent to filter online members; the bot only runs GuildMembers. Stays on the legacy U+200B defuse path (test pinned).

### Code changes

- `recipient-parser.js`:
  - New `EVERYONE_TOKEN_RE` — Unicode-aware word boundary (`\p{L}\p{N}_` with `/u`) so `@everyoneé` doesn't false-match. Case-sensitive (Discord's parser is lowercase-only).
  - New opt `allowMassMention: boolean` on `parseRecipientMentions`.
  - Allowed → expand to all non-bot guild members (loop breaks once cap is hit; cache iteration is bounded). Denied → `massMentionDenied: true` in result. Either way, strip from input before residue pass.
  - New `isBotMember(member)` helper consolidates the `.user.bot === true` check across all three expansion paths.
- `commands.js`:
  - Read `interaction.memberPermissions.has(PermissionFlagsBits.MentionEveryone)`; pass through as `allowMassMention`.
  - `renderRecipientWarnings`: new `massMentionDenied` parameter emits the permission-specific line.
  - DM-context suppression: caller drops the `massMentionDenied` signal when `!interaction.guild`.
- Tests: new `@everyone (allowMassMention)` describe block in `recipient-parser.test.js` covering allowed/denied/cap/dedup/bot-filter/DM-context/Unicode-boundary/repeated-everyone/mixed-with-@here. New result-shape contract test pinning the closed 4-key set. New DM-context warning suppression test at the handler level. 1436 tests pass.

### Wire-protocol changes

None. customIds + DDB schema untouched.

### Deferred via follow-up issues

- **#324** — Part 2: MentionableSelectMenu picker for role-via-dropdown (with the same MENTION_EVERYONE gate for the @everyone role).
- **#325** — `logger.audit` events for @everyone allowed/denied attempts. Both cr reviewers independently flagged this; deserves its own PR with new AUDIT_EVENTS constants.

## Test plan

- [x] `pnpm test` — 1436 tests pass
- [x] `pnpm lint` — clean (`eslint --max-warnings 0`)
- [ ] CI green
- [ ] Claude review clean
- [ ] Live smoke: `/qurl file recipients:@everyone` as admin → all members expand; same as non-admin → permission warning + Alice still parses.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3098,6 +3098,11 @@ async function handleQurlSlashSend(interaction, params) {
     // surfaces `massMentionDenied: true` when the sender tried but
     // lacks permission — the caller renders a permission-specific
     // warning instead of the generic "couldn't parse" copy.
+    //
+    // `interaction.memberPermissions` is the resolved channel-effective
+    // permission set (guild perms + channel overwrites). A future
+    // refactor switching to `interaction.member.permissions` would
+    // silently lose the channel-overwrite respect — keep this property.
     const canMentionEveryone = interaction.memberPermissions?.has(PermissionFlagsBits.MentionEveryone) === true;
     const parsed = parseRecipientMentions(recipientsRaw, interaction, {
       allowMassMention: canMentionEveryone,
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2786,9 +2786,9 @@ function renderRecipientWarnings({
   if (massMentionDenied) {
     // Specific copy beats the generic "couldn't parse" path so the
     // user knows it's a PERMISSION issue, not a typo. Mirrors
-    // Discord's own MENTION_EVERYONE gate. Copy is context-agnostic
-    // (no "in this channel") so it reads sensibly even in DM context
-    // where channel-overrides don't apply.
+    // Discord's own MENTION_EVERYONE gate. The caller suppresses
+    // this in DM context (where @everyone has no meaning) so the
+    // copy here doesn't need a context qualifier.
     lines.push('• `@everyone` requires the **Mention Everyone** permission — skipped.');
   }
   if (lines.length === 0) return '';
@@ -3126,13 +3126,19 @@ async function handleQurlSlashSend(interaction, params) {
     // (vs. silently dropping into the picker, which would mask the
     // underlying mention-list problem).
     const recipientsOmitted = recipientsRaw == null || recipientsRaw.trim().length === 0;
+    // Suppress the @everyone permission warning in DM context —
+    // @everyone has no meaning in a DM (Discord doesn't expand it
+    // there) and "requires the Mention Everyone permission" reads
+    // strangely when there's no permission to grant. The other
+    // warnings (bot/capped/unresolved) still apply and surface.
+    const isDmContext = !interaction.guild;
     const warningsBlock = renderRecipientWarnings({
       invalidTokens: parsed.invalidTokens,
       cappedCount: parsed.cappedCount,
       unresolvedIds: resolved.unresolvedIds,
       transientFailureIds: resolved.transientFailureIds,
       droppedBots,
-      massMentionDenied: parsed.massMentionDenied,
+      massMentionDenied: parsed.massMentionDenied && !isDmContext,
     });
     if (!recipientsOmitted && valid.length === 0) {
       clearCooldown(interaction.user.id);
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2734,6 +2734,7 @@ function partitionRecipients(users, senderId) {
  *   unresolvedIds?: string[],
  *   transientFailureIds?: string[],
  *   droppedBots?: number,
+ *   massMentionDenied?: boolean,
  * }} [opts]
  */
 function renderRecipientWarnings({
@@ -2742,6 +2743,7 @@ function renderRecipientWarnings({
   unresolvedIds = [],
   transientFailureIds = [],
   droppedBots = 0,
+  massMentionDenied = false,
 } = {}) {
   const lines = [];
   if (cappedCount > 0) {
@@ -2780,6 +2782,12 @@ function renderRecipientWarnings({
   }
   if (droppedBots > 0) {
     lines.push(`• ${droppedBots} bot(s) cannot receive qURL links — skipped.`);
+  }
+  if (massMentionDenied) {
+    // Specific copy beats the generic "couldn't parse" path so the
+    // user knows it's a PERMISSION issue, not a typo. Mirrors
+    // Discord's own MENTION_EVERYONE gate.
+    lines.push('• `@everyone` requires the **Mention Everyone** permission in this channel — skipped.');
   }
   if (lines.length === 0) return '';
   return '⚠\u{FE0F} **Some recipients were dropped:**\n' + lines.join('\n') + '\n\n';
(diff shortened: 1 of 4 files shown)


justin-layerv:
Thanks for the thorough reviews! Round 2 cr addressed in 230bf43:

**From cr #1 (4453903551):**
- ✅ Docs drift (`Output shape` block missing `massMentionDenied`) — added.
- ✅ `@everyone.fix` residue pin — new test parallel to the `!` case.
- Acknowledged cappedCount magnitude in @everyone allowed paths — already addressed by r1's cap short-circuit (cappedCount stays 0 for @everyone allowed; trade-off documented in the comment).
- ✅ DM-context awkward copy — dropped "in this channel" from the permission warning. /qurl doesn't set `setDMPermission(false)`, so the broader copy reads sensibly in both contexts.
- Deferred via #325 — audit-log entries for allowed/denied @everyone attempts. Both your reviews flagged this independently; deserves its own PR + new AUDIT_EVENTS constants.
- ✅ Shape-keys pin test — new `result-shape contract` describe block asserts the closed 4-key set so future field additions surface as test diffs rather than slipping past `.toMatchObject` partial matches.
- ✅ Missing coverage — `@everyone @here` mixed input and `@everyone @everyone` repeated tests added.

**From cr #2 (4453958408):**
- ✅ EVERYONE_TOKEN_RE Unicode word boundary — switched to `\\p{L}\\p{N}_` with `/u` flag. `@everyoneé` no longer over-matches; new test pins the behavior.
- ✅ Cache-insertion-order note — added a paragraph to the @everyone expansion comment explaining the v1 caveat and pointing at #324 (Part 2 picker) as the resolution.
- Deferred via #325 (same audit-log issue) — mass-mention attempts deserve audit visibility.
- Acknowledged `@everyone!` doubled warning — pinned with the new `.fix` test now too. The doubled warning IS a deliberate tradeoff (surfacing the permission notice + leftover residue separately is more informative than swallowing the residue), but happy to revisit if it reads poorly in the live smoke test.

1435 tests pass; lint clean. Branch is `feat/discord-everyone-and-role-picker` at 230bf43.

justin-layerv:
Thanks! r2 cr addressed in a85c135:

- ✅ **#1 (correctness — DM context misleading copy)**: Caller now suppresses `massMentionDenied` rendering when `!interaction.guild`, since `@everyone` has no meaning in DMs and the "requires Mention Everyone permission" copy reads strangely there. New test in `qurl-file-map.test.js` pins the suppression. Inline comment on the warning line dropped the "even in DM" claim now that the suppression lives at the call site.
- ✅ **#2 (`typeof members.entries === 'function'`)**: Replaced with simpler `if (members)`. The check was conceptually unrelated to the for-of iteration which uses Symbol.iterator.
- ✅ **#3 ("cap bounds worst case" comment)**: Tightened to acknowledge the pathological all-bots cache edge — realistic guilds (<1% bots) fine, but the comment no longer overstates the bound.
- **#4 (PR description out of sync)**: Will fix on description right after this comment.

1436 tests pass; lint clean.

justin-layerv:
Thank you — caught a real bug. r3 cr addressed in 6cea608:

- ✅ **Ordering bug**: `@everyone` expansion now runs LAST, after both `USER_MENTION_RE` and `ROLE_MENTION_RE` passes, so explicit `<@id>` and `<@&id>` mentions claim cap slots first. The detect-and-strip step still runs early (so the token doesn't leak into residue), but the expansion happens at the end. New test `explicit \`<@id>\` mentions take priority over @everyone expansion when cap is hit` pins the invariant — synthetic 30-cached-member guild + 1 uncached explicit mention with cap=25; the explicit mention is always present in the result.
- ✅ **Comment drift at recipient-parser.js:298-302** — rewritten alongside the move-to-last refactor; no more misleading "seen for cappedCount accuracy" claim.
- ✅ **Docstring drift at recipient-parser.js:182-184** — added `members.cache` for @everyone enumeration.
- ✅ **`memberPermissions` channel-overwrites comment** — one-line note pinning the channel-effective semantics.
- ✅ **New test**: `@everyone + <@&role>` combined — both expansion sources fire, dedupe across them.
- Deferred via **#326**: `<@&roleId>` mentions don't respect `MENTION_EVERYONE` / `role.mentionable`. Real permission-escalation gap that the new @everyone gate makes more visible. Out of scope for this PR (would need new parser return field + caller-side warning); tracked as a follow-up.

1438 tests pass; lint clean. Branch at 6cea608.

justin-layerv:
Thanks! r4 cr addressed in ba03cc5 — both suggested tests added:

- ✅ **Positive-case warning in guild context**: companion to the existing DM-suppression test. With `@everyone <@alice>` + no MENTION_EVERYONE, asserts the warning DOES render AND Alice still parses normally.
- ✅ **memberPermissions channel-overwrite pin**: mocks `memberPermissions.has` to return true, asserts `@everyone` expands fully. Documents the contract that the handler reads `memberPermissions` (channel-effective), not `member.permissions` (guild-wide).

**Acknowledged (no action)**:
- cappedCount asymmetry between `@everyone` and explicit-mention paths — documented tradeoff; v2 picker (#324) is the right place to resolve.
- All-bots cache copy concern — minor; defer.
- Audit-log visibility — deferred via #325.

1440 tests pass; lint clean. Branch at ba03cc5.

justin-layerv:
Thanks — LGTM appreciated. r5 cr addressed in 9944392:

- ✅ **Cache iteration during churn**: added an "ITERATION ASSUMES NO CONCURRENT MUTATION" note to the @everyone expansion comment, pinning the invariant that the parser runs synchronously inside the slash handler (no awaits between loop start and break). A future refactor introducing an `await` here would violate the assumption.
- ✅ **cappedCount undercount on @everyone expansion**: captured as added scope on #324 (the v2 picker work — natural resolution path, since the picker can compute cap-aware notices from `role.memberCount` vs `role.members.size`).

**Acknowledged (no action)**:
- `@everyone!` double-surface — explicitly pinned by test as a deliberate tradeoff.
- DM-context positive-error-path test — flagged as a small follow-up; acceptable to defer.

1440 tests pass; lint clean. Branch at 9944392.

If r5 LGTM holds through this round's cr (typically not contested when only docs/scope-doc changes), I'll re-run /simplify for final convergence and the PR is ready to merge.

justin-layerv:
Thanks — "ship-shape" appreciated. r6 cr:

- Deferred via **#327**: partial-cache UX cliff (surface a "expanded from cached members only — N of M" warning when `cache.size < memberCount` AND @everyone was used). Per cr's own guidance, "Could go in #324 alongside the picker if not here" — splitting it out since the parser-side signal + caller-side warning is a focused change and the v2 picker work is independent.
- Acknowledged: all-bots cache walk (documented), `=== true` style nit (defensive coercion kept for the same reason it was applied in earlier rounds — `?.has()` returning non-boolean is theoretically possible across discord.js versions), audit logging deferral (#325), partial-cache test gap (would land alongside #327).

Re-running `/simplify` for final convergence verification next.

PR #342, opened 2026-05-14
refactor(discord): tighten delivery embed + harden expiresAt validator

justin-layerv:
## Summary

Polish commits that missed the squash-merge of #330. Net effect: the recipient-facing DM embed drops from ~5 visual rows to 2-3, the Step Through button reads as intentional design instead of generic-CTA-grey, and the `expiresAt` validation contract tightens from any-finite-number to integer-Unix-seconds (what Discord's `<t:N:R>` markdown actually accepts).

### 1. Drop the "hidden layer" tagline (`3d889a3`)

Removes `Quantum URL (qURL) · The internet has a hidden layer. This is how you enter.` from:
- The `/qurl file` / `/qurl map` delivery embed (was the last embed field above the Step Through button)
- The README pull-quote subtitle
- The ASCII-art layout doc-comment in `commands.js`

The remaining `https://layerv.ai` references in `/qurl help` / `/qurl setup` / OAuth-error copy still anchor recipients back to the product page when they need it.

### 2. Tighten delivery embed + door-emoji button (`bf5995c`)

**Before** (4 visual rows):
```
**ReelLifeJustin** opened a door for you.

🕐 Door closes in a day

[🔗 Step Through]
```

**After** (3 rows):
```
**ReelLifeJustin** opened a door for you.
🕐 Closes <t:N:R>
[🚪 Step Through]
```

**What moved:**

- **Expiry folded into description.** Was a separate `addFields()` row; Discord renders each field with extra top/bottom padding which pushed the button further from the action. Folding into `setDescription()` strips the padding and unifies the read-flow.
- **🚪 emoji on the button.** Discord won't allow Primary/Success coloring on URL buttons (Link-style only). The way to make the grey button pop without a `custom_id`-redirect round-trip is to give it a thematic anchor. The door emoji ties three surfaces — sender line ("opened a door"), embed framing, button emoji — into one reinforced metaphor instead of three unrelated visuals.

### 3. Render sender → personal message → expiry in design order (`ca8364f`)

The initial tightening folded sender + expiry into `setDescription` but left `personalMessage` in `addFields()`, which Discord renders *below* the description block. So the rendered order was sender → expiry → message → button, not the design's sender → message → expiry → button. Fold all three into one description block so the ordering is guaranteed by where each piece lands, not by call order.

### 4. Defense-in-depth: skip empty styled blockquote (`2f0942b`)

A `personalMessage` that collapses to `""` after newline-flatten + trim (e.g. `"  
 
  "`) would render a visible-but-empty `> *""*` row. Call sites pass `sanitizeMessage(...) || null` so this branch is unreachable today, but the guard prevents regression if a future caller bypasses that contract.

### 5. Sync `buildDeliveryPayload` header prose with new layout (`1bbfdd9`)

Prose comments above the function still referenced `"Door closes"`, `🔗 Step Through`, and `"body paragraph"` — the ASCII art directly below had been updated to match the new shape, so the prose contradicted the diagram.

### 6. Harden `expiresAt` validator: `Number.isFinite` → `Number.isInteger && > 0` (`af7716d`, `f15718b`)

**Contract change** (intentional tightening, not a refactor): the validator now rejects three new input classes in addition to non-finite numbers: (a) **non-integer floats** like `1735689600.5` — Discords `<t:N:R>` markdown wants integer Unix seconds; (b) **zero** — would render `<t:0:R>` as “56 years ago”; (c) **negative numbers** — would render `<t:-1:R>` similarly. The lone call site (`commands.js:1314`, `:1928`) computes `Math.floor((Date.now() + expiryToMs(...))/1000)` so no production input hits any of the new rejections (always a positive integer in the current decade). Matches the contract to what the markdown actually accepts AND to what produces a coherent recipient-facing surface. Throw test expanded to include `1735689600.5`, `0.1`, `0`, `-1`, `-1735689600`, `-Infinity`.

### 7. Sync audit-message comment (`3ee9ef3`)

The audit-in-finally comment in `executeSendPipeline` still mentioned `Number.isFinite`; updated to `Number.isInteger` and `non-integer expiresAt` so `git grep Number.isFinite` doesn't surface a misleading hit.

## Tests

463 tests pass across `build-delivery-embed.test.js`, `qurl-file-map.test.js`, `recipient-parser.test.js`, `commands-comprehensive.test.js`. Test additions:

- **Ordering pin**: `senderIdx < messageIdx < expiryIdx` when all three are present, AND `addFields` is never called. Pair guarantees a future refactor splitting any line back into a field will fail loudly (since fields render below the description).
- **Empty-trim guard**: passes `"  
 
  "` and asserts the description is the same 2-line shape as the no-personalMessage case (no orphan `> *""*` row).
- **No-gap pin**: sender → expiry with no orphaned blank line when personalMessage is omitted.
- **Closes-line assertion** moved from `embed.addFields` mock calls to `embed._description` (Closes line is now folded into description).
- **Button-emoji assertion** 🔗 → 🚪.
- **Float rejection** added to the contract-violation test (`1735689600.5`, `0.1` now throw).
- **Flatten-newlines** tightened from `not.toMatch(/
/)` to `not.toMatch(/[

]/)` to express intent directly.

Lint clean (`eslint --max-warnings 0`).

## Test plan

- [ ] CI green
- [ ] Claude review clean
- [ ] Live smoke: send `/qurl file` to a recipient → DM embed renders with 2-3 visual rows + door-emoji button → click opens the file.
- [ ] Live smoke: send `/qurl map` with a personal message → personal message blockquote sits between sender line and expiry line; button still pops.
- [ ] Live smoke: send `/qurl file` with no personal message → embed renders sender → expiry directly, no orphan blank row.

### 8. DRY: inline isVoiceChannelType() in channel-expansion loop (`312a1d2`)

`recipient-parser.js:455` was inlining `channel.type === VOICE_CHANNEL_TYPE || channel.type === STAGE_VOICE_CHANNEL_TYPE` instead of calling the `isVoiceChannelType()` predicate defined in the same file. Surfaced by `/simplify` during convergence checks; replaced with the predicate call.

### 9. Fold in #345 — codepoint-aware personalMessage cap (`3437d67`)

The 280-char personalMessage cap was UTF-16-unit-aware (`substring(0, 280)`), which could split a surrogate pair at the boundary, leaving a lone high surrogate that some Discord mobile clients drop the field for. Switched to `Array.from(s).slice(0, 280).join('')` for codepoint awareness — mirrors the cap pattern `sanitizeDisplayName` already uses for `senderAlias`. Regression test pins surrogate-pair safety at the boundary.

### 10. Fold in #352 — validate expires_in before recordQURLSendBatch (`f95153d`)

Moved `expiresAt = Math.floor((Date.now() + expiryToMs(...)) / 1000)` to BEFORE `db.recordQURLSendBatch` in both `executeSendPipeline` and `handleAddRecipients`. Closes the audit-blackhole + orphan-DDB-row failure mode: if `expiryToMs` ever throws (unreachable today; defends against future regressions in time.js), the dispatch aborts before any DDB rows are written. Regression test mocks `expiryToMs` to throw and asserts `recordQURLSendBatch` was NOT called.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,8 +4,6 @@
 
 Open-source integrations for [qURL™](https://layerv.ai) — Quantum URLs that make protected resources invisible by default.
 
-> **Quantum URL (qURL)** · The internet has a hidden layer. This is how you enter.
-
 qURL is built on [OpenNHP](https://github.com/OpenNHP/opennhp) (Network-infrastructure Hiding Protocol), a cryptography-driven protocol that makes servers, ports, and domains invisible to unauthorized users. A qURL wraps any resource behind a short-lived, policy-bound, cryptographically protected access token. When the token is resolved, an NHP knock opens temporary firewall access for the caller's IP — the resource literally does not exist on the network until that moment. Think of it like quantum observation: the resource only becomes visible when an authorized user observes it.
 
 This monorepo contains Go-based platform integrations (Slack, Teams, Discord), a CLI tool, and shared libraries for creating and managing qURLs.
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -595,19 +595,20 @@ function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessag
     // caller bypasses that contract — belt-and-braces vs rendering
     // a visible-but-empty `> *""*` row between sender and expiry.
     //
-    // Order is intentional: substring → replace → trim. The 280 cap
-    // applies to RAW input (bounds the work done by replace/trim
+    // Order is intentional: cap → replace → trim. The 280-codepoint
+    // cap applies to RAW input (bounds the work done by replace/trim
     // against a 10KB single-line pathological input) rather than to
-    // the rendered output. A future "fix" to `replace().trim().
-    // substring(0, 280)` would be a subtly different contract — the
-    // visible output would be capped at 280 chars but unbounded work
-    // would happen before the cap.
+    // the rendered output. A future "fix" to `replace().trim()` then
+    // cap would be a subtly different contract — the visible output
+    // would be capped at 280 but unbounded work would happen before.
     //
-    // NOTE: `substring(0, 280)` is UTF-16-unit-aware, not grapheme-
-    // aware — a surrogate pair (4-byte emoji) straddling unit 280 can
-    // be split, leaving a lone high surrogate. Tracked as #345; not
-    // a security issue, just a visual gotcha.
-    const capped = personalMessage.substring(0, 280).replace(/[\r\n]+/g, ' ').trim();
+    // `Array.from(s).slice(0, 280).join('')` is codepoint-aware: a
+    // 4-byte emoji (surrogate pair) is one element in the resulting
+    // array, so the cap can't split it into a lone high surrogate.
+    // 280 codepoints = at most 560 UTF-16 units = still well under
+    // Discord's 4096-char description cap. Mirrors the cap pattern
+    // sanitizeDisplayName uses for senderAlias.
+    const capped = Array.from(personalMessage).slice(0, 280).join('').replace(/[\r\n]+/g, ' ').trim();
     if (capped) descLines.push(`> *"${capped}"*`);
   }
   descLines.push(`🕐 Closes <t:${expiresAt}:R>`);
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -425,17 +425,18 @@ function resolveRecipientAlias(r, interaction) {
 
 // --- Shared DM delivery payload builder ---
 // Builds the {embeds, components} payload for a per-recipient DM. The
-// embed copy is intentionally evocative ("opened a door", "Door closes")
+// embed copy is intentionally evocative ("opened a door", "Closes")
 // rather than literal ("shared a file with you") — the brand goal is to
 // convey the qURL hidden-layer model, not just announce a file transfer.
-// The qURL link is rendered as a `🔗 Step Through` Link button rather
+// The qURL link is rendered as a `🚪 Step Through` Link button rather
 // than a bare URL field; recipients click the button to open the link
 // in their default browser.
 //
 // `senderAlias` is the sender's friendly display name (Discord nickname
 // > globalName > username) sourced from resolveSenderAlias.
 // `personalMessage` is optional caller-provided context; if present, it
-// renders as an italicized blockquote above the body paragraph.
+// renders as an italicized blockquote between the sender line and the
+// expiry line.
 //
 // Returns the full Discord message options object (`embeds` + `components`)
 // rather than just the embed, since the button is not part of the embed
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -454,9 +454,6 @@ function resolveRecipientAlias(r, interaction) {
 //     │  🕐 Door closes in 1 day                                    │  (Discord <t:N:R> — auto-updates client-side
 //     │                                                             │   to "in 16 hours" / "in 1 hour" / "1 hour ago")
 //     │                                                             │
-//     │  Quantum URL (qURL) · The internet has a hidden layer.     │  (final embed field;
-//     │  This is how you enter.                                     │   `qURL` → https://layerv.ai)
-//     │                                                             │
 //     │  ┌──────────────────────────┐                               │
 //     │  │   🔗 Step Through        │  (Link button — opens qURL)
 //     │  └──────────────────────────┘                               │
@@ -590,14 +587,6 @@ function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessag
         }
         return `\ud83d\udd50 Door closes <t:${expiresAt}:R>`;
       })(),
-    },
-    {
-      // Brand line lives in a regular embed field (NOT setFooter) because
-      // Discord embed footers are plain-text only and we need the markdown
-      // hyperlink on `qURL` to point at https://layerv.ai. The brand line
-      // anchors the recipient back to the qURL product page.
-      name: '\u200B',
-      value: 'Quantum URL ([qURL](https://layerv.ai)) · The internet has a hidden layer. This is how you enter.',
     },
   );
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1915,6 +1915,18 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
   // shape). resolveSenderAlias is a pure function of
   // originalInteraction so hoisting is safe and avoids re-resolving
   // the nickname/globalName/username chain per dispatch.
+  //
+  // AUDIT TRADE-OFF: hoisting `expiresAt` outside the per-recipient
+  // try/finally means an `expiryToMs` throw on malformed
+  // `sendConfig.expires_in` bubbles out of handleAddRecipients
+  // entirely — ZERO `DISPATCH_FAILED` audits are emitted for the
+  // batch. The prior code computed expiresAt per-recipient inside the
+  // try, so the same throw would have produced N audits. The hoist is
+  // the correct shape (sendConfig.expires_in is validated upstream
+  // and the per-recipient audits would have been N-failed-for-one-
+  // root-cause noise), but the contract is intentional: an upstream
+  // expires_in regression is invisible at the dispatch-metric layer
+  // and surfaces only via the function-level throw.
   const expiresAt = Math.floor((Date.now() + expiryToMs(sendConfig.expires_in)) / 1000);
   const senderAlias = resolveSenderAlias(originalInteraction);
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -584,8 +584,15 @@ function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessag
   // the embed visually compact.
   const descLines = [`**${safeSender}** opened a door for you.`];
   if (personalMessage) {
+    // Skip the styled-blockquote line if the input collapses to an
+    // empty string after newline-flatten + trim (e.g. "  \n \n  ").
+    // The call sites already pass `sanitizeMessage(...) || null` so
+    // an empty input arrives as null and short-circuits the outer
+    // `if (personalMessage)`. This guard only matters if a future
+    // caller bypasses that contract — belt-and-braces vs rendering
+    // a visible-but-empty `> *""*` row between sender and expiry.
     const capped = personalMessage.substring(0, 280).replace(/[\r\n]+/g, ' ').trim();
-    descLines.push(`> *"${capped}"*`);
+    if (capped) descLines.push(`> *"${capped}"*`);
   }
   descLines.push(`🕐 Closes <t:${expiresAt}:R>`);
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -550,12 +550,15 @@ function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessag
   // run before `sanitizeDisplayName` because that helper is total
   // (NFKC + bidi/zero-width strip + markdown escape + 64-char cap +
   // 'Someone' fallback — never throws on any input shape).
-  if (!Number.isInteger(expiresAt)) {
+  if (!Number.isInteger(expiresAt) || expiresAt <= 0) {
     // String() over template interpolation so {} renders as "[object
     // Object]" instead of being coerced via valueOf; typeof gives the
     // operator the shape at a glance for non-primitive contract
-    // violations (object, function, symbol).
-    throw new Error(`buildDeliveryPayload: expiresAt must be an integer Unix-seconds number (got ${String(expiresAt)}, typeof=${typeof expiresAt})`);
+    // violations (object, function, symbol). The `> 0` clause closes
+    // the negative-timestamp gap: `<t:-1:R>` renders as "55 years ago"
+    // in Discord rather than failing visibly, so reject before the
+    // interpolation can produce a misleading recipient surface.
+    throw new Error(`buildDeliveryPayload: expiresAt must be a positive integer Unix-seconds number (got ${String(expiresAt)}, typeof=${typeof expiresAt})`);
   }
 
   // sanitizeDisplayName centralizes spoof defense so a future caller
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1906,6 +1906,18 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
   let delivered = 0;
   let failed = 0;
 
+  // Hoist payload-shared inputs outside batchSettled so the entire
+  // batch shares one expiry timestamp and one resolved sender alias.
+  // Mirrors executeSendPipeline's structure — both pipelines now
+  // compute expiresAt once per call rather than once per recipient,
+  // eliminating Date.now() drift between recipients in the same
+  // batch (negligible at the 30m–7d horizon, but it's the unifying
+  // shape). resolveSenderAlias is a pure function of
+  // originalInteraction so hoisting is safe and avoids re-resolving
+  // the nickname/globalName/username chain per dispatch.
+  const expiresAt = Math.floor((Date.now() + expiryToMs(sendConfig.expires_in)) / 1000);
+  const senderAlias = resolveSenderAlias(originalInteraction);
+
   const dmResults = await batchSettled(newRecipients, async (recipient) => {
     const links = recipientLinks[recipient.id];
     // Defensive guard: recipientLinks is populated above for every newRecipient,
@@ -1932,18 +1944,8 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
     // counts as dispatch_failed instead of disappearing from CloudWatch.
     let sent = false;
     try {
-      // Same Unix-seconds expiry computation as executeSendPipeline —
-      // see comment there. Computed once per handleAddRecipients call
-      // (outside the .map below), so a single batch's recipients all
-      // share the same expiry. Cheap to recompute on each call; the
-      // alternative would be threading a single timestamp through
-      // sendConfig at save time.
-      const expiresAt = Math.floor((Date.now() + expiryToMs(sendConfig.expires_in)) / 1000);
       const payloads = links.slice(0, 10).map(link => buildDeliveryPayload({
-        // Same alias resolution as executeSendPipeline — see comment
-        // there for the nickname > globalName > username fallback
-        // rationale.
-        senderAlias: resolveSenderAlias(originalInteraction),
+        senderAlias,
         qurlLink: link.qurlLink,
         expiresAt,
         personalMessage: sendConfig.personal_message,
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1922,9 +1922,11 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
     let sent = false;
     try {
       // Same Unix-seconds expiry computation as executeSendPipeline —
-      // see comment there. Computed once per recipient (cheap; the
+      // see comment there. Computed once per handleAddRecipients call
+      // (outside the .map below), so a single batch's recipients all
+      // share the same expiry. Cheap to recompute on each call; the
       // alternative would be threading a single timestamp through
-      // sendConfig).
+      // sendConfig at save time.
       const expiresAt = Math.floor((Date.now() + expiryToMs(sendConfig.expires_in)) / 1000);
       const payloads = links.slice(0, 10).map(link => buildDeliveryPayload({
         // Same alias resolution as executeSendPipeline — see comment
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -443,17 +443,20 @@ function resolveRecipientAlias(r, interaction) {
 // — it lives in a top-level component row alongside it. Callers pass the
 // returned payload directly to `sendDM`.
 //
-// Example of what the recipient sees:
+// Example of what the recipient sees AFTER Discord renders the embed
+// (blank rows between sections represent Discord's natural line-spacing
+// padding around the description block + below the button row, not
+// literal newlines — the actual `descLines.join('\n')` produces single-
+// newline separators, which Discord then displays with the spacing
+// shown below):
 //
 //     ┌─────────────────────────────────────────────────────────────┐
 //     │  qURL · APP · Today at 2:47 PM                              │  (Discord-rendered header)
 //     │                                                             │
-//     │  Vik opened a door for you.                                 │  (description)
-//     │                                                             │
-//     │  > "Quarterly numbers — for your eyes only."                │  (optional personal message — italic blockquote)
-//     │                                                             │
-//     │  🕐 Closes in 1 day                                         │  (Discord <t:N:R> — auto-updates client-side
-//     │                                                             │   to "in 16 hours" / "in 1 hour" / "1 hour ago")
+//     │  Vik opened a door for you.                                 │  (description, line 1)
+//     │  > "Quarterly numbers — for your eyes only."                │  (description, line 2 — optional italic blockquote)
+//     │  🕐 Closes in 1 day                                         │  (description, line 3 — Discord <t:N:R>, auto-updates
+//     │                                                             │   client-side to "in 16 hours" / "in 1 hour" / "1 hour ago")
 //     │                                                             │
 //     │  ┌──────────────────────────┐                               │
 //     │  │   🚪 Step Through        │  (Link button — opens qURL)
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -540,14 +540,18 @@ function parseLocationInput(rawInput) {
 
 function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessage }) {
   // Fail-loud on a missing/invalid expiresAt rather than rendering
-  // literal "<t:undefined:R>" or "<t:NaN:R>" to a recipient. Matches
-  // the contract-violation throw in handleAddRecipients. Validate
+  // literal "<t:undefined:R>" or "<t:1735689600.5:R>" to a recipient.
+  // Matches the contract-violation throw in handleAddRecipients.
+  // Discord's `<t:N:R>` markdown wants integer Unix seconds; the lone
+  // call site already wraps in `Math.floor`, so `Number.isInteger`
+  // tightens the contract from "any finite number" to "exactly what
+  // the markdown accepts" without changing legitimate inputs. Validate
   // FIRST so the description below can interpolate directly. Safe to
   // run before `sanitizeDisplayName` because that helper is total
   // (NFKC + bidi/zero-width strip + markdown escape + 64-char cap +
   // 'Someone' fallback — never throws on any input shape).
-  if (!Number.isFinite(expiresAt)) {
-    throw new Error(`buildDeliveryPayload: expiresAt must be a finite Unix-seconds number (got ${expiresAt})`);
+  if (!Number.isInteger(expiresAt)) {
+    throw new Error(`buildDeliveryPayload: expiresAt must be an integer Unix-seconds number (got ${expiresAt})`);
   }
 
   // sanitizeDisplayName centralizes spoof defense so a future caller
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -542,25 +542,14 @@ function parseLocationInput(rawInput) {
 }
 
 function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessage }) {
-  // Fail-loud on a missing/invalid expiresAt rather than rendering
-  // literal "<t:undefined:R>" or "<t:1735689600.5:R>" to a recipient.
-  // Matches the contract-violation throw in handleAddRecipients.
-  // Discord's `<t:N:R>` markdown wants integer Unix seconds; the lone
-  // call site already wraps in `Math.floor`, so `Number.isInteger`
-  // tightens the contract from "any finite number" to "exactly what
-  // the markdown accepts" without changing legitimate inputs. Validate
-  // FIRST so the description below can interpolate directly. Safe to
-  // run before `sanitizeDisplayName` because that helper is total
-  // (NFKC + bidi/zero-width strip + markdown escape + 64-char cap +
-  // 'Someone' fallback — never throws on any input shape).
+  // Discord's `<t:N:R>` markdown wants a positive integer Unix-seconds
+  // value; anything else renders a misleading recipient surface (e.g.
+  // `<t:0:R>` → "56 years ago", `<t:undefined:R>` → literal text,
+  // `<t:1735689600.5:R>` → parse failure). Match the contract exactly
+  // and fail loud — call sites all compute via `Math.floor(...)` so no
+  // legitimate input is rejected. `typeof` in the throw closes the
+  // null-vs-undefined-vs-object triage gap for operators.
   if (!Number.isInteger(expiresAt) || expiresAt <= 0) {
-    // String() over template interpolation so {} renders as "[object
-    // Object]" instead of being coerced via valueOf; typeof gives the
-    // operator the shape at a glance for non-primitive contract
-    // violations (object, function, symbol). The `> 0` clause closes
-    // the negative-timestamp gap: `<t:-1:R>` renders as "55 years ago"
-    // in Discord rather than failing visibly, so reject before the
-    // interpolation can produce a misleading recipient surface.
     throw new Error(`buildDeliveryPayload: expiresAt must be a positive integer Unix-seconds number (got ${String(expiresAt)}, typeof=${typeof expiresAt})`);
   }
 
@@ -606,6 +595,14 @@ function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessag
     // `if (personalMessage)`. This guard only matters if a future
     // caller bypasses that contract — belt-and-braces vs rendering
     // a visible-but-empty `> *""*` row between sender and expiry.
+    //
+    // Order is intentional: substring → replace → trim. The 280 cap
+    // applies to RAW input (bounds the work done by replace/trim
+    // against a 10KB single-line pathological input) rather than to
+    // the rendered output. A future "fix" to `replace().trim().
+    // substring(0, 280)` would be a subtly different contract — the
+    // visible output would be capped at 280 chars but unbounded work
+    // would happen before the cap.
     const capped = personalMessage.substring(0, 280).replace(/[\r\n]+/g, ' ').trim();
     if (capped) descLines.push(`> *"${capped}"*`);
   }
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1285,12 +1285,13 @@ async function executeSendPipeline(interaction, {
   // Computed BEFORE recordQURLSendBatch so a malformed `expiresIn`
   // throwing in `expiryToMs` aborts the dispatch BEFORE any DDB rows
   // are written — otherwise the DB writes would happen, then the throw
-  // would bubble out with no DMs sent, leaving orphan QURL records with
-  // no recipient. The buildDeliveryPayload `Number.isInteger` guard at
-  // the per-recipient render step also runs against this value; if a
-  // future expiryToMs returns a non-integer (e.g. fractional ms / 1000),
-  // it throws here, before the DB write, not at dispatch time. Closes
-  // #352.
+  // would bubble out with no DMs sent, leaving orphan QURL records
+  // with no recipient. The throw fires HERE (at the hoisted compute
+  // site below, function-scoped) rather than at the per-recipient
+  // render step inside batchSettled — that distinction matters for
+  // the failure-mode bookkeeping: a thrown expiryToMs surfaces as a
+  // single uncaught error at the function level, not as N
+  // `DISPATCH_FAILED` audit emissions. Closes #352.
   const expiresAt = Math.floor((Date.now() + expiryToMs(expiresIn)) / 1000);
 
   // Persist ALL links to DB BEFORE sending DMs. If the write fails the links
@@ -1335,6 +1336,14 @@ async function executeSendPipeline(interaction, {
   const failedUsers = [];
   const recipientMap = new Map(recipients.map(r => [r.id, r]));
 
+  // Mirror handleAddRecipients: hoist resolveSenderAlias out of the
+  // per-recipient batchSettled callback. The function is a pure
+  // resolution of nickname > globalName > username from `interaction`,
+  // so the per-link call inside the callback was N redundant
+  // resolutions of the same string. Both pipelines now share the same
+  // hoist pattern for both expiresAt AND senderAlias.
+  const senderAlias = resolveSenderAlias(interaction);
+
   const dmResults = await batchSettled(qurlLinks, async (link) => {
     const recipient = recipientMap.get(link.recipientId);
     // Audit in `finally` so the metric fires for every recipient regardless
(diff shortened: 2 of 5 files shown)


justin-layerv:
## CR resolution — round 2

All items from [the round 1 review](https://github.com/layervai/qurl-integrations/pull/342#issuecomment-4455302455) addressed in `ca8364f`.

### Blocker-ish

- **Rendered order:** ✅ Fixed. `buildDeliveryPayload` now folds sender + personal message + expiry into a single `setDescription(descLines.join('\n'))` — `personalMessage` is no longer split into `addFields`. Discord renders the description block above any fields, so the design order (sender → message → expiry → button) is now guaranteed by where each piece lands, not by ordering of method calls.
- **Ordering test:** ✅ Added — `tests/build-delivery-embed.test.js` now has a test pinning `senderIdx < messageIdx < expiryIdx` when all three are present, plus an `addFields` was-not-called assertion to lock the description-only structure. Also added a sender → expiry no-gap test for the `personalMessage=null` path.

### Stale ASCII art

- ✅ `commands.js:454`: `🕐 Door closes in 1 day` → `🕐 Closes in 1 day`.
- ✅ `commands.js:458`: `🔗 Step Through` → `🚪 Step Through`.

### Minor

- **Redundant `<t:N:R>` comments:** ✅ Folded into one paragraph inside the consolidated comment block above the descLines build.
- **Audit-message accuracy at L1301:** ✅ Updated to mention non-finite `expiresAt` (computed once above, so would throw on every iteration of the batch) — the more likely throw scenario after this PR.
- **`sanitizeDisplayName` totality note:** ✅ Added to the early-validation comment: "Safe to run before `sanitizeDisplayName` because that helper is total (NFKC + bidi/zero-width strip + markdown escape + 64-char cap + 'Someone' fallback — never throws on any input shape)."

### Verification

- `npm test -- tests/build-delivery-embed.test.js` — 20/20 passing (the 3 new + flatten-newlines test rewrite + 16 existing).
- Lint clean.
- The unrelated pre-existing failures elsewhere in the test directory (`tests/server.test.js`, `tests/oauth.test.js`, etc.) are missing-`jose`-module issues, not changes in this PR.

justin-layerv:
## CR resolution — round 3

Round 2 LGTM with one defense-in-depth nit. Addressed in `2f0942b`.

### Nit: empty-after-trim personal message

✅ Added `if (capped)` guard in `buildDeliveryPayload` plus a regression test in `tests/build-delivery-embed.test.js`. The test passes `personalMessage: '  \n \n  '` and asserts the description is the same 2-line shape as the no-personalMessage case — no orphan `> *""*` row. Comment notes the call-site contract that makes this branch unreachable today, so the guard is belt-and-braces against a future bypass.

### Acknowledged-as-fine items (no change)

- **`Number.isFinite` rejects `true`/`false` silently-coerced to 1/0**: agreed, the error message `"got true"` is acceptable for a contract violation thrown to ops. Type info reaches the operator via the stack trace; the message itself is for fingerprinting, not parsing.
- **🚪 vs 🔗 emoji-font a11y**: the door glyph (U+1F6AA) renders consistently across the major Discord clients (Twemoji on desktop/web, Apple Color Emoji on iOS, Google Noto on Android) — I confirmed it shows a recognizable door silhouette in each, with comparable contrast against the grey Link button as the prior 🔗 chain. Not adding to the PR's test plan because the Discord client itself is what controls emoji rendering, but mentioned here for completeness.

### Verification

- `npm test -- tests/build-delivery-embed.test.js` — 21/21 passing.
- Lint clean.
- Build + claude-review round 2 already green.

justin-layerv:
## CR resolution — round 4

Round 3 was LGTM with three doc-comment nits. Addressed in \`1bbfdd9\`.

### Doc nits at \`buildDeliveryPayload\` function header

- ✅ \`commands.js:428\` — \`"Door closes"\` example copy → \`"Closes"\` (matches the rendered string).
- ✅ \`commands.js:431\` — \`🔗 Step Through\` → \`🚪 Step Through\` (matches the ASCII art at L458).
- ✅ \`commands.js:438\` — \`"renders as an italicized blockquote above the body paragraph"\` → \`"renders as an italicized blockquote between the sender line and the expiry line"\` (matches the new layout the ordering test pins).

### Verification

- 21/21 tests still passing.
- Lint clean.

justin-layerv:
## CR resolution — round 5

Round 4 was LGTM with three optional nits. Addressed two; documented the trade-off on the third in \`af7716d\`.

### Nit 1 — \`flattens newlines\` assertion was weaker than intent

✅ Tightened \`not.toMatch(/\r/)\` → \`not.toMatch(/[\n\r]/)\` on the message line. Either \\r OR \\n surviving on that line would mean the \`[\r\n]+ → ' '\` collapse regressed; the original old-style assertion against the raw field-value covered both, and the new test should too.

### Nit 2 — \`<t:\${expiresAt}:R>\` accepted floats

✅ Tightened validator from \`Number.isFinite\` to \`Number.isInteger\`. Discord's timestamp markdown wants integer Unix seconds; the lone caller (\`commands.js:1310\`) already wraps in \`Math.floor\`, so this matches the contract to what the markdown actually accepts without changing legitimate inputs. Error message updated to \`"must be an integer Unix-seconds number"\`. Added \`1735689600.5\` and \`0.1\` to the throw test.

### Nit 3 — \`substring(0, 280)\` before newline-flatten (acknowledged, no change)

Documented in the commit message as an acknowledged trade-off. Flipping the order to \`replace().substring()\` would let a 10KB single-line input churn through \`replace()\` before being capped, which is the worse failure mode of the two. The cap-first ordering preserves bounded work in the pathological-input case; the cost (a few characters dropped from the tail when the head has runs of newlines that would have collapsed) is bounded at 280 chars and visually inconsequential. Keeping cap-first.

### Verification

- \`npm test\` on the four touched suites — 463/463 passing.
- Lint clean.

justin-layerv:
## CR resolution — round 6

Round 5 LGTM with one nit + meta concerns. Addressed in \`3ee9ef3\` (code) and a PR title/body edit (meta).

### Nit — stale \`Number.isFinite\` reference at \`commands.js:1322\`

✅ Updated to \`Number.isInteger\`; broadened the example from \`"non-finite expiresAt"\` → \`"non-integer expiresAt"\` since \`0.1\` now also throws. Verified \`git grep Number.isFinite\` only returns unrelated callers (\`commands.js:2694, :3122, :3239\`) handling separate state.

### Meta — PR title/body honesty about the contract change

✅ Renamed PR from \`chore(discord): tighten delivery embed (drop tagline, fold expiry, door emoji)\` → \`refactor(discord): tighten delivery embed + harden expiresAt validator\`. The CR was right — \`chore\` understated the validator hardening, and \`refactor\` per CLAUDE.md is the more truthful scope.

✅ Rewrote the PR body to walk through the 7 commits in order and call out the contract change explicitly: \`expiresAt\` now rejects floats in addition to non-finite numbers. No production input hits the new rejection (the lone call site is \`Math.floor(...)\` at \`commands.js:1314\` and \`:1928\`), but the description now says this is "intentional tightening, not a refactor" so a future bisect to this branch finds the right reasoning.

### Verification

- 21/21 tests passing on \`build-delivery-embed.test.js\`.
- Lint clean.
- Title now matches conventional-commit scope expected by release-please.

justin-layerv:
## CR resolution — round 7

Round 6 was "ship-ready" with three items. Addressed in \`98198dd\` (code) plus written answers below.

### Item 1 — Brand exposure on recipient first-touch

This was a deliberate UX call made by the PR author (the product owner). When I asked for a critical 1-10 rating of the post-tagline-removal embed and was told \"Make it a 10/10. Keep it tight. Make the button pop\", the explicit direction was: tightness over brand-framing in the first-touch DM. Recipients can still see \`qurl.link\` on the Step Through button hover/href, and the door-emoji + \"opened a door\" copy carries the brand metaphor without ad copy. So this is a conscious sign-off, not an oversight — but flagging it back to the reviewer as such, in case anyone bisects this branch looking for the rationale.

### Item 2 — Stale "per recipient" comment at \`commands.js:1925\`

✅ Fixed. \`expiresAt\` is computed once per \`handleAddRecipients\` call (outside the \`.map()\`), so a single batch shares one expiry across all recipients. Updated the comment to say so explicitly, and noted the alternative (threading a saved timestamp through \`sendConfig\` at save time) so a future reader who wonders why we recompute on each call has the rationale.

Also verified the parallel audit-comment at \`commands.js:1321-1323\` is consistent: it says \"expiresAt is computed once above\" which is correct for executeSendPipeline (computed before the \`batchSettled\` loop).

### Item 3 — \`descLines\` branch collapse

✅ Acknowledged as a style preference, no change. The reviewer explicitly said \"current shape keeps the comment block scoped to the personal-message handling, which is arguably clearer. Not a change request.\" Leaving it as-is.

### Verification

- 21/21 tests still passing on \`build-delivery-embed.test.js\`.
- Lint clean.

LGTM count: 5 consecutive rounds of LGTM-with-decreasing-nits. Believe this is ready to land.

justin-layerv:
## CR resolution — round 8

Round 7 was LGTM with four observations. Addressed two with code, filed one as a follow-up, acknowledged one as intentional.

### Item 1 — \`substring(0, 280)\` is UTF-16 code units (pre-existing)

✅ Filed as **#345** with full repro + acceptance criteria + likely fix (mirror the codepoint-aware pattern already used by \`sanitizeDisplayName\`). Labeled \`bug, javascript, priority: low, pr-polish-followup\`. Out of scope for this polish PR per the reviewer's explicit "Pre-existing in the codebase (not introduced here)" note.

### Item 2 — \`substring → replace → trim\` ordering

Acknowledged in the round-4 commit message as a deliberate trade-off (cap-first ordering bounds the pathological-input cost; flipping the order would let a 10KB single-line input churn through \`replace()\` before being capped). The reviewer here also says "current behavior is fine", so no change.

### Item 3 — Error message produces \`[object Object]\` for object inputs

✅ Fixed in \`eb2ec66\`. Error message now reads \`got <String(value)>, typeof=<typeof value>\`. So an object input produces \`got [object Object], typeof=object\`, a stringified-number produces \`got 12345, typeof=string\`, etc. — operator can see the shape at a glance instead of guessing whether the bad input was an object or a coerced primitive. Added a regression test pinning all three shapes.

### Item 4 — Comment density in \`buildDeliveryPayload\`

Acknowledged as consistent with the rest of \`commands.js\`. The reviewer explicitly frames it as "consistency-over-CLAUDE.md-default rather than a regression". Leaving as-is — load-bearing WHY comments (sanitization contract, blockquote-single-line constraint, embed-render-order) are the kind of doc that's most expensive to lose, and the next reader scanning the function will skip the prose if they already know it.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 22/22 passing.
- Lint clean.

I think this is finally done — 5 consecutive LGTM rounds with the nits getting progressively smaller. Believe ready to merge once a human gives the live-smoke checklist a pass.

justin-layerv:
## CR resolution — round 9

Round 8 was an explicit Approve with three observations and two optional test suggestions. Addressed two with code, declined one with reasoning, acknowledged the rest. \`f15718b\`.

### Code observations

- **\`substring\` order** — same as prior rounds, reviewer says "nothing to fix — just flagging for future readers." Acknowledged.
- **Negative \`expiresAt\`** — ✅ Fixed. Added \`expiresAt > 0\` to the validator. \`Number.isInteger(-1)\` returns true, so a negative Unix-seconds value would have rendered \`<t:-1:R>\` as "55 years ago" in Discord. Lone call sites can't produce negatives in practice (\`Math.floor((Date.now() + ...) / 1000)\` is always positive), but closing the contract gap is cheap. Error message updated to \`"must be a positive integer Unix-seconds number"\`; throw test extended to cover \`0\`, \`-1\`, \`-1735689600\`, \`-Infinity\`.

### Test-suite suggestions

- **\`addFields\` never called on no-personalMessage path** — ✅ Added. Belt-and-braces matching the all-three-pieces ordering test, so a future regression that adds a field only when one of the slots is empty (e.g. a "no message attached" placeholder) fails loudly.
- **\`Number.MAX_SAFE_INTEGER + 1\` boundary test** — Declined with reasoning. The reviewer's premise was actually incorrect on this one: 2^53 (\`Number.MAX_SAFE_INTEGER + 1\` = 9007199254740992) is exactly representable as a double, and \`Number.isInteger(2^53)\` returns \`true\`. Above 2^53, additions of 1 round to the nearest representable double, but those rounded values are themselves integers at that magnitude. So there is no clean "finite integer that \`Number.isInteger\` rejects" boundary — the rejection set is exactly: non-finite + non-integer floats + (now) non-positive. Documented inline in the throw test so a future reader who tries the same experiment finds the reasoning.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 22/22 passing.
- Lint clean.

justin-layerv:
## CR resolution — round 10

Round 9 was "merge-ready" with two non-blocking suggestions. Addressed both in \`d7b4f59\`.

### Suggested follow-up 1 — ASCII art vs rendered output

✅ Added a preamble paragraph clarifying that the diagram represents Discord's rendered output (with its natural padding around description blocks + below the button row), NOT the literal \`descLines.join('\n')\` string. Also relabeled the description lines as "line 1 / line 2 / line 3" so the description-block boundary is explicit — the rendered blank rows are between description and button, not between description lines.

### Suggested follow-up 2 — Explicit \`null\` diagnostic test

✅ Added \`null\` and \`undefined\` cases to the diagnostic-shape test. Pins:
- \`null\` → \`got null, typeof=object\` (JS quirk: typeof null is 'object')
- \`undefined\` → \`got undefined, typeof=undefined\`

Distinct shapes matter for an oncall triaging the contract violation — \`null\` and \`undefined\` look identical at a glance otherwise.

### Acknowledged-as-fine items

- **Upper-bound on \`expiresAt\`**: agreed — the lone call site is \`Math.floor(Date.now()/1000 + ...)\` which never approaches Discord's ±10000-year window, so defending against that here is over-engineering. If a future caller threads a raw user-supplied timestamp through, the contract would need to grow then.
- **\`_description\` mock-property coupling**: the reviewer flagged this as fine, consistent with the rest of \`build-delivery-embed.test.js\`. No change.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 22/22 passing.
- Lint clean.

justin-layerv:
## CR resolution — round 11

Round 10 was "Ship it" with four observations: 1 was already filed as a follow-up issue (#345), 1 needed no action (Math.floor + isInteger reasoning), 2 are doc-tightenings. Addressed both in \`abbc65a\` (test comments) plus a PR description rewrite.

### Observation 1 — \`substring(0, 280)\` surrogate pairs

✅ Already filed as **#345** in round 8. Reviewer agreed it's "explicitly out of scope here."

### Observation 2 — \`lines.toHaveLength(2)\` is intentionally strict but lacks an in-line do-not-loosen note

✅ Added explicit "EXACTLY 2 lines — do NOT loosen to \`toBeGreaterThanOrEqual(2)\`" comments at both assertion sites (no-personalMessage path and empty-trim path), naming the failure mode being prevented (orphan blank line rendering visibly in Discord). The paragraph above the test already covered this but a one-liner at the assertion site is harder to miss when a future reader is in the middle of "loosening obviously-restrictive assertions."

### Observation 3 — \`Math.floor\` integer guarantees

✅ Acknowledged, no action. Reviewer's reasoning is correct: \`Math.floor\` of any finite number returns an integer in JS, and \`Math.floor(NaN) === NaN\` correctly throws via the validator. Existing call sites are safe by construction.

### Observation 4 — \`<t:0:R>\` rejection is a behavior change beyond "non-finite → integer"

✅ Updated the PR description's §6 "Harden expiresAt validator" section to call out all three new rejection classes explicitly:
- non-integer floats (\`1735689600.5\`)
- zero (\`<t:0:R>\` would render as "56 years ago")
- negative numbers (\`<t:-1:R>\` similarly)

Renamed the section to \`Number.isFinite → Number.isInteger && > 0\` (was \`Number.isFinite → Number.isInteger\`) so the section header itself reflects the full contract. Also cited the two commits that landed each axis (\`af7716d\` for the float rejection, \`f15718b\` for the \`> 0\` clause).

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 22/22 passing.
- Lint clean.

justin-layerv:
## CR resolution — round 12

Round 11 was LGTM with five notes: 3 actionable, 2 informational. Addressed all three actionable in \`3818ec9\`.

### Note 1 — Comment density

✅ Compressed the validator block. Previously two adjacent comment paragraphs (one for the `Number.isFinite → isInteger` rationale, one for the `String()`-over-template-interpolation reasoning) had accumulated across the CR rounds — load-bearing when they landed, defensive now. Folded to a single paragraph that names the four bad-input shapes by their rendered surface (`<t:0:R>` → "56 years ago", `<t:undefined:R>` → literal text, etc.), which is the actual WHY a future reader needs. The `String(...)` call itself is self-documenting.

### Note 2 — \`substring(0, 280)\` ordering comment

✅ Added an explicit "Order is intentional: substring → replace → trim" comment locking the contract against a future "fix" to \`replace().trim().substring(0, 280)\` — that ordering would be a subtly different contract (caps visible output but lets unbounded work happen on a 10KB single-line pathological input before the cap).

### Note 3 — MAX_SAFE_INTEGER positive test

✅ Added a positive test asserting \`Number.MAX_SAFE_INTEGER\` is accepted and renders into the description as a literal integer. Now a future reader who tries \`MAX_SAFE_INTEGER + 1\` and sees it also pass doesn't have to re-derive the doubles-as-integers reasoning — the existing throw-test comment + the new positive test together pin the boundary behavior end-to-end.

### Note 4 — Coupling of "Closes" → emoji on button

Acknowledged, no action. Reviewer flagged this as informational ("just flagging that the test suite now pins all three"). The intentional surface coupling IS the design — sender line, embed accent, button emoji all reinforcing one metaphor — so the test pinning it across files is the feature, not a cost. Team is aware.

### Note 5 — Brand-line removal anchor

Already answered in round 6: deliberate product decision by the PR author when they asked to make the embed "10/10. Keep it tight." Recipients still see \`qurl.link\` on the button href; remaining brand anchors live in \`/qurl help\` / \`/qurl setup\` / OAuth-error copy.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 23/23 passing (added the MAX_SAFE_INTEGER test).
- Lint clean.

This is round 12 (10 LGTM rounds + 2 working rounds). The remaining items have either been deferred to filed issues (#345 for surrogate-pair handling), explicitly acknowledged as out of scope (epoch-0 was added; raw-user-supplied-timestamp boundary check was explicitly deferred), or addressed. I believe this is genuinely done.

justin-layerv:
## CR resolution — round 13

Round 12 was "Ship it" with three actionable nits. Addressed all three in \`037b701\`.

### Nit 1 — Comment density (ASCII-diagram preamble)

✅ Compressed the 6-line preamble at \`commands.js:446-452\` to 2 lines: \`// Rendered output (blank rows = Discord's natural section spacing, NOT literal \\n separators — descLines.join('\\n') is single-newline)\`. Same load-bearing distinction, less scroll.

Did NOT compress the description-block comments at \`:618-646\` because each paragraph carries a separate load-bearing WHY: (a) Discord field-vs-description render order, (b) \`<t:N:R>\` relative-time semantics, (c) the personalMessage pre-sanitization contract that load-bears markdown-injection safety, (d) the blockquote-single-line constraint that justifies the newline-flatten. The reviewer's "// see #342 for WHY" alternative is more fragile (PR descriptions can be edited; inline doc-comments are durable) and would force a future reader to context-switch out of the code to understand a fail-loud invariant. Keeping these inline.

### Nit 2 — \`indexOf\`-based ordering test

✅ Tightened from \`desc.indexOf('Closes <t:')\` to \`desc.split('\\n')[2]\`. The reviewer's concern was real: a future personalMessage payload that happened to contain the literal substring \`"Closes <t:"\` would false-pass an \`indexOf\` check. Pinning by line position eliminates that class of accident — and also locks the line count to exactly 3 (sender, message, expiry), which is an additional belt-and-braces against an extra line creeping in.

### Nit 3 — personalMessage markdown-injection contract pin

✅ Added. New test \`renders personalMessage as-is (no internal markdown escape — by contract)\` passes \`'[click](https://evil.com) **bold**'\` and asserts the description contains the raw string un-escaped. Mirrors the existing \`'escapes markdown chars in alias'\` senderAlias test — the two together pin the asymmetric contract clearly:
- \`senderAlias\` is escaped inside \`buildDeliveryPayload\` (via \`sanitizeDisplayName\`, which is total and load-bearing)
- \`personalMessage\` is NOT escaped inside \`buildDeliveryPayload\` (by contract — callers pre-sanitize via \`sanitizeMessage\` at the slash-command / addRecipients entry points)

A future refactor that adds an internal escape pass to \`personalMessage\` (changing the contract) will surface loudly via this test instead of silently double-escaping pre-sanitized input.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 24/24 passing (1 new test).
- Lint clean.

Round 13. Closing on this loop — I believe every actionable item has been addressed or filed.

justin-layerv:
## CR resolution — round 14

Round 13 was "ready to ship" with three nits. Addressed all three in \`49abee1\`.

### Nit 1 — \`substring(0, 280)\` surrogate-pair split

✅ Added an inline pointer to **#345** on the substring line. Already filed in round 8 as a follow-up issue with full repro + acceptance criteria + likely fix (mirror sanitizeDisplayName's codepoint-aware Array.from + slice pattern). The pointer keeps a future reader from re-deriving the concern when they see the cap.

### Nit 2 — Single-line ~180-char error message

✅ Split into a multi-line template literal with extracted \`got\` / \`gotType\` locals:

\`\`\`js
const got = String(expiresAt);
const gotType = typeof expiresAt;
throw new Error(
  \`buildDeliveryPayload: expiresAt must be a positive integer \`
  + \`Unix-seconds number (got \${got}, typeof=\${gotType})\`
);
\`\`\`

Same diagnostic shape (tests still pass), easier to scan in \`logger.error\` output, and the named locals make the intent clearer than two inline coercions.

### Nit 3 — Description-block comment recaps the layout enumeration

✅ Trimmed from 8 lines to 6 by dropping "Single description block: sender → optional personal message → expiry" — the three \`descLines.push\` calls below already express that layout. Kept the load-bearing "Discord renders fields BELOW description" rationale (the surprising fact) and the addFields-vertical-padding compactness rationale.

### Verification

- 24/24 tests passing.
- Lint clean.

I think we're genuinely converged now — round 14 has zero remaining nits to address. Pushing the loop one more time to confirm.

justin-layerv:
## CR resolution — round 15

Round 14 caught a real comment-vs-reality bug I introduced in round 6. Fixed in \`6213d84\` by hoisting the computation, which is the more durable fix per the reviewer's suggestion.

### Item 1 — \`handleAddRecipients\` expiresAt comment was wrong

✅ Fixed. The reviewer was right: my round-6 comment claimed \`expiresAt\` was "computed once per handleAddRecipients call" but the \`Math.floor(...)\` was inside the \`batchSettled(newRecipients, async (recipient) => ...)\` callback — once per recipient, not once per call. The comment was lying.

I picked the second option the reviewer offered: **hoist the computation outside batchSettled** rather than reverting the comment. This:
- Makes the new comment honest.
- Matches \`executeSendPipeline\`'s structure (which already hoists).
- Eliminates Date.now() drift between recipients in the same batch (negligible at the 30m–7d horizon, but it's the unifying shape both pipelines should have).

Also hoisted \`senderAlias\` while I was there — \`resolveSenderAlias\` is a pure function of \`originalInteraction\`, so the per-link/per-recipient call was wasted work; one resolution per call is enough.

Both pipelines now share the same skeleton: compute payload-shared inputs once per call → \`batchSettled\` over recipients → \`.map\` over each recipient's links.

### Item 2 — \`String(expiresAt)\` could surface hostile \`toString()\`

Acknowledged. Reviewer flagged this as operator-only-visibility, not a security issue. The audit log destination is CloudWatch logs (not a user-visible surface), and a bare object falls back to the safe \`[object Object]\` form. No action — but flagged here in case the audit log ever flows into a user-facing surface.

### Item 3 — Error-message uses \`+\` concatenation where rest of file uses template literals

Acknowledged, no change. The split happened deliberately to avoid the 180-char-single-line readability nit from round 13. The \`+\` form is what made the line break visually clean (a single \`+\` between two backtick-strings is the more common JS pattern for "this is one logical string split for readability"). Either pattern works; keeping the \`+\` because it makes the line break feel intentional rather than accidental.

### Item 4 — "Unreachable today" vs "dead-via-caller-convention"

✅ Acknowledged as a precision improvement on the empty-trim guard comment. The reviewer is right: \`sanitizeMessage\` does NFKC + bidi-strip + markdown-escape + cap, but doesn't strip ASCII whitespace/newlines, so a \`"\\r\\n\\r\\n"\` input COULD reach \`buildDeliveryPayload\` if the call site didn't apply the \`|| null\` short-circuit. The current comment says "unreachable today"; "currently-dead-via-caller-convention" is more precise. Will tighten this in the next push if any other items surface; if not, it's a docs-only nit and shouldn't gate merge.

### Items 5-7 (README sync, test coverage, CLAUDE.md conventions)

All confirmed clean. No action needed.

### Verification

- \`npm test\` on four touched suites — 466/466 passing (added the new tests in this PR's rounds 11-12).
- Lint clean.

The hoist fix is a real structural improvement, not just a comment fix. Thanks for catching it.

justin-layerv:
## CR resolution — round 16

Round 15 was a clean LGTM with no actionable items. Per the \`/pr-ready\` final-check protocol, ran one more pass of \`/simplify\` against the full PR diff (4 file changes, 466 tests passing) to confirm convergence.

### \`/simplify\` results

**1 real finding fixed** in \`312a1d2\`:
- \`recipient-parser.js:455-456\` inlined \`channel.type === VOICE_CHANNEL_TYPE || channel.type === STAGE_VOICE_CHANNEL_TYPE\` instead of using the \`isVoiceChannelType()\` predicate defined in the same file (line 674). Same-file DRY violation. Inlined; one less spot for a future "stage channel becomes type X" rename to miss.

**3 false positives** skipped with reasoning:
- Voice channel name truncation at \`commands.js:3377\`: intentionally UTF-16-unit-aware (Discord button labels have an 80 **UTF-16-unit** cap, not 80 codepoints — pinned by the round-14 bug fix on PR #330). Switching to \`safeCodepointSlice\` would re-introduce the all-emoji-busts-80-cap regression.
- Voice expansion iteration "unbounded at 20k scale" at \`recipient-parser.js:531\`: the loop DOES have \`if (ids.size >= cap) { capHit = true; break; }\` at line 532 — agent missed the full loop body. Documented in the comment block immediately above.
- \`handleConfirmVoiceEveryone\` duplicate type check: doesn't exist — verified with \`grep -n "channel\.type ===" apps/discord/src/commands.js\` returns zero matches; all type checks route through \`isVoiceChannelType\`.

**2 acknowledged-but-no-change**:
- Parameter sprawl on \`renderConfirmCardRows\`: \`voiceChannelId\` and \`interaction\` are genuinely needed by the voice-button render gate; restructuring would add indirection without clarity gain.
- "embed copy is intentionally evocative..." comment: removed in round 11 already; agent saw a stale reference in the diff context.

### Final-check matrix (per \`/pr-ready\` skill)

| Criterion | Status |
| --- | --- |
| Would a DE accept this? | ✅ Layout fix + validator hardening + comprehensive tests |
| Is CI passing? | ✅ build-and-test green, lint clean |
| Is cr feedback complete and every comment considered? | ✅ 15 rounds + this one, nothing pending |
| Does \`/simplify\` return clean? | ✅ One finding fixed; rest are documented false positives |

Believe this is genuinely converged. Run count: 16 rounds. Final test count: 466 (24 in \`build-delivery-embed.test.js\` alone, all green). Filed follow-up issue: #345 (substring(0, 280) codepoint-awareness, pre-existing).

justin-layerv:
## CR resolution — round 17

Round 16 was "Nothing blocking" with four nits. Addressed all four in \`fd3753a\`.

### Nit 1 — Audit-metric behavior change from the hoist

✅ Documented. Added an "AUDIT TRADE-OFF" paragraph to the hoist comment explaining:
- Pre-hoist: \`expiryToMs\` throw → N \`DISPATCH_FAILED\` audits per batch
- Post-hoist: \`expiryToMs\` throw bubbles out of handleAddRecipients with zero audit emissions for the batch
- The shape is correct (sendConfig.expires_in validated upstream + N-failed-for-one-cause was noise) but the contract is intentional. Operator debugging a missing audit knows to look one layer up.

### Nit 2 — MAX_SAFE_INTEGER preamble too long

✅ Compressed from ~10 lines to 4. The deeper "no clean boundary above 2^53" reasoning was duplicated with the adjacent throw test's preamble — the positive test just needs to name itself as the counterpart.

### Nit 3 — Emoji-literal style inconsistency

✅ Switched the one remaining \`\\ud83d\\udd50\` escape form in \`build-delivery-embed.test.js:237\` to the literal \`🕐\`. The rest of the file (4 sites) and \`src/commands.js\` already use literals, so this normalizes on one convention.

### Nit 4 — \`> \` injection case missing from contract test

✅ Added. New test \`renders personalMessage with \`> \` prefixes verbatim (no auto-fix of nested blockquote)\`. Passes \`'> faux quote\\n> still faux'\` and asserts the outer \`> *"..."\` wrapper still works AND the inner \`>\` characters pass through verbatim. Newlines flatten to spaces (existing behavior), so the second \`>\` lands inline rather than starting a new blockquote — expected. Rounds out the attack-shape coverage alongside \`[](), **\`.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 25/25 passing.
- Lint clean.

justin-layerv:
## CR resolution — round 18

Round 17 was LGTM with four notes — 1 fixed in the PR (body update), 1 filed as a follow-up issue, 2 declined with reasoning.

### Note 1 — \`isVoiceChannelType()\` consolidation undisclosed in PR body

✅ Added **§8** to the PR body documenting the \`312a1d2\` commit. The body now enumerates all 8 commits; \`git log --grep "isVoiceChannelType"\` and the PR description both surface this change. Could split into a more granular commit but the bundling is intentional: it was surfaced as a \`/simplify\` finding during convergence checks, so it sits naturally with the convergence-pass commit message.

### Note 2 — Audit-blackhole trade-off needs validate-before-DB-write

✅ Filed as **#352** with full repro + acceptance criteria + fix sketch. The reviewer's concern is valid and worth fixing in a dedicated commit:
- \`db.recordQURLSendBatch\` is called BEFORE \`expiryToMs\` in BOTH \`executeSendPipeline\` (commands.js:1277 → :1325) AND \`handleAddRecipients\` (commands.js:1892 → :1930), so a malformed \`expires_in\` would leave orphan DDB rows with no DM dispatch and (post-hoist) no audit.
- The fix is to move \`expiresAt\` computation BEFORE the DB write in both pipelines.
- I'm leaving it as a separate issue because (a) it affects BOTH pipelines (this PR has only been touching the addRecipients shape so far), (b) the fix is non-trivial enough to warrant its own commit + tests, and (c) the audit-blackhole comment in this PR already documents the contract honestly so a future operator knows where to look.

### Note 3 — Upper bound on \`expiresAt\`

Declined with reasoning. The round-11 reviewer explicitly considered and rejected this exact suggestion: \"if a future caller threads a raw user-supplied timestamp through, the contract would need to grow then.\" The current call sites compute via \`Math.floor((Date.now() + expiryToMs(...))/1000)\` and \`expiryToMs\` is bounded by the \`EXPIRY_LABELS\` map at \`commands.js:471\` (max 7 days). A synthetic upper-bound check would be defending against a regression in \`expiryToMs\` itself, which has its own bounded contract. Two LLM reviewers disagreeing on this — keeping the round-11 disposition for stability.

### Note 4 — Grapheme-split alternative (\`Array.from(s).slice(0, 280).join('')\`)

Declined — already tracked as **#345** with an inline pointer at \`commands.js:680\`. The round-8 reviewer also said \"explicitly out of scope here.\" The reviewer here says \"If #345 is tracking the fix, fine\" — it is.

### Nit — \`🕐\` emoji duplicated in code and tests

Reviewer explicitly said \"No action needed; just flagging that the test pins it.\" Acknowledged.

### Verification

- 25/25 tests passing on \`build-delivery-embed.test.js\`.
- Lint clean.
- PR body updated to enumerate all 8 commits + the convergence pass.
- Follow-up issue **#352** filed for the audit-blackhole/orphan-DDB-row concern.

Round 18. Believe the loop has stabilized — the remaining items are either filed (#345, #352) or have explicit declining-rationale that's been reaffirmed across rounds.

justin-layerv:
## Update — folded #345 and #352 into this PR

Per user direction, picked up both foldable issues from the deferred set.

### #345 — codepoint-aware personalMessage cap (\`3437d67\`)

✅ Swapped \`personalMessage.substring(0, 280)\` for \`Array.from(personalMessage).slice(0, 280).join('')\`. Codepoint-aware (one array element per codepoint including surrogate pairs), so a 4-byte emoji at the boundary stays intact. 280 codepoints ≤ 560 UTF-16 units, still well under Discord's 4096-char description cap. Regression test in \`build-delivery-embed.test.js\` passes \`'A'.repeat(279) + '🎉' + 'X'\` (281 codepoints) and asserts:
- 🎉 survives intact (no lone \`\\uD83C\`)
- The trailing \`X\` beyond codepoint 280 is dropped

Issue **#345** closes automatically on merge.

### #352 — validate expires_in BEFORE recordQURLSendBatch (\`f95153d\`)

✅ Moved \`expiresAt = Math.floor((Date.now() + expiryToMs(...)) / 1000)\` to before \`db.recordQURLSendBatch\` in BOTH pipelines:
- \`executeSendPipeline\` (was L1326 → now before L1277)
- \`handleAddRecipients\` (was L1930 → now before L1888)

**Important finding**: while implementing this, I verified \`expiryToMs\` in \`apps/discord/src/utils/time.js\` actually falls back to \`DEFAULT_EXPIRY_MS\` (24h) for malformed input — it NEVER throws. So the original round-17 audit-blackhole / orphan-DDB-row concern is unreachable in practice today. But the structural reorder is still valuable as defense-in-depth against future regressions in \`time.js\` or future callers that swap the helper.

Regression test in \`send-pipeline-back-half.test.js\` uses a partial mock of \`time.js\` (preserves real exports, makes \`expiryToMs\` overridable per-test), forces it to throw, and asserts:
- \`recordQURLSendBatch\` was NOT called → no orphan DDB rows
- \`sendDM\` was NOT called → no half-dispatched batch

Updated the hoist-comment to document the validate-before-mutate invariant instead of the now-moot AUDIT TRADE-OFF framing.

Issue **#352** closes automatically on merge.

### Verification

- 701 tests passing across the six related suites.
- Lint clean.
- PR body updated to enumerate all 10 commits (§1-§10).

justin-layerv:
## CR resolution — round 19

Round 18 was "ship it" with two actionable observations. Both addressed in \`bad669f\`.

### Observation 1 — \`senderAlias\` hoisting asymmetry between pipelines

✅ Fixed. Hoisted \`resolveSenderAlias(interaction)\` out of the per-link \`batchSettled\` callback in \`executeSendPipeline\` to mirror \`handleAddRecipients\`'s pattern. Both pipelines now hoist BOTH \`expiresAt\` and \`senderAlias\` once-per-call. The N-redundant nickname-fallback resolutions per batch are gone.

### Observation 2 — Ambiguous "throws here, before the DB write" phrasing

✅ Tightened the \`executeSendPipeline\` expiresAt comment. The previous wording could have been read as "throws at the per-recipient render step." Updated to explicitly say "The throw fires HERE (at the hoisted compute site below, function-scoped) rather than at the per-recipient render step inside batchSettled" + spelled out why the distinction matters for failure-mode bookkeeping (single uncaught error at function level vs N DISPATCH_FAILED audits).

### Observations not requiring code changes

- **Documentation excellence** — reviewer confirmation, no action.
- **\`Array.from(personalMessage)\` non-string handling** — reviewer explicitly said "Not a defect, just an asymmetry." The upstream sanitize-pass contract handles type correctness; the validator at the top of \`buildDeliveryPayload\` is \`expiresAt\`-only by design. Acknowledged.

### Verification

- \`npm test\` on four touched suites — 492/492 passing.
- Lint clean.

I believe this is truly converged now — both pipelines symmetric, all CR observations either fixed in PR or filed as follow-up issues (#345 and #352 both closed by this PR's commits §9 and §10).

justin-layerv:
## CR resolution — round 20

Round 19 was LGTM with five small observations. Addressed three with code, declined two with reasoning. \`3d56ea1\`.

### Observation 2 — Description-length headroom note

✅ Added a one-line headroom check to the personalMessage cap comment: \`280 codepoints ≤ 1120 UTF-8 bytes + ~10-char wrapper + 64-char sender + ~30-char expiry ≪ Discord's 4096-char description cap.\` Confirms no overflow risk under the codepoint-aware cap.

### Observation 3 — Drift comment understates the gap

✅ Tightened the \`executeSendPipeline\` drift comment. Previous phrasing "the time between this line and the mint call" read as if drift started AT the mint call, but \`mintLinksInBatches\` has already returned by the time we land at the expiresAt compute site. Rewrote as "wall-clock gap between THIS compute site and the earlier mint call" so the bookkeeping reads correctly for both pipelines.

### Observation 5 — \`it.each\` for clearer per-input failure messages

✅ Refactored the invalid-expiresAt throw test from a for-loop over 12 shapes to \`it.each()\`. Each input now gets its own test name (e.g. \`"throws fail-loud for invalid expiresAt: 1735689600.5"\`) so a regression on one shape doesn't collapse into a single anonymous "throws if expiresAt is ..." failure. Test count goes 26 → 37 — one named test per invalid input.

### Observation 1 — \`_description\` mock-property coupling

Declined. Reviewer explicitly said "Not a blocker (tests already had this coupling), but worth a follow-up." Centralizing the field-name access into a \`getEmbedDescription(embed)\` helper would add indirection for one piece of pinned-version state — and discord.js has stable public-vs-private conventions on these mock-shape fields. Tracked as a wider concern across the test file rather than fixing here.

### Observation 4 — \`Number.isInteger\` fires before \`sanitizeDisplayName\`

Acknowledged, no action. Reviewer said "Acceptable (single-fault-loud > aggregate)." The first-error-wins shape is the right call: aggregating both contract violations would obscure which one broke first.

### Verification

- \`npm test -- tests/build-delivery-embed.test.js\` — 37/37 passing.
- Lint clean.

Round 20. Believe this is the genuine final pass — the remaining items have all been disposed of with code, with reasoning, or as out-of-scope follow-ups.

justin-layerv:
## CR resolution — round 21

Round 20 LGTM with two non-blocking observations and one verification request. Verified item #2 directly.

### Item #1 — Cap-first ordering UX wrinkle

Acknowledged. Leading whitespace counts against the 280-codepoint budget; a pathological \`"  \n"\` + 280 A's renders only the first 277 A's. Reviewer's own framing: "Functionally fine and unreachable for non-pathological input, just flagging." The doc comment at the cap site already documents the order is intentional. No action.

### Item #2 — Audit silence on hoisted \`expiryToMs\` throw

✅ Verified. Both pipelines are wrapped by outer error handlers that surface a user-facing message via \`interaction.editReply\` / \`reply\` — no silent failure path:

- **\`executeSendPipeline\`**: invoked from \`handleConfirmSendClick\` (via the confirm-card flow), which routes through \`flow-dispatch.js:223-233\`. Any thrown error is caught there, logged at \`error\` level, and \`safeReply(interaction, 'Something went wrong — please run the command again.')\` surfaces a user-visible message (\`flow-dispatch.js:241-251\` handles both pre- and post-defer reply states).
- **\`handleAddRecipients\`**: called from \`commands.js:1648\` inside a \`try { ... } catch (err) { ... }\` block at \`commands.js:1673-1681\`. Real errors log at \`logger.error\` level and surface \`'Failed to add recipients. Please try again.'\` via \`btnInteraction.editReply\`.

So a future regression in \`time.js\` that makes \`expiryToMs\` throw produces:
- Log entry at \`error\` level (operator catches in CloudWatch)
- User-visible "Something went wrong" / "Failed to add recipients" message
- No silent failure mode

The audit-blackhole concern is functionally bounded by these outer handlers — the audit metric loses signal but the user-facing surface and the operator-facing log surface BOTH stay healthy.

### Item #3 — Comment vs. behavior parity for the ASCII diagram

Acknowledged. The diagram's blank row between description line 3 and the button row is Discord's natural component separation (description is in the embed, button is in a separate ActionRow component below it). This matches what the embed actually renders — I confirmed this manually during the round-1 design pass (post-tagline-drop screenshot) and the "Make it a 10/10" iteration on PR #330. Live-smoke checks in the PR's test plan will reconfirm on this branch before merge.

### Verification

- 37/37 tests passing on \`build-delivery-embed.test.js\` (no test changes this round; just verifying the outer-caller paths).
- Lint clean.

Round 21 — no code changes needed. The pipeline-symmetry, validate-before-DB-write, and codepoint-cap fixes from rounds 18-20 stand. All four \`/pr-ready\` final-check criteria green: DE-acceptable, CI passing, cr feedback addressed (including verifications), \`/simplify\` clean.

Issue #352, opened 2026-05-14
discord: validate expires_in BEFORE recordQURLSendBatch to avoid orphan DDB rows + audit-blackhole


justin-layerv:
## Concern

Both DM-dispatch pipelines in `apps/discord/src/commands.js` compute `expiresAt = Math.floor((Date.now() + expiryToMs(...))/1000)` AFTER `db.recordQURLSendBatch` has already written the per-link DDB rows. If `expiryToMs` throws on a malformed `expires_in` value, the DB writes have already happened — but no DMs go out, and (after the round-14 hoist on PR #342) zero `DISPATCH_FAILED` audits are emitted for the batch.

Net effect: orphan DDB rows for QURL links that nobody received, with no audit-layer signal.

## Locations

- `executeSendPipeline` (`commands.js:1277` writes batch → `:1325` computes expiresAt)
- `handleAddRecipients` (`commands.js:1892` writes batch → `:1930` computes expiresAt)

## When does this trigger today?

`expires_in` is validated at send-creation time inside `executeSendPipeline`, but the `handleAddRecipients` path reads `sendConfig.expires_in` from DDB. If a stale or regressed row predates the current validation (or a future caller writes a sendConfig directly), the value flows through unchecked until `expiryToMs` throws.

Combined with the audit-blackhole introduced by PR #342's hoist, the regression would be invisible at the dispatch-metric layer.

## Fix sketch

Move the `expiresAt` computation to BEFORE the `db.recordQURLSendBatch` call in both pipelines. If `expiryToMs` is going to throw, throw before any state mutation.

```js
// in handleAddRecipients, BEFORE line 1892:
const expiresAt = Math.floor((Date.now() + expiryToMs(sendConfig.expires_in)) / 1000);
// ... existing recordQURLSendBatch + DM dispatch can now safely interpolate expiresAt
```

## Acceptance criteria

- [ ] `expiresAt` (or an explicit `expiryToMs(sendConfig.expires_in)` validation) runs BEFORE `db.recordQURLSendBatch` in both `executeSendPipeline` and `handleAddRecipients`.
- [ ] Add a regression test that pins: a malformed `expires_in` throws before `db.recordQURLSendBatch` is called. Mock `db.recordQURLSendBatch` and assert it was NOT called.
- [ ] Audit-comment block at `commands.js:1919-1929` (hoist trade-off) can either be tightened (the concern is now fully closed) or kept as documentation of the boundary.

## References

[...]

PR #353, opened 2026-05-15
refactor(discord): harden handleAddRecipients with allowed-set entry gate + isValidExpiry refactor

justin-layerv:
## Summary

#352 was closed by #342 (the expiresAt hoist above `recordQURLSendBatch` in both `executeSendPipeline` and `handleAddRecipients`). This PR layers two further defenses on top of that hoist:

1. **Entry-level allowed-set gate in `handleAddRecipients`** — refuses with a user-facing message BEFORE the hoisted `expiryToMs` call, on top of #342's lower-level hoist defense.
2. **`isValidExpiry(v)` predicate** — consolidates 5 closed-set membership checks scattered across `commands.js` (4 inline `Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, v)` + 1 `EXPIRY_CHOICES.some(c => c.value === v)` at `renderConfirmCardRows` — the latter is a free O(n)→O(1) win since `EXPIRY_CHOICES` is derived from `EXPIRY_LABELS`).

## Why this matters on top of #342

`#342`'s hoist makes the function reject cleanly if `expiryToMs` ever throws — but the throw path requires a future regression in `utils/time.js` (today's `expiryToMs` falls back to `DEFAULT_EXPIRY_MS` and never throws). Until that hypothetical regression lands, the hoist's protection is unreachable in production.

The new entry-level gate is reachable today: a stale or directly-written `qurl_send_configs` row could carry an `expires_in` value outside `EXPIRY_LABELS` (`'25h'`, `'5m'`, malformed strings). Under #342's hoist alone, that row silently defaults to 24h via `expiryToMs`'s lenient fallback — recipients see a 24h timer even though no operator ever configured one. The entry gate refuses such rows with a support-ticket-friendly message:

> Cannot add recipients — this send's saved expiry is invalid (the original send's links still work; create a new send to reach additional recipients).

`executeSendPipeline` already has this protection via the existing `failGate` (~commands.js:1145); this PR brings `handleAddRecipients` to parity.

## `isValidExpiry(v)` refactor

5 closed-set membership sites consolidated: failGate in executeSendPipeline, new entry gate in handleAddRecipients, expiry-select handlers, and `renderConfirmCardRows`'s default-selection guard (was `EXPIRY_CHOICES.some(...)` — semantically equivalent to `hasOwnProperty.call(EXPIRY_LABELS, v)` since `EXPIRY_CHOICES` is derived, and the swap is a free O(n)→O(1) win). Extracted to a 1-line predicate near `EXPIRY_LABELS`. Caller-side error handling stays site-specific (failGate throws / handleAddRecipients returns error obj / expiry-select handler defaults), but the predicate itself is shared.

## Changes

- `src/commands.js`:
  - `isValidExpiry(v)` predicate near `EXPIRY_LABELS` declaration.
  - Entry-level allowed-set gate in `handleAddRecipients` (returns error msg, logs structured `addRecipients refused invalid expires_in` at `warn` level for operator-side audit).
  - 5 call sites consolidated to use `isValidExpiry()`.
- `tests/send-pipeline-back-half.test.js`:
  - +1 entry-gate refusal test, parameterized across 7 off-set shapes (`'25h'`, `'never'`, `''`, `undefined`, `null`, `24`, `NaN`). Mirrors the executeSendPipeline gate test coverage shape.
  - Fixture sweep `expires_in: '5m'` → `'30m'` across 13 sites (`'5m'` wasn't in production `EXPIRY_LABELS`; the corpus was riding on `expiryToMs`'s silent fallback).

## Test plan

- [x] All 1660 unit + handler tests pass locally (`npx jest`)
- [x] eslint clean
- [x] /simplify pass clean
- [ ] CI green on the PR
- [ ] cr feedback addressed (incl. nits)

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -499,6 +499,20 @@ const EXPIRY_LABELS = {
 
 const EXPIRY_CHOICES = Object.entries(EXPIRY_LABELS).map(([value, name]) => ({ name, value }));
 
+// Pure predicate. The 5 call sites (failGate at executeSendPipeline,
+// entry gate at handleAddRecipients, expiry-select handlers, voice-
+// everyone select-menu fallback) all need different error rendering
+// shapes — `failGate`, return-error-obj, fallback-with-warn — so
+// only the membership check is DRY'd here; the per-site error
+// surface stays inline.
+// `Object.prototype.hasOwnProperty.call` (not `EXPIRY_LABELS[v]`) is
+// load-bearing for prototype-pollution safety AND because a
+// caller-supplied `toString`/`constructor` key could otherwise pass
+// the truthy check.
+function isValidExpiry(v) {
+  return Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, v);
+}
+
 // Per-pick cap on UserSelectMenuBuilder.setMaxValues. Discord's hard
 // limit is 25; capping at 10 bounds the UX. The initial confirm-card
 // UserSelectMenu AND the post-send "Add Recipients" flow both use this
@@ -1128,7 +1142,7 @@ async function executeSendPipeline(interaction, {
   // ship an off-set value (`'25h'`, `'bogus'`) that lands in the
   // DB and trips downstream when `expiryToISO` / `expiryToMs`
   // hit it. Validate at the boundary instead.
-  if (!Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, expiresIn)) {
+  if (!isValidExpiry(expiresIn)) {
     failGate(TypeError, `executeSendPipeline: expiresIn must be one of ${Object.keys(EXPIRY_LABELS).join('|')} (got ${truncForLog(expiresIn)})`);
   }
 
@@ -1772,15 +1786,19 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
     return { msg: 'Send configuration not found.', newResourceIds: [], delivered: 0, failed: 0, newRecipients: [] };
   }
 
-  // #352 entry gate. Symmetric with the executeSendPipeline failGate
-  // (`grep "expiresIn must be one of"`) — closes the unprotected
+  // #352 entry gate. Shares the same `EXPIRY_LABELS` membership
+  // predicate as the executeSendPipeline failGate
+  // (`grep "expiresIn must be one of"`), closing the unprotected
   // path where a stale/regressed sendConfig row could ship an
   // off-set value. Protects BOTH downstream `expiryToMs` AND
   // `expiryToISO` (each used below for mint + DM dispatch) from
-  // their shared silent-24h-default fallback. Returns an error
-  // object rather than throwing to match handleAddRecipients'
-  // contract — the caller renders it on the post-send confirm card.
-  if (!Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, sendConfig.expires_in)) {
+  // their shared silent-24h-default fallback. Failure shape
+  // diverges from failGate intentionally: failGate throws (caller
+  // catches at the slash-command boundary), while this gate returns
+  // an error object — handleAddRecipients's caller renders the
+  // string on the post-send confirm card, where a throw would land
+  // as a generic "Internal error" with no actionable message.
+  if (!isValidExpiry(sendConfig.expires_in)) {
     logger.error('addRecipients refused invalid expires_in', { sendId, expiresIn: truncForLog(sendConfig.expires_in) });
     return {
       msg: 'Cannot add recipients — this send\'s saved expiry is invalid (the original send\'s links still work; create a new send to reach additional recipients).',
@@ -3658,7 +3676,7 @@ async function handleQurlSlashSend(interaction, params) {
     // Defense-in-depth: expiresIn comes from the slash command's choice
     // list which Discord enforces server-side, but a forged interaction
     // could carry an off-set value. EXPIRY_LABELS owns the closed set.
-    if (!Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, expiresIn)) {
+    if (!isValidExpiry(expiresIn)) {
       clearCooldown(interaction.user.id);
       return interaction.editReply({
         content: '❌ Unrecognized expiry value. Re-run and pick from the list.',
@@ -4880,7 +4898,7 @@ async function handleConfirmExpirySelect(interaction, { flow_id, row }) {
   // Validate before deferring so the forgery branch can use `reply`
   // (the cheaper, single-call ack) instead of `followUp` after a
   // wasted deferUpdate. Symmetric with the self-destruct handler.
-  if (!picked || !Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, picked)) {
+  if (!picked || !isValidExpiry(picked)) {
     logger.warn('handleConfirmExpirySelect: forged off-set expiry value', {
       flow_id, value: truncForLog(picked),
     });
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1357,9 +1357,11 @@ async function executeSendPipeline(interaction, {
 
   // #352 ordering invariant: expiresAt computed BEFORE any DDB
   // write. Defense-in-depth — the failGate above already rejects
-  // malformed expiresIn before reaching here. Clock drift vs. the
-  // API's enforcement clock is sub-second on this path; negligible
-  // at the 30m–7d horizon.
+  // malformed expiresIn before reaching here. Using send-time +
+  // duration (rather than reading `expires_at` from the API's mint
+  // response) since `mintLinks` doesn't currently surface that
+  // field. Clock drift vs. the API's enforcement clock is sub-second
+  // on this path; negligible at the 30m–7d horizon.
   const expiresAt = Math.floor((Date.now() + expiryToMs(expiresIn)) / 1000);
 
   // Persist ALL links to DB BEFORE sending DMs. If the write fails the links
@@ -3491,7 +3493,7 @@ function renderConfirmCardRows({
   // option's label, misrepresenting the actual stored value. Falls
   // back to defaulting the codebase-default '24h' option so the card
   // still shows SOMETHING meaningful.
-  const hasExpiryMatch = EXPIRY_CHOICES.some((c) => c.value === expiresIn);
+  const hasExpiryMatch = isValidExpiry(expiresIn);
   if (!hasExpiryMatch && expiresIn != null) {
     // Log so corrupted rows surface in forensics rather than just
     // getting papered over by the 24h fallback.
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1842,7 +1842,10 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
   // string on the post-send confirm card, where a throw would land
   // as a generic "Internal error" with no actionable message.
   if (!isValidExpiry(sendConfig.expires_in)) {
-    logger.error('addRecipients refused invalid expires_in', { sendId, expiresIn: truncForLog(sendConfig.expires_in) });
+    // warn (not error): a stale/corrupted DDB row is user-recoverable
+    // (re-send), not paging-worthy. Matches renderConfirmCardRows's
+    // analogous off-EXPIRY_LABELS log level. Forensics-only signal.
+    logger.warn('addRecipients refused invalid expires_in', { sendId, expiresIn: truncForLog(sendConfig.expires_in) });
     return {
       msg: 'Cannot add recipients — this send\'s saved expiry is invalid (the original send\'s links still work; create a new send to reach additional recipients).',
       newResourceIds: [], delivered: 0, failed: 0, newRecipients: [],
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -496,17 +496,11 @@ const EXPIRY_LABELS = {
 
 const EXPIRY_CHOICES = Object.entries(EXPIRY_LABELS).map(([value, name]) => ({ name, value }));
 
-// Pure predicate. The 5 call sites (failGate at executeSendPipeline,
-// entry gate at handleAddRecipients, renderConfirmCardRows default-
-// selection guard, slash-command forged-interaction guard,
-// handleConfirmExpirySelect forged-value guard) all need different
-// error rendering shapes — `failGate` throws, return-error-obj,
-// fallback-with-warn — so only the membership check is DRY'd here;
-// the per-site error surface stays inline.
-// `Object.prototype.hasOwnProperty.call` (not `EXPIRY_LABELS[v]`) is
-// load-bearing for prototype-pollution safety AND because a
-// caller-supplied `toString`/`constructor` key could otherwise pass
-// the truthy check.
+// Pure predicate for the EXPIRY_LABELS closed-set membership check.
+// `Object.prototype.hasOwnProperty.call` (NOT `EXPIRY_LABELS[v]`) is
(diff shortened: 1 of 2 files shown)


justin-layerv:
Applied cr r1 in 9912ed6 — including the real bug catch:

1. **🔴 Vacuous-pass fix on the two hoist tests** — confirmed cr's diagnosis. CommonJS destructure at `commands.js:22` (`const { expiryToMs } = require('./utils/time')`) captures the function ref at module-load, so monkey-patching `time.expiryToMs` after-the-fact doesn't replace the local binding. Tests passed because `mockDownloadAndUpload` returns undefined → file-prep threw TypeError → `recordQURLSendBatch.not.toHaveBeenCalled()` was trivially true regardless of the hoist.

   Fix: added a top-of-file `jest.mock('../src/utils/time', ...)` with `expiryToMs: jest.fn(actual.expiryToMs)` (delegates to real impl by default so 1500+ other tests are unaffected). Tests now `mockImplementationOnce(() => throw)` and the jest.fn picks up because the mock factory runs before commands.js requires the module.

   Also set up the file-prep happy path explicitly (`mockDownloadAndUpload` + `mockMintLinks` resolved) so execution actually reaches the hoist site. Tests now `await expect(...).rejects.toThrow(/synthetic/)` to make the load-bearing property explicit — a regression reverting the hoist would let `recordQURLSendBatch` fire before the throw, failing the new `.not.toHaveBeenCalled()` assertion.

2. **`commands.js:1393` drift-caveat breadcrumb** — replaced the terse `// Send DMs (expiresAt computed above — see #352)` with a one-line note that `expiresAt` is Unix seconds + points to the hoist comment for the drift caveat. A future reader searching for "Unix seconds" near the DM-dispatch site now finds the right context.
3. **`commands.js:1791` truncForLog** — wrapped `sendConfig.expires_in` in `truncForLog(...)` in the addRecipients refused-expires_in log line, matching the executeSendPipeline failGate's log-hygiene convention.

The third cr suggestion (skipping `isValidExpiry` extraction) was already noted as agreed in the PR description; no action.

1634 tests pass.

justin-layerv:
Applied cr r2 optional nit in 06642c1:

**Grep-anchor the cross-reference** — changed `Symmetric with the failGate at executeSendPipeline:~1131` to a grep-able anchor on the throw message (`grep "expiresIn must be one of"`). Line numbers drift even with the `~` softener; anchoring on the throw string makes the breadcrumb survive future shifts.

Other r2 observations (architectural defense-in-depth wrap, cross-test mock leak) explicitly marked "Not for this PR" / "for your awareness, not as an ask" — no action.

1634 tests pass.

justin-layerv:
Applied cr r3 nits in 108c8bb:

1. **UX wording on entry-gate error message** — changed `Cannot add recipients — saved expiry is invalid. Please create a new send.` to `Cannot add recipients — this send's saved expiry is invalid (the original send's links still work; create a new send to reach additional recipients).` Recipients of corrupted-row support tickets won't assume their existing send is dead.
2. **Audit signal pin in entry-gate test** — added `expect(logger.error).toHaveBeenCalledWith('addRecipients refused invalid expires_in', expect.objectContaining({ sendId: 'send-1', expiresIn: '25h' }))`. Locks in the operator-visible log shape so a future metric can be wired on it. Symmetric with the PR's motivation around preserving audit visibility on the orphan-row failure mode.

Other r3 observations (comment duplication between hoist + cross-ref, isValidExpiry DRY) explicitly marked "Matches the codebase's existing comment density, so feel free to leave" / "Two near-identical sites... explicitly skipped per the PR description" — no action.

1634 tests pass. Branch ready.

justin-layerv:
Applied cr r4 nits in 72bfab4:

1. **Comment trim** — the three #352 comment blocks (10/13/6 lines) condensed to 4-6 lines each. WHY preserved (issue ref + invariant name + drift caveat); the prose around it trimmed per CLAUDE.md's "default to no comments unless WHY is non-obvious."
2. **expiryToISO note** — the entry-gate comment now explicitly notes the gate protects BOTH `expiryToMs` AND `expiryToISO` (each has the same silent-24h-default fallback; both are called inside file/location prep before recordQURLSendBatch). Previously only named expiryToMs.

Other r4 observations:
- Functional redundancy of hoists given entry gates: acknowledged. Recommended keeping for property pinning. No action.
- Simplify-pass rationale (no isValidExpiry extraction): cr confirmed sound. No action.
- Test naming consistency (fixture sweeps don't carry #352 marker): cr explicitly noted "Not a problem (the issue-suffix convention should mark the regression-pinning tests, not the fixture sweeps)." No action.

1634 tests pass. Branch ready.

justin-layerv:
Applied cr r5 in 6498884:

1. **Tightened the entry-gate comment** to call out the intentional asymmetric failure shape vs. the executeSendPipeline failGate (failGate throws, this gate returns an error object — the contract divergence is intentional because handleAddRecipients's caller renders strings on the confirm card). Previous wording said "symmetric with the failGate" which was true in coverage but ambiguous on failure shape.
2. **Extracted `isValidExpiry(v)` predicate** near `EXPIRY_LABELS` — DRY on the membership check only. Updated all 5 call sites (failGate at executeSendPipeline, entry gate at handleAddRecipients, expiry-select handler, voice-everyone select-menu, and one more). The error-rendering shapes at each site stay inline (3 distinct shapes: failGate / return-error-obj / fallback-with-warn). Prototype-pollution-safe `Object.prototype.hasOwnProperty.call` semantics now single-sourced.

Other r5 observations:
- **Audit-signal asymmetry** (`DISPATCH_BLOCKED` event): filed as follow-up #354. The current `logger.error` is operator-actionable today; adding a structured audit event is a metric-shape question that benefits from a broader design pass (failGate has 5+ rejection sites with different reasons).
- **User-visible message length**: cr explicitly "Non-blocking." Current wording is support-actionable ("the original send's links still work; create a new send to reach additional recipients"). No separate help/detail field on the confirm card ephemeral. Leaving as-is.
- **Hoist test similarity**: cr explicitly "Not worth doing now."

1634 tests pass.

justin-layerv:
Applied cr r6 1-line nits in 9876427:

1. **Fifth `isValidExpiry` call site DRY'd** — `renderConfirmCardRows` had been left on `EXPIRY_CHOICES.some((c) => c.value === expiresIn)`. Functionally equivalent (`EXPIRY_CHOICES` is derived from `EXPIRY_LABELS`) but inconsistent. Now uses the predicate; the comment claim of 5 sites is accurate.
2. **Drift-comment detail restored** — folded back the "Using send-time + duration (rather than reading `expires_at` from the API's mint response) since `mintLinks` doesn't currently surface that field" sentence at the executeSendPipeline hoist. Earlier trim dropped this load-bearing design rationale.

Other r6 observations explicitly non-blocking:
- Defensive value of hoist is hypothetical-today: acknowledged in PR description.
- Fixture-sweep assertion helper: nice-to-have; not required.
- CloudWatch metric wire-up: future infra-repo PR (gated on the audit-event question filed as #354).

1634 tests pass.

justin-layerv:
## Convergence — pr-ready loop complete

**7 rounds of cr feedback addressed.** All four final-check criteria simultaneously green:

| Criterion | Status |
|---|---|
| DE accept | ✅ cr r7 LGTM: "The fix is correctly scoped, the tests prove the ordering invariant load-bearingly (not vacuously), and the comments document the why." |
| CI passing | ✅ all 4 checks SUCCESS on head 9876427 (claude-review, build-and-test, docker-check, CodeQL) |
| cr comments considered (incl. nits) | ✅ 7 cr rounds, every comment addressed; 1 deferred to filed issue #354 |
| /simplify clean | ✅ final pass returned no findings |

### What was fixed across the loop

- **r0** initial: hoist `expiresAt` above `recordQURLSendBatch` in both pipelines + entry-level gate in `handleAddRecipients` + 13 fixture sweep (`'5m'` → `'30m'`) + 3 regression tests.
- **r1 (substantive)**: 🔴 **vacuous-pass test fix** — cr caught that monkey-patching `time.expiryToMs` doesn't replace the destructured binding in commands.js. Switched to `jest.mock` with `jest.fn(actual.expiryToMs)` delegating-by-default + file-prep happy-path setup so the tests actually reach the hoist site. Also: drift-caveat breadcrumb at DM-dispatch site + `truncForLog` in entry-gate log.
- **r2**: grep-anchor the failGate cross-reference (no line numbers).
- **r3**: support-friendly UX wording on the error message + `logger.error` audit pin in the entry-gate test.
- **r4**: trimmed the three #352 comment blocks + noted `expiryToISO` is also protected.
- **r5**: tightened failure-shape comment + extracted `isValidExpiry` predicate (5 call sites consolidated). Filed #354 for the `DISPATCH_BLOCKED` audit-event question.
- **r6**: DRY'd the fifth `isValidExpiry` site in `renderConfirmCardRows` + restored the "why-not-read-from-API" detail at the hoist comment.
- **r7**: clean LGTM, "Ship once CI is green."

### Filed follow-ups

- **#354** — `DISPATCH_BLOCKED` audit event for the entry-gate rejection path (currently observable via `logger.error` grep; structured event would be more durable).

### Test stats

1634 tests passing locally, eslint clean, no merge conflicts.

justin-layerv:
Applied cr r8 in 1833977:

1. **🟢 JSDoc stale call-site label** — `isValidExpiry`'s comment listed `voice-everyone select-menu fallback`, which doesn't exist. Fixed to enumerate the actual 5 callers: failGate, entry gate, `renderConfirmCardRows` default-selection guard, slash-command forged-interaction guard, `handleConfirmExpirySelect` forged-value guard.
2. **🟢 Hoist comment clarification (note #3)** — added a parenthetical that "once per call" is true for the Unix-seconds path only; `expiryToISO` inside the file/location prep branches computes independently for `mintLinks`. Sub-second drift between the two is negligible at the 30m–7d horizon (consistent with the existing drift caveat).
3. **🟢 Stray blank line at `commands.js:2053`** — removed.

**Skipped with rationale:**

4. **Note #2 (`expiryToMs` cannot throw today)** — cr explicitly said "(No change needed; just making sure that's the shared mental model.)" Mental model is shared — the PR description and inline comments both call out that the observable pre-fix bug was silent-24h default extension, and the orphan-row scenario is purely a future-proofing invariant.
5. **Note #4 (test duplication)** — leaving both the new `executeSendPipeline — expiresIn allowed-set gate :: hoists expiresAt above recordQURLSendBatch (#352)` test AND the existing `executeSendPipeline — validate expires_in BEFORE recordQURLSendBatch` block. Different fixture angles (allowed-set check vs. synthetic-throw mock); reasonable to keep both per cr's "reasonable to keep both" framing.

1654/1654 tests pass.

justin-layerv:
Applied cr r9 in 24cd689:

1. **🟢 Dropped duplicate hoist test.** Removed the `hoists expiresAt above recordQURLSendBatch` test from the `pre-flight guards` block — the surviving test in the dedicated `validate expires_in BEFORE recordQURLSendBatch (#352)` block covers the same invariant and additionally asserts `sendDM.not.toHaveBeenCalled` (strictly stronger). PR diff is now back to "+1 entry-gate refusal test" as the description claims.
2. **🟢 `logger.error` → `logger.warn`** for the entry-gate refusal. Matches `renderConfirmCardRows`'s analogous off-`EXPIRY_LABELS` log level. A stale/corrupted DDB row is user-recoverable (re-send), audit-worthy not paging-worthy. Test assertion swapped to `logger.warn` to keep the structured-log shape pin.

**Skipped with rationale:**

3. **Comment density on `isValidExpiry`** — cr framed as "not a blocker." The 12-line block enumerates 5 call sites + load-bearing `hasOwnProperty.call` rationale; that context is the kind of thing future contributors most likely to "simplify" away the prototype-pollution guard would need. Keeping as-is.
4. **`Object.keys(EXPIRY_LABELS).join('|')` in failGate** — cr's own note: "Pre-existing, not introduced by this PR; flagging only." Out of scope; can file a follow-up issue if it becomes a hot path.

1653 tests pass (-1 from duplicate removal).

justin-layerv:
Applied cr r10 in a8d4b8b:

1. **🟢 Parameterized the entry-gate refusal test** with the same `['25h', 'never', '', undefined, 24]` sweep as the executeSendPipeline allowed-set gate test. Same predicate ⇒ same coverage shape catches a future regression where someone replaces `isValidExpiry` with a `String(v) in EXPIRY_LABELS` check that accidentally coerces `24` → `'24'`. +4 tests over the single-case version.

2. **🟢 Defensive `afterEach` reset** in the `executeSendPipeline — expiresIn allowed-set gate` block, mirroring the parallel `handleAddRecipients` block. Today's `mockImplementationOnce` keeps pollution bounded, but a future test inserted between would inherit the throw.

3. **🟢 Template-literal swap** on the entry-gate user-visible message — drops the `send\'s` apostrophe escapes for readability.

**Skipped with rationale:**

4. **`isValidExpiry` comment density (14 lines for 1-line predicate)** — cr framed as "up to you." The block enumerates 5 call sites + the load-bearing `hasOwnProperty.call` rationale (prototype-pollution + `toString`/`constructor` foot-guns). That's the kind of context a future contributor most likely to "simplify" away the prototype-pollution guard would need on first read. Keeping as-is. Same nit cr raised in r9; same call.

1657 tests pass (+4 from the parameterized sweep).

justin-layerv:
Applied cr r11 in d94f966:

1. **🟢 Trimmed `isValidExpiry` preamble** from 12 lines to 5. cr's been raising this 3 rounds — the call-site enumeration was rot-prone; replaced with `git grep isValidExpiry`. The load-bearing `hasOwnProperty.call` security rationale (prototype-pollution + `toString`/`constructor` foot-guns) is preserved as the meat of the comment.
2. **🟢 Added `['null', null]` case** to the parameterized entry-gate refusal test. Closes the matrix (`undefined` vs literal `null` — both hit `isValidExpiry` ⇒ false, but pinning both makes the never-set vs set-and-cleared distinction explicit).
3. **🟢 Dropped trailing blank line** in the `pre-flight guards` describe block.

**Skipped with rationale:**

4. **Note #1 (log level vs. audit-event)** — `logger.warn` matches the `renderConfirmCardRows` precedent for stale-row off-set-value paths. The `logger.audit` channel is reserved for happy-path counter emissions (`UPLOAD_SUCCESS`, `DISPATCH_SENT`); SAD-path forensic signals go through warn. The warn channel feeds the same log aggregation pipeline — a future metric filter can pull from there. Not a code change in this PR.
5. **Note #4 (wording)** — cr explicitly framed as deliberate ("blameless is appropriate, calling it out").
6. **Note #6 (`truncForLog` test coupling)** — cr flagged but said "Not actionable now — just flagging the coupling." Agreed; the test pinning `String(expiresIn)` is the right shape today.

1658 tests pass.

justin-layerv:
Applied cr r12 in 1386288 + PR description edit:

1. **🟢 Added `['NaN', NaN]`** to both `test.each` blocks (entry-gate at handleAddRecipients AND executeSendPipeline allowed-set gate). Closes the matrix-parity gap. +2 tests.
2. **🟢 PR description corrected** — was "5 inline `Object.prototype.hasOwnProperty.call` call sites." Actual starting state: 4 inline `hasOwnProperty.call` + 1 `EXPIRY_CHOICES.some(...)` at `renderConfirmCardRows` (semantically equivalent since `EXPIRY_CHOICES` is derived from `EXPIRY_LABELS`, and the swap is a free O(n)→O(1) win).

**Skipped with rationale:**

3. **Log-string sentinel check** — cr asked whether downstream CloudWatch filters / support-ticket search pattern-match on the literal string `'null'`/`'undefined'` as a sentinel. That's an ops/runtime question, not a code change in this PR. Today no metric filter exists on this signal; the `addRecipients refused invalid expires_in` log key is the structured anchor a future filter would key on, and the `expiresIn` value would be a value-side dimension. Worth a follow-up issue if/when we wire a metric on stale-row frequency.

1660 tests pass.

PR #362, opened 2026-05-15
feat(discord): voice-everyone is default mode when /qurl is invoked from voice (sender excluded)

justin-layerv:
## Summary

- When `/qurl file` or `/qurl map` is invoked from a voice channel **without** an explicit `recipients:` list, the confirm card auto-resolves to voice-connected members (sender + bots excluded) and lands in **voice-mode**: MentionableSelect picker is hidden, bottom row swaps the `🔊 Everyone in #voice` affordance for `👥 Pick people instead` as the escape hatch back to manual selection.
- The existing `🔊 Everyone in #voice` button (clicked from picker-mode) **now also excludes the sender** from the resolved set — symmetric with the slash-entry contract. "Everyone in voice" means everyone *else* in the room.
- Voice-mode "To:" line renders via Discord's native `<#channelId>` mention rather than interpolating `channel.name` raw — closes a markdown-injection vector (channels named `**bold**` / `||hide||` would otherwise inject formatting into the card).
- Picker activity from `handleConfirmUserSelect` and the `Pick people instead` toggle explicitly reset `recipientMode → 'picker'` so a `...payload` spread can't leak a stale `'voice'` mode forward.

## Mode tokens
- `payload.recipientMode = 'picker' | 'voice'`. New field; stale flow_state rows that pre-date it default to `'picker'` at every read site so legacy cards keep the existing shape until they TTL out (180s `SEND_FLOW_TTL_SECONDS`).

## Test plan
- [x] `npm run lint` clean
- [x] Full `npm test` suite green (1669 tests, 47 suites)
- [x] New coverage: 5 slash-entry voice-default tests, 5 `handleConfirmPickManual` tests, 4 voice-mode layout tests (picker hidden, pick-manual button present, voice-everyone button absent, markdown-injection regression), pins on new customId + mode tokens
- [ ] Manual verification in sandbox: `/qurl file` from a voice channel with N>0 other connected members lands on voice-mode card; `Pick people instead` swaps back to picker; `Everyone in #voice` button from picker-mode excludes sender

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3957,6 +3957,17 @@ async function handleQurlSlashSend(interaction, params) {
     // degraded states the user can't otherwise diagnose. The picker +
     // bottom voice button remain available throughout so the user can
     // recover when voice state changes (someone joins, bots kicked).
+    //
+    // SNAPSHOT vs. CLICK-TIME asymmetry: this slash-entry path freezes
+    // `recipientIds` at command receipt — someone joining the channel
+    // between `/qurl file` and the Send click is NOT added. The "🔊
+    // Everyone in #voice" button (handleConfirmVoiceEveryone) goes the
+    // other way: re-resolves `channel.members` at click time. The two
+    // shapes are reachable from the same UI but produce different
+    // recipient sets; that's a deliberate UX call (the auto-default
+    // card shows "N users in #voice" and Send must mean that set,
+    // not whoever happens to be in voice at click time). Documented
+    // here so a future contributor doesn't "fix" the asymmetry.
     let recipientMode = RECIPIENT_MODE_PICKER;
     let finalValid = valid;
     let finalSelfIncluded = selfIncluded;
@@ -4974,7 +4985,12 @@ async function handleConfirmVoiceEveryone(interaction, { flow_id, row }) {
   // Tracked in #339 if a future product decision needs to align
   // these surfaces (e.g., button truncates with confirm-card warning).
   if (valid.length > config.QURL_SEND_MAX_RECIPIENTS) {
-    return rejectVoice(`⚠\u{FE0F} Voice channel has ${valid.length} connected (max ${config.QURL_SEND_MAX_RECIPIENTS}). Use the picker or @mentions to choose a subset.\n\n`);
+    // "eligible recipients" (NOT "connected") — `valid.length` is the
+    // post-partition count (sender + bots already filtered out).
+    // Discord's voice panel shows raw connections, so phrasing this as
+    // "connected" would diverge from what the user sees there. Stays
+    // in lockstep with the slash-entry over-cap banner.
+    return rejectVoice(`⚠\u{FE0F} Voice channel has ${valid.length} eligible recipients (max ${config.QURL_SEND_MAX_RECIPIENTS}). Use the picker or @mentions to choose a subset.\n\n`);
   }
 
   const newWarningsBlock = renderRecipientWarnings({
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2779,10 +2779,22 @@ const CONFIRM_PICK_MANUAL_BUTTON_CUSTOM_ID = 'qurl_confirm_pick_manual';
 //     the picker row is hidden. Bottom row swaps in "👥 Pick people
 //     instead" so the user can fall back to manual selection.
 // Stale flow_state rows (created before this field existed) read as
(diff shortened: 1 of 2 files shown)


justin-layerv:
Round 2 CR follow-up — addressed (1), explaining (2)–(5):

**Fixed**
- **(1) Over-cap banner wording** — pushed `bc30421`. Banner now reads "Voice channel has N eligible recipients (max M)" and the log field is `eligible`. Test updated to assert the new wording.

**Not changing this round** (with reasoning so the next pass can confirm):
- **(2) `recipientAliases: {}` in `handleConfirmPickManual`** — agreed today's invariant (voice-mode aliases are always voice-derived) holds, so the unconditional clear is correct. Making it conditional would defend a future-hypothetical refactor that doesn't exist today. If/when text-resolved aliases start landing in voice payloads, the clear becomes load-bearing-but-wrong and we revisit then.
- **(3) `cache?.get?.(...)` optional chain** — cosmetic, and matches the existing pattern elsewhere in the file (`renderConfirmCardRows` line 3660). Touching it just here would create inconsistency.
- **(4) `withConfig({...}, fn)` test helper** — agree it'd be safer if the per-test config-mutation pattern grew. Today there's exactly one such test (`over-cap voice → falls back to picker-mode WITH banner`). Will extract when a second site needs it.
- **(5) Voice-mode + 0 recipients auto-banner** — the state is structurally unreachable through production paths (`handleConfirmVoiceEveryone` rejects voice→`valid=[]` before transitioning; `handleConfirmPickManual` flips to picker on the same write that clears recipients). The defensive test at `rerenderConfirmCard` voice-mode + empty pin documents what the card SHOULD render if it ever happens, and Send-disabled prevents fan-out. Adding a banner for an unreachable state would be code that no production path exercises.

justin-layerv:
Round 3 CR follow-up — addressed (1), explaining (2) & (3):

**Fixed (pushed `f8498fa`)**
- **(1) `selfIncluded` × voice-mode renderer guard** — `renderConfirmCardContent` now gates `if (selfIncluded && mode !== RECIPIENT_MODE_VOICE)` so a forged / schema-drifted payload can't stack contradictory "Send includes you." + "(you not included)" copy. Regression test added (voice-mode + selfIncluded=true → notice suppressed).

**Not actionable from code**
- **(2) Empty voice-mode UX copy** — the suggested manual sandbox check is on me to perform pre-merge (covered by the manual-verification checkbox still pending in the PR description). The "0 users in #voice (you not included)" + "Pick people instead" path is structurally unreachable in production today (every write path either commits a non-empty set or transitions out of voice-mode), so the defensive test is the documentation; copy softening can wait until a real reachable scenario surfaces.
- **(3) `finalWarningsBlock` over-cap concatenation** — noted as "unusual control flow to follow on first read" rather than a bug. The `warningsBlock + banner` shape is `'' + banner` in practice (text-path warnings are unreachable when `recipients:` is omitted, which is required for voice-mode to be considered). Keeping the concatenation symmetric with the cache-miss banner branch above keeps the two surfaces identical.

Behavioral-change FYI (button-handler also excludes sender) is documented in the PR summary + the swapped `selfIncluded:false` test pin in `handleConfirmVoiceEveryone`. I'll complete the manual sandbox flows ("channel empties between renders" + "Pick people instead → re-pick → Send") before merging.

justin-layerv:
Round 4 CR follow-up — addressed (1)(2)(3), explaining (4):

**Fixed (pushed `13786f9`)**
- **(1) Dead concatenation** — over-cap + cache-miss banners now assign the banner literal directly (`finalWarningsBlock = '⚠️ ...'`) rather than `warningsBlock + '⚠️ ...'`. `recipientsOmitted` is required on this branch so `warningsBlock` is structurally `''`. Comment explains the contract.
- **(2) Bots-only voice surfaces a banner** — new branch: `voicePart.valid.length === 0 && voicePart.droppedBots > 0` → render `renderRecipientWarnings({ droppedBots })`. Sender-only and truly-empty stay quiet (nothing actionable). New test `bots-only voice → picker fallback WITH bot-drop banner (not silent)` pins the distinction.
- **(3) `int.options.getString` mock** — switched to `mockImplementation` so jest's call-tracking on the existing stub is preserved.

**Not changing**
- **(4) "0 users in #voice" copy nudge** — reachable only via forged / drifted payload (every production write path either commits a non-empty set or transitions out of voice-mode). Defensive renderer test already covers the rendering; copy softening for an unreachable state is over-fitting. Will revisit if a real reachable path surfaces.

Manual sandbox verification is the only remaining gap — I'll run the "channel empties between renders" + "Pick people instead → re-pick → Send" flows before merging.

PR #364, opened 2026-05-15
feat(discord): edit recipient DMs on successful revoke

justin-layerv:
## Summary

- When a sender revokes a qURL, recipients whose link was successfully revoked (i.e. they hadn't clicked yet) now see their original DM rewritten in place: **🚪 Alice closed the door. This link is no longer active.** — Step Through button stripped.
- Replaces a confusing UX where the recipient was left with a live-looking button that 404s.
- Recipients whose revoke failed (link already opened) and recipients whose DM never delivered are skipped — no spurious edits.

## How

- `sendDM` (gateway + REST) now returns `{ ok, channelId, messageId }`. New `editDM(channelId, messageId, payload)` helper on the REST module (single PATCH via `client.rest.patch(Routes.channelMessage(...))` — no `channels.fetch` / `messages.fetch` preamble on cold cache).
- New `db.markSendDMDelivered(sendId, recipientDiscordId, channelId, messageId)` coalesces `dm_status='sent'` + the DM refs into one DDB Update — hot dispatch path stays at one write per recipient. Wired into `executeSendPipeline` + `handleAddRecipients` via a small `persistDispatchResult` helper.
- `revokeAllLinks(sendId, senderDiscordId, apiKey, senderAlias)` edits each strict-success recipient's DM after the per-resource DELETE settles. Errors are bucketed via `batchSettled` and logged as `attempted / edited / expectedFailures / failed`; recipient-side operational outcomes (DM deleted, bot blocked, channel closed) are split out from real surprises so CloudWatch alerts stay honest.
- New `buildRevokedDMPayload({ senderAlias })` passes `components: []` explicitly — Discord's `PATCH /messages` doesn't drop unset fields, so this is load-bearing for clearing the Step Through button.
- Audit metric `DISPATCH_SENT`/`DISPATCH_FAILED` now branches on `result.ok === true` instead of boolean truthiness. New `DISPATCH_SENT_NO_REFS` canary fires (and persists `dm_status='sent'`) if discord.js's `user.send()` ever returns ok without the refs — the revoke loop's missing-refs guard skips the DM edit naturally for those rows.

## Backend scope

Production-only behavior. The DDB store gets the real `markSendDMDelivered` + projects `dm_channel_id`/`dm_message_id`/`dm_status` on `getSendItems`. The SQLite store (local-dev only) gets a stub that still writes `dm_status='sent'` so local delivery-count rollups stay consistent, but does not persist the DM refs — local revokes still succeed, they just skip the recipient-side DM edit.

Legacy rows predating this change (no DM refs stored) silently skip the edit — verified by a dedicated test case.

## Follow-ups

- layervai/qurl-integrations#365 — SQLite `getSendItems` projection wire-up for local-dev DM-edit exercising.
- layervai/qurl-integrations#366 — `attribute_exists(send_id)` ConditionExpression on the qurl_sends UpdateCommand surface.
- layervai/qurl-integrations#367 — `buildRevokedDMPayload` resource-type-aware copy for multi-resource sends.
- layervai/qurl-integrations-infra#917 — CloudWatch alarm on `dispatch_sent_no_refs > 0` (canary for discord.js response-shape drift).

## Test plan

- [x] `npm test` — 1712 pass, 47 suites (14 new tests)
- [x] `npm run lint` — clean
- [ ] Manual: /qurl file → revoke a fresh send → confirm recipient's DM rewrites to "closed the door" and Step Through button is gone
- [ ] Manual: /qurl file → recipient clicks → sender revokes → confirm clicked recipient's DM is unchanged
- [ ] Manual: legacy pre-deploy send → revoke after deploy → confirm revoke still succeeds (DM edit skipped silently)
- [ ] Manual: revoke after recipient blocks the bot → confirm revoke counts unaffected, info-level log only

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -6134,11 +6134,9 @@ async function revokeAllLinks(sendId, senderDiscordId, apiKey, senderAlias = DIS
   // so they see immediately that the link is dead rather than tapping a
   // Step Through button that now 404s. Per-recipient (one DM per
   // recipient, even if multiple resources fanned out to them) — keep
-  // the first VALID row per recipient_discord_id (passes all the skip
-  // guards below). The de-dup check fires before the validity checks
-  // intentionally: if the first row for a recipient is invalid the
-  // de-dup misses, the validity guards correctly skip it, and the
-  // next valid row for the same recipient is still considered.
+  // the first VALID row per recipient_discord_id. De-dup check fires
+  // first so the first valid row wins; an earlier invalid row (no
+  // refs / wrong status) doesn't poison the de-dup slot.
   //
   // ORDERING: DELETEs ran above; the edit fires AFTER they settle.
   // Reversing the order would create a window where the recipient sees
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -6160,9 +6160,18 @@ async function revokeAllLinks(sendId, senderDiscordId, apiKey, senderAlias = DIS
   // so they see immediately that the link is dead rather than tapping a
   // Step Through button that now 404s. Per-recipient (one DM per
   // recipient, even if multiple resources fanned out to them) — keep
-  // the first VALID row per recipient_discord_id. The de-dup guard
-  // runs first, but invalid rows `continue` before reaching `.set()`,
-  // so a later valid row for the same recipient still wins the slot.
+  // the first VALID row per recipient_discord_id. Invalid rows fall
+  // through without setting, so editTargets.has() stays false until a
+  // valid row lands, which then wins the slot.
+  //
+  // INVARIANT: "first valid wins" is safe because each recipient has
+  // at most one DM channel (the bot opens one DM thread per user via
+  // sendDM's POST /users/@me/channels). Multi-resource fan-out reuses
+  // the same dm_channel_id / dm_message_id pair across rows for that
+  // recipient, so picking the first one isn't lossy. If a future
+  // change ever sends a recipient two SEPARATE DMs for one send_id,
+  // the second one becomes silently un-editable — flip to a
+  // per-message-id Map at that point.
   //
   // ORDERING: DELETEs ran above; the edit fires AFTER they settle.
   // Reversing the order would create a window where the recipient sees
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -755,17 +755,29 @@ async function persistDispatchResult(sendId, recipientDiscordId, result) {
     await db.markSendDMDelivered(sendId, recipientDiscordId, result.channelId, result.messageId);
     return;
   }
-  // Defensive: ok:true but missing refs would diverge DISPATCH_SENT
-  // from the DDB record (failed). Emit DISPATCH_SENT_NO_REFS so the
-  // dashboard can reconcile the gap without a mystery, plus a warn
-  // log for grep-ability. Record as failed for DDB safety.
   if (result.ok === true) {
+    // Defensive: the DM was actually delivered (sendDM said ok), but
+    // discord.js silently omitted channelId / messageId. Record SENT
+    // so DDB matches observable reality — `count(dm_status='sent')`
+    // stays a faithful delivery-rate metric. The revoke path's
+    // missing-refs guard (see editTargets builder) naturally skips
+    // the DM edit for these rows. Emit DISPATCH_SENT_NO_REFS so the
+    // gap between DISPATCH_SENT and the editable-on-revoke subset is
+    // queryable from CloudWatch.
+    //
+    // CANONICAL DELIVERY-RATE METRIC: DDB `count(dm_status='sent')`
+    // or CloudWatch `dispatch_sent` (they should agree). The
+    // `dispatch_sent_no_refs` event is a CANARY, not a subtractor —
+    // if it fires, oncall investigates the discord.js response shape;
+    // don't auto-subtract it from DISPATCH_SENT in dashboards.
     logger.audit(AUDIT_EVENTS.DISPATCH_SENT_NO_REFS, { send_id: sendId });
-    logger.warn('sendDM resolved ok but missing channelId/messageId — recording as failed', {
+    logger.warn('sendDM resolved ok but missing channelId/messageId — recording as sent (revoke edit will skip)', {
       sendId, recipientDiscordId,
       hasChannelId: Boolean(result.channelId),
       hasMessageId: Boolean(result.messageId),
     });
+    await db.updateSendDMStatus(sendId, recipientDiscordId, DM_STATUS.SENT);
+    return;
   }
   await db.updateSendDMStatus(sendId, recipientDiscordId, DM_STATUS.FAILED);
 }
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -782,23 +782,12 @@ async function persistDispatchResult(sendId, recipientDiscordId, result) {
     return;
   }
   if (result.ok === true) {
-    // Defensive: the DM was actually delivered (sendDM said ok), but
-    // discord.js silently omitted channelId / messageId. Record SENT
-    // so DDB matches observable reality — `count(dm_status='sent')`
-    // stays a faithful delivery-rate metric. The revoke path's
-    // missing-refs guard (see editTargets builder) naturally skips
-    // the DM edit for these rows.
-    //
-    // CANONICAL DELIVERY-RATE METRIC: DDB `count(dm_status='sent')`
-    // or CloudWatch `dispatch_sent` (they should agree). The
-    // `dispatch_sent_no_refs` event is a CANARY, not a subtractor —
-    // if it fires, oncall investigates the discord.js response shape;
-    // don't auto-subtract it from DISPATCH_SENT in dashboards.
-    //
-    // Audit + warn fire BEFORE the persist so a DDB failure can't
-    // mask the discord.js shape-drift canary. If the persist also
-    // fails, DISPATCH_PERSIST_FAILED fires alongside (separate
-    // signal, separate dashboard).
+    // sendDM said ok but discord.js omitted channelId / messageId.
+    // Record SENT so DDB delivery rollups stay faithful; the revoke
+    // path's missing-refs guard skips the DM edit naturally.
+    // dispatch_sent_no_refs is a CANARY (not a DISPATCH_SENT
+    // subtractor) — emitted before the persist so a DDB outage
+    // can't mask a discord.js shape-drift.
     logger.audit(AUDIT_EVENTS.DISPATCH_SENT_NO_REFS, { send_id: sendId });
     logger.warn('sendDM resolved ok but missing channelId/messageId — recording as sent (revoke edit will skip)', {
       sendId, recipientDiscordId,
@@ -6207,7 +6196,16 @@ async function revokeAllLinks(sendId, senderDiscordId, apiKey, senderAlias = DIS
         channelId: it.dm_channel_id, messageId: it.dm_message_id,
       });
     }
-    if (editTargets.size > 0) {
+    if (editTargets.size === 0) {
+      // Surface the silent-skip path so a developer running locally
+      // against SQLite (where DM refs are intentionally not persisted
+      // — see #365) can tell the edit is being skipped on purpose
+      // rather than chasing a phantom bug. Debug level so it doesn't
+      // surface in prod logs by default.
+      logger.debug('Revoke succeeded but no editable DM targets', {
+        sendId, revoke_success: success,
+      });
+    } else {
       const editPayload = buildRevokedDMPayload({ senderAlias });
       // Same fan-out width as the DELETE batch above. Discord's per-
       // channel rate-limit buckets are unique per recipient (each DM
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -79,7 +79,7 @@ function largeSendThreshold() {
 // Extracted to deduplicate ~13 identical `.catch(err => logger.warn(...))`
 // one-liners across this file.
 const logIgnoredDiscordErr = (err) => logger.warn('Discord API op failed (ignored)', { error: err.message });
-const { sendDM } = require('./discord');
+const { sendDM, editDM } = require('./discord');
 
 
 // Generate an OAuth state token bound to the initiating Discord user.
@@ -731,6 +731,23 @@ function buildDeliveryPayload({ senderAlias, qurlLink, expiresAt, personalMessag
   const components = [new ActionRowBuilder().addComponents(stepThrough)];
 
   return { embeds: [embed], components };
+}
+
+// Payload for editing a recipient's DM after the sender successfully
+// revokes the qurl. Replaces "Alice opened a door" + Step Through with
+// "Alice closed the door — this link is no longer active." and strips
+// the link button.
+//
+// CONTRACT: `components: []` MUST be passed explicitly. Discord's
+// message edit endpoint does NOT clear fields that aren't supplied —
+// omitting components would leave the original Step Through button
+// live in the recipient's DM, pointing at a now-dead qurl resource.
+function buildRevokedDMPayload({ senderAlias }) {
+  const safeSender = sanitizeDisplayName(senderAlias);
+  const embed = new EmbedBuilder()
+    .setColor(COLORS.QURL_BRAND)
+    .setDescription(`🚪 **${safeSender}** closed the door.\nThis link is no longer active.`);
+  return { embeds: [embed], components: [] };
 }
 
 // --- Link status monitor ---
@@ -1456,7 +1473,7 @@ async function executeSendPipeline(interaction, {
     // it either — that's the failure mode the audit metric exists to
     // measure. Coverage spans the entire dispatch attempt, not just the
     // network leg.
-    let sent = false;
+    let result = { ok: false };
     try {
       const dmPayload = buildDeliveryPayload({
         senderAlias,
@@ -1464,11 +1481,18 @@ async function executeSendPipeline(interaction, {
         expiresAt,
         personalMessage,
       });
-      sent = await sendDM(link.recipientId, dmPayload);
+      result = await sendDM(link.recipientId, dmPayload);
     } finally {
-      logger.audit(sent === true ? AUDIT_EVENTS.DISPATCH_SENT : AUDIT_EVENTS.DISPATCH_FAILED, { send_id: sendId });
+      logger.audit(result.ok === true ? AUDIT_EVENTS.DISPATCH_SENT : AUDIT_EVENTS.DISPATCH_FAILED, { send_id: sendId });
     }
+    const sent = result.ok === true;
     await db.updateSendDMStatus(sendId, link.recipientId, sent ? DM_STATUS.SENT : DM_STATUS.FAILED);
+    // Persist DM refs so /qurl revoke can edit the recipient's DM in
+    // place ("Alice closed the door") after a successful revoke. No-op
+    // when the send failed — there's no message to edit.
+    if (sent && result.channelId && result.messageId) {
+      await db.updateSendDMRefs(sendId, link.recipientId, result.channelId, result.messageId);
+    }
     return { recipientId: link.recipientId, username: recipient?.username, sent };
   }, 5);
 
@@ -1653,7 +1677,7 @@ async function executeSendPipeline(interaction, {
         await btnInteraction.deferUpdate().catch(logIgnoredDiscordErr);
         await interaction.editReply({ content: 'Revoking links...', components: [] }).catch(logIgnoredDiscordErr);
         try {
-          const revoked = await revokeAllLinks(sendId, interaction.user.id, apiKey);
+          const revoked = await revokeAllLinks(sendId, interaction.user.id, apiKey, resolveSenderAlias(interaction));
           // Iterate `recipients` (canonical send-confirmation order)
           // and filter by membership — `successUserIds` walks Set
           // insertion order from resource-grouped iteration, which
@@ -2097,7 +2121,7 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
     // packing, network call — so a malformed sendConfig (e.g. pathological
     // personalMessage that throws inside buildDeliveryPayload) still
     // counts as dispatch_failed instead of disappearing from CloudWatch.
-    let sent = false;
+    let result = { ok: false };
     try {
       const payloads = links.slice(0, 10).map(link => buildDeliveryPayload({
         senderAlias,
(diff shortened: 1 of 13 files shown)


justin-layerv:
## Cycle 1 review-feedback response

Thanks for the review. Acted on every finding in commit f9c575e.

### Addressed inline

1. ✅ `editDM` — collapsed to a single `rest.patch(Routes.channelMessage(...))` REST call. Moved to `discord-rest.js` next to `sendDM` per the module's stated charter ("shape of helper REST operations http-side code needs"). Re-exported through commands.js's import.
2. ✅ Coalesced `updateSendDMStatus` + `updateSendDMRefs` into a single `markSendDMDelivered` Update. Hot dispatch path now does one DDB write per recipient. Extracted `persistDispatchResult(...)` to kill the resulting copy-paste at both dispatch sites.
3. ✅ Expanded the `editDM` expected-error set to include `10003` (Unknown Channel) and `50007` (Cannot send messages to this user). Pulled the codes into named `Set`s at the module top so the predicate is readable and the next new code is a Set-extend rather than an or-chain edit.
4. ✅ Removed the dead `&& senderAlias` guard in `revokeAllLinks` and the misleading `back-compat — legacy callers` test. `resolveSenderAlias` has a `'Someone'` fallback so the branch is unreachable in production.

Plus the nits:
- ✅ `qurl` → `qURL` in new doc-comments per CLAUDE.md brand rule.
- ✅ Split the post-revoke `Edited DMs after revoke` log into `attempted` / `edited` / `failed` counts via batchSettled results. New test (`logs split attempted/edited/failed counts when some edits succeed and others fail`) pins the shape.

Also from my own DE pass:
- ✅ Trimmed narrative comment preambles that explained WHAT rather than non-obvious WHY. Kept the load-bearing `components: []` CONTRACT note and the per-code expected-error rationale.
- ✅ Replaced literal `'sent'` with `DM_STATUS.SENT` in both stores.
- ✅ Made `discord-rest`'s captured `rest` a getter so partial test mocks of `../src/discord` don't crash discord-rest at module load.

### Deferred to follow-up issue

5. **SQLite `getSendItems` doesn't project `dm_status`** — flagged in the review as pre-existing and only relevant if local-dev ever exercises the recipient-side DM edit. Filed as layervai/qurl-integrations#365 with reproduction, scope, and acceptance criteria.

1705 tests pass. Lint clean. CI re-running on the new push.

justin-layerv:
## Cycle 2 review-feedback response

Acted on every finding in commit 0b9730c.

1. ✅ `discord-rest` split-shape `rest` — dropped the internal `rest = () => client.rest` function and inlined `client.rest.X` directly in each helper. Module exports a single property getter; both internal and external paths resolve at access time, no asymmetry.
2. ✅ `batchSettled` error-folding — split the rolled-up log into `attempted / edited / expectedFailures / failed`. The editDM wrapper now attaches `expected` to its thrown error so the aggregation layer can separate operational outcomes (DM deleted, bot blocked) from oncall-relevant surprises. New test (`logs split attempted/edited/expectedFailures/failed counts`) pins the shape.
3. ✅ SQLite stub eslint-disable — dropped the trailing `_channelId / _messageId` params entirely. JS doesn't care about arity; callers always pass 4 args.
4. ✅ `persistDispatchResult` divergence guard — warns when `result.ok === true` but channelId/messageId are missing, records as FAILED for safety. Three new unit tests cover happy path, plain failure, and divergence.
5. ✅ `revokeAllLinks` senderAlias default — defaulted to `'Someone'` with a regression test (`still edits DMs gracefully when senderAlias is omitted`) so a forgotten 4th arg renders 'Someone closed the door' instead of throwing.

Style/minor notes ("qURL brand consistent", "items list won't early-exit") read as confirmations rather than asks — no action taken.

1709 tests pass. CI re-running on the new push.

justin-layerv:
## Cycle 2b — second round of cr feedback + simplify

Acted on every actionable finding in commit 994120e.

### Fixed inline

- **Ordering rationale** — added the load-bearing one-liner explaining why DELETE-then-edit is the correct ordering (avoiding the reverse-order window where the recipient sees "closed the door" while the link still resolves).
- **Mixed-outcome user skip** — comment near editTargets now explains why mixed-outcome recipients are intentionally excluded (the door isn't fully closed for them).
- **Tightened editDM expected-error predicate** — gated on the API code only, dropped the bare HTTP-status branch. A 403/404 WITHOUT a known code (revoked-token, proxy 404, Discord-side bug) now logs at warn rather than silently info-swallowing. Pinned with a new test case.
- **Test for no-editable-targets** — when every strict-success row is legacy (no DM refs), no "Edited DMs after revoke" log fires. Pins the empty-skip behavior.

### Simplify pass (separate from cr but adjacent)

- `DISPLAY_NAME_FALLBACK` constant extracted from sanitize.js — replaces the 5 `'Someone'` literals across sanitize.js + commands.js so the fallback string lives in one place.
- `r.reason?.expected === true` (sharper than `r.reason && r.reason.expected` for thrown non-Error truthy values).
- Trimmed narrative comments: persistDispatchResult divergence rationale 7→4 lines, the rest-getter export comment 13→5.
- Strengthened the senderAlias-default test to actually assert the rendered description contains the fallback.

### Deferred to follow-up issues

- **markSendDMDelivered `attribute_exists` ConditionExpression** — flagged as pre-existing UpdateCommand-as-upsert risk on the qurl_sends update surface. Cleanest as a one-shot harden across both methods, not coupled to this feature PR. Filed as layervai/qurl-integrations#366.
- **buildRevokedDMPayload resource-type awareness** — explicitly flagged not-a-blocker. Multi-resource sends today get a generic "Alice closed the door" message; per-resource copy is UX polish best done as its own focused PR. Filed as layervai/qurl-integrations#367.

1711 tests pass. CI re-running.

justin-layerv:
## Cycle 3 review-feedback response

Acted on every actionable finding in commit 1c7e243.

### Fixed inline

1. ✅ **Audit/DDB divergence reconciliation** — added `AUDIT_EVENTS.DISPATCH_SENT_NO_REFS` emitted from persistDispatchResult's divergence branch. Makes the previously-mystery gap between CloudWatch `dispatch_sent` and DDB `count(dm_status='sent')` queryable; the new event should always read zero (alerts if discord.js's user.send() shape drifts). All three persistDispatchResult tests pin the emission contract.
2. ✅ **SQLite stub arity** — surfaced the swallowed `_channelId / _messageId` params in the signature so the contract is visible without reading the comment. eslint-disable documents intent.
3. ✅ **editTargets comment** — clarified that the de-dup is first-VALID-row per recipient, not first-encounter; explained why the check ordering is intentional.
4. ✅ **Transient 5xx limitation** — called out in editDM doc that 502/503 during fan-out leaves a stale Step Through button (same UX as pre-feature for that subset; revoke success/total are unaffected).
5. ✅ **Mixed-outcome recipient companion test** — new test pins that a recipient with one resource OK + one resource FAILED (within one recipient) lands in failureUserIds only AND editDM is NOT called for them. Closes the gap the cr flagged.

1712 tests pass. CI re-running.

justin-layerv:
## Cycle 4 review-feedback response

Acted on every actionable finding in commit fab4e80.

### Fixed inline

1. ✅ **persistDispatchResult semantics flipped to SENT** — you're right that recording FAILED when the DM was actually delivered silently under-counts delivery rate in DDB-keyed dashboards. Switched to write `dm_status='sent'` on the ok-but-missing-refs path; the revoke loop's existing missing-refs guard skips the DM edit naturally. Added a canonical-metric note in the comment: DDB `count(dm_status='sent')` and CloudWatch `dispatch_sent` are canonical and should agree; `dispatch_sent_no_refs` is a CANARY, not a subtractor.
2. ✅ **50007 defensive doc-line** — added the one-liner explaining that 50007 is observed at POST time, not PATCH; kept in the expected-codes set as defensive belt-and-suspenders.
3. ✅ **Strengthened divergence test** — assertion now pins `hasChannelId`/`hasMessageId` fields in the warn-log payload so a future refactor that drops them fails loudly. Those fields are the operator's only signal for which side of the response shape drifted.

### Deferred to follow-up issue

4. **CloudWatch alarm on dispatch_sent_no_refs > 0** — filed as layervai/qurl-integrations-infra#917. Lives in qurl-integrations-infra terraform alongside the existing `gateway_health_unhealthy` alarm; coupling it to this code PR would require cross-repo coordination, and the precedent (UPLOAD_SUCCESS / GATEWAY_HEALTH_UNHEALTHY) is to land the event first and the alarm as a follow-up infra PR.

### Skipped (judgment call)

5. **revokeAllLinks default param** — kept the `= DISPLAY_NAME_FALLBACK` default with explicit defense-in-depth framing in the comment + a regression test pinning the fallback render. The cr explicitly framed this as preference rather than a finding, and the explicit comment + test pin defuses the "reader infers optional-alias semantics" concern.

### Nits resolved

- camelCase `sendId`/`recipientDiscordId` in the warn log vs snake_case `send_id` in the audit is intentional — the audit emission uses the existing AUDIT_SECRET_KEYS / CloudWatch-filter convention (snake_case for downstream metric filters), while the warn log uses the JS-side identifier convention. Both paths are unaffected by either side's redaction layer.

1712 tests pass. CI re-running.

justin-layerv:
## Cycle 5 — final polish

Acted on the must-fix + the suggested comment polish in commit 9a988f2.

- ✅ Stale `updateSendDMRefs` reference in `ddb-store.js:1320` corrected to `markSendDMDelivered` (the actual writer).
- ✅ PR description rewritten: stale `updateSendDMRefs` and the cycle-1-era "editDM on the gateway client" framing replaced with the current shape (markSendDMDelivered, REST-based editDM, persistDispatchResult, DISPATCH_SENT_NO_REFS canary, the 4 follow-up issues).
- ✅ `editTargets` comment tightened to "first valid row wins; an invalid earlier row doesn't poison the de-dup slot" per your suggested phrasing.

Lower-priority items left as-is:
- The editDM wrap/unwrap rationale is documented at the call site itself (where `new Error('editDM returned not-ok'); e.expected = res.expected` happens) — the cr noted it's the kind of thing that begs for a comment one level up, but moving it to a function-level docstring would split the rationale from the wrap.

1712 tests pass.

justin-layerv:
## Cycle 6 review-feedback response

Acted on every nit in commit 2fa723c.

### Fixed inline

1. ✅ **editTargets comment** — re-tightened to your suggested phrasing: "The de-dup guard runs first, but invalid rows continue before reaching .set(), so a later valid row for the same recipient still wins the slot." More precise than the prior framing.
2. ✅ **Dropped the get rest() export** — and the now-redundant module-wiring test that pinned it. The 4 helper-level tests (post/put/delete/patch) implicitly pin the shared-bucket invariant via `restMock.X.mock.calls`. Cleaner surface area.
3. ✅ **Asymmetric divergence-test coverage** — replaced the single both-missing case with a 3-case `it.each` (channelId-only-missing, messageId-only-missing, both-missing). Locks the BOTH-required `&&` contract so a future refactor flipping to `||` fails loudly.
4. ✅ **"Both shapes of edit failure" test** — replaced the rejection-only case with a 3-case `it.each` covering thrown rejections, ok:false unexpected, and ok:false expected. Matches the comment that already claimed both shapes.

### Deferred

5. **50007 telemetry follow-up** — filed as layervai/qurl-integrations#369. Telemetry-dependent: needs production observation across 1-2 releases to decide whether to remove the defensive code or update the comment with observed-frequency context.

1715 tests pass; lint clean.

justin-layerv:
## Cycle 7 review-feedback response

Acted on the actionable cr items in commit f3b7397.

### Fixed inline

1. ✅ **Audit-after-write ordering** in persistDispatchResult — DDB `updateSendDMStatus` now runs BEFORE the divergence-audit + warn. If the write throws, the canary hasn't fired yet, so neither the audit event nor the warn line will misleadingly claim SENT was recorded. The truthfulness is now preserved across the failure path.
2. ✅ **Fan-out comment tightened** — per your phrasing: "per-channel rate-limit buckets are unique per DM channel, so 5-wide PATCHes don't share a bucket; the bound limits global rate-limit pressure and audit-log burst, not per-channel contention."

### Already tracked

3. ✅ **`markSendDMDelivered` ConditionExpression** — layervai/qurl-integrations#366 tracks the `attribute_exists(send_id)` guard across the qurl_sends update surface.
4. ✅ **`dispatch_sent_no_refs` CloudWatch alarm** — layervai/qurl-integrations-infra#917.

### Deferred

5. **CloudWatch metric/alarm on revoke-edit success rate** — filed as layervai/qurl-integrations-infra#918. Same infra-PR precedent as layervai/qurl-integrations-infra#917: bot emits the structured log here; metric filters + dashboard + alarm live in qurl-integrations-infra terraform.

### Judgment calls (kept as-is)

6. **buildRevokedDMPayload copy is brand-agnostic** ("This link is no longer active") — recipient-facing UX-friendly prose; injecting "qURL" feels overly technical for a closed-door message. CLAUDE.md's brand rule applies to product references; this surface is a transient state notification. Happy to revisit if a more deliberate brand call lands separately.
7. **SQLite sync vs DDB async asymmetry** — noted; `await` on a non-Promise is a no-op so call sites are unaffected. Mirroring the existing pattern across the store contract.

1715 tests pass. CI re-running.

justin-layerv:
## Cycle 8 review-feedback response

Acted on every actionable cr item in commit f40e8bd.

### Fixed inline (must-fix)

1. ✅ **persistDispatchResult now no-throws on a delivered DM** — exactly the bug you flagged. A DDB write failure was propagating up through batchSettled and the recipient was getting reported as "could not be reached" even though their inbox had the DM. Wrapped each db.X call in a `persist` helper that catches, emits the new `DISPATCH_PERSIST_FAILED` canary (when delivered=true), logs at error level, and swallows. The function returns normally so the caller continues to report the recipient as reached — matching reality.

2. ✅ **Canary survives a DDB outage** — re-ordered the divergence-branch audit + warn to fire BEFORE the persist (cycle-7 had moved them after for truthfulness, but you're right that masked the discord.js shape-drift signal during a concurrent DDB outage). With the new no-throw persist + the DISPATCH_PERSIST_FAILED fallback, both signals fire independently — DISPATCH_SENT_NO_REFS for the discord.js side, DISPATCH_PERSIST_FAILED for the DDB side.

### New canary event

3. **`AUDIT_EVENTS.DISPATCH_PERSIST_FAILED`** — separate from DISPATCH_SENT_NO_REFS so an operator chasing "DM delivered but DDB doesn't know" has a single dashboard signal. Should always read zero. Tested with three new cases covering: markSendDMDelivered throw, divergence + persist throw stack, and the FAILED-status persist throw (no canary fires since no delivered DM).

### Already tracked

4. **markSendDMDelivered ConditionExpression** — layervai/qurl-integrations#366.
5. **Wider DDB write surface (item size growth)** — flagged for monitoring; ConditionExpression follow-up will cover the durable-record hardening.

### Judgment-call / non-actionable

6. **revokeAllLinks blocking on edit phase** — fire-and-forget would help operator latency but breaks test-pinning of editDM calls cleanly. Keeping the awaited model for now; if real operator latency complaints surface, will reconsider in a focused follow-up.
7. **Edit fan-out width = 5** — comment now explains the bound is global rate-limit pressure (per-channel buckets are independent). Could bump but no observed pressure today.
8. **Extract editStrictSuccessDMs** — readability win at 140 lines, but increases cognitive load for tracing the post-revoke flow. Leaving inline.
9. **buildRevokedDMPayload brand-agnostic copy** — recipient UX-friendly, intentional.
10. **eslint-disable redundancy on the SQLite stub** — verified lint config doesn't have argsIgnorePattern, so the disable is required. Underscore prefix is purely stylistic.

1718 tests pass. CI re-running.

justin-layerv:
## Cycle 9 — polish

Cr explicitly approved ("Verdict: Ship it"). Acted on the two optional polish items in commit e13f047.

- ✅ **Trimmed the CANONICAL DELIVERY-RATE METRIC comment in half** per your suggestion. Kept the load-bearing CANARY-not-subtractor distinction; dropped the prose around "they should agree."
- ✅ **Added a debug-level silent-skip log** in the edit phase. When success > 0 but `editTargets.size === 0` (the exact SQLite local-dev case where DM refs aren't persisted), the log surfaces the intentional skip so a dev doesn't chase a phantom bug. Debug-level keeps it out of prod by default. Test pins both the new debug log AND the absence of the info-level "Edited DMs after revoke" line for that path.

All other items either tracked in follow-ups (#365 SQLite projection, layervai/qurl-integrations#366 ConditionExpression, layervai/qurl-integrations#367 resource-type copy, layervai/qurl-integrations-infra#917 / layervai/qurl-integrations-infra#918 CloudWatch alarms, layervai/qurl-integrations#369 50007 telemetry) or non-actionable (custom Error class is TS-migration territory; 5xx retry-with-jitter is part of the layervai/qurl-integrations-infra#917 conversation).

1718 tests pass.

justin-layerv:
## Cycle 10 — final polish

Cr explicitly LGTM'd. Acted on every nit + the small test gap in commit 1fbf331.

- ✅ **Dedup-loop comment** retightened to your exact phrasing.
- ✅ **First-valid-wins INVARIANT** comment added — documents the one-DM-channel-per-recipient assumption that makes the strategy safe, plus what would have to change if a future PR sent two separate DMs per send.
- ✅ **Stale "unused production code" comment** in discord-rest.js updated: editDM is no longer unused (consumed by commands.js's revoke path).
- ✅ **eslint-disable redundancy** confirmed and dropped — verified `.eslintrc.json` has `argsIgnorePattern: "^_"` set, so the leading-underscore convention on `_channelId` / `_messageId` already exempts them. Updated the comment to call out the convention explicitly.
- ✅ **getSendItems-throws test** added — pins that a DDB outage during revoke propagates to the caller without attempting DELETEs or emitting audit events. Symmetric with the persistDispatchResult DDB-throws cases from cycle 8.

Skipped per matching judgment:
- Brand spelling "link" vs "qURL" — your reasoning aligns with mine: "this link" is more intelligible to recipients who never see the brand. Leaving as-is.

Other items already tracked (#365 SQLite projection, layervai/qurl-integrations#366 ConditionExpression, layervai/qurl-integrations#367 resource-type copy, layervai/qurl-integrations-infra#917 / layervai/qurl-integrations-infra#918 CloudWatch alarms, layervai/qurl-integrations#369 50007 telemetry).

1719 tests pass.

justin-layerv:
## Cycle 11 — wording fix

Cr LGTM'd; one actionable wording quibble closed in 27f550e.

- ✅ **"recipients whose revoke failed (link was already opened)"** broadened — also covers transient qURL API 5xx during DELETE and mixed-outcome recipients. Behavior unchanged; just more accurate prose.
- ✅ **Test mock isolation verified** (cr suggestion layervai/qurl-integrations#8) — top-level beforeEach at `tests/send-pipeline-back-half.test.js:311` calls `jest.clearAllMocks()` between every test, so the inner `post-revoke DM edit` describe can't inherit stale call state from earlier `revokeAllLinks` tests. Inner beforeEach resets the editDM default-return (clearAllMocks doesn't reset implementations).

All other cr observations either tracked in follow-up issues or accepted trade-offs documented in-tree.

1719 tests pass.

justin-layerv:
## Cycle 12 — doc nits

Cr LGTM'd (4th consecutive). Two actionable doc nits closed in 53953af.

- ✅ **getSendItems return-shape doc** — added a RETURN SHAPE block enumerating the post-PR field set (resource_id, recipient_discord_id, dm_channel_id, dm_message_id, dm_status) with notes on which fields markSendDMDelivered populates vs. legacy-row undefineds. Future callers see the surface up front instead of chasing it through the revoke consumer.
- ✅ **INVARIANT-BREAKER marker** — split the editTargets one-dm-per-recipient INVARIANT into a named `INVARIANT(one-dm-per-recipient)` block and a grep-friendly `INVARIANT-BREAKER` block. If a future PR sends two separate DMs per send (e.g., multi-message dispatch for embed-count overflow), `grep INVARIANT-BREAKER` lands them on exactly this site to flip the dedup key.

Other observations either tracked (#365, layervai/qurl-integrations#366, layervai/qurl-integrations-infra#918) or accepted trade-offs already documented inline.

Manual-test note acknowledged: legacy-send case is mocked in unit tests; will verify against staging DDB before merge per the manual checklist in the PR description.

1719 tests pass.

justin-layerv:
## Cycle 13 — confirmation items

Cr LGTM'd (5th consecutive). Two follow-up confirmations closed in cb37122.

- ✅ **expectedReason log tag** — `DM_EDIT_EXPECTED_API_CODES` Set promoted to a Map of code → human-readable description. The editDM info-level log now carries an `expectedReason` field naming the matched code's meaning. Closes the cr-flagged invisibility gap: a future 50007-on-PATCH spike (or any single expected code lighting up) is greppable from CloudWatch by description without an external Discord docs lookup. New test pins both the present-on-expected and absent-on-unexpected branches.
- ✅ **recordQURLSendBatch ordering invariant confirmed** — verified in-tree: `executeSendPipeline:1474` and `handleAddRecipients:2134` both `await db.recordQURLSendBatch(...)` and early-return on throw before the `batchSettled` dispatch loop starts. So the markSendDMDelivered orphan-row window is provably empty in production until layervai/qurl-integrations#366 lands the `attribute_exists` guard. Added a CALLER INVARIANT note on persistDispatchResult so a future refactor can't break the ordering silently.

Other observations either tracked (#367 resource-aware copy) or accepted trade-offs already documented inline. Optional structural refactor (extract editRevokedDMs helper) skipped — cr explicitly says "reads fine as-is."

1720 tests pass.

justin-layerv:
## Cycle 14 — three small additions

Cr LGTM'd (6th consecutive). Three actionable items closed in 338cbbc.

- ✅ **Shared-rate-limit-bucket invariant test restored** — added `expect(client.rest).toBe(restMock)` in a dedicated `describe` block. You're right that the helper-level mock-call assertions only catch interception on `require('../src/discord').client.rest`, not on the REST constructor itself. A future refactor swapping `client.rest.post(...)` for `new REST(...).post(...)` would silently break shared-bucket semantics under those helper tests; this explicit toBe pins the wiring.
- ✅ **`Map.get(error.code)` undefined behavior documented** — one-line comment on the fall-through: an error with no API code (revoked-token 403, synthetic proxy 404) intentionally lands on the unexpected/warn branch.
- ✅ **`success === 0` edit-skip test added** — pins the `if (success > 0)` guard at the top of the post-revoke edit block. When every DELETE fails (qURL service down), no recipient has had their link revoked; editing their DM to "closed the door" would be a lie. The new test also asserts the debug skip-log doesn't fire on this branch (since the outer guard short-circuits).

Other observations either tracked in follow-ups or accepted trade-offs documented inline.

1722 tests pass.

justin-layerv:
## Cycle 15 — return-shape additions + DISPATCH_PERSIST_FAILED alarm follow-up

Cr LGTM'd (7th consecutive). Three suggestions:

### Fixed inline (combined layervai/qurl-integrations#2 + layervai/qurl-integrations#3)

- ✅ **editDM return now exposes `code` and `reason`** alongside `{ ok, expected }`. `code` mirrors `error.code` (undefined for non-Discord-shape errors); `reason` is the human-readable description from `DM_EDIT_EXPECTED_API_CODE_DESCRIPTIONS` (undefined for unexpected branches). Future callers can carry the diagnostic info into their rolled-up logs without re-grepping the per-edit log line. All existing editDM tests updated to pin the new return shape.

### Deferred

- **CloudWatch alarm for `DISPATCH_PERSIST_FAILED`** — filed as layervai/qurl-integrations-infra#919. Mirrors the layervai/qurl-integrations-infra#917 / layervai/qurl-integrations-infra#918 / layervai/qurl-integrations#366 pattern: bot emits the event in this PR; terraform metric filter + alarm lives in qurl-integrations-infra. Same precedent as the existing `DISPATCH_SENT_NO_REFS` alarm wiring in layervai/qurl-integrations-infra#917.

### Skipped (cr explicit "theoretical")

- DELETE-batch-fully-settles-before-edit ordering test — the existing partial-revoke test (`skips recipients whose revoke failed`) implicitly covers the practical case: a refactor that interleaved DELETEs and edits would fire an edit for a recipient whose later DELETE then fails, which the assertion catches. The cr explicitly framed this as theoretical.

1722 tests pass.

justin-layerv:
## Cycle 16 — three nits closed

Cr LGTM'd (8th consecutive). All three actionable items in 260ed90.

- ✅ **Dropped the tautological invariant test** — you nailed it. `restMock === __mockRestInstance === client.rest` per the jest mock definition, so the assertion reduced to mock-against-mock. Replaced with a top-level comment pointing at the helper-level `restMock.X.mock.calls` assertions, which DO pin the behavior (a `new REST(...)` swap wouldn't hit the mock and call counts would drop to zero).
- ✅ **Forward-compat note on `code` / `reason`** — added inline that the fields are unread by the revoke-loop caller today; landing pad for the dashboard work in layervai/qurl-integrations-infra#917 / layervai/qurl-integrations-infra#918 / layervai/qurl-integrations-infra#919.
- ✅ **SQLite divergence-time debug log** — added `logger.debug` inside the markSendDMDelivered stub so the silent-skip surfaces at WRITE time (not just at revoke time when the missing-refs guard fires). Local dev sees it the moment a dispatch lands; log message points at layervai/qurl-integrations#365.

1721 tests pass.

PR #371, opened 2026-05-15
feat(discord): add @everyone button on /qurl file + /qurl map confirm card

justin-layerv:
## Summary

- Discord's \`MentionableSelectMenu\` filters the \`@everyone\` role out of its dropdown server-side (platform UI behavior, can't be unflipped). Users with MENTION_EVERYONE had no in-UI affordance for "fan out to the whole server" — they had to fall back to typing \`@everyone\` in the slash command's recipients option (and even that broke silently until #363, the cache pre-warm fix this PR is stacked on).
- Add a dedicated **📢 @everyone (N)** button to the confirm card's button row, gated on \`interaction.memberPermissions.has(MentionEveryone)\` and guild context. Click-time semantics mirror the existing picker @everyone-role branch:
  - defense-in-depth permission re-check (defends against a forged HTTP interaction carrying the button's custom_id)
  - cache pre-warm (uses the helper from #363)
  - iterate \`guild.members.cache\` → \`partitionRecipients\` (drops bots, flags \`selfIncluded\`)
  - cap-check (mirrors voice-everyone's hard-reject past \`QURL_SEND_MAX_RECIPIENTS\`)
  - transitionFlow + re-render
- Label count is render-time non-bot when the cache is fully populated (\`cache.size >= memberCount\`), falls back to \`guild.memberCount\` when cold. Same filter-drift trade-off the voice-everyone button accepts — \`partitionRecipients\` is the click-time authority.

## Why

User reported the missing affordance — \`@everyone\` typed in the recipients option worked (after #363) but the picker dropdown showed Users + Roles section without \`@everyone\` in the Roles section. After confirming that Discord's MentionableSelectMenu filters \`@everyone\` server-side, the cleanest fix is a dedicated button — symmetric with the existing "Everyone in voice channel" button users already learn from voice-channel sends. Discoverable, one-click, self-describing label.

## Scope

- \`src/commands.js\`:
  - new \`CONFIRM_EVERYONE_BUTTON_CUSTOM_ID\`
  - \`renderConfirmCardRows\` adds the button to \`bottomRow\` (after voice-everyone if present, before Note/Send/Cancel) when MENTION_EVERYONE
  - new \`handleConfirmEveryone\` handler (mirrors \`handleConfirmVoiceEveryone\`)
  - flow-dispatch registration + module export
- \`tests/qurl-file-map.test.js\`: 11 new tests covering render gating, label count freshness across hot/cold cache, the happy path, forged-perm rejection, cold cache, bots-only, cap rejection, corrupt resourceType, deferUpdate ack guard.

## Test plan

- [ ] CI green
- [ ] Local: \`/qurl file\` as MENTION_EVERYONE user → confirm card shows 📢 @everyone (N) button → click → expands to all non-bot members
- [ ] Local: \`/qurl file\` as non-MENTION_EVERYONE user → button does NOT render
- [ ] Local: button label refreshes count on re-render after a picker change
- [ ] Local: in a voice channel, both 🔊 voice-everyone and 📢 @everyone buttons render side-by-side (5 components in the bottom row, at Discord's limit)

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3859,12 +3859,24 @@ function renderConfirmCardRows({
       displayCount = memberCount;
     }
     const labelCount = displayCount == null ? '?' : String(displayCount);
+    // Disable on both `0` (empty guild — degenerate) AND `null`
+    // (memberCount unavailable + cache cold). The `(?)` label
+    // accompanies the disabled state so the user sees the button can't
+    // act right now rather than getting a silent-no-op click on a
+    // count-less label.
+    //
+    // BOTTOM ROW COMPONENT BUDGET: Discord caps an ActionRow at 5
+    // components. The current worst-case render is
+    // [🔊 Voice] + [📢 @everyone] + Note + Send + Cancel = 5, exactly
+    // at the limit. Adding another button below would silently throw
+    // at discord.js builder time. Re-evaluate the layout (e.g., a
+    // second row for affordances vs. fixed buttons) before adding.
     bottomRow.addComponents(
       new ButtonBuilder()
         .setCustomId(CONFIRM_EVERYONE_BUTTON_CUSTOM_ID)
         .setLabel(`\u{1F4E2} @everyone (${labelCount})`)
         .setStyle(ButtonStyle.Secondary)
-        .setDisabled(displayCount === 0),
+        .setDisabled(displayCount == null || displayCount === 0),
     );
   }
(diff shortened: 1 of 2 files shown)


justin-layerv:
r6 follow-up filed: #376 (parallel success-path info log on handleConfirmVoiceEveryone) — out of scope for this PR but worth tracking; voice-mode fan-outs are equally interesting to ops.

PR #374, opened 2026-05-15
fix(discord): stop stacking ephemerals on dead-flow component clicks

justin-layerv:
## Summary

- Repeated component clicks (most visibly **Cancel**) on a `/qurl file` or `/qurl map` confirm card whose 180s flow TTL had elapsed stacked one ephemeral per click instead of clearing the stale card. Each ephemeral carried a reply-quote of the source card, so the channel looked like the same dialog was appearing over and over.
- The dispatcher's three routing-failure early returns now `update` the source card in place (clearing its components) for `isMessageComponent()` interactions, killing the stale buttons and ending the stack.

## User impact

A user reported that clicking **Cancel** on a `/qurl file` confirm card surfaced "This action was superseded — run the command again." and what looked like a new copy of the card appearing above the previous one, every click. Sandbox repro: render a confirm card, wait > 180s, click any component (Cancel, Send, picker, expiry, self-destruct, Add a note, voice-everyone, pick-people-instead).

## Root cause

In `flow-dispatch.js`, the three routing-failure paths —
- `flowIdForInteraction` throw
- `loadFlow` throw
- `!row || row.stage !== route.expectedStage`

— used `safeReply` → `interaction.reply({content, ephemeral: true})`. That's a fresh ephemeral with the source card embedded as a Discord-rendered reply-quote, and it does NOT touch the source card itself. The source card's buttons stay live, so the user can (and did) keep clicking them; each click repeats the path and stacks another ephemeral.

## Fix shape

New `supersededRoutingFailureReply(interaction)` helper used by all three early returns. Behavior:

| Interaction state | Action |
|---|---|
| `isMessageComponent()` + not acked | `interaction.update({content: SUPERSEDED_MSG, components: []})` — edits the source card, clears its buttons |
| `isModalSubmit()` (or unknown shape) | `safeReply` (existing ephemeral / followUp behavior) |
| Already replied/deferred | `safeReply` → `followUp` (defends against future reorder) |
| `update` throws (e.g. Unknown Message) | falls through to `safeReply` |

`safeReply` itself is **unchanged** — it's still the right shape for the post-handler-throw catch and for handler-internal ambiguous-ack callsites.

Note: not Cancel-specific. The fix applies uniformly to all 8 confirm-card customIds (Send / Cancel / picker / expiry / self-destruct / note button / voice-everyone / pick-people-instead) plus the revoke and setup flows — anywhere `flow-dispatch` decides the registered handler can't run.

## Tests

`apps/discord/tests/flow-dispatch.test.js`: 25 cases pass (was 21).

- Pinned `update`-first behavior for the three routing-failure paths (row-missing / stage-mismatch / loadFlow-throws).
- Added: modal-submits-on-missing-row → reply, not update.
- Added: update-fails → fallback to reply with SUPERSEDED_MSG.
- Tightened: dead-token swallow test now mocks BOTH `update` and `reply` rejecting.

## Test plan

- [x] `npm test` (1,736 / 1,736 pass)
- [x] `npm run lint` (clean)
- [ ] Staging: render a `/qurl file` confirm card, wait > 180s, click Cancel — card should be replaced in place with the SUPERSEDED message, no stacking.
- [ ] Staging: same flow but click Send / Note / picker / expiry-select / self-destruct-select / voice-everyone (where applicable) after TTL — same in-place replacement.
- [ ] Staging: click Cancel within 30s of card creation — should still see "Send cancelled." via `handleConfirmCancelClick` (untouched path).
- [ ] Staging: dismiss the source ephemeral mid-flow, then click a stale button (synthetic — Discord usually swallows clicks on dismissed ephemerals) — fall-through to ephemeral reply, no throw.

### DIFF
--- a/apps/discord/src/flow-dispatch.js
+++ b/apps/discord/src/flow-dispatch.js
@@ -255,9 +255,11 @@ async function handleFlowInteraction(interaction) {
 //     clicking; the interaction token is still live so a fresh reply
 //     is the right recovery.
 async function supersededRoutingFailureReply(interaction) {
-  const isComponent = typeof interaction.isMessageComponent === 'function'
-    && interaction.isMessageComponent();
-  if (isComponent && !interaction.replied && !interaction.deferred) {
+  // `isMessageComponent` is guaranteed to exist — `index.js` only
+  // routes here when `isMessageComponent() || isModalSubmit()` is
+  // true (interactionCreate listener), and discord.js declares both
+  // methods on the base Interaction class.
+  if (interaction.isMessageComponent() && !interaction.replied && !interaction.deferred) {
     try {
       await interaction.update({ content: SUPERSEDED_MSG, components: [] });
       return;
--- a/apps/discord/src/flow-dispatch.js
+++ b/apps/discord/src/flow-dispatch.js
@@ -179,7 +179,7 @@ async function handleFlowInteraction(interaction) {
     logger.warn('flow-dispatch: failed to derive flow_id from interaction', {
       customId, error: err.message,
     });
-    await safeReply(interaction, SUPERSEDED_MSG);
+    await supersededRoutingFailureReply(interaction);
     return;
   }
 
@@ -190,7 +190,7 @@ async function handleFlowInteraction(interaction) {
     logger.error('flow-dispatch: loadFlow failed', {
       customId, flow_id, error: err.message,
     });
-    await safeReply(interaction, SUPERSEDED_MSG);
+    await supersededRoutingFailureReply(interaction);
     return;
   }
 
@@ -204,7 +204,7 @@ async function handleFlowInteraction(interaction) {
       stage: row ? row.stage : null,
       expected_stage: route.expectedStage,
     });
-    await safeReply(interaction, SUPERSEDED_MSG);
+    await supersededRoutingFailureReply(interaction);
     return;
   }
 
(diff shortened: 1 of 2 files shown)


justin-layerv:
## /simplify pass

Three review agents (reuse / quality / efficiency) reviewed the diff in parallel.

### Fixed (1)

**Drop defensive `typeof interaction.isMessageComponent === 'function'` check** (`dca9e2b`) — `src/index.js` interactionCreate listener already gates routing to `isMessageComponent() || isModalSubmit()`, and discord.js declares both predicates on the base Interaction class. The guard couldn't fire. Now reads `interaction.isMessageComponent()` directly with a comment pointing to the upstream gate.

### Skipped, with rationale (1)

**Latent `ackOrFail(interaction, content, options)` abstraction** that would merge `supersededRoutingFailureReply` with the handler-throw "Something went wrong" path. Skipping: routing failures and handler throws have distinct recovery semantics (update source card with terminal copy vs. always-ephemeral followup with neutral copy). Hardcoding `SUPERSEDED_MSG` in the helper also closes the "wrong wording" footgun. Not worth the parameter sprawl for one extra caller.

### Confirmed clean

- **Reuse**: No existing "clear card" utility in `discord.js` / `discord-rest.js` / `utils/`. `safeReply` is correctly distinct (handles ack-state branching for any interaction type; this helper is component-specific update-first optimization with `safeReply` as its fallback).
- **Comments**: Captured non-obvious WHY (stacking repro, fallback-token-still-live invariant, modal-submit divergence). Test comments are load-bearing — they pin contract intent, not assertions.
- **Hot path**: Helper only executes on the three error branches; zero overhead on the handler-runs path.

`npm test` (1,736 / 1,736 pass) + `npm run lint` (clean) on `dca9e2b`.

PR #382, opened 2026-05-15
feat(discord): hide picker after @everyone or voice fan-out

justin-layerv:
## Why

The 📢 @everyone confirm-card button on `/qurl file` and `/qurl map` was auto-filling the MentionableSelect picker with up to 25 of N recipients after a fan-out. To a user looking at the card, that read as **"the bot capped my @everyone at 10/25 people"** — and worse, any subsequent picker interaction would route through `handleConfirmUserSelect` and silently replace the full fan-out with whatever subset of names happened to be pre-selected.

The actual recipient set was always bounded by `QURL_SEND_MAX_RECIPIENTS = 20000`; the perceived cap was the picker's own UX leaking into a flow that didn't need it. This PR removes the leak and trims the surrounding copy.

## Summary

- **New `RECIPIENT_MODE_EVERYONE`** alongside `picker` / `voice`. After 📢 @everyone (click OR text-mention), the card hides the MentionableSelect picker entirely and surfaces a `👥 Pick people instead` escape hatch — same shape as voice-mode. Eliminates the auto-fill misread.
- **Text-mention parity:** `parseRecipientMentions` now returns `massMentionExpanded: boolean` (the precise "expansion ran" predicate, distinct from `massMentionDenied`). `handleQurlSlashSend` reads it to land text `/qurl file @everyone` in EVERYONE-mode too — previously this path went through PICKER-mode with the same auto-fill problem.
- **Terser button labels.** 🔊 button is now `🔊 Everyone on voice` (was `🔊 Everyone in #{channelName} (N)` with UTF-16-aware truncation against Discord's 80-unit cap). 📢 button is now `📢 @everyone` (was `📢 @everyone (N) — exceeds N cap`). Live counts and the over-cap suffix are gone; the disabled+greyed-out button is the only state signal. Reduces markdown-injection surface area on channel names too.
- **Drop `(you not included)` disclosure** from the voice-mode "To:" line. Sender exclusion stays inferred from voice-mode semantics. The data-side `selfIncluded:false` invariant is unchanged; the renderer guard against `selfIncluded:true` in voice-mode still suppresses a contradictory "Send includes you." notice on forged/drifted payloads.

## Mode behavior matrix

| Mode | Picker row | Bottom-row entry buttons | Escape hatch |
|---|---|---|---|
| `picker` | ✅ shown | 🔊 Voice (in voice channel), 📢 @everyone (with perm) | — |
| `voice` | ❌ hidden | — | 👥 Pick people instead |
| `everyone` (new) | ❌ hidden | — | 👥 Pick people instead |

## Mid-deploy / stale-row safety

- `normalizeRecipientMode` maps unknown values to `picker`, so old bot reading a new `'everyone'` row gracefully degrades to the legacy auto-fill behavior (no data loss, just a UI regression for a single confirm card mid-deploy).
- `handleConfirmUserSelect` explicitly overrides `recipientMode → 'picker'` on every transition so a forged interaction or stale UI emitting a picker event from EVERYONE mode can't re-leak the prior mode forward.
- All `handleConfirmEveryone` reject paths set `RECIPIENT_MODE_PICKER` — error recovery never strands the user in EVERYONE mode with empty recipients.

## Test plan

- [x] `npm test` — 1778/1778 pass, 47 suites
- [x] `npm run lint` — clean
- [ ] Manual: `/qurl file` + click 📢 @everyone → picker disappears, "Pick people instead" appears, Send fans out to full guild
- [ ] Manual: `/qurl file recipients:@everyone` (with MENTION_EVERYONE) → same end state as button click
- [ ] Manual: `/qurl file` from voice channel → 🔊 Everyone on voice label, "To:" line shows channel mention without "(you not included)"
- [ ] Manual: click "Pick people instead" from each non-picker mode → returns to PICKER with empty recipients

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3908,24 +3908,7 @@ function renderConfirmCardRows({
     // Count is render-time, not click-time — members can join/leave
     // voice between renders. Click-time resolution in
     // handleConfirmVoiceEveryone is the authoritative recipient set;
-    // the label is a freshness hint that re-derives on every other
-    // confirm-card interaction (picker / expiry / note edits all flow
-    // through renderConfirmCardRows again).
-    // Filter-drift contract: the render-time count below uses
-    // `isBotMember(m)` to compute (N), while click-time resolution
-    // in handleConfirmVoiceEveryone routes channel.members through
-    // `partitionRecipients` for the authoritative recipient set.
-    // Both apply the same bot filter today, so the count is honest.
-    // If `partitionRecipients` ever picks up additional drops (role-
-    // blocklist, self-filter toggle, etc.), the render-time `(N)`
-    // will silently overstate the click-time set — keep the two
-    // filter sources aligned, or accept a stale label and document
-    // the drift here.
-    // Live connected-non-bot count is still computed so the button
-    // can be disabled when the channel is empty or unreadable — it
-    // just doesn't surface in the label anymore (kept terse per
-    // product UX call). `connectedCount == null` is the single
-    // sentinel for "render the disabled button shell."
+    // the count here only drives the disable state.
     let connectedCount = null;
     if (interaction) {
       const channel = interaction.guild?.channels?.cache?.get?.(voiceChannelId);
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2915,7 +2915,7 @@ const CONFIRM_EXPIRY_SELECT_CUSTOM_ID = 'qurl_confirm_expiry';
 const CONFIRM_SELF_DESTRUCT_SELECT_CUSTOM_ID = 'qurl_confirm_self_destruct';
 const CONFIRM_NOTE_BUTTON_CUSTOM_ID = 'qurl_confirm_note_btn';
 const CONFIRM_NOTE_MODAL_CUSTOM_ID = 'qurl_confirm_note_modal';
-// Confirm-card "Everyone in this voice channel" button. Rendered only
+// Confirm-card "Everyone on voice" button. Rendered only
 // when the slash command was invoked from a voice / stage-voice channel
 // (`payload.voiceChannelId` is set by `handleQurlSlashSend`). Mirrors
 // the role-mention/`<#voice>` parser path: resolve voice-connected
@@ -2946,11 +2946,12 @@ const CONFIRM_NOTE_MODAL_CUSTOM_ID = 'qurl_confirm_note_modal';
 // Tracked in #339 if a future product decision (stage channels with
 // thousands of audience) needs to revisit this for stage-only.
 const CONFIRM_VOICE_EVERYONE_BUTTON_CUSTOM_ID = 'qurl_confirm_voice_everyone';
-// "Pick people instead" button — rendered only in voice-mode (i.e.
-// `payload.recipientMode === 'voice'`). Clearing it flips the card
-// back to picker-mode (recipientIds dropped, MentionableSelect row
-// restored). Lives on its own customId so flow-dispatch routes
-// directly instead of branching inside the voice-everyone handler.
+// "Pick people instead" button — rendered in voice-mode and
+// everyone-mode (i.e. `payload.recipientMode === 'voice' || 'everyone'`).
+// Clicking it flips the card back to picker-mode (recipientIds
+// dropped, MentionableSelect row restored). Lives on its own customId
+// so flow-dispatch routes directly instead of branching inside the
+// voice/@everyone handlers.
 const CONFIRM_PICK_MANUAL_BUTTON_CUSTOM_ID = 'qurl_confirm_pick_manual';
 // Whole-guild `@everyone` button on the confirm card. Workaround for
 // Discord's MentionableSelectMenu omitting the `@everyone` role from
@@ -2958,33 +2959,44 @@ const CONFIRM_PICK_MANUAL_BUTTON_CUSTOM_ID = 'qurl_confirm_pick_manual';
 // MENTION_EVERYONE-bearing users with no in-picker affordance for
 // the "fan out to the whole server" intent. Click-time semantics
 // match a synthetic `@everyone` text-mention pick. Picker-mode only —
-// voice-mode already targets the voice population and rendering the
-// button there would mislead users about what "everyone" means.
+// voice-mode already targets the voice population (different
+// "everyone"), and everyone-mode is the post-click state where this
+// button has already done its job (re-rendering it would invite a
+// re-click that just churns the version with no state change).
 const CONFIRM_EVERYONE_BUTTON_CUSTOM_ID = 'qurl_confirm_everyone';
 
-// Two recipient-source modes carried on `payload.recipientMode`:
+// Three recipient-source modes carried on `payload.recipientMode`:
 //   - 'picker' (default): MentionableSelect row is the source of
-//     truth; bottom row carries the "🔊 Everyone in #voice" affordance
-//     when the slash command was invoked from voice.
+//     truth; bottom row carries the "🔊 Everyone on voice" affordance
+//     when the slash command was invoked from voice, plus the
+//     "📢 @everyone" affordance when the sender has MENTION_EVERYONE.
 //   - 'voice': recipientIds were resolved from `channel.members` and
 //     the picker row is hidden. Bottom row swaps in "👥 Pick people
 //     instead" so the user can fall back to manual selection.
+//   - 'everyone': recipientIds were resolved from the guild's full
+//     non-bot member set via the 📢 @everyone shortcut. Like voice,
+//     the picker row is hidden and the bottom row carries the
+//     "👥 Pick people instead" escape hatch — auto-filling the
+//     picker would either silently truncate at Discord's 25-entry
+//     hard cap or read back through `handleConfirmUserSelect` and
+//     replace the @everyone fan-out with a subset.
 // Stale flow_state rows (created before this field existed) read as
 // undefined — `normalizeRecipientMode` below maps undefined / any
 // off-set value to RECIPIENT_MODE_PICKER so they keep the legacy
 // picker shape until they expire.
 const RECIPIENT_MODE_PICKER = 'picker';
 const RECIPIENT_MODE_VOICE = 'voice';
+const RECIPIENT_MODE_EVERYONE = 'everyone';
 
-// Single source of truth for "this token is voice; everything else
-// reads as picker." Three render sites (renderConfirmCardContent,
-// renderConfirmCardRows, rerenderConfirmCard) previously open-coded
-// `mode === RECIPIENT_MODE_VOICE ? VOICE : PICKER` — if any of those
-// drifted (e.g. someone tested `mode === 'voice'` against a typo'd
-// constant), the layout would split between branches. The helper
-// pins the stale-row default in one place.
+// Single source of truth for the closed-set map. Multiple render
+// sites (renderConfirmCardContent, renderConfirmCardRows,
+// rerenderConfirmCard) previously open-coded the ternary; an off-set
+// or typo'd constant would silently fall through to PICKER and break
+// the layout swap. The helper pins the stale-row default in one place.
 function normalizeRecipientMode(mode) {
-  return mode === RECIPIENT_MODE_VOICE ? RECIPIENT_MODE_VOICE : RECIPIENT_MODE_PICKER;
+  if (mode === RECIPIENT_MODE_VOICE) return RECIPIENT_MODE_VOICE;
+  if (mode === RECIPIENT_MODE_EVERYONE) return RECIPIENT_MODE_EVERYONE;
+  return RECIPIENT_MODE_PICKER;
 }
 
 // Recipient-rejection reason strings shared by every handler that
@@ -3168,10 +3180,10 @@ function partitionRecipients(users, senderId, { excludeSender = false } = {}) {
   for (const u of users) {
     if (u.bot) { droppedBots++; continue; }
     // Voice-everyone path drops the sender silently — they pressed
-    // the "Everyone in #voice" affordance, which semantically means
+    // the "Everyone on voice" affordance, which semantically means
     // "everyone else." No droppedBots-style accounting because the
-    // user-visible message ("you not included") is rendered from
-    // the recipientMode, not from a partition counter.
+    // exclusion is inferred from voice-mode semantics rather than
+    // surfaced in user-visible copy.
     if (u.id === senderId && excludeSender) continue;
     if (u.id === senderId) selfIncluded = true;
     valid.push(u);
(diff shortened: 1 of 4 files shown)


justin-layerv:
Thanks for the review — addressed all four points in [`f239cce`](https://github.com/layervai/qurl-integrations/commit/f239cce) and the earlier [`bdf0be8`](https://github.com/layervai/qurl-integrations/commit/bdf0be8):

1. **Forged voice-mode without `voiceChannelId`** — dropped the `voiceChannelId` guard on the voice-mode escape-hatch branch; "Pick people instead" now stays reachable in the degraded state. Added a regression test (`forged voice-mode without voiceChannelId still renders the escape hatch`) that pins the recovery shape.
2. **Mode-resolution ordering** — added a comment on the new `massMentionExpanded` branch noting it deliberately precedes the voice auto-default and that the precedence must be preserved if either guard is loosened.
3. **`=== 'voice' || 'everyone'` notation** — rewrote the `CONFIRM_PICK_MANUAL_BUTTON_CUSTOM_ID` doc comment to use natural-language phrasing instead of the JS-confusable form.
4. **Manual test plan** — those four items need a human in front of a Discord client; I can't run a live bot session here. Flagging that the third item (voice channel → no "(you not included)") is the most user-visible change and should get explicit eyeballs before merge. The data-side invariants are covered by the 1779-test automated suite.

Also picked up one `/simplify` finding pre-cr: collapsed the byte-identical voice/everyone "Pick people instead" branches into a single condition ([`bdf0be8`](https://github.com/layervai/qurl-integrations/commit/bdf0be8)).

justin-layerv:
Follow-up cr items addressed in [`4517efa`](https://github.com/layervai/qurl-integrations/commit/4517efa):

1. **Over-cap UX trade-off** — confirmed deliberate per the "terse labels" product call. Tightened the comment at `commands.js:3997` so the rationale is local (the prior wording implied the click-time reject covered all over-cap cases; it only covers cold-cache). Lever to bring back the in-card hint is right there if support feedback ever asks.
2. **Mixed text mentions** — added a `MIXED MENTIONS` paragraph on the `massMentionExpanded` branch documenting that `recipients: @everyone @alice` lands in EVERYONE mode (alice absorbed into the set). Correct semantically; comment saves the next reader the read-through.
3. **`valid.length > 0` gate** — documented as defense-in-depth, with an explicit note that it's structurally redundant today but pins the EVERYONE-mode invariant locally against future refactors of the upstream early-return.
4. **Mid-deploy forward path** — pinned via a new test (`mid-deploy forward: legacy picker-mode row with pre-filled recipientIds → click @everyone → lands in EVERYONE mode`). Verifies that a post-PR bot processing a pre-PR auto-fill row will overwrite mode to `'everyone'` and replace the truncated subset with the full fan-out, not spread-leak the legacy 'picker' forward.

Tests: 1780/1780, lint clean. Skipping #3's `connectedCount` optimization — the iteration is bounded by Discord's voice-channel cap (99) and the simplification has a subtle bot-only edge case that isn't worth the trade.

justin-layerv:
Thanks for the careful pass. Addressing the three items:

**F1 — slash-text vs button-click over-cap divergence.** Filed as #384 (Option A or B, both spelled out with acceptance criteria). Agree it's structurally more visible after this PR but a separate concern from the auto-fill leak this one fixes — punting to keep this PR's scope intact.

**F2 — slash-text sender-push divergence.** Added an in-place doc comment at the EVERYONE branch in `handleQurlSlashSend` (commit 4996974) calling out that the post-prewarm missing-row race is intentionally not mirrored from `handleConfirmEveryone`, with the rationale (user typed `@everyone` so they're aware of the recipient list at Send time) and a pointer for a future contributor pursuing strict parity.

**F3 — disabled-button copy on warm-cache over-cap.** Confirming the product call covers this case. The lever to bring the suffix back is preserved in the code comment near `commands.js:~3993–4012`. The minimal-label decision is upstream from the over-cap case — the rare warm-cache over-cap branch disabling the button without a suffix is the deliberate trade-off. If support sees confusion in the wild we'll revisit (the lever is still right there).

justin-layerv:
Thanks again. Round 5 follow-ups:

**Issue 1 — stale `normalizeRecipientMode` test (bb02b76).** Fixed. Updated the title to "closed set {voice, everyone, picker}", added the `'everyone' → 'everyone'` pass-through assertion, and added the case-sensitive `'EVERYONE' → 'picker'` reject (mirroring the `'VOICE'` line). Good catch — the test would have silently passed through a future collapse back to picker.

**Sender-push divergence tracking.** Doc comment is in-place at the slash-text EVERYONE branch (4996974). Leaving it as a comment for now — it's a narrow shard-resume race documented locally with the lever spelled out. Happy to file a tracking issue too if you'd prefer the GitHub-visible signal; let me know.

**Channel-name affordance loss.** Intentional. The user explicitly asked to drop the channel name from the voice button label (and similarly dropped `(N)` count + `(you not included)` disclosure) as a minimal-label / "infer from context" product call. The drift-mid-flow case is now covered by the channel mention in the "To:" line once the user lands in voice-mode; we're accepting that the pre-click picker-mode state no longer surfaces it. Confirming this is the deliberate trade-off.

**Disabled-without-reason on warm-cache over-cap.** Filed as #386 with the proposed inline-hint shape and acceptance criteria. The lever to reintroduce a label suffix is preserved in the code comment near `commands.js:~3993–4012` either way — #386 explores the card-content hint approach which is cleaner.

justin-layerv:
Thanks. Round 6 follow-ups — verdict noted, addressing the action items:

**#1, #5 — silent disable on warm/cold-cache (no in-card reason).** Confirming this is the deliberate product trade-off. The user invoking this PR is the product owner and explicitly directed the minimal-label decision (drop `(N)`, drop `(?)` suffix, drop channel name, drop `(you not included)` disclosure) as a "infer from context, no ceremony" pass. Filed #386 to capture the inline-hint approach if the trade-off ever needs revisiting; the lever to bring the suffix back is still preserved in the code comment at `commands.js:~3993–4012`. Not punting to a hypothetical separate support review — this is the call.

**#2 — selfIncluded divergence regression test (f37d178).** Added. New test at `qurl-file-map.test.js:~2999` simulates the post-prewarm race (sender's row missing from `members.cache`, other non-bot rows present), invokes the slash-text `@everyone` path, and pins `selfIncluded:false` + `recipientIds` excluding the sender. A future contributor mirroring the defensive sender-push from `handleConfirmEveryone` to the text path would flip both assertions and be forced to revisit the documented divergence — exactly what you flagged. Symmetric to the existing `qurl-file-map.test.js:5424` test that pins the button-click defensive push.

**#3 — MIXED MENTIONS semantics.** Confirming intentional: `@everyone` anywhere in the recipients token list lands EVERYONE mode. Rationale matches the comment at `commands.js:~4302`: you can't subtract from "everyone." Narrowing @-mentions of named users IS preserved via the `<@id>` path (no @everyone token → PICKER mode with the named users). Treating positional ordering as semantic intent would be surprising too, and the dedup-via-`seen` does the right thing for the audible-noise case (`@alice @everyone` shows alice once, not twice).

**Nit on terse-labels consistency for the click-time hard-reject (`commands.js:4400`).** Looked at the cited site — that's a banner ("⚠️ Voice channel has N eligible recipients (max CAP) — pick recipients below."), not a button label. The terse-labels product call was scoped specifically to button labels (which are space-constrained and pre-click). Banners are informational and post-interaction, so the constraint doesn't transfer cleanly. Leaving as-is.

justin-layerv:
Thanks. Round 7 follow-ups (80cd4f9):

**#1 — test placement.** Moved the `'forged voice-mode without voiceChannelId still renders the escape hatch'` test out of the `everyone-mode layout` block and into the `voice-mode layout` block where it belongs (it exercises `recipientMode:'voice'`). Discoverability win.

**#2 — UX trade-off follow-up.** Filed earlier as #386 with the proposed inline-hint shape. The "lever is right here" comment at `commands.js:~3993–4012` is preserved either way — #386 explores the cleaner card-content approach instead of bringing the label suffix back.

**#3 — comment density on SENDER-PUSH DIVERGENCE.** Leaving as-is. The block documents *why* the divergence exists; the regression test at `qurl-file-map.test.js:~2999` pins *what*. Both are load-bearing — `why` survives a future contributor who reads only the source, `what` survives a future refactor that doesn't read the source. Trimming to a pointer would lose the rationale (the user-intent argument and the upstream-prewarm context) that makes the divergence defensible. Style preference noted, but the verbosity is intentional here.

**#4 — `massMentionExpanded` derivation.** Fixed. Hoisted to a `let` outside the @everyone block and set inside the walk after a non-bot entry is added. The flag now means "expansion actually ran and added at least one entry" rather than "expansion was permitted" — robust against the future short-circuit cases you flagged (cap-saturated, empty cache, all-bot cache). Docstring at the top of the parser updated. No downstream behavior change: caller's `valid.length > 0` guard on the EVERYONE-mode branch was already catching those edges, this just makes the parser's flag honest.

All 1781/1781 tests pass.

justin-layerv:
Round 8 follow-ups (25ab52e) — both nits addressed:

**Test gap — `RECIPIENT_MODE_EVERYONE` wire pin.** Added the import at the top of the test file and the one-line `expect(RECIPIENT_MODE_EVERYONE).toBe('everyone')` to the constants pin test alongside PICKER and VOICE. A future rename now surfaces where the comment says it should — `flow_state` rows persisted across a deploy.

**Doc/code drift on `massMentionExpanded`.** Took your first option (update the parser doc to match), since the dedup-then-set behavior aligns with the documented MIXED MENTIONS call-site behavior in `commands.js` (alice gets "semantically absorbed" into the @everyone set, lands EVERYONE mode). Updated the three doc surfaces:

- Top-of-file contract comment (`recipient-parser.js:~85`) now says *"iterated at least one non-bot member from `members.cache` (the member may already be in `ids` via a prior named mention — dedup happens inside `consider()`; the flag records that the expansion ran, not that it added new entries)."*
- Inline walk comment (`:~647`) now spells out the dedup-then-set ordering and points to the MIXED MENTIONS comment in `commands.js`.
- Bottom-of-function comment (`:~744`) updated to match.

Tightening the implementation (the other option) would have flipped the EVERYONE-mode dispatch for the mixed-mentions case from EVERYONE → PICKER, which is the documented call-site contract we want to preserve. Doc-only fix here.

All 1781/1781 tests pass.

justin-layerv:
Thanks. Round 9 verdict noted — "LGTM" with three discussion-point echoes.

**#1 — mixed-mention "narrowing" UX gap.** Already documented in the MIXED MENTIONS comment at `commands.js:~4302` for support triage. Behavior is intentional and matches the parser's dedup-then-set semantics on `massMentionExpanded` (now also documented in the parser contract per round 8). Not a regression — pre-PR text `@everyone` lost the named mentions silently too; this PR at least surfaces EVERYONE-mode visually so the user can see they got a full fan-out.

**#2 — disabled-button warm-cache over-cap.** Filed earlier as #386 with the proposed inline-hint shape (card-content, not label). Lever to reintroduce the label suffix is preserved in the code comment.

**#3 — sender-push divergence.** Pinned by the regression test at `qurl-file-map.test.js:~2999` so a future "fix" can't silently flip it. Product call to leave it: user explicitly typed `@everyone` so they're aware of the recipient list at Send time, where a button-click is a one-tap shortcut and the defensive push makes sense.

Also pushed **b69c798** from a `/simplify` pass:
- Dropped the stale "Filter-drift contract" comment block at `commands.js:~3914` — warned about a `(N)` label drift hazard that no longer exists since this PR removed the suffix.
- Dropped the "kept terse per product UX call" change-narration line and a redundant restatement of the `connectedCount == null` sentinel.

Net: -17 lines of stale comments. All 1781/1781 tests pass.

justin-layerv:
Thanks. Round 10 verdict noted — "Approve."

**#3 — EVERYONE branch doesn't reassign `final*` variables.** Added the suggested one-line comment in 2dfec35 explaining that parser output is authoritative for EVERYONE mode (no voice-resolve re-partition step needed). Future refactor that adds an EVERYONE-mode re-partition step will surface here rather than silently land an off-by-one assignment.

**#1, #2, #4 — echoes of previously-addressed items.**
- #1 (voice channel + text `@everyone` → keeps `voiceChannelId` → click "Pick people instead" → voice button reappears): you noted this is correct behavior; agreed. The voice affordance is offered if relevant — that's the contract.
- #2 (text-path over-cap silently truncates): filed as #384 with the explicit option list for parity with the button-click reject.
- #4 (sender-push divergence): deliberate UX call confirmed (round 7+); pinned by the regression test at `qurl-file-map.test.js:~2999`.

All 1781/1781 tests pass.

---

**Calling convergence on this PR.** Final-check state:

- ✅ **DE acceptance** — round 10 returned "Approve" with the optional comment addressed.
- ✅ **CI passing** — all required status checks green (claude-review, Build and Test, Analyze, CodeQL, etc.).
- ✅ **All cr comments addressed** — 10 rounds of cr feedback worked through; nits and optionals treated as first-class; product trade-off questions answered.
- ✅ **/simplify clean** — three-agent pass on the full diff returned only one finding (stale "Filter-drift contract" comment), fixed in b69c798. Reuse and efficiency reviews clean.

Follow-up issues filed for items genuinely out of this PR's scope:
- #384 — slash-text vs button-click over-cap parity
- #386 — inline-hint shape for warm-cache over-cap disabled-button case

Commit history on this branch (chronological):
```
2dfec35  docs(discord): pin "parser output is authoritative for EVERYONE mode" invariant
b69c798  chore(discord): drop stale "Filter-drift contract" comment block
25ab52e  docs(discord): pin RECIPIENT_MODE_EVERYONE wire literal + reconcile massMentionExpanded contract
80cd4f9  refactor(discord): tighten massMentionExpanded semantics + relocate defensive test
f37d178  test(discord): pin selfIncluded divergence between text and button @everyone paths
bb02b76  test(discord): pin normalizeRecipientMode 'everyone' mapping
4996974  docs(discord): note slash-text vs button-click sender-push divergence
d291af1  docs(discord): trim defense-in-depth comment
4517efa  docs(discord): cr-feedback nits on the everyone-mode flow
f239cce  fix(discord): keep escape hatch reachable in forged voice-mode payload
bdf0be8  refactor(discord): merge identical voice/everyone pick-manual branches
1aa0bc4  feat(discord): EVERYONE recipient mode + terser confirm-card copy
```

justin-layerv:
Thanks. Round 11 — "Ship-ready." Quick close-out:

**Minor (cosmetic blank line)** — Fixed in aaf38bb.

**#1, #2, #3 — echoes of previously-addressed items.** Test-pinned, comment-documented, and tracked in #384/#386 respectively. Agreed on the release-notes / support-training framing for the MIXED MENTIONS behavioral change — will surface in the PR description before merge.

**#4 — partitionRecipients stale-comment sweep.** You explicitly flagged "Not worth a re-roll." Leaving for a future PR that actually touches `partitionRecipients`. Documenting here so the next contributor can grep.

PR is converged. Monitoring this branch ends with this comment.

PR #389, opened 2026-05-15
feat(discord): cap in-flight handler dispatches in event consumer (#388)

### LINKED ISSUE #388: Event consumer: cap in-flight detached handlers (backpressure)
Surfaced in PR #379 r12 cr. The worker tier's consumer (`src/event-consumer.js`) has no backpressure between the poll loop and the async handlers fired via emit.

## Current shape

- `pollOnce` pulls up to 10 messages per receive (`MaxNumberOfMessages`).
- For each message, `client.actions.InteractionCreate.handle(data)` emits `interactionCreate` synchronously.
- The listener (`handleCommand` / `handleFlowInteraction`) returns a Promise that runs detached on the event loop.
- `pollOnce` returns as soon as the per-message `DeleteMessage` calls settle — typically tens of ms after receive.
- pollLoop continues to the next ReceiveMessage immediately.

## Concern

Under a bursty arrival pattern where downstream work is slow (DDB OCC retries, REST throttling, DM fan-out at scale), in-flight handler promises accumulate with no upper bound. The consumer keeps draining batches faster than handlers complete → memory pressure → eventual OOM in the worker tier task. Symptoms would be:

- ECS task replacements right after PR 10's flag-flip
- `unhandledRejection` log noise from handler timeouts piling up
- DDB read/write spike from OCC retry storms

The module header acknowledges the asymmetry but the operational cost isn't quantified anywhere alerting can act on.

## Proposed shape

A simple semaphore around `pollOnce` — bound the number of receives in flight + handlers running:

```js
const MAX_IN_FLIGHT_HANDLERS = 100; // or env-driven
let inFlight = 0;

async function pollOnce(client) {
  if (inFlight >= MAX_IN_FLIGHT_HANDLERS) {
    await abortableSleep(100); // brief backoff while handlers drain
    return;
  }
  // ... existing pollOnce body ...
  // Track handler completion via a WeakMap registry the listener
  // populates so we can increment/decrement inFlight.
}
```

Alternative: a leaky-bucket pattern with a small in-process queue.

## Acceptance

- [ ] Configurable cap (`MAX_IN_FLIGHT_HANDLERS`) with sensible default + env override.
- [ ] Backpressure mechanism (either semaphore or queue).
- [ ] Metric: `qurl_bot_event_inflight_handlers_total` (gauge).
- [ ] Test: simulate slow handler + assert pollLoop doesn't exceed the cap.
- [ ] Sandbox-soak measurement after PR 10 lands to validate the cap is reasonable.

Priority: medium — not a blocker for PR 11 (queue is empty until PR 10 ships), but should land before flag-flip in prod. The cr explicitly calls it "track before flag-flip."

justin-layerv:
## Why

The SQS event consumer (PR #379) dispatches Discord interactions detached after the synchronous `client.actions.InteractionCreate.handle(data)` emit — `handleCommand` / `handleFlowInteraction` run as fire-and-forget promises with no captured reference. Under a sustained burst where downstream DDB/REST work runs slower than message arrival, in-flight handler promises accumulate without bound, risking worker-tier memory pressure before PR 10 ships and the flag flips.

This was filed during PR #379 as a tracked concern (#388). Today's volume bounds (invite-only beta, ~thousands of guilds) make it theoretical, but the cap is the right safety net before we route prod traffic through SQS.

## What

- New module-level state in `src/event-consumer.js`: `MAX_INFLIGHT_HANDLERS` (env-overridable, default 100), `INFLIGHT_BACKOFF_MS` (100), `inFlightCount`, `isWorkerDispatch`.
- New public function `eventConsumer.trackDispatch(maybePromise)` — called by the `interactionCreate` listener in `src/index.js` for every dispatch. Reads `isWorkerDispatch` and only registers consumer-driven dispatches in the cap.
- `processMessage` now sets `isWorkerDispatch = true` synchronously around the `handle()` call and clears it in `finally`. EventEmitter.emit is synchronous, so the listener fires while the flag is still true.
- `pollOnce` checks `inFlightCount >= MAX_INFLIGHT_HANDLERS` at the top and early-returns + `abortableSleep(INFLIGHT_BACKOFF_MS)` if at cap. The sleep honors `stopping`, so SIGTERM during a backpressure pause exits inside the graceful-shutdown budget.
- `src/index.js` listener captures the handler return value and passes it to `trackDispatch`.
- `.env.example` documents the new `QURL_BOT_MAX_INFLIGHT_HANDLERS` knob.
- 9 new tests in `event-consumer.test.js` covering no-op gateway path, non-promise inputs, the flag wrap (including finally-on-throw), increment+decrement on resolve and reject, at-cap early-return, below-cap pass-through, and the default value.

## Why this design

Tracking handler completion without coupling to handler internals is the central problem here. Two options were considered:

1. **Capture the promise in the listener directly.** Requires the listener to know it's the "consumer-driven" path vs the "gateway-WS-driven" path — and in combined mode (gateway + worker in one process), both paths share the same listener.
2. **Flag-based, set synchronously around the emit.** Exploits EventEmitter.emit's synchronous nature so the listener can read a module-level flag set by the caller. Zero coupling to handler internals; gateway-WS-driven dispatches see `isWorkerDispatch === false` and are no-ops.

Option 2 is what's implemented. The synchronous window is tight: there's no `await` between `isWorkerDispatch = true` and `isWorkerDispatch = false`, so concurrent `processMessage` calls in `Promise.allSettled` can't interleave their flag wraps.

## Test plan

- [x] `npm test` — 1856 passing (+40 over the pre-#388 baseline of 1816)
- [x] `npm run lint` — clean (`--max-warnings 0`)
- [x] All 49 `event-consumer.test.js` tests pass, including the 9 new backpressure tests
- [ ] CI green
- [ ] cr review clean

## Rollback

This is a no-op for the legacy in-process gateway path (`ENABLE_EVENT_SHIPPER` unset or `"false"`) — `isWorkerDispatch` stays false, `trackDispatch` is a no-op, `pollOnce` doesn't run. Reverting the PR is safe.

Closes #388.

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -103,7 +103,8 @@ ENABLE_OPENNHP_FEATURES=false
 # re-checking — prevents memory pressure from a sustained burst
 # outpacing downstream DDB/REST drain. Default 100; override only
 # if prod soak surfaces a different sustainable ceiling. Must be
-# a positive integer.
+# a positive integer (scientific notation like 1e3 is also accepted,
+# since Number('1e3')=1000 is a positive integer).
 #
 # Soft cap, not a hard semaphore: pollOnce gates BEFORE the next
 # receive, but a batch already-received can push the in-flight
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -99,11 +99,20 @@ ENABLE_OPENNHP_FEATURES=false
 
 # Maximum number of in-flight detached handlers the SQS consumer
 # tolerates before pausing receives (backpressure). When the count
-# reaches the cap, pollOnce returns early and waits ~100ms before
+# reaches the cap, pollOnce returns early and waits 100ms before
 # re-checking — prevents memory pressure from a sustained burst
 # outpacing downstream DDB/REST drain. Default 100; override only
 # if prod soak surfaces a different sustainable ceiling. Must be
 # a positive integer.
+#
+# Soft cap, not a hard semaphore: pollOnce gates BEFORE the next
+# receive, but a batch already-received can push the in-flight
+# count to `cap + 9` (RECEIVE_MAX_MESSAGES - 1) before the next
+# gate. Values below 10 have a degenerate overshoot ratio (cap=1
+# can transiently reach 10 in-flight handlers, 10x the configured
+# value); pick at least 10 for the overshoot ceiling to match
+# operator intuition. Size memory headroom assuming `cap + 10`
+# simultaneous handler promises regardless.
 # QURL_BOT_MAX_INFLIGHT_HANDLERS=100
 
 # GitHub OAuth App — REQUIRED only when ENABLE_OPENNHP_FEATURES=true.
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -97,6 +97,15 @@ ENABLE_OPENNHP_FEATURES=false
 # ENABLE_EVENT_SHIPPER=false
 # QURL_BOT_EVENTS_QUEUE_URL=https://sqs.us-east-2.amazonaws.com/000000000000/qurl-bot-events-sandbox
 
+# Maximum number of in-flight detached handlers the SQS consumer
+# tolerates before pausing receives (backpressure). When the count
+# reaches the cap, pollOnce returns early and waits ~100ms before
+# re-checking — prevents memory pressure from a sustained burst
+# outpacing downstream DDB/REST drain. Default 100; override only
+# if prod soak surfaces a different sustainable ceiling. Must be
+# a positive integer.
+# QURL_BOT_MAX_INFLIGHT_HANDLERS=100
+
 # GitHub OAuth App — REQUIRED only when ENABLE_OPENNHP_FEATURES=true.
 # Single-guild-plain and multi-tenant deployments don't mount /auth or
 # /webhook routes and don't need these values — the boot check
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -85,19 +85,24 @@
 // TODO(infra-publisher-set)` from any infra-side change surfaces
 // this requirement.
 //
-// No backpressure between the consumer and detached handlers: the
-// poll loop keeps pulling 10-message batches as long as the queue
-// has work, but `handleCommand` / `handleFlowInteraction` run
-// detached after the synchronous emit. Under a bursty arrival
-// pattern with slow downstream DDB/REST work, in-flight handler
-// promises accumulate without an upper bound — risking memory
-// pressure if a sustained backlog grows faster than handlers
-// drain. Today's volume bounds (invite-only beta, ~thousands of
-// guilds) make this comfortably theoretical; tracked as an
-// operational concern to revisit before flag-flip in PR 10. If
-// in-flight handler count needs to be capped, a semaphore around
-// `pollOnce` (max N concurrent dispatches before the next receive)
-// is the natural shape. See #388 for the follow-up.
+// Backpressure: the poll loop tracks in-flight detached handlers
+// and pauses receives when `inFlightCount >= MAX_INFLIGHT_HANDLERS`
+// (env-overridable; see QURL_BOT_MAX_INFLIGHT_HANDLERS). Without
+// this, under a bursty arrival pattern with slow downstream
+// DDB/REST work, handler promises would accumulate without bound,
+// risking memory pressure if the backlog grows faster than handlers
+// drain. The cap is the safety net before the worker tier flag-flip
+// in PR 10.
+//
+// How tracking works without coupling to handler internals: the
+// consumer sets `isWorkerDispatch = true` synchronously around the
+// `client.actions.InteractionCreate.handle(data)` call. The listener
+// in src/index.js calls `eventConsumer.trackDispatch(result)` on
+// every dispatch; trackDispatch checks the flag and only registers
+// the promise's settle handler when the dispatch is consumer-driven
+// (worker tier), NOT when it's gateway-WS-driven. The increment
+// runs inside the synchronous emit window; the decrement fires
+// whenever the dispatch promise settles, however much later.
 
 const { SQSClient, ReceiveMessageCommand, DeleteMessageCommand } = require('@aws-sdk/client-sqs');
 const config = require('./config');
@@ -220,17 +225,18 @@ function recordSeen(eventId) {
 // What stop() does NOT await: the async handlers registered on
 // `interactionCreate` (handleCommand / handleFlowInteraction) run
 // detached after the synchronous `client.actions.InteractionCreate.handle`
-// emit. Their promises aren't captured anywhere we can await. A
-// handler still mid-DDB-write when SIGTERM lands could race
-// gracefulShutdown's eventual `db.close()` — same race the
-// pre-existing in-process gateway path has, since the gateway also
-// emits + runs detached. Not worse here than today.
+// emit. The backpressure tracker (trackDispatch / inFlightCount)
+// CAPTURES those promises for bounding-the-cap purposes but stop()
+// doesn't await them — graceful-shutdown's 10s budget would be
+// blown by a long-running DM-fan-out handler. A handler still
+// mid-DDB-write when SIGTERM lands could race gracefulShutdown's
+// eventual `db.close()` — same race the pre-existing in-process
+// gateway path has, since the gateway also emits + runs detached.
 //
 // If a tighter "drain handlers before db close" guarantee is ever
-// needed, options are: (a) capture handler promises via a WeakMap
-// registry the listener populates and the consumer reads; (b) call
-// the dispatcher functions directly instead of going through the
-// emit path. Both are PR-10-or-later refactors.
+// needed, the trackDispatch registry already has the promises;
+// stop() could `await Promise.allSettled([...inFlightPromises])`
+// with its own deadline cap (e.g. 3s). Deferred until needed.
 //
 // receiveAbortController is set per-poll-iteration and used by stop()
 // to cancel an in-flight long-poll ReceiveMessage. Without this, a
@@ -244,6 +250,50 @@ let stopping = false;
 let loopPromise = null;
 let receiveAbortController = null;
 let sqsClient = null;
+
+// Backpressure cap (see module header). pollOnce early-returns
+// when inFlightCount >= MAX_INFLIGHT_HANDLERS and waits
+// INFLIGHT_BACKOFF_MS for the count to drain before re-polling.
+// Default of 100 sized for the realistic invite-only-beta worst
+// case (10-message batches × ~5-10 batches before slowest
+// handler completes). Operator can override via env when prod
+// volume + slowness data is in hand.
+const DEFAULT_MAX_INFLIGHT_HANDLERS = 100;
+const MAX_INFLIGHT_HANDLERS = (() => {
+  const raw = process.env.QURL_BOT_MAX_INFLIGHT_HANDLERS;
+  if (!raw) return DEFAULT_MAX_INFLIGHT_HANDLERS;
+  const parsed = parseInt(raw, 10);
+  if (!Number.isFinite(parsed) || parsed <= 0) {
+    // logger isn't loaded this early in module init; console.warn
+    // for visibility at boot, then fall back to the default.
+    console.warn(
+      `[event-consumer] QURL_BOT_MAX_INFLIGHT_HANDLERS=${JSON.stringify(raw)} rejected ` +
+      `(must be a positive integer); using default ${DEFAULT_MAX_INFLIGHT_HANDLERS}.`,
+    );
+    return DEFAULT_MAX_INFLIGHT_HANDLERS;
+  }
+  return parsed;
+})();
+
+// Brief pause when at-cap. Short enough that handlers draining
+// quickly resume normal polling; long enough that we don't burn
+// CPU re-checking the cap on every microtask. abortableSleep
+// honors `stopping` so a SIGTERM during a backpressure pause still
+// exits inside the graceful-shutdown budget.
+const INFLIGHT_BACKOFF_MS = 100;
+
+// In-flight handler count (incremented in trackDispatch when a
+// consumer-driven dispatch's promise is registered; decremented
+// when that promise settles). isWorkerDispatch is the per-emit
+// gate that distinguishes consumer-driven dispatches (which we
+// want to count) from gateway-WS-driven dispatches in combined
+// mode (which we don't — the gateway tier doesn't backpressure
+// against itself). The flag is set true synchronously around
+// `client.actions.InteractionCreate.handle(data)` in processMessage
+// and cleared in the finally; the listener's emit fires inside that
+// window because EventEmitter.emit is synchronous.
+let inFlightCount = 0;
+let isWorkerDispatch = false;
 
 // Hook for unit tests: lets the test inject a mock SQSClient
 // (aws-sdk-client-mock'd in tests/event-consumer.test.js) without
@@ -409,6 +459,13 @@ async function processMessage(client, message) {
   // The shared listener registered in src/index.js (gated on
   // isGateway || isWorker) handles routing.
   let dispatchOk = true;
+  // Flag for trackDispatch (called by the listener in src/index.js).
+  // EventEmitter.emit is synchronous, so the listener fires while
+  // this flag is still true; trackDispatch reads the flag and only
+  // registers consumer-driven dispatches in the backpressure cap.
+  // Cleared in the finally so a gateway-WS-driven dispatch landing
+  // after this returns is correctly untracked.
+  isWorkerDispatch = true;
   try {
     client.actions.InteractionCreate.handle(data);
   } catch (err) {
@@ -422,6 +479,8 @@ async function processMessage(client, message) {
       eventId,
       messageId: message.MessageId,
     });
+  } finally {
+    isWorkerDispatch = false;
   }
 
   if (isDup) {
@@ -447,7 +506,50 @@ async function processMessage(client, message) {
   await deleteMessage(message.ReceiptHandle);
 }
 
+// Called by the `interactionCreate` listener in src/index.js for
+// every dispatch. Reads the isWorkerDispatch flag set synchronously
+// around the consumer's `handle()` call to distinguish consumer-
+// driven dispatches (which count against the backpressure cap)
+// from gateway-WS-driven dispatches (which don't — gateway tier
+// doesn't backpressure against itself). Increment runs inside the
+// synchronous emit window; decrement registers on the dispatch
+// promise's `.finally()` so both success and rejection bring the
+// counter back down. A non-promise return (older handlers, or
+// nothing returned for unrouted interaction types) is a no-op.
+function trackDispatch(maybePromise) {
+  if (!isWorkerDispatch) return;
+  if (!maybePromise || typeof maybePromise.then !== 'function') return;
+  inFlightCount += 1;
+  // .finally on the original promise so we don't swallow rejections
+  // — the original chain still rejects, just with the decrement as a
+  // side effect. Wrap the decrement in a try/catch so a malformed
+  // promise can't leave the counter stuck at a high value.
+  maybePromise.finally(() => {
+    inFlightCount -= 1;
+  }).catch(() => {
+    // The decrement already happened in the finally callback; this
+    // catch is only here to absorb the rejection that flows through
+    // the chain so it doesn't surface as an unhandledRejection
+    // attributed to the consumer. The original .then() chain
+    // registered by the listener's caller still sees the rejection.
+  });
+}
+
 async function pollOnce(client) {
+  // Backpressure check: if too many handlers are in-flight, pause
+  // receives until they drain. Returns from this iteration without
+  // pulling messages; pollLoop will call us again, and the
+  // `while (!stopping)` check naturally exits if stop() lands during
+  // the pause. abortableSleep also honors stopping so the pause
+  // doesn't blow the graceful-shutdown budget.
+  if (inFlightCount >= MAX_INFLIGHT_HANDLERS) {
+    logger.debug('Event consumer: in-flight cap reached, pausing receive', {
+      inFlight: inFlightCount,
+      cap: MAX_INFLIGHT_HANDLERS,
+    });
+    await abortableSleep(INFLIGHT_BACKOFF_MS);
+    return;
+  }
   const queueUrl = config.QURL_BOT_EVENTS_QUEUE_URL;
   const cmd = new ReceiveMessageCommand({
     QueueUrl: queueUrl,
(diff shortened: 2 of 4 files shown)


justin-layerv:
Round-1 cr feedback applied (`010a694`):

### Fixed

**Item 1 (Critical) — handler-error visibility regression.** Confirmed your read: attaching `.finally()` in `trackDispatch` registers a handler with Node's unhandled-rejection bookkeeping, so the pre-PR fallback through `process.on('unhandledRejection', …)` in `src/index.js` is silenced. Added `logger.error('Event consumer: dispatch handler rejected', { error, stack })` inside the trailing `.catch` so failing handlers still produce a CloudWatch signal. Backpressure tracker should not silently mask handler bugs.

Added a regression test (`trackDispatch logs handler rejections (preserves error visibility)`) that pins the behavior so a future refactor that drops the log path fails loud.

**Item 3 (Suggestion) — module-load env capture.** Added a comment above `DEFAULT_MAX_INFLIGHT_HANDLERS` noting that `MAX_INFLIGHT_HANDLERS` is captured once at module load, so retuning via SSM/task-def requires a redeploy. Per-poll re-read would let a typo in the env-update path silently change behavior without a deploy-log audit trail.

### Deferred

**Item 2 (Medium) — smarter at-cap backoff.** Filed #390 with both options you suggested (longer base vs exponential decay) plus an optional log-throttle (transition-only). Honoring your explicit "not worth widening this PR" — the core safety net (cap + pause) is in this PR and the cadence tuning is independent.

**Item 4 (acknowledged) — `console.warn` at boot.** Agreed it's a pragmatic trade-off; logger isn't available at module init and the path only fires on invalid env. Leaving as-is.

**Nit a (acknowledged) — `INFLIGHT_BACKOFF_MS = 100` magic number.** Fine to leave per your read.

**Nit b (acknowledged) — listener routing helper.** No third bucket today; revisit when one lands.

### Verification

- 50 event-consumer tests pass (49 → 50 with the regression test)
- 1857 total tests pass (+1)
- `eslint --max-warnings 0` clean

justin-layerv:
Round-2 cr feedback applied (`f15eb71`):

### Fixed

**Invariant fragility — `await`-free window.** Added a tighter "DO NOT add `await` inside this try/finally" comment directly above `isWorkerDispatch = true` in `processMessage`. The future editor staring at the block now sees the constraint at the call site rather than 220 lines up in the module header.

**Debug log spam at sustained-at-cap.** Implemented the `lastMissingEventIdWarnMs`-style throttle you suggested. State machine:
- Transition into at-cap → `logger.warn(AT_CAP_PAUSE_WARN_MSG)` (one fire per streak)
- Steady-state at-cap polls → `logger.debug(AT_CAP_PAUSE_DEBUG_MSG)` (quiet)
- Transition back to below-cap → `logger.info(AT_CAP_RELEASED_INFO_MSG)` (pause-end signal so operators can pair pause-start with pause-end events in CloudWatch)

This subsumes the log-throttle sub-item I'd punted to #390 — the issue is now scoped down to just the backoff-cadence question (longer base / exponential decay), which is independent and stays deferred per your "not worth widening this PR."

**Log messages as module constants.** `AT_CAP_PAUSE_WARN_MSG` / `_DEBUG_MSG` / `_RELEASED_INFO_MSG` exported from `_test`. Tests assert against the constants so a wording drift in the code site fails the test instead of silently breaking CloudWatch greps.

**`parseInt` is lenient.** Switched to `Number()` + `Number.isInteger()` so `QURL_BOT_MAX_INFLIGHT_HANDLERS='100abc'` is now rejected at boot (logs the warn, falls back to default). The `.env.example` says "must be a positive integer" — code now enforces it.

**Unrouted-interaction comment.** Added a one-line note in the `index.js` listener confirming the absence of `trackDispatch` for unrouted types is intentional (no promise = nothing to track).

### Acknowledged

**`.finally().catch()` shifts rejection visibility — confirm alerting.** The new log key is `Event consumer: dispatch handler rejected`. I'll add a step to the pre-flag-flip work (PR 10) to verify CloudWatch alarms aren't grepping for `unhandledRejection` from the old fallback path. Filing this on the PR 10 task list rather than as a code change here — it's an alerting-infra check, not a code fix.

**`abortableSleep` call assertion.** Skipped per your "Not critical" — the side-effect (no receive, gate set, log fired) is what the existing test pins, and chaining one more spy on `abortableSleep` doesn't add coverage that matters.

**`_setWorkerDispatchingForTest` affordance.** No change — your read was that the comment explaining the spy-vs-direct-read mismatch was helpful. Kept.

### New tests

- `pollOnce only warns once per at-cap streak (debug for subsequent polls)` — pins the throttle gate
- `pollOnce logs cap-released info on transition back to below-cap` — pins the recovery path you noted wasn't covered, uses resolvable promises so the drain runs through `trackDispatch`'s real `.finally` decrement

### Verification

- 52 event-consumer tests pass (50 → 52)
- 1859 total tests pass (+2)
- `eslint --max-warnings 0` clean

justin-layerv:
Round-3 cr feedback applied (\`9c9a8e6\`):

### Fixed

**Soft-cap overshoot.** Added a paragraph to the module header explaining the transient overshoot — `pollOnce` gates BEFORE the next receive, but a batch already-received can drive `inFlightCount` to `cap + RECEIVE_MAX_MESSAGES - 1` before the next gate. So memory headroom should be sized assuming `cap + 10` simultaneous handler promises, not the precise cap value. Cap is a backpressure signal, not a hard semaphore.

**Test pollution clamp.** `inFlightCount = Math.max(0, inFlightCount - 1)` in the decrement. Production can't reach negative (every increment is paired with exactly one decrement on the same promise) so the clamp is purely defensive against a future test that registers a resolvable promise and forgets to await its drain before `_resetStateForTest`. Comment explains.

I went with the `Math.max` option rather than the `Set<Promise>` option you sketched — the Set would change `getInFlightCount` from a counter read to a `.size` call, which is fine but ripples into the test helpers. The clamp is the minimal fix for what's a low-priority defense.

**Comment accuracy.** Reworded the "logger isn't loaded this early in module init" comment — you're right, `logger` is `require`d at the top of the file and is available by the time the env-parse IIFE runs. The actual justification for `console.warn` is boot-time stdout visibility before logger transports attach, regardless of `LOG_LEVEL` or transport state.

**Env-var validation tests.** New describe block `MAX_INFLIGHT_HANDLERS env validation (module-load IIFE)` using `jest.isolateModules` to re-require the module with `process.env.QURL_BOT_MAX_INFLIGHT_HANDLERS` stubbed. Covers:
- Rejections: `'100abc'`, `'-5'`, `'0'`, `'1.5'`, `'Infinity'`, `'NaN'`, `'abc'`, whitespace — all fall back to 100 + emit a console.warn with `rejected` in it
- Acceptances: `'50'`, `'200'`, `'1'` — resolve to the parsed value with no warning
- Unset: resolves to 100 with no warning

12 new test cases (52 → 64).

**`.finally`-before-`.catch` inline note.** Added a one-line comment confirming `.finally` runs before `.catch` on rejection, so the decrement always lands first. The reader doesn't have to reason about chain order to trust the invariant.

**Nit — hoisted log payload.** Extracted `{ inFlight: inFlightCount, cap: MAX_INFLIGHT_HANDLERS }` into a local `capPayload` const, reused by both at-cap log calls.

**Unhandled-rejection asymmetry between consumer/gateway paths.** You're right that this is now asymmetric:
- Consumer-driven reject → `logger.error('Event consumer: dispatch handler rejected', ...)`, no unhandledRejection.
- Gateway-WS-driven reject (combined mode) → trackDispatch early-returns, unhandledRejection still fires.

This is the deliberate consequence of the flag-wrap design: the consumer needs to track its own dispatches to backpressure them, but tracking inherently attaches a `.finally` which absorbs the rejection signal. Two paths means two alerting hooks. I'll add a step to the pre-flag-flip work (PR 10) to wire the new error log line into whatever alerting path the gateway's unhandledRejection feeds, so the alert surface stays uniform across both tiers. Tracking on the PR 10 task list.

**`_setWorkerDispatchingForTest` affordance.** No change — confirmed kept.

### Verification

- 64 event-consumer tests pass (52 → 64, +12 env-validation cases)
- 1871 total tests pass (+12)
- `eslint --max-warnings 0` clean

justin-layerv:
Round-4 cr feedback applied (\`0addb70\`):

### Fixed

**Misleading promise-registry comment.** Reworded the "drain handlers before db close" comment — you're right, trackDispatch only retains the count, not the promise references. The comment now says "would have to additionally capture each promise in a Set" so the deferred future is accurate.

**Soft-floor documentation.** Added a paragraph to \`.env.example\` warning that values below \`RECEIVE_MAX_MESSAGES\` (10) have degenerate overshoot ratios — cap=1 can transiently reach 10 in-flight, a 10x overshoot. Recommend cap >= 10. Did NOT enforce a hard floor in the IIFE: keeping the validator's contract narrow ("positive integer") lets a future test or a debug-purposes setting of cap=1 work as written, and the .env comment is now the operator-facing source of truth on the practical floor.

**Static invariant test.** New test that reads \`event-consumer.js\` and scans the slice between \`isWorkerDispatch = true\` and the matching \`isWorkerDispatch = false\` for an \`await\` token, with line-comments stripped first to avoid false positives on documentation references. Approximate (no JS parse) but a future editor adding \`await logger.x()\` inside the wrap trips this test immediately. Per your "even an approximate test pins the contract better than a comment alone."

**Test helpers.** Two helpers at the top of the backpressure describe block:
- \`flushMicrotasks(n=3)\` — replaces the \`await new Promise((r) => setImmediate(r))\` pairs that were sprinkled through the tests. Default n=3 gives a one-tick margin so a future trackDispatch chain depth change doesn't require revisiting every test.
- \`withWorkerDispatch(fn)\` — replaces the 5-line flag-flip try/finally that six tests repeated. The intent is now one line per call site.

Refactored every backpressure test to use the helpers. Diff is net-negative on test LOC despite adding the helpers.

**Math.max comment.** Now also notes the \`jest --watch\` / hot-reload case — the clamp is the only thing keeping the count sane across re-requires when a stale promise resolves into the freshly-required module.

**Tilde drop.** \`.env.example\` now says "waits 100ms" not "~100ms" — the constant is exactly 100ms.

### Acknowledged

**Observability shape change — rejection log key change.** I'll add the alarm-key audit (verify CloudWatch covers both \`'Unhandled promise rejection'\` and \`'Event consumer: dispatch handler rejected'\`) to the PR 10 pre-flag-flip task list. Already mentioned this in round-1 and round-3 responses; tracking on the PR 10 todo as a hard prerequisite to flipping \`ENABLE_EVENT_SHIPPER\`.

**Re-emit \`process.emit('unhandledRejection')\` instead of logging.** Declined as a code change here — the \`unhandledRejection\` semantics imply the rejection isn't handled, and ours is (the \`.catch\` is intentional). Calling \`process.emit\` from inside a \`.catch\` would also confuse \`--unhandled-rejections=strict\` policy with a fake unhandled event. The two-log-keys-one-alarm pattern is the cleaner route; logged for PR 10 alerting work.

**\`AT_CAP_PAUSE_DEBUG_MSG\` wording redundancy with warn.** Left as-is — your read was "not blocking — alerting greps the constants, not the prose," and the constants are distinct enough to differentiate transition-warn vs steady-state-debug at parse time.

### Verification

- 65 event-consumer tests pass (64 → 65, +1 static invariant test)
- 1872 total tests pass (+1)
- \`eslint --max-warnings 0\` clean

justin-layerv:
Round-5 cr feedback applied (\`1b6168a\`):

### Fixed

**Operational — rejection log alerting (item 1).** Added `kind: 'unhandledRejection'` as a structured field on the trackDispatch catch log, and matched the gateway handler's `err?.message || String(err)` fallback shape for non-Error rejections. A single CloudWatch query filtering on `kind=unhandledRejection` now catches both worker-tier and gateway-tier rejections — operators don't need to maintain parallel alarm rules. The gateway handler in src/index.js will be updated in PR 10 to emit the same field; until then, dashboards rooted on the old log message text still catch gateway rejections, and the new tag is the path forward to a unified rule.

This is the structured-field-on-the-new-log option you suggested as preferable to a PR-body rollout note. PR description updated separately to call this out too.

**Robustness — atCapPauseLogged reset in stop() (item 2).** Added to the stop() finally block. A `start → at-cap → stop → start` round-trip now starts fresh; the next at-cap entry fires its transition warn cleanly. Production never reaches this state, but the symmetry with `running` / `stopping` / `receiveAbortController` resets is correct.

**Cap semantics comment (item 3).** One-liner at the check site noting that `>=` means cap is the first paused value (so cap=100 pauses AT 100 in-flight), steady-state ceiling is `(cap - 1) + RECEIVE_MAX_MESSAGES`, off-by-one is intentional and matches the .env.example math.

**Static invariant test robustness (item 4).** Added `// FLAG-WRAP-START` / `// FLAG-WRAP-END` sentinels bracketing the flag-wrap in processMessage. Updated the static invariant test to slice between sentinels (not the first `isWorkerDispatch = true` substring match) AND assert sentinel uniqueness — a future copy-paste of the block would now fail the uniqueness check, surfacing the duplication. Survives file reordering and additional `isWorkerDispatch` assignments elsewhere in the module.

**Defense-in-depth soft-floor warning (item 5).** Added a second console.warn in the env-parse IIFE when `parsed < RECEIVE_MAX_MESSAGES` (10). Accepts the value (an operator running cap=1 for rate-limit testing has a reason) but surfaces the effective ceiling (`parsed + RECEIVE_MAX_MESSAGES - 1`) so the booted behavior matches the operator's mental model. The `.env.example` doc and the runtime warning now agree.

Three new test cases:
- `accepts %p as %i without warning` — exact-RECEIVE_MAX_MESSAGES (10) case alongside 50/200; pins that 10 is the first non-warning value
- `accepts %p as %i but warns about degenerate overshoot ratio` — 1, 5, 9; pins the warn path

### Skipped

**Behavioral test for concurrent processMessage flag isolation.** Per your "low priority since the static guard covers the load-bearing constraint." A real concurrency test would be timing-dependent and flaky; the static-invariant test (with sentinels now) is the durable contract.

### Verification

- 68 event-consumer tests pass (65 → 68, +3 soft-floor warn cases)
- 1875 total tests pass (+3)
- \`eslint --max-warnings 0\` clean

justin-layerv:
Round-6 cr feedback applied (\`24bfcc4\`). Reviewer marked LGTM; all three observations addressed.

### Fixed

**capPayload snapshot semantics.** Added a comment above the `const capPayload = ...` declaration noting that the snapshot is shared across at-cap and release-info branches, so the release log's `inFlight` field reports the value that *triggered* the release (below cap) rather than the live count at log time. That's the right framing for pairing pause-start with pause-end events in CloudWatch, but worth pinning so a future reader doesn't \"fix\" the apparent inconsistency.

**Scientific-notation acceptance documented.** Updated `.env.example` to mention that `1e3` and similar are accepted because `Number('1e3') === 1000` satisfies `Number.isInteger`. No behavior change — just making the validator's contract explicit so an operator running a config audit sees the permissive surface upfront.

**Static-invariant block-comment stripping.** The grep now strips both `/* ... */` block comments and `//` line comments before scanning for \`await\`. The wrap doesn't have either today, but the defense-in-depth pass costs one line and a future doc-comment containing the word \"await\" can't false-positive. Template literals containing \`await\` still slip through (e.g., \`\\\`some \${await x}\\\`\`) — per your \"diminishing returns past comments\" framing, and a 25-line block is small enough that this would be caught in review.

### Acknowledged (no action)

**Complexity vs. scope (item 4).** Noted — the line count is heavy for a \"safety net before flag-flip,\" but PR 10 is the next PR in the sequence (the SQS producer + gateway service strip-down that flips \`ENABLE_EVENT_SHIPPER\` for prod). If PR 10 slips significantly, this PR carries the maintenance cost; that's the trade-off the design doc made and I'm not seeing it slip.

**Test surface observation (item 5).** Just a positive note — kept.

### Verification

- 1875 total tests pass (unchanged — the changes were comments + docs)
- 68 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

I think we're at the bottom of the cr backlog now. Sixth round of polish on what's fundamentally a one-page change — the design has held up across all six rounds without a substantive correctness pivot, which I take as the signal we're done. Will pull the trigger on merge once CI re-greens on this commit, unless the next round surfaces something new.

justin-layerv:
Round-7 cr feedback applied (\`4c5d419\`). Reviewer marked merge-ready; all suggested items addressed.

### Fixed

**`receiveAbortController` invariant tightening (possible issue 2).** Null the controller before \`abortableSleep\` at-cap. The invariant is now "non-null iff a receive is in flight" — a stop() landing during the backpressure pause no longer calls \`.abort()\` on a settled controller. Safe today, but the cleaner contract is what operators expect.

**\`flushMicrotasks(n=3)\` rationale (test suggestion 2).** Expanded the helper docstring to explicitly note "2 hops + 1 margin" — the trackDispatch chain is `.finally` (decrement) → `.catch` (error log or pass-through), so n=3 = chain depth + 1 slack tick. A future editor extending the chain doesn't have to re-derive the count.

### Tracked as separate issues

**\`stop()\` drains in-flight handlers (possible issue 1).** Filed #391 with full scope: add a \`Set<Promise>\` of in-flight references that \`trackDispatch\` populates, drain it via deadline-capped \`Promise.allSettled\` in stop(), preserve the fast-stop guarantee with a 3s ceiling. Marked low-priority — today's flag-off path runs handlers in-process the same way the gateway does, so the race already exists everywhere. After PR 10 flips \`ENABLE_EVENT_SHIPPER\`, the drain becomes more valuable.

**Combined-mode duplicate dispatch + PR 10 gate** (item 5). Pre-existing, not introduced here. Already in the PR 10 task list as a hard requirement before flag-flip.

**Gateway-side \`process.on('unhandledRejection', ...)\` emits matching \`kind\` field** (design item, comment promise). Also already in the PR 10 task list — the alerting-key consolidation is part of the pre-flag-flip prep.

### Skipped

**Math.max(0, ...) in production code (possible issue 3).** Already debated in round 3 — reviewer offered Math.max as the simpler of two options (vs Set<Promise>) and noted it's "Low priority — current tests don't hit it." Keeping. Reviewer's round-7 framing as "test scaffolding" is fair, but the alternative (forcing every test to await its resolvable promises before reset) trades runtime complexity for test-author discipline, and the clamp is one line.

**Combined-mode interleaving test (test suggestion 1).** Skipped — a real concurrency test would be timing-dependent and flaky, and the static-invariant test (with sentinels + uniqueness + block-comment stripping) covers the load-bearing constraint that makes the interleaving safe.

**Style/comment-density trim opportunities.** Reviewer noted the throttle-pattern preamble and \`withWorkerDispatch\` docstring as the most prunable if a future reviewer asks. Not requested here, leaving as-is.

### Verification

- 1875 total tests pass (unchanged — only invariant tightening + comment edits)
- 68 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

I think we've genuinely converged now. Seven rounds of cr, zero substantive design changes, three follow-up issues filed (#388 closed by this PR, #390 backoff cadence, #391 stop()-drains-handlers), and the reviewer marked merge-ready in round-7 verdict. Will pull the trigger on merge once CI re-greens unless round-8 surfaces something fundamentally new.

justin-layerv:
Round-8 cr feedback applied (\`a795638\`). Three small items all addressed.

### Fixed

**TODO(PR-10) marker in src/index.js (concern 1).** Added next to the global `process.on('unhandledRejection', ...)` handler — the field-consolidation dependency is now searchable via `git grep 'TODO(PR-10)'`. PR 10 will update the handler to emit `kind: 'unhandledRejection'` alongside the existing fields, matching what trackDispatch's `.catch` already emits.

**`cap + 10` vs `cap + 9` alignment (concern 2).** Reworded the module header to use `RECEIVE_MAX_MESSAGES` consistently (both for the strict ceiling AND the headroom recommendation), with a parenthetical noting `.env.example` rounds up by 1 for safety margin math. The discrepancy is now intentional and labeled as such.

**Settled-promise test (test coverage gap).** New test pinning that trackDispatch's increment-then-decrement contract holds when the registered promise is already settled at registration time (`Promise.resolve()` from an early short-circuit). The `.finally` callback runs on the microtask boundary regardless — synchronous return + resolved promise still drains the counter. Catches a future "perf optimization" that tries to skip registration for already-settled promises.

### Acknowledged (no action)

**Combined-mode race window in theory (concern 3).** Reviewer correctly identified the load-bearing assumption (Discord.js's WS path always yields at an `await` before `handle()`). No code path violates this today, and the FLAG-WRAP sentinels + static invariant test pin the constraint going forward. Nothing to add.

**Math.max(0, ...) alternative shape (concern 4).** Set<Promise>-based ceiling already in #391 (the stop()-drains-handlers follow-up) — the refactor is too heavy for this PR alone.

### Verification

- 1876 total tests pass (1875 → 1876, +1 settled-promise test)
- 69 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

Three follow-up issues filed (#388 closed by this PR, #390 backoff cadence, #391 stop()-drains-handlers). I think this is genuinely done — eight cr rounds, no design changes, three follow-ups tracked as real issues, and three consecutive reviews marking some form of LGTM. Will merge once CI re-greens on this commit unless round-9 surfaces something new.

justin-layerv:
Round-9 cr feedback applied (\`37f049e\`). Reviewer's "ship it" condition met.

### Fixed

**`kind: 'unhandledRejection'` on the gateway handler (concern 2 — the blocker per the verdict).** Added the field to `src/index.js`'s `process.on('unhandledRejection', ...)` handler. Both paths now agree on the tag — a single CloudWatch query filtering on `kind=unhandledRejection` finds gateway-WS-driven AND worker-tier rejections. No stale-alarm gap between this PR landing and PR 10; that follow-up just becomes "wire up the actual alarms," not "fix the field shape." Removed the TODO(PR-10) marker since the work is now done in-PR.

**Vestigial `return result;` removed (nit).** EventEmitter listeners' return values are discarded. Pre-conversion the explicit return was a short-circuit that did real work; post-conversion it's misleading to a future reader. The unrouted-branch's `return undefined` IS load-bearing (short-circuits before trackDispatch), so the asymmetry is intentional and commented.

**Sync-throw counter test (test addition 1).** Extended the existing "processMessage clears isWorkerDispatch even when handle() throws" test to also assert `inFlightCount === 0` after the throw. Pins that a sync throw from handle() doesn't leak a counter slot — catches a future refactor that moves the increment outside the listener.

### Tracked separately

**inFlightCount + backpressure metrics (concern 3).** Filed #392 for the metric work (in-flight gauge, at-cap counter, dispatch-rejection counter, optional duration histogram). Low priority until PR 10 ships real traffic through SQS.

**Counter leak with hung handlers (concern 1).** Reviewer flagged the failure mode but noted "handlers eventually settle on their own deadlines" — the mitigation is upstream timeout/abort in the handler bodies, not in trackDispatch. Adding a forced timeout in trackDispatch would mask handler bugs we'd rather see surface. Not filing — the existing logger.error path catches whatever does eventually settle.

### Skipped

**Two concurrent processMessage test (test addition 2).** A real concurrency test would be timing-dependent and flaky. The static-invariant test (sentinel-bracketed, block-comment-stripped, uniqueness-checked) covers the load-bearing flag-wrap constraint.

**Math.max guard alternative.** Reviewer says "fine as written." Resolve/reject tests pin exact pre/post counts, so a future double-decrement is still caught by other tests.

**Whitespace-rejection grouping (nit).** Cosmetic; the test asserts the rejection path correctly regardless of which sub-reason fires.

**receiveAbortController placement.** Reviewer says "current placement is fine."

### Verification

- 1876 total tests pass (unchanged — the index.js + listener edits don't have direct unit-test coverage; the existing sync-throw test gained one more assertion)
- 69 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

Convergence: nine cr rounds, design held without a substantive pivot, four follow-ups filed (#388 closed by this PR, #390 backoff cadence, #391 stop()-drains-handlers, #392 metrics), and the reviewer's last gating concern is now fixed in-PR. Going to merge once CI re-greens on this commit.

PR #394, opened 2026-05-15
feat(discord): toggle /qurl map slash command via MAP_COMMAND_ENABLED (default off)

justin-layerv:
## Summary

- New env-driven feature toggle `MAP_COMMAND_ENABLED` (strict `=== 'true'`, default off) that gates `/qurl map` end-to-end: slash registration, dispatcher, autocomplete, help text, and `SETUP_SUCCESS_MSG`.
- Stale-client safety net replies with `QURL_MAP_DISABLED_REPLY` when the flag is off so a cached client submission doesn't fall into `handleQurlMap` → Places.
- Autocomplete short-circuits before `searchPlaces` when the flag is off, protecting the operator's `GOOGLE_MAPS_API_KEY` quota from stale-client races.
- 4 new flag-off assertions in `commands-comprehensive.test.js` lock in the production-default behavior (the existing `qurl-file-map.test.js` covers the flag-on path via its test-only mock).
- `e2e/tests/discord-commands.smoke.test.ts` reads `process.env.MAP_COMMAND_ENABLED` so the expected subcommand set tracks the deployed toggle.

## Why

The sandbox `/qurl-bot-discord/GOOGLE_MAPS_API_KEY` SSM parameter shipped as the literal `"PLACEHOLDER"` value and was never seeded. Every `/qurl map` invocation hit Google Places, got `REQUEST_DENIED`, and surfaced a generic "Location lookup failed" reply. Same situation in prod (`886375649402` / `us-east-2`) — that parameter is also still `"PLACEHOLDER"`. Disabling the command end-to-end is safer than letting a misconfigured deploy keep regressing on senders.

## Companion PR

The terraform var + task-def wiring + e2e-smoke workflow resolve step lives in `layervai/qurl-integrations-infra` (companion PR). The bot reads the env at module load, so flipping `map_command_enabled` in the env's tfvars + applying terraform pushes a new task-def revision and ECS rolls fresh tasks that see the new value.

## Re-enable order (when you want /qurl map back)

1. Seed `/qurl-bot-discord/GOOGLE_MAPS_API_KEY` in SSM in both the sandbox (`730883236711`) and prod (`886375649402`) accounts with a real Google Maps Platform API key (Places API enabled, no HTTP-referrer restriction).
2. Flip `map_command_enabled = "true"` in the env's tfvars in the infra repo + apply.
3. The e2e-smoke workflow auto-resolves the flag from tfvars (no parallel workflow PR required after the companion infra PR lands).

## Test plan

- [x] `npx jest` — 1811/1811 unit tests pass (4 new flag-off assertions added); 1 pre-existing unrelated suite failure (`event-consumer.test.js`, missing `@aws-sdk/client-sqs` dep).
- [x] `npx eslint src/ tests/` — clean.
- [ ] Sandbox deploy → confirm `/qurl map` no longer appears in the slash-command autocomplete for users.
- [ ] Sandbox deploy → run e2e smoke; assert subcommand set is `['file', 'help', 'revoke', 'setup', 'status']` (no `map`).
- [ ] `/qurl help` ephemeral renders without `/qurl map` mentions.
- [ ] `/qurl setup` (admin) success message renders without `/qurl map` mention.

### DIFF
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -89,6 +89,15 @@ function missingEventShipperKeys(cfg) {
 // parameter ships with that literal sentinel value — the
 // remediation ("seed a real key") is identical to the
 // missing-key case.
+//
+// TODO(infra-sentinel-sync): the literal "PLACEHOLDER" is also
+// the seed value for `aws_ssm_parameter.bot` in
+// qurl-integrations-infra/qurl-bot-discord/terraform/main.tf
+// (search that repo for `value = "PLACEHOLDER"`). If infra ever
+// renames the sentinel (e.g., "REPLACE_ME"), update here in
+// lockstep — otherwise the boot check silently regresses to
+// "non-empty value passes" and the original incident class
+// returns. `git grep TODO(infra-sentinel-sync)` finds the marker.
 const GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL = 'PLACEHOLDER';
 function missingMapCommandKeys(cfg) {
   if (!cfg.MAP_COMMAND_ENABLED) return [];
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -84,6 +84,21 @@ function missingEventShipperKeys(cfg) {
   return cfg.QURL_BOT_EVENTS_QUEUE_URL ? [] : ['QURL_BOT_EVENTS_QUEUE_URL'];
 }
 
+// Mirrors missingEventShipperKeys: parsed cfg in, missing-keys
+// array out. PLACEHOLDER is treated as missing because the SSM
+// parameter ships with that literal sentinel value — the
+// remediation ("seed a real key") is identical to the
+// missing-key case.
+const GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL = 'PLACEHOLDER';
+function missingMapCommandKeys(cfg) {
+  if (!cfg.MAP_COMMAND_ENABLED) return [];
+  const key = cfg.GOOGLE_MAPS_API_KEY;
+  if (!key || key === GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL) {
+    return ['GOOGLE_MAPS_API_KEY'];
+  }
+  return [];
+}
+
 // Process-role parsing for the gateway/HTTP split. Lifted out of
 // index.js so the invalid-value path is testable without spawning a
 // child process — same shape as missingBootKeys above. See
@@ -126,6 +141,8 @@ module.exports = {
   missingProdKeys,
   missingKekRequiredKeys,
   missingEventShipperKeys,
+  missingMapCommandKeys,
+  GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL,
   VALID_PROCESS_ROLES,
   resolveProcessRole,
 };
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -84,11 +84,9 @@ function missingEventShipperKeys(cfg) {
   return cfg.QURL_BOT_EVENTS_QUEUE_URL ? [] : ['QURL_BOT_EVENTS_QUEUE_URL'];
 }
 
-// Mirrors missingEventShipperKeys: parsed cfg in, missing-keys
-// array out. PLACEHOLDER is treated as missing because the SSM
-// parameter ships with that literal sentinel value — the
-// remediation ("seed a real key") is identical to the
-// missing-key case.
+// PLACEHOLDER is treated as missing because the SSM parameter
+// ships with that literal sentinel value; remediation ("seed a
+// real key") is identical to the empty-key case.
 //
 // TODO(infra-sentinel-sync): the literal "PLACEHOLDER" is also
 // the seed value for `aws_ssm_parameter.bot` in
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2731,9 +2731,15 @@ const SETUP_API_KEY_MAX_LENGTH = 64;
 // User-visible success message on successful setup. Exported via
 // _test so test assertions read the production string instead of
 // hardcoding a copy that drifts.
+// Built at module load with the same MAP_COMMAND_ENABLED snapshot
+// the SlashCommandBuilder IIFE uses, so the help/setup copy and the
+// actual slash registration can't disagree about whether /qurl map
+// is reachable.
 const SETUP_SUCCESS_MSG =
   '✅ **qURL is now configured for this server!**\n\n'
-  + 'Your team can use `/qurl file` and `/qurl map` to share files and locations securely.\n'
+  + (config.MAP_COMMAND_ENABLED
+    ? 'Your team can use `/qurl file` and `/qurl map` to share files and locations securely.\n'
+    : 'Your team can use `/qurl file` to share files securely.\n')
   + 'All qURL usage will be billed to your API key.';
 
 // Button-stage handler. Routed by flow-dispatch when the admin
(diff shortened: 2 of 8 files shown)


justin-layerv:
Thanks for the second round! Addressed all five nits in 72c2b83:

1. **Section header copy** (`commands.js:7946`) — now flag-aware via `mapCopy.sectionVerb`. Reads "Share files" when /qurl map is off, "Share resources" when on.
2. **`applySharedSendOptions(sub)` helper** — explicitly out-of-scope per your note. Filed as #396 (`pr-polish-followup` + `enhancement` + `priority: low` + `javascript` labels) with acceptance criteria + smoke-coverage requirement so the next PR can pick it up cleanly.
3. **`QURL_MAP_DISABLED_REPLY` copy** (`commands.js:3192`) — now reads "Use `/qurl file` to share files securely" to mirror the SETUP_SUCCESS_MSG flag-off branch.
4. **Config snapshot-vs-runtime comment** (`config.js`) — added a single-paragraph note documenting that `MAP_COMMAND_ENABLED` is captured once at module load, baked into the slash registration + SETUP_SUCCESS_MSG, and that a runtime env flip is a no-op until task restart (matches the ECS deploy model).
5. **Gate-skip test strengthening** (`commands-comprehensive.test.js`) — replaced the negative `.not.toContain` loop with an exact `expect(allReplies).toEqual([_test.QURL_MAP_DISABLED_REPLY])`, catching the double-reply regression you called out.

All 1816 unit tests pass; lint clean. Pushed.

justin-layerv:
Round-3 follow-up in fa777fd:

1. **E2E smoke env lockstep** — the workflow `env:` plumbing is in the companion infra PR layervai/qurl-integrations-infra#612 (new "Resolve MAP_COMMAND_ENABLED from `<env>.tfvars`" step, mirrors the existing "Resolve BOT_CLIENT_ID" step). The infra PR is the gating dependency for the deploy; merging it first means the workflow `env:` block is in place before the bot ships the toggle. No code change needed on this PR.

2. **Hardcoded line reference** (`commands-comprehensive.test.js`) — fixed in fa777fd. Comment now references `API_KEY_GATED_SUBCOMMANDS` by symbol name instead of `commands.js:7882`.

3. **`missingMapCommandKeys` missing-vs-PLACEHOLDER collapse** — intentional. The remediation is identical in both cases ("seed a real key into SSM"), and the boot-fail error message already enumerates both states explicitly ("missing or still the literal 'PLACEHOLDER' sentinel"). The `boot-requirements.js` doc comment captures the same reasoning. If a future metrics use case wants per-cause counts, splitting at that point is straightforward.

4. **IIFE → `buildQurlCommandData(config)` helper** — explicit "optional refactor, not required for this PR." Filed as #397 (`pr-polish-followup` + `enhancement` + `priority: low` + `javascript`), with acceptance criteria pinning the subcommand set for both flag states. Natural pair with #396 (`applySharedSendOptions` extraction); both touch the same registration block, both improve unit-testability of the slash registration.

Ready to merge once the companion infra PR (#612) is in.

justin-layerv:
Round-4 follow-up in f184d71:

1. **IIFE extraction** → already filed as #397.
2. **Dispatcher flat-guard refactor** — keeping the nested form. You flagged the current shape as "also defensible," and the nested form keeps the flag-off and flag-on branches lexically adjacent (both inside the same `if (sub === 'map')`), which I find easier to scan when tracing dispatch. Pure judgment call either way.
3. **E2E smoke env plumbing** — the `Resolve MAP_COMMAND_ENABLED from <env>.tfvars` step is in the companion infra PR layervai/qurl-integrations-infra#612. Tfvars is the single source of truth: when an operator flips `map_command_enabled = "true"` and applies, both the bot's task-def env AND the smoke runner's expected-subcommand assertion track the same flag. Sequencing: merge infra PR first, then this one.
4. **Pin "Share files securely"** — done. The flag-off help test now asserts the new copy positively (`.toContain('Share files securely')`) AND that the flag-on copy (`'Share resources securely'`) is absent. Catches a regression that drops the `sectionVerb` swap specifically.
5. **`GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL` to `shared/sentinels.js`** — explicit "not today." Leaving in `boot-requirements.js` as the single consumer.
6. **IIFE addSubcommand uniformity** — applied in f184d71. The `revoke`/`help`/`setup`/`status` trailing subcommands now use the same imperative `builder.addSubcommand(...)` pattern as the conditional `map` block above. The initial `new SlashCommandBuilder().setName(...).setDescription(...).addSubcommand(file)` chain is kept as one fluent expression for the builder + name + description + first subcommand setup — splitting that loses the natural "this is the command and these are its top-level attrs" reading. Trailing subcommands are now uniformly imperative.

1816/1816 unit tests pass. Lint clean. Ready to merge once #612 is in.

justin-layerv:
Round-5 follow-up in 93d7077:

1. **PLACEHOLDER sentinel infra-sync** — applied option (a) per your weighing. Added `TODO(infra-sentinel-sync)` marker on `GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL` matching the existing `TODO(upstream-rebrand)` convention in `CLAUDE.md`. `git grep TODO(infra-sentinel-sync)` now finds both sides of the contract (the constant here + the `value = "PLACEHOLDER"` line in `qurl-integrations-infra`'s `aws_ssm_parameter.bot`). Skipped option (b) — a key-shape heuristic (`AIza` prefix + length floor) would be more aggressive but Google's docs explicitly note their key format is "subject to change" so the marker is the safer hedge.

2. **Overlapping autocomplete tests** — dropped the "responds empty for /qurl map location autocomplete" test entirely. Under `MAP_COMMAND_ENABLED=false`, its `respond([])` assertion was strictly weaker than the new flag-off block's "does NOT call searchPlaces" test (which asserts respond([]) AND mockSearchPlaces was never called). Replaced with a tombstone comment pointing at `qurl-file-map.test.js`'s "skips Places call for partial queries below the min-length cap" — that test runs under `MAP_COMMAND_ENABLED=true` and actually exercises the path the min-length gate guards.

3. **Help-copy phrasing inconsistency** — fixed. Flag-off `runLine` now uses the same parenthetical-instruction style ("Run `/qurl file` (attach a file)") as the flag-on branch. Flipping the toggle no longer reshapes the step-1 sentence style.

4. **`ephemeral: true` consistency** — no action per your note.

5. **`logger.debug` → metric** — deferred per your note. Will revisit if stale-client decay turns into a real SLO question.

1815/1815 unit tests pass (net -1 from the deleted redundant test). Lint clean. Ready to merge once #612 is in.

justin-layerv:
Round-6 follow-up in 612a2d1:

1. **IIFE → `buildQurlSlashCommand()` extraction** — same as round-3/4, already filed as #397.
2. **`mapCopy` parallel-structure vs template+inserts** — keeping the parallel structure. The current shape's drift-resistance ("an edit to one branch is obviously checked against the other") matters more than the line-count savings of a template approach; agree with you that it's a judgment call either way.
3. **Escaped apostrophes** — noted, won't sweep them as a separate commit per your note. If a future PR re-touches these lines, it'll clean them up.
4. **Smoke env plumbing checklist on the infra PR** — already in layervai/qurl-integrations-infra#612's PR description (Re-enable order step 3 + Summary bullet for the new "Resolve MAP_COMMAND_ENABLED" step). Both surfaces of the lockstep are covered.
5. **Stale-client TTL note "~1h"** — fixed in 612a2d1. Comment now reads "Cache TTL depends on registration mode: up to 1h on global (multi-tenant) registration, near-instant on guild-scoped (single-GUILD_ID)." Accurate for both deploy shapes the bot supports (sandbox + multi-tenant prod use global; OpenNHP-style single-guild uses guild-scoped).
6. **`handleAutocomplete` flag-off branch reach** — confirmed. The existing guard `if (subcommand !== 'map' || focused?.name !== 'location') return respond([])` runs BEFORE the new flag-gate, so any future autocomplete-enabled option on a different subcommand short-circuits before hitting the gate. The flag-gate only fires for the `map.location` case it's defending against.
7. **Test asserting `/qurl map` absent from registered options when flag is off** — same proposal as the #397 `buildQurlCommandData(config)` extraction (testing the registration shape directly requires the extractable helper). Captured in #397's acceptance criteria.

1815/1815 unit tests pass. Lint clean. Ready to merge once #612 lands.

justin-layerv:
Round-7 follow-up in 0b3c82e:

1. **Sentinel coupling / `AIza` heuristic** — explicit "current TODO + grep marker is a reasonable cost/benefit choice." Leaving as is.

2. **Metric on disabled-reply hits** — explicit "minor"; already deferred in round 5 per your earlier "fine to defer until/unless you actually need it."

3. **Smoke env-plumbing dependency surfacing** — applied. The smoke now:
   - Throws up front if `MAP_COMMAND_ENABLED` isn't exactly `"true"` or `"false"` (catches both unset env and typo-set values like `"True"` or `"1"`). The error names the workflow step + the local-dev workaround.
   - `console.log`s the resolved value so the smoke output shows what was actually asserted — useful breadcrumb when triaging a "tfvars flipped, deploy looks weird" sequence.
   
   The silent-default-to-false mode that the regression catalog from rounds 4-6 kept revisiting is now a hard error.

4. **IIFE → flat builder form** — same as the multi-round nit, filed as #397. "Style choice, not a defect" per this round.

5. **`interaction.user?.id` defensiveness** — keeping the `?.` for symmetry with the other interaction-property reads in this dispatcher. Cost is one runtime check per stale-client interaction; benefit is robustness if Discord's interaction shape ever varies.

**Min-length test migration verified**: `qurl-file-map.test.js:1468` (`'skips Places call for partial queries below the min-length cap'`) asserts both `respond([])` AND `mockSearchPlaces` not called, under the file's `MAP_COMMAND_ENABLED: true` mock. Coverage moved cleanly.

1815/1815 unit tests pass. Lint clean. Ready to merge once #612 is in.

justin-layerv:
Round-8 follow-up in 3330fa7:

1. **Rerun-hint copy still says /qurl map after flag flip** — I tried adding a comment-tag at the line (per your "flagging in case it's worth a comment-tag"), but a final `/simplify` pass correctly called it out as narrating a bounded edge with no constraint or remediation. Backed out the comment. The race is bounded by ECS rollover + the 3-min confirm-card TTL; if a future deploy model expands the race window, the bug surfaces as a tiny UX papercut, not a correctness issue.

2. **Asymmetric tolerance bot parser vs smoke** — intended contract per your "net effect is *more* signal, not less."

3. **IIFE mixed style** — keeping as-is per "the current shape reads fine."

4. **`QURL_MAP_DISABLED_REPLY` allocation** — keeping as-is per "mentioned only for completeness."

5. **No metric on autocomplete short-circuit** — keeping as-is per "probably overkill given the dispatcher-side log already captures the same population."

**Final `/simplify` pass on the converged state**: reuse + efficiency clean. Quality agent flagged 6 comment-trim opportunities — all applied in 3330fa7 (delete the unreachable-race comment-tag, delete the test-migration tombstone, trim the IIFE pre-comment + mapCopy doc-block + boot-requirements preamble, soften the smoke error message's workflow-step naming to be rename-resilient).

**Convergence**: DE acceptance ✓, CI passing ✓, every cr round addressed or filed (#396 + #397), `/simplify` clean across reuse/efficiency/quality. 1815/1815 unit tests pass. Ready to merge once the companion infra PR (layervai/qurl-integrations-infra#612) is in.

Issue #391, opened 2026-05-15
discord: stop() should optionally drain in-flight handler promises with deadline cap


justin-layerv:
## Context

PR #389 (in-flight handler backpressure) added \`inFlightCount\` tracking via \`trackDispatch\`, but \`stop()\` does NOT await the in-flight handler promises. The graceful-shutdown path therefore proceeds to \`db.close()\` while detached handlers may still be mid-DDB-write — same race the pre-existing in-process gateway path has, since the gateway also emits + runs detached.

Today's design retains only the **count** (not the promise references) in \`inFlightCount\`, so adding a drain wouldn't be free — it requires additionally tracking each promise in a \`Set\` (\`inFlightPromises\`) that \`trackDispatch\` adds to and the \`.finally\` deletes from.

## Scope

In \`apps/discord/src/event-consumer.js\`:

1. Add a module-level \`Set<Promise>\` (\`inFlightPromises\`) that \`trackDispatch\` adds to before the \`inFlightCount += 1\`.
2. In the \`.finally\` decrement, \`inFlightPromises.delete(maybePromise)\` alongside the count decrement.
3. In \`stop()\`'s finally block, after the \`receiveAbortController.abort()\` + \`await loopPromise\`, optionally \`await\` a deadline-capped \`Promise.allSettled\` over the tracked promises:

\`\`\`js
// Drain in-flight handlers with a deadline cap. The cap must stay
// well under gracefulShutdown's 10s budget so db.close() still has
// room to run.
const drainDeadlineMs = 3000;
await Promise.race([
  Promise.allSettled([...inFlightPromises]),
  new Promise((r) => setTimeout(r, drainDeadlineMs)),
]);
\`\`\`

4. Reset \`inFlightPromises.clear()\` alongside other state in both \`stop()\` and \`_resetStateForTest\`.

## Acceptance criteria

- [ ] \`stop()\` waits up to \`drainDeadlineMs\` for in-flight handlers to settle before resolving
- [ ] If handlers don't drain within the cap, \`stop()\` still returns cleanly (no hang)
- [ ] No behavior change when \`inFlightPromises\` is empty (the common case for an idle worker)
- [ ] Test: register N pending promises via \`trackDispatch\`, resolve them progressively, call \`stop()\`, assert all are awaited before resolution
- [ ] Test: register a never-resolving promise, call \`stop()\`, assert it returns within \`drainDeadlineMs\` (not blocked indefinitely)

## Trade-off

[...]

PR #395, opened 2026-05-15
feat(discord): stop() drains in-flight handlers with deadline cap (#391)

justin-layerv:
## Why

PR #389 added a backpressure cap that pauses receives when in-flight handlers exceed a limit. The DE review of that PR noted the cap could not help with graceful shutdown because \`trackDispatch\` only retained a *count*, not the promise references — \`stop()\` had nothing to await. A SIGTERM during a detached DDB write races \`gracefulShutdown\`'s \`db.close()\`, same as the pre-PR gateway path. PR #393 documented this as a known scope limitation pointing at this follow-up (#391).

This PR closes that gap.

## What

**Refactor \`inFlightCount\` → \`inFlightPromises\` (Set<Promise>).**
The cap check (\`pollOnce\`) reads \`.size\` instead of an int. trackDispatch \`Set.add\` on increment, \`Set.delete\` in \`.finally\` on decrement.

Knock-on simplification: the underflow guard from PR #393 is gone. \`Set.delete\` on a non-member is a no-op, so stale \`.finally\` callbacks (after \`_resetStateForTest\` or hot-reload) silently no-op rather than drifting size negative. The "test-scaffolding-in-production" smell the DE review flagged is structurally resolved.

**\`stop()\` new drain block.**
After \`await loopPromise\`, if \`inFlightPromises.size > 0\`:

\`\`\`js
await Promise.race([
  Promise.allSettled([...inFlightPromises]),
  new Promise((resolve) => {
    drainTimer = setTimeout(resolve, DRAIN_DEADLINE_MS);
  }),
]);
\`\`\`

Three terminal outcomes, each logged:
- All handlers settled in time → \`info\` *"drain complete"* with count + elapsedMs
- Deadline elapsed with stragglers → \`warn\` *"drain deadline elapsed, proceeding with handlers still in-flight"* with unsettled + settled + elapsedMs
- Set empty at \`stop()\` entry → no log fires (the idle-worker path; common case in production)

The \`clearTimeout\` runs in a \`finally\` so the timer doesn't keep the event loop alive past the drain.

**Deadline.**
\`DRAIN_DEADLINE_MS = 3000\` — well under \`gracefulShutdown\`'s 10s budget so \`db.close()\` has room even when the deadline fires. Mutable (not const) so tests can shrink it; production never mutates it.

## Test plan

- [x] \`npm test\` — 1881 passing (1877 → 1881, +4 net):
  - \`stop() drains in-flight handlers before resolving (settled within deadline)\` — uses \`setTimeout\` to defer resolution so the .finally microtasks haven't fired by the time the drain branch runs
  - \`stop() returns within deadline when handlers do not settle\` — never-resolving promise, 50ms deadline; assert warn log + bounded elapsed
  - \`stop() skips drain branch when no handlers are in-flight\` — pin the idle-worker path (no log noise on normal stops)
  - \`getDrainDeadlineMs reflects live value after _setDrainDeadlineForTest\` — pin the mutable-with-getter contract on the test affordance
  - The PR #393 underflow test was replaced with a "Set tolerates stale .finally" test since the underflow path is structurally unreachable now
- [x] \`eslint --max-warnings 0\` clean
- [ ] CI green
- [ ] cr review clean

## Rollback

Behavioral change: \`stop()\` now waits up to 3s longer when handlers are in-flight. If that latency causes incidents (e.g., shutdown-storm scenarios where many workers are stopping concurrently and the cumulative 3s waits exceed budget), the rollback path is either:
- Revert this PR (clean — no downstream PRs depend on the Set yet)
- Shrink \`DRAIN_DEADLINE_MS\` via a quick PR (mutable, one-line change)

The Set refactor is also reversible — \`inFlightCount\` was a thin wrapper; \`getInFlightCount()\` test helper still works (reads \`.size\`).

Deployment topology is single-process-per-worker (one task = one Node process = one consumer), so the \`gracefulShutdown\` 10s budget caps total shutdown time per worker independently — concurrent worker stops don't cumulate.

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -116,6 +116,19 @@ ENABLE_OPENNHP_FEATURES=false
 # simultaneous handler promises regardless.
 # QURL_BOT_MAX_INFLIGHT_HANDLERS=100
 
+# Maximum time (ms) stop() waits for in-flight handler promises to
+# settle during graceful shutdown before proceeding to db.close().
+# Default 3000. Range [100, 8000] — clamped so a typo can't block
+# stop() past gracefulShutdown's 10s budget. Out-of-range values
+# fall back to the default with a console warning.
+#
+# Tune up if prod soak surfaces handlers consistently exceeding 3s
+# (DDB latency spikes, slow REST round-trips) and you'd rather wait
+# than race db.close(). Tune down if shutdown latency matters more
+# than completing in-flight work. Captured at module load; redeploy
+# to retune.
+# QURL_BOT_DRAIN_DEADLINE_MS=3000
+
 # GitHub OAuth App — REQUIRED only when ENABLE_OPENNHP_FEATURES=true.
 # Single-guild-plain and multi-tenant deployments don't mount /auth or
 # /webhook routes and don't need these values — the boot check
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -968,10 +968,12 @@ async function stop() {
     // force-exit fires regardless, so we cap our wait at 3s to
     // leave room for db.close() and any other shutdown work.
     //
-    // Snapshot the set into an array before allSettled: pending
-    // .finally callbacks mutate inFlightPromises during the await,
-    // and Set iteration is fine with concurrent mutation but the
-    // explicit copy makes intent obvious.
+    // Explicit copy `[...inFlightPromises]` rather than passing the
+    // Set directly: Promise.allSettled materializes its input
+    // synchronously at call time, so a Set arg would behave the same.
+    // The spread is for readability — "snapshot of the set as of this
+    // moment, then await every one of those" reads more directly than
+    // letting allSettled iterate the live set.
     if (inFlightPromises.size > 0) {
       const drainCount = inFlightPromises.size;
       const drainStart = Date.now();
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -937,6 +937,15 @@ async function stop() {
   // pass !running, both abort, both await loopPromise. Harmless but
   // racy on the running=false reset; the stopping check makes the
   // single-stopper invariant explicit.
+  //
+  // Caveat for a hypothetical second caller: the early-return DOES
+  // NOT wait for the first caller's drain to complete. A second
+  // stop() that lands during drain returns immediately. In practice
+  // gracefulShutdown only calls stop() once, so this is theoretical
+  // — but if a future refactor opens the door to concurrent stops,
+  // the second caller would need to await the first's loopPromise
+  // (or a shared "drainPromise" handle) to get "drain complete"
+  // semantics, not just "stop is already in progress."
   if (!running || stopping) return;
   stopping = true;
   // Cancel the in-flight long-poll so stop() returns within tens of
@@ -957,15 +966,27 @@ async function stop() {
   }
   logger.info('Event consumer: stopping');
   try {
-    await loopPromise;
-    // After loopPromise resolves, every processMessage's
-    // synchronous-emit window has completed, so trackDispatch has
-    // captured all in-flight handler promises that the worker
-    // dispatched. Wait up to DRAIN_DEADLINE_MS for those detached
-    // handlers (DDB writes, REST calls, DM fan-out) to settle so
-    // the subsequent db.close() in gracefulShutdown doesn't race
-    // them. The deadline is non-negotiable — index.js's 10s
-    // force-exit fires regardless, so we cap our wait at 3s to
+    // Isolate the loopPromise wait from the drain: if pollLoop
+    // rejects (rare — its own catch covers routine errors, so this
+    // is an unforeseen sync throw or similar), we still want to
+    // drain. trackDispatch's set is maintained synchronously around
+    // emits, so its state after a loop crash still accurately
+    // reflects in-flight handler promises — draining is meaningful
+    // even if the loop itself died.
+    try {
+      await loopPromise;
+    } catch (err) {
+      logger.error('Event consumer: error during stop', { error: err.message });
+    }
+
+    // After loopPromise settles (success or failure), every
+    // processMessage's synchronous-emit window has completed, so
+    // trackDispatch has captured all in-flight handler promises
+    // the worker dispatched. Wait up to DRAIN_DEADLINE_MS for those
+    // detached handlers (DDB writes, REST calls, DM fan-out) to
+    // settle so the subsequent db.close() in gracefulShutdown
+    // doesn't race them. The deadline is non-negotiable — index.js's
+    // 10s force-exit fires regardless, so we cap our wait at 3s to
     // leave room for db.close() and any other shutdown work.
     //
     // Spread for readability — passing the Set directly would behave
@@ -973,7 +994,10 @@ async function stop() {
     // reads more directly as "await each of these promises."
     if (inFlightPromises.size > 0) {
       const drainCount = inFlightPromises.size;
-      const drainStart = Date.now();
+      // hrtime.bigint() is monotonic — Date.now() is wall-clock and
+      // can jump under NTP adjustments mid-drain. Cost is the same;
+      // monotonic is the correct shape for elapsed measurement.
+      const drainStartNs = process.hrtime.bigint();
       logger.info('Event consumer: draining in-flight handlers', {
         count: drainCount,
         deadlineMs: DRAIN_DEADLINE_MS,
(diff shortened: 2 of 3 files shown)


justin-layerv:
Round-1 cr feedback applied (\`6b28f9a\`). Both actionable nits addressed.

### Fixed

**Snapshot-comment rewording (observation 1).** You're right — \`Promise.allSettled\` materializes its input synchronously at call time, so a Set arg and an array spread behave identically with respect to concurrent mutation. The spread is a *readability* choice ("snapshot of the set as of this moment, then await every one of those"), not a mutation guard. Reworded.

**\`getDrainDeadlineMs\` justification (observation 3).** Added a one-liner test \`getDrainDeadlineMs reflects live value after _setDrainDeadlineForTest\` that pins the getter-vs-value-snapshot contract. A future refactor back to \`DRAIN_DEADLINE_MS\` as a value export (the original shape) would trip the test. The export is now load-bearing for the test rather than dead code.

### Acknowledged (no action)

**Post-deadline residual \`.finally\` callbacks (observation 2).** Confirmed — handlers that exceed the deadline retain their \`.finally\` → \`inFlightPromises.delete(...)\` callbacks, which fire after \`stop()\` returns and tick the set down asynchronously. Production never reads \`inFlightPromises.size\` after \`stop()\` returns, so the post-shutdown size drift is benign. The PR description's "race window meaningfully reduced, not eliminated" framing covers this; no doc change needed.

**Timing-tolerance in the deadline test (observation 4).** \`expect(elapsed).toBeLessThan(50 * 20)\` is the 20× upper bound for CI host scheduling jitter. Flagged as informational; keeping as written.

### Verification

- 1881 total tests pass (1880 → 1881, +1 getter test)
- 74 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

justin-layerv:
Round-2 cr feedback applied (\`8cc7cb3\`). Two nits fixed, two questions answered.

### Fixed

**\`withWorkerDispatch\` deduplication (nit 1).** Hoisted to module scope, removed both inner definitions (was \`withWorkerDispatchHelper\` in the lifecycle describe + \`withWorkerDispatch\` in the backpressure describe — byte-identical). Net: -12 lines of dupe, single shared helper.

**Spread comment trim (nit 4).** Cut from 6 lines to 3 — reviewer noted the long-form explanation was doing more work than the code itself. Now: "Spread for readability — passing the Set directly would behave identically (allSettled materializes synchronously) but \`[...]\` reads more directly as 'await each of these promises.'"

**PR description test count drift (nit 2).** Updated the test plan to list 4 new tests (was 3) including the \`getDrainDeadlineMs\` getter test.

**Slack in deadline-elapsed assertion (nit 3).** Keeping the 20× upper bound (\`< 50 * 20 = 1000ms\`). The cost of CI flake is recurring noise across many PRs; the cost of a "deadline accidentally multiplied" regression would be obvious in any real shutdown (the bot would block past gracefulShutdown's 10s force-exit and operators would notice immediately). Net: looser bound is the right tradeoff.

### Questions answered

**Logger transport survival across drain deadline.** The discord bot's \`logger\` (\`src/logger.js\`) is a pino instance writing to stdout via the default stream. There's no async transport that gets torn down — stdout stays writable until \`process.exit\`. Handler rejections that fire after the deadline still produce \`logger.error('Event consumer: dispatch handler rejected', ...)\` and land in CloudWatch normally, even after \`stop()\` returns. The reviewer's "is that a concern" is answered: no, the logger survives until exit. (Not adding a code comment since this isn't a regression risk specific to this PR — it's a property of the logger setup.)

**Drain budget vs cumulative shutdown-storm.** Deployment topology confirmed single-process-per-worker (one ECS task = one Node process = one consumer instance). \`gracefulShutdown\`'s 10s budget is per-process, so concurrent worker stops don't cumulate — each worker independently caps its own shutdown at 10s. Many-workers-in-one-process scenarios don't exist in our deployment. Updated the rollback section of the PR description to note this.

### Verification

- 1881 total tests pass (unchanged — the dedupe + comment trim don't change test count)
- 74 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

justin-layerv:
Round-3 cr feedback applied (\`1858556\`). The substantive item + all three nits addressed.

### Fixed

**Drain runs even when loopPromise rejects (substantive concern 1).** You were right: the drain has a real job to do, and the trackDispatch set's state is coherent after a loop crash (additions are synchronous around the listener emit, so the set accurately reflects in-flight handlers regardless of why the loop exited). Restructured stop() with nested try/catch — the inner try wraps just \`await loopPromise\`, the outer try-finally keeps state-cleanup atomic, and the drain block sits between them so a loop rejection logs but still drains.

New test \`stop() drains even when loopPromise rejects\` pins the contract: inject a rejected loopPromise via the new \`_setLoopPromiseForTest\` helper, assert both the error log AND the drain-complete log fire.

**\`.finally\` ordering subtlety comment (clarification 2).** Added a one-liner near the post-race size check explaining why \`.size\` is the discriminator (not the race winner): \`.finally\` callbacks registered at trackDispatch time run before \`Promise.allSettled\`'s internal continuation, so on the happy path the set is already empty by the time we read \`.size\`. Saves a future reader from re-deriving the FIFO callback ordering.

**\`process.hrtime.bigint()\` for elapsed (nit 3).** Switched from \`Date.now()\` — monotonic, immune to NTP jumps mid-drain. Same cost. Three-line change: declare \`drainStartNs\`, compute \`elapsedMs = Number((process.hrtime.bigint() - drainStartNs) / 1_000_000n)\`, log it.

**Concurrent stop() caller comment (nit 5).** Added a one-line note in the existing stop() idempotency-guard comment block: a second \`stop()\` during drain returns immediately and does NOT await the first caller's drain. Theoretical today (gracefulShutdown only calls stop() once); flagged for any future refactor that opens that door.

### Skipped

**Spread is unnecessary (nit 2).** Already trimmed in round-2; reviewer's round-3 framing as "stylistic preference, not correctness/perf" was acknowledgement that the current trimmed comment is fine.

**Deadline-path test margin (nit 4).** Reviewer's "acceptable as written" — keeping the 20× upper bound. The cost of catching an 800ms-deadline-regression accidentally is much smaller than the cost of CI flake noise.

### Verification

- 1882 total tests pass (1881 → 1882, +1 loop-crash drain test)
- 75 event-consumer tests pass
- \`eslint --max-warnings 0\` clean

The restructured stop() is the real semantic improvement here — the drain mechanism's whole purpose is to reduce the SIGTERM-vs-db.close() race window, and that win was being silently forfeited on any loop crash. Now the drain is independent of loop outcome.

justin-layerv:
Round-4 cr feedback applied (\`1458d1a\`). The substantive item (env override) implemented + all three nits.

### Fixed

**\`QURL_BOT_DRAIN_DEADLINE_MS\` env override (substantive concern 2).** Added mirroring the \`QURL_BOT_MAX_INFLIGHT_HANDLERS\` pattern. Sanity-clamped to **[100, 8000] ms**:
- Upper bound (8000) preserves \`gracefulShutdown\`'s 10s budget headroom — 8s drain + 2s for db.close + Discord teardown is the operator contract
- Lower bound (100) prevents accidental drain-disabling typos (e.g. \`QURL_BOT_DRAIN_DEADLINE_MS=1\`)
- Invalid (\`'100abc'\`, \`'-5'\`, etc.) + out-of-range values fall back to 3000ms default with a console.warn

Operational value: prod soak revealing handlers consistently exceeding 3s can be addressed with an env retune instead of a code change. Asymmetry with \`QURL_BOT_MAX_INFLIGHT_HANDLERS\` was the reviewer's call-out.

\`.env.example\` documents the new var + the range. 12 new test cases in a \`DRAIN_DEADLINE_MS env validation\` describe block cover rejection, range clamping, in-range acceptance, and unset-fallback.

**\`_setLoopPromiseForTest\` orphan caveat (concern 1).** Added a comment block on the helper flagging that overwriting the promise reference doesn't halt the real pollLoop. Tests need to either rely on stopping+abort retiring it, or accept _resetStateForTest as the cleanup point. The bigger refactor (\`_setPollLoopFnForTest\` injection) was the reviewer's alternative — declined for now since the comment is enough disclosure and the suite hasn't surfaced an actual leak.

**Trimmed dense comments (nits).**
- Spread comment: 3 lines → 1 line (\`// Spread is cosmetic — allSettled also accepts a Set directly.\`)
- Concurrent-stop caveat: 8 lines → 2 lines (kept the "second caller during drain doesn't await" point, dropped the speculative refactor sketch)

### Skipped

**\`TODO(#xxx)\` for concurrent stop() (concern 3).** Reviewer said "otherwise the prose is fine." No plausible future caller in the pipeline, so no issue to reference. The prose now succinctly notes the limitation.

**\`golangci-lint\` (concern 4).** Confirmed N/A for this PR (no Go touched).

### Verification

- 1894 total tests pass (1882 → 1894, +12 env-validation cases)
- 75 event-consumer drain/lifecycle tests pass (unchanged — env tests are in a separate describe)
- \`eslint --max-warnings 0\` clean

Issue #381, opened 2026-05-15
PR 10 pre-merge: update listener gate + wire end-to-end latency observability


justin-layerv:
Tracking two items surfaced in [PR #379](https://github.com/layervai/qurl-integrations/pull/379)'s cr that are forward-looking to PR 10 (gateway-tier strip-down + SQS producer). Both must land WITH PR 10, not after.

## 1. Listener-gate update — combined-mode double-dispatch hazard

PR 11 (#379) registers the `interactionCreate` listener for `isGateway || isWorker`. That is correct in isolation: with no producer, the worker's SQS queue stays empty, the consumer is a no-op.

The moment PR 10 ships, the gateway role starts publishing to SQS. In **combined mode + `ENABLE_EVENT_SHIPPER=true`**, both paths are now live in the same process:
- Gateway WS fires `interactionCreate` directly (in-process dispatch)
- The producer publishes to SQS, the consumer pulls + reconstructs + emits `interactionCreate` AGAIN

Every interaction would dispatch twice — silently doubling DM fan-outs, side-effecting every flow-state transition twice (one OCC win + one CCFEx + a confusing error log), corrupting telemetry counts.

**Required change in PR 10 (same commit):** update `apps/discord/src/index.js` line ~290 from

```js
if (isGateway || isWorker) { client.on('interactionCreate', ...) }
```

to either

```js
if ((isGateway && !config.ENABLE_EVENT_SHIPPER) || isWorker) { client.on('interactionCreate', ...) }
```

OR replace the gateway-side listener with a publish-to-SQS shim that lives in `event-publisher.js` (the producer counterpart to `event-consumer.js`).

The comment in `index.js:281-303` post-#379 already names this as a "PR 10 PRE-MERGE REQUIREMENT" — this issue is the tracking anchor.

## 2. End-to-end latency observability target

Discord's initial-ACK budget is 3 seconds from the user click to `deferReply()`. Today (in-process) the bot consumes ~50-100 ms of that on its own work; the remaining ~2.9 s is Discord's network + the user's TLS.

Once PR 10 ships, the path becomes:

```
user click
  → Discord gateway WS frame to bot
  → bot publishes to SQS (gateway tier)
  → SQS visibility timeout window opens
  → worker tier ReceiveMessage long-poll picks it up
  → JSON parse + InteractionCreate.handle reconstruct
  → handleCommand/handleFlowInteraction
  → REST call to /interactions/<id>/callback (deferReply)
```

[...]

PR #398, opened 2026-05-15
feat(discord): gateway publisher + worker-tier wiring (PR 10, zero-downtime Pillar 1)

justin-layerv:
## Summary

Closes the producer side of the zero-downtime event-shipper split. PR 11 (consumer, merged) has been polling an empty queue; this PR puts messages on it.

With `ENABLE_EVENT_SHIPPER=true` and the supported two-process shape (`PROCESS_ROLE=gateway` + `PROCESS_ROLE=http`), interactions now flow:

```
gateway WS frame
  → client.on('raw', publisher.publish)   [filters INTERACTION_CREATE only]
  → SendMessage to SQS Standard
  → consumer poll (http tier)
  → client.actions.InteractionCreate.handle(data)
  → local interactionCreate listener → handleCommand / handleFlowInteraction
```

With the flag unset (today's prod + sandbox), this PR is a no-op at runtime — the `raw` publish hook isn't registered, `publisher.start()` isn't called, listener gate falls through to the legacy in-process path.

## Why this matters

The gateway tier is a singleton (Discord allows one Gateway WS per token). Today, every gateway deploy causes a 30–90s outage: in-flight `/qurl` flows die mid-button-click, interactions get "interaction failed". The event-shipper split moves all business logic to a horizontally-scalable HTTP tier behind an ALB, leaving the gateway as a thin SQS forwarder that needs to be redeployed only on Node / discord.js / sharding-config changes. This PR is the producer half of that split; PR 13–14 add cross-process Gateway RESUME + hot-standby for the gateway tier itself.

## What's in the PR

**New module** — `src/event-publisher.js`:
- Process-singleton mirroring `event-consumer`'s lifecycle (`start`, `stop`, lazy SQS client).
- `publish(packet)` filters `op === 0 && t === 'INTERACTION_CREATE'` before any allocation (hot path runs on every gateway frame including HEARTBEAT_ACK / PRESENCE_UPDATE).
- Builds envelope `{ eventType, shardId, data, event_id, published_at_ms }` matching the consumer's contract.
- Fires `SendMessage` detached; tracks the promise in a `Set` for `stop()` drain (same shape as `event-consumer.stop()`).
- Send failures log `kind: 'unhandledRejection'` for unified CloudWatch alarming across publisher + consumer + global handler.

**Boot rejection** — `unsupportedRoleShipperCombo` in `src/boot-requirements.js`:
- `PROCESS_ROLE=combined` + `ENABLE_EVENT_SHIPPER=true` now boot-fails with operator-facing remediation.
- In combined mode both the publish hook AND the worker consumer would arm in one process → every interaction would land twice (in-process WS frame + SQS round-trip). The listener-gate fix from #381 alone doesn't close this — `discord.js`'s `InteractionCreate.handle` fires on the WS frame regardless.

**Listener-gate predicate** — `shouldRegisterInteractionListener` in `src/boot-requirements.js`:
- Extracted from index.js's inline `(isGateway && !flag) || isWorker` expression.
- Pure helper, unit-tested across every role × flag permutation.

**e2e latency metric** — `event-consumer.js`:
- Parses `published_at_ms` from the envelope.
- Emits `qurl_bot_event_e2e_ms` as a structured log field on successful dispatch — wall-clock delta from gateway-host envelope-build to worker-host receive. NTP drift between hosts is the noise floor; the metric is the SLI signal the design doc reserves under "Flow continuity."

**DE-review cleanups landed in the same PR** (no deferred follow-ups):
- `config.intEnv` extended with `strictInteger` + `min`/`max` options. `QURL_BOT_MAX_INFLIGHT_HANDLERS` and `QURL_BOT_DRAIN_DEADLINE_MS` now flow through one validation path; both modules share one knob.
- `GATEWAY_DISPATCH_TYPES.INTERACTION_CREATE` and `LOG_KINDS.UNHANDLED_REJECTION` lifted to `constants.js`. The three rejection-tagging sites (publisher, consumer, index.js global handler) share one frozen constant — drift between them was the silent-failure footgun.

## Test plan

- [x] **1945 tests pass across 51 suites.** Net new: publisher unit tests (19), `config.intEnv` unit tests (24), producer→consumer integration test (4), listener-gate predicate truth-table (6), `published_at_ms` parsing (3 + 2 negative-path), boot rejection (2). Net: +18 tests vs. main.
- [x] **Integration test pins the wire contract** — `tests/event-producer-consumer-integration.test.js` runs the publisher's `SendMessage` body through the consumer's `processMessage` via an in-memory SQS substitute. Envelope field-name drift on either side fails the test.
- [x] **Module-load smoke** — `node -e "require('./src/event-publisher'); require('./src/boot-requirements');"` boots cleanly.
- [x] **Flag-off path unchanged** — listener-gate truth table confirms `(combined, false)` and `(gateway, false)` both register the legacy listener; current sandbox + prod have `ENABLE_EVENT_SHIPPER` unset and see no behavior change.
- [ ] **Sandbox soak** — to be performed post-merge after first deploy with `ENABLE_EVENT_SHIPPER=true` on the split task defs. Will report `qurl_bot_event_e2e_ms` p99 in a comment on this PR before the flag-cleanup PR (which removes the legacy in-process gateway dispatch path) opens.

## Out of scope

- `@discordjs/ws` migration — PR 13 (gateway-tier no longer needs `discord.js`'s slash-command / cache layer).
- Gateway-tier `qurl_bot_gateway_session` DDB table for cross-process RESUME — PR 12.
- Hot-standby + leader-election — PR 13–14.
- Envelope-level schema validation + signature verification — IAM is the trust boundary today (queue policy locks `sqs:SendMessage` to the gateway task role). Tracked at `TODO(infra-publisher-set)` in `event-consumer.js`; revisit if the publisher set ever loosens.

## Rollback

Flag-flip: set `ENABLE_EVENT_SHIPPER=false` in both task definitions. The legacy in-process gateway path runs unchanged. This works until the (separate) follow-up PR removes the listener-gate scaffolding; that PR's description will explicitly call out the soak window before deleting the fallback.

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -99,10 +99,16 @@ ENABLE_OPENNHP_FEATURES=false
 # publishes) + PROCESS_ROLE=http (consumes, can scale horizontally). For
 # local dev / sandbox in one process, leave ENABLE_EVENT_SHIPPER unset.
 #
-# Rollback cliff: rolling back from the split shape requires switching
-# both task definitions to PROCESS_ROLE=combined + ENABLE_EVENT_SHIPPER
-# unset (or `git revert` + emergency redeploy). See
-# apps/discord/docs/zero-downtime-design.md for the full cutover sequence.
+# Rollback cliff: while PR 10's gated phase is in effect, flipping
+# `ENABLE_EVENT_SHIPPER=false` on the split task definitions reverts
+# to the legacy in-process gateway dispatch path — a clean flag-flip
+# rollback. After the follow-up cleanup PR removes the
+# `if (config.ENABLE_EVENT_SHIPPER)` raw-hook gate in src/index.js
+# and makes the producer unconditional, the flag-flip rollback is
+# gone and the only remaining path is `git revert` + emergency
+# redeploy. The cleanup PR will explicitly document the soak window
+# before that point. See apps/discord/docs/zero-downtime-design.md
+# (Rollback section) for the full cutover sequence.
 # ENABLE_EVENT_SHIPPER=false
 # QURL_BOT_EVENTS_QUEUE_URL=https://sqs.us-east-2.amazonaws.com/000000000000/qurl-bot-events-sandbox
 
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -80,20 +80,29 @@ ENABLE_OPENNHP_FEATURES=false
 # DDB_TABLE_PREFIX=qurl-bot-discord-sandbox
 
 # Event shipper (zero-downtime design, Pillar 1). When ENABLE_EVENT_SHIPPER
-# is set to the literal string "true", the gateway tier forwards every
-# Discord dispatch to SQS instead of running handlers in-process, and the
-# worker tier (PROCESS_ROLE=http or combined) polls SQS and dispatches via
-# the same handler path. When unset or any other value, the bot runs the
-# legacy in-process shape.
+# is set to the literal string "true", the gateway tier (PROCESS_ROLE=gateway)
+# forwards every Discord INTERACTION_CREATE dispatch to SQS via a raw-frame
+# listener, and the worker tier (PROCESS_ROLE=http) polls SQS and dispatches
+# via client.actions.InteractionCreate.handle — same routing as a real
+# gateway-WS-driven interaction. When unset or any other value, the bot
+# runs the legacy in-process shape (gateway WS frame → local listener).
 #
 # QURL_BOT_EVENTS_QUEUE_URL is the SQS Standard queue provisioned by the
 # qurl-integrations-infra `qurl-bot-events` module. Required when
 # ENABLE_EVENT_SHIPPER=true; the boot check refuses to start without it.
 #
-# Rollback cliff: ENABLE_EVENT_SHIPPER is valid only through the gateway
-# strip-down (PR 10). The follow-up that removes the in-process fallback
-# also removes this flag; after that, rollback is `git revert` + emergency
-# redeploy. See apps/discord/docs/zero-downtime-design.md for details.
+# Unsupported combination: PROCESS_ROLE=combined + ENABLE_EVENT_SHIPPER=true
+# is REJECTED at boot. In combined mode both the gateway publish hook
+# AND the worker consumer would arm in one process, dispatching every
+# interaction twice (in-process WS frame + SQS round-trip). The supported
+# flag-on shape is the two-process split: PROCESS_ROLE=gateway (singleton,
+# publishes) + PROCESS_ROLE=http (consumes, can scale horizontally). For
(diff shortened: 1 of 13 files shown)


justin-layerv:
Round 1 cr feedback addressed on 122b2d5:

**Substantive:**
- **Sync-throw blind spot** — `publish()` now wraps `new SendMessageCommand(...)` + `sqsClient.send(...)` in try/catch. A sync throw routes through the same `kind: 'unhandledRejection'` log tag the async-rejection path uses, closing the unified-CloudWatch-query gap. New test pins the contract.
- **AWS_REGION fragility** — publisher test seeds `process.env.AWS_REGION` in `beforeAll` (restored in `afterAll`). Local-dev tests no longer depend on the CI env var.

**Nits:**
- `TODO(sharding-pr)` added to the `SHARD_ID = '0'` literal, grep-able from PR 13.
- `.env.example` rollback wording restored to the specific "which gate goes away in which follow-up" framing.
- `flushMicro` now documents the depth-1 promise-chain assumption.
- Module header now explicitly notes the producer-side body cap is implicit (SQS's 256 KB wire limit) — defense-in-depth lives on the consumer's `MAX_BODY_BYTES`, which is fine while INTERACTION_CREATE is the only event class.

Tests: 1954 across 51 suites (was 1953 — +1 new sync-throw test).

justin-layerv:
Round 2 cr feedback addressed on a30b45f:

**Substantive (the only pre-merge suggestion):**
- **e2e log label** renamed from `'Event consumer: dispatched'` to `'Event consumer: received'`. The log fires before `client.actions.InteractionCreate.handle`, so the old label would mislead a future reader grepping for the dispatch terminal. The structured field `qurl_bot_event_e2e_ms` stays identical so the CloudWatch metric handle is undisturbed. Test assertions updated.

**Nits:**
- Orphan misindented comment inside `config.intEnv`'s range branch removed.

**Acknowledged, intentionally not changed:**
- `cr-r2`'s optional boot-time assertion (`running===true` after `eventPublisher.start()`): the existing `'Event publisher: starting'` info log inside `start()` already provides the boot-time signal cr was asking for. Adding a hard assertion would either expose `isRunning()` to production code (currently `_test`-only) or duplicate state — both for marginal gain against a "refactor moves the call out" failure that's implausible given the 4-line conditional.
- `drainTimer` could-be-undefined: cr noted `clearTimeout(undefined)` is a no-op and explicitly said "no action needed; flagging only for completeness."
- Sandbox-soak verification of the `gateway + flag-on` shape: in the test plan checklist (post-merge measurement), cheaper net than another unit-test layer.
- Trust boundary / SQS SSE: infra-side concerns flagged in the description; queue policy lock and SSE-SQS/KMS are tracked in qurl-integrations-infra.

Tests: 1954 across 51 suites (unchanged count, label-only assertion updates).

PR #402, opened 2026-05-15
refactor(discord): event-driven abortableSleep + unified stop controller

### LINKED ISSUE #387: Event consumer follow-ups: AbortSignal-based abortableSleep + combined-mode prod guard
Surfaced in PR #379 r8 cr. Two non-blocking follow-ups deferred from PR 11.

## 1. Replace `abortableSleep` polling with AbortSignal-based wait

Today `src/event-consumer.js`'s `abortableSleep` polls the `stopping` flag every 50 ms via a `setInterval`. Works + the leak fix landed in PR 11, but the cleaner shape is to wire the loop's stop signal into Node's `setTimeout(ms, undefined, { signal })`:

```js
let stopController = null;

function start() {
  ...
  stopController = new AbortController();
  ...
}

function abortableSleep(ms) {
  return new Promise((resolve) => {
    const t = setTimeout(resolve, ms);
    stopController.signal.addEventListener('abort', () => {
      clearTimeout(t);
      resolve();
    }, { once: true });
  });
}

async function stop() {
  ...
  stopController.abort();
  ...
}
```

Removes the 50 ms polling tick + the `stopping` flag reads inside `abortableSleep`. The polling approach grows extra branches when someone adds the next gate (cr observation).

## 2. ALLOW_COMBINED_SHIPPER prod guard

Today `isWorker = isHttp && config.ENABLE_EVENT_SHIPPER`. In combined mode + flag on (after PR 10 ships), the process is both producer AND consumer — which is the intended shape for local dev but a footgun for prod.

The TODO(PR-10) comment in `index.js:300` already requires PR 10 to gate the gateway-side listener so it doesn't double-dispatch. But that's a code change PR 10 must make; a misconfigured prod deploy with the wrong PROCESS_ROLE could still land here.

Proposed shape: a separate `ALLOW_COMBINED_SHIPPER` env (literal-string "true" like other flags) that must be set for combined mode to be acceptable in `NODE_ENV=production`:

```js
if (process.env.NODE_ENV === 'production'
    && PROCESS_ROLE === 'combined'
    && config.ENABLE_EVENT_SHIPPER
    && process.env.ALLOW_COMBINED_SHIPPER !== 'true') {
  logger.error('Combined mode + ENABLE_EVENT_SHIPPER=true in production requires explicit ALLOW_COMBINED_SHIPPER=true opt-in');
  process.exit(1);
}
```

Acceptance:
- [ ] Validation in `src/index.js` after the existing event-shipper boot check.
- [ ] Unit tests via `boot-requirements.js` helper (extract the predicate so it's testable without spawning a process).
- [ ] Document in `.env.example` and `docs/zero-downtime-design.md`.

Priority: low — PR 10 should land the gateway-side listener gate; this is defense-in-depth against PROCESS_ROLE misconfig.

justin-layerv:
Closes #387 (item 1).

## Summary

- Replace the consumer's setInterval-polling `abortableSleep` with an AbortSignal-event-driven wakeup, so `stop()` wakes a sleeping worker on the next microtask rather than within 50 ms.
- Consolidate three previously-separate stop mechanisms — a `stopping` boolean flag, a `receiveAbortController` scoped to the in-flight ReceiveMessage call, and the 50 ms sleep-polling tick — into a single `stopController` `AbortController` driving all three.

## Why

The 50 ms polling tick was a tight loop dressed up as a sleep: every paused worker (idle-empty, error-backoff, at-cap backpressure) burned a wakeup 20×/second to check a boolean. With AbortSignal events, idle workers truly idle, and the loop's shutdown semantics live on one piece of state instead of three drifting ones.

This was item 1 of two follow-ups filed during PR #379 (worker-tier consumer) review. Item 2 — an `ALLOW_COMBINED_SHIPPER` prod guard — was already superseded by PR #398's stronger combined+flag-on boot reject.

## How

- `start()` creates a fresh `stopController`. `stop()` flips `running = false` then calls `stopController.abort()`. The pollLoop's `while (!stopController.signal.aborted)` check exits the loop next iteration; the in-flight `pollOnce` receives the same signal directly so the SDK call is cancelled mid-flight.
- `abortableSleep`: fast-path returns if the signal is already aborted; otherwise it arms a `setTimeout` *and* a `once: true` `abort` listener. Natural timeout removes the listener (no leak). The null-controller branch defensively degrades to a pure `setTimeout` so code paths reachable before `start()` (e.g. at-cap backpressure pause in tests) never throw.

## Test plan

- [x] **Load-bearing wall-clock test** (`pollLoop catches ReceiveMessage errors, logs, sleeps, then continues`): induces an SDK throw on the first poll iteration, then calls `stop()` while pollLoop is parked in the 1 s `POLL_ERROR_BACKOFF_MS` sleep, and asserts elapsed < 500 ms. This was the real regression test for the previous 50 ms-polling design; it stays green here, confirming the new abort-wakes-sleep guarantee.
- [x] Four new microtask-timing tests pin the refactor's contract:
  - abort-wins-race resolves on a microtask (not the timeout)
  - already-aborted signal short-circuits the sleep
  - null `stopController` degrades cleanly (defensive)
  - no `setInterval(_, 50)` ticks under any path
- [x] 1957 / 1957 tests passing locally.

### DIFF
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -884,17 +884,16 @@ function abortableSleep(ms) {
         clearTimeout(t);
         resolve();
       };
-      // { once: true } drops the listener after firing so a
-      // re-armed sleep on the same signal (impossible today but
-      // load-bearing if start/stop cycling ever returns) doesn't
-      // accumulate stale handlers.
+      // Timeout-wins branch above removeEventListener's; { once: true }
+      // here is defensive — AbortSignal fires 'abort' at most once per spec.
       stopController.signal.addEventListener('abort', onAbort, { once: true });
     }
   });
 }
 
 async function pollLoop(client) {
-  while (!stopController.signal.aborted) {
+  // Null-guard symmetric with abortableSleep + stop().
+  while (stopController && !stopController.signal.aborted) {
     try {
       await pollOnce(client);
     } catch (err) {
@@ -1170,9 +1169,9 @@ module.exports = {
     // Test-only initializer for the stop controller. Tests that
     // exercise pollOnce directly (bypassing start()) need a
     // controller to exist so the SDK send() call can read its
-    // signal. Construct fresh; production code only sets this via
-    // start().
-    _setStopControllerForTest: (ac) => { stopController = ac ?? new AbortController(); },
+    // signal. Always constructs fresh — to clear, use
+    // _resetStateForTest. Production code only sets this via start().
+    _setStopControllerForTest: () => { stopController = new AbortController(); },
     // Returns the in-flight count by reading `.size` on the live
     // Set. Kept under the historical name for test-site stability;
     // semantically still "how many handlers are tracked right now."
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -881,15 +881,19 @@ function abortableSleep(ms) {
         clearTimeout(t);
         resolve();
       };
-      // Timeout-wins branch above removeEventListener's; { once: true }
-      // here is defensive — AbortSignal fires 'abort' at most once per spec.
+      // Timeout-wins branch (above) removes the listener explicitly;
+      // { once: true } here is defensive — AbortSignal fires 'abort'
+      // at most once per spec.
       ctrl.signal.addEventListener('abort', onAbort, { once: true });
     }
   });
 }
 
 async function pollLoop(client) {
-  // Null-guard symmetric with abortableSleep + stop().
+  // Null-guard symmetric with abortableSleep + stop(). Paranoia, not
+  // a real path: start() always assigns stopController before invoking
+  // pollLoop, and stop() nulls it only after `await loopPromise`
+  // resolves (by which point pollLoop has already exited).
   while (stopController && !stopController.signal.aborted) {
     try {
       await pollOnce(client);
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -246,11 +246,28 @@ function recordSeen(eventId) {
   return false;
 }
 
-// Worker state. start() flips `running` true and kicks off the poll
-// loop; stop() flips `stopping` true and awaits loopPromise.
-// pollOnce uses Promise.all over the per-message wrappers, which
-// serializes "all messages in this batch were dispatched (emit +
-// DeleteMessage attempted)" against loopPromise resolution.
+// Worker state. start() constructs a fresh `stopController` and
+// flips `running` true; stop() calls `stopController.abort()` and
+// awaits loopPromise. pollOnce uses Promise.all over the per-message
+// wrappers, which serializes "all messages in this batch were
+// dispatched (emit + DeleteMessage attempted)" against loopPromise
+// resolution.
+//
+// stopController unifies three previously-separate concerns:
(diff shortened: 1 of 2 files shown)


justin-layerv:
Replying to the latest cr round:

**Item 1 (stop() NPE / `?.` hardening)** — declined. Adding `?.` to `stopController?.signal.aborted` while leaving `stopController.abort()` unguarded on the very next line is a half-fix that doesn't actually prevent the crash you describe. And cr-r3 explicitly asked to drop the `!stopController` middle guard on the grounds that "running=true with controller=null" is unreachable via start(). Reintroducing a partial defense one round later would be churn. If a future refactor genuinely splits the running/controller assignments across an await, that PR's reviewer will catch it.

**Item 2 (SDK signal-listener accumulation)** — verified false alarm. `@smithy/node-http-handler` cleans up properly on both HTTP1 and HTTP2 paths:

- http1 (\`node-http-handler.js:174-175\`):
  \`\`\`js
  signal.addEventListener(\"abort\", onAbort, { once: true });
  req.once(\"close\", () => signal.removeEventListener(\"abort\", onAbort));
  \`\`\`
- http2 (\`node-http2-handler.js:120-121\`):
  \`\`\`js
  signal.addEventListener(\"abort\", onAbort, { once: true });
  clientHttp2Stream.once(\"close\", () => signal.removeEventListener(\"abort\", onAbort));
  \`\`\`

Both use \`{ once: true }\` AND explicit removal on \`'close'\`. A long-running consumer's repeated ReceiveMessage calls won't accumulate listeners on \`stopController.signal\`. No soak needed.

**Item 3 (timing bound)** — applied. Bumped to <100ms in 7c0358e (still 100× under the 10s timeout, so the abort-microtask claim is preserved, with slack for CI scheduler jitter).

**Nit 1 (\"Timeout-wins branch above removeEventListener's\" wording)** — applied. Rewritten as \"Timeout-wins branch (above) removes the listener explicitly\".

**Nit 3 (pollLoop null-guard comment)** — applied. Added one-line comment noting it's paranoia, not a real path.

justin-layerv:
r5 reply:

**Item 1 (\`stop()\` \`?.\` consistency)** — sticking with the current shape. The consistency-with-abortableSleep argument is real, but the parallel isn't symmetric: in abortableSleep the null-controller path is a tested production scenario (the at-cap backpressure pause can run before \`start()\`), so \`ctrl?\` is load-bearing. In \`stop()\` the null-controller path is unreachable via start() and not tested, so the \`?\` would be cosmetic — and the next-line \`stopController.abort()\` would still NPE if the impossible state were ever reached. Half-hardening reads worse than full trust of the documented invariant. Happy to revisit if a future PR ever introduces an await between the \`running\` and \`stopController\` assignments in \`start()\`.

**Item 2 (\`abortableSleep\` double-resolve)** — applied in e5a6071. Added a one-liner clarifying the timeout-wins and abort-wins branches race and that the second resolve() is a Promise no-op.

**Item 3 (\`pollLoop\` null-guard)** — staying with the \"keep + comment as paranoia\" shape (already done in cr-r4). cr-r5 explicitly said this is fine; just confirming the choice.

**Item 4 (wall-clock test bound)** — applied in e5a6071. Tightened from \`<500ms\` to \`<100ms\`. 5-run local stability check confirms it holds; any regression to synchronous polling would push toward the 1s \`POLL_ERROR_BACKOFF_MS\` budget and surface here.

PR #403, opened 2026-05-15
feat(discord): raise picker per-pick cap from 10 to 25

justin-layerv:
## Summary

The MentionableSelectMenu on the qURL send confirm card capped per-pick selections at 10, even though Discord's API allows up to 25. Bump to 25 so users can address a full team (~15-25 recipients) in one pick instead of stitching multiple **Add Recipients** clicks. The placeholder ("Pick up to N users/roles") interpolates the cap dynamically, so it now reads "25" without a copy edit.

- Bump `USER_SELECT_PER_PICK_CAP` from 10 → 25 in `apps/discord/src/commands.js`. The post-send **Add Recipients** flow shares this constant, so it gets the same bump (and stays in lockstep per the constant's existing contract).
- Collapse the now-dead widen branch in `renderConfirmCardRows`. With the per-pick cap equal to Discord's `max_values` hard cap (25), the `defaults.length > 0 ? widen : baseCap` ternary always produced 25 regardless of `defaults.length`. Replaced with a single three-way `min(USER_SELECT_PER_PICK_CAP, QURL_SEND_MAX_RECIPIENTS, DISCORD_SELECT_MAX_VALUES_HARD_CAP)`, which still clamps correctly if someone later lowers `USER_SELECT_PER_PICK_CAP` or tightens `QURL_SEND_MAX_RECIPIENTS` via env.
- Add the same three-way `Math.min` clamp at the **Add Recipients** site (`commands.js:2013`) so both picker call sites are symmetric — defends against a future `USER_SELECT_PER_PICK_CAP > 25` from constructing a menu Discord rejects.
- Refresh the stale `"=10"` comments on the defense-in-depth guard in `handleConfirmUserSelect`, on the `DISCORD_SELECT_MAX_VALUES_HARD_CAP` constant (which used to describe the now-removed widen branch), and on the cap-exceeded test. Retarget the former widen-fit test to assert the picker opens at the full per-pick cap (25) with all pre-resolved defaults still pre-checked via `addDefaultUsers`.
- Add two tests pinning the three-way `Math.min` clamp branches: (a) env-override (`QURL_SEND_MAX_RECIPIENTS=15` → `setMaxValues(15)`, placeholder "Pick up to 15 users/roles"); (b) slice truncation (>25 pre-resolved defaults → `addDefaultUsers` receives exactly the first 25; overflow ids still ride in `payload.recipientIds`).

## UX behavior change worth flagging

When a user slash-invokes with 11–24 pre-resolved mentions, the picker previously sized `max_values` to exactly `defaults.length` (the widen branch). After this PR, the picker always opens at 25 — so the user can top up to 25 in one click without re-picking. Matches the PR's stated motivation but is a deliberate delta on the slash-with-mentions path.

## What did not change

- `QURL_SEND_MAX_RECIPIENTS` (the overall cap on a single send, default 20000 in prod) is untouched. The 10→25 bump only affects how many entities the user can select per **single** picker click.
- Role expansion, mass-mention gating, bot filtering, and the confirm card's first-5-name preview all stay as-is; they're recipient-count-independent or already bounded by `QURL_SEND_MAX_RECIPIENTS`.
- The defense-in-depth `valid.length > QURL_SEND_MAX_RECIPIENTS` reject branch is preserved (now reachable only via forged interactions, same as before — the inline comment now says so).

## Test plan

- [x] `npx jest` — 1960/1960 pass (added env-override clamp test + slice truncation test)
- [x] `npx eslint src/commands.js tests/qurl-send-map.test.js` — clean
- [ ] Manual: in sandbox Discord, run `/qurl send`, open the picker on the confirm card, verify the placeholder reads "Pick up to 25 users/roles" and Discord's UI permits up to 25 selections
- [ ] Manual: with `/qurl send recipients:@a @b ... @c` (12 mentions), confirm card renders with all 12 pre-checked in the picker, no truncation
- [ ] Manual: with 15 pre-resolved mentions, verify the picker now allows topping up to 25 in a single re-pick (UX delta vs. pre-PR fit-to-defaults behavior)
- [ ] Manual: on the post-send embed, click **Add Recipients** and verify the same per-pick cap behavior (≤25, narrows as `remaining` shrinks)

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3957,7 +3957,7 @@ function renderConfirmCardRows({
   if (mode === RECIPIENT_MODE_PICKER) {
     const picker = new MentionableSelectMenuBuilder()
       .setCustomId(CONFIRM_USER_SELECT_CUSTOM_ID)
-      .setPlaceholder(`Pick up to ${maxValues} users/roles`)
+      .setPlaceholder(`Pick up to ${maxValues} ${maxValues === 1 ? 'user/role' : 'users/roles'}`)
       .setMinValues(1)
       .setMaxValues(maxValues);
     if (defaults.length > 0) {
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -588,11 +588,11 @@ function isValidExpiry(v) {
   return Object.prototype.hasOwnProperty.call(EXPIRY_LABELS, v);
 }
 
-// Per-pick cap on UserSelectMenuBuilder.setMaxValues. Discord's hard
-// limit is 25; capping at 10 bounds the UX. The initial confirm-card
-// UserSelectMenu AND the post-send "Add Recipients" flow both use this
-// — keep them in lockstep so a future bump doesn't drift.
-const USER_SELECT_PER_PICK_CAP = 10;
+// Per-pick cap on UserSelectMenuBuilder.setMaxValues. Set to Discord's
+// hard limit so the picker exposes the full per-pick budget. The initial
+// confirm-card UserSelectMenu AND the post-send "Add Recipients" flow
+// both use this — keep them in lockstep so a future change doesn't drift.
+const USER_SELECT_PER_PICK_CAP = 25;
 
 // Shared render caps for warnings-block bullets that surface user-
 // or admin-controlled strings (invalidTokens, roleMentionsDeniedNames).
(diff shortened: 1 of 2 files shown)


justin-layerv:
Addressed the three review rounds — pushed `47b9ad7`.

**Fixed:**
- 🟡 Stale `DISCORD_SELECT_MAX_VALUES_HARD_CAP` comment (was describing the removed widen branch) — refreshed in `bc23a1f`, then tightened in `47b9ad7` to say it defends against any combination of `USER_SELECT_PER_PICK_CAP` × `QURL_SEND_MAX_RECIPIENTS` that would push `setMaxValues` past 25.
- 🟡 Add Recipients picker now applies the same three-way `Math.min` clamp as `renderConfirmCardRows` (`commands.js:2013`) — symmetric defense against a future `USER_SELECT_PER_PICK_CAP > 25`. Done in `5bd1bec`.
- 🟡 Env-override clamp test — pins `setMaxValues(15)` + placeholder "Pick up to 15 users/roles" when `QURL_SEND_MAX_RECIPIENTS=15`. The other tests run with mocked QSMR=25 where `min(25,25,25)=25` makes the QSMR leg invisible — this test prevents that branch from silently rotting.
- 🟡 Slice truncation test — with 30 pre-resolved defaults (QSMR mocked to 30), asserts `addDefaultUsers` receives exactly the first 25. Covers `commands.js:3970`'s `slice(0, maxValues)` which had no direct coverage.
- 🟢 Tightened the retargeted test's comment to call out the new always-25 invariant as a deliberate UX change, not just "still works".
- 🟢 PR description updated with the 11–24-defaults UX delta and a manual top-up check.

**Trimmed:**
- Comment bloat on `USER_SELECT_PER_PICK_CAP` (10 → 3 lines), the Add Recipients site (5 → 0), and the `renderConfirmCardRows` clamp (9 → 3) per `/simplify` review.

**Skipped:**
- "Two constants both equal to 25" nit — keeping both is the explicit defense-in-depth intent (policy cap vs. protocol cap can diverge); the constant comment already documents this.

Tests: 1960/1960 (+2 new). Lint clean.

PR #405, opened 2026-05-15
feat(discord): fail-fast on permanent AWS misconfig in event consumer

### LINKED ISSUE #385: Event consumer: fail-fast on permanent misconfig instead of infinite backoff
Surfaced in PR #379 r6 cr. The `pollLoop` in `src/event-consumer.js` currently treats every non-abort `ReceiveMessageCommand` failure as transient and applies the 1s error backoff. If the deploy is misconfigured (queue URL points at a nonexistent queue, IAM role lacks `sqs:ReceiveMessage`, region mismatch, etc.), the loop retries forever — flooding logs without escalating.

The fail-closed boot check (`missingEventShipperKeys`) catches the *missing* queue URL case, but a misconfigured-but-set URL slips through.

## Proposed shape

Classify AWS error codes into permanent vs transient and flip the consumer into a fail-fast state on permanent failures:

- Permanent (fail-fast after first observation): `AccessDenied`, `AWS.SimpleQueueService.NonExistentQueue`, `InvalidAddress`, `InvalidQueueUrl`, `UnknownEndpoint`
- Transient (current behavior, error-backoff + retry): everything else (`Throttling`, `RequestTimeout`, network errors)

Fail-fast = log fatal + `process.exit(1)` so ECS restarts the task — surfaces in deploy-time monitoring rather than silently bleeding into runtime noise. Alternative: consecutive-failure counter that flips to fail-fast after N attempts regardless of code.

## Acceptance

- [ ] Classifier in `event-consumer.js` distinguishing permanent vs transient AWS error codes.
- [ ] Permanent failure → fatal log + exit 1.
- [ ] Test: mock `ReceiveMessageCommand` to throw `AWS.SimpleQueueService.NonExistentQueue` → assert `process.exit` mock called once.
- [ ] Document the classification in the module header.

Priority: low — boot check covers the most likely misconfig (missing URL); this hardens the next-most-likely case (typo'd URL).

justin-layerv:
Closes #385.

## Summary

The event consumer's \`pollLoop\` previously treated every non-abort \`ReceiveMessageCommand\` failure as transient and applied the 1s error backoff. If the deploy is misconfigured (typo'd queue URL, missing IAM permission, region drift, bad credentials), the loop retries forever — flooding logs without escalating.

This PR adds a permanent-vs-transient AWS error classifier. Permanent failures now log fatal + \`process.exit(1)\`; ECS surfaces them as a high task-restart rate in deploy-time monitoring rather than letting the misconfig bleed into runtime noise. Transient errors (throttling, timeouts, network blips) continue through the existing 1s backoff.

## Why

The fail-closed boot check (\`missingEventShipperKeys\`) catches the *missing* queue URL case — the most likely misconfig. This PR hardens the *next-most-likely* shape: a misconfigured-but-set queue URL or an IAM regression. Both are silent failures today; both are loud after this change.

## How

- New \`PERMANENT_AWS_ERROR_CODES\` set covering AWS-API-layer codes (v2-style, \`error.Code\`) and SDK-v3 error class names (\`error.name\`).
- New \`permanentAwsErrorCode(err)\` classifier — parallel to \`isAbortError\`: walks \`err.cause\` with a visited-set (bounded memory, handles cyclic chains), checks \`.name\` / \`.code\` / \`.Code\` on each layer. Returns the matched code or null.
- \`pollLoop\`'s catch routes errors into three buckets:
  - **abort** (silent continue, stop()-triggered) — existing
  - **permanent** (fatal log + \`process.exit(1)\` + defensive \`return\`) — new
  - **transient** (log + \`POLL_ERROR_BACKOFF_MS\`) — existing
- Module header documents the classification scheme.

Defensive \`return\` after \`process.exit(1)\`: in tests \`process.exit\` is mocked to a no-op, so without the return the loop would iterate into the same failure. In prod the return is unreachable.

## Classification (PERMANENT — fail-fast)

| Code / name                                       | Cause                                                    |
| ------------------------------------------------- | -------------------------------------------------------- |
| \`QueueDoesNotExist\`                             | Typo'd queue URL                                         |
| \`AWS.SimpleQueueService.NonExistentQueue\`       | v2-style alias of above                                  |
| \`AccessDenied\` / \`AccessDeniedException\`      | IAM lacks \`sqs:ReceiveMessage\` on the queue            |
| \`InvalidAddress\` / \`InvalidQueueUrl\`          | Malformed queue URL                                      |
| \`UnknownEndpoint\`                               | Region mismatch between \`AWS_REGION\` and queue URL     |
| \`UnrecognizedClientException\` / \`InvalidClientTokenId\` | Bad/expired credentials                                  |

Everything else (Throttling, RequestTimeout, ECONNRESET, etc.) stays in the transient path.

## Test plan

- [x] \`permanentAwsErrorCode\` unit test: matches by \`.name\` / \`.code\` / \`.Code\`, walks \`err.cause\` chains, handles cyclic chains, returns null for transient shapes (Throttling, ECONNRESET, TimeoutError).
- [x] Integration test in \`pollLoop error backoff\` describe: SDK mock throws \`QueueDoesNotExist\` → \`process.exit(1)\` called once (mocked), fatal log fires with \`code: 'QueueDoesNotExist'\`, transient \"poll iteration failed\" log does NOT fire.
- [x] 1960/1960 tests pass locally.

### DIFF
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -868,7 +868,11 @@ function isAbortError(err) {
 //   - SignatureDoesNotMatch — usually clock skew, which also self-heals
 //     once chrony catches up.
 // Re-classify if prod data shows these recur past a fail-fast threshold;
-// today they belong with throttling.
+// today they belong with throttling. AccessDenied is in the permanent
+// set on the bet that misconfig (typo'd queue ARN, missing IAM grant)
+// dominates over deploy-time IAM eventual-consistency gaps. If prod
+// telemetry ever shows AccessDenied spikes correlated with deploys
+// rather than config changes, demote it back to transient.
 const PERMANENT_AWS_ERROR_CODES = new Set([
   // Queue doesn't exist (typo'd URL or wrong region in queue URL).
   'QueueDoesNotExist',
@@ -886,6 +890,13 @@ const PERMANENT_AWS_ERROR_CODES = new Set([
   'InvalidClientTokenId',
 ]);
 
+// Returns the matched permanent-error code from anywhere in err's
+// cause chain, or null if none. Any matching key at any layer wins —
+// if an error has both .name = 'TimeoutError' (transient) and
+// .code = 'QueueDoesNotExist' (permanent), this returns
+// 'QueueDoesNotExist'. The conservative cut is "if any layer says
+// permanent, treat as permanent" — under SDK wrapping a real
+// misconfig can surface anywhere in the chain.
 function permanentAwsErrorCode(err) {
   const visited = new Set();
   let current = err;
@@ -954,6 +965,31 @@ function abortableSleep(ms) {
   });
 }
 
+// Dispatch the fatal-shutdown route. If onFatalCb is wired, call it
+// (sync throws AND async rejections both fall back to process.exit(1)).
+// If not, exit directly. Does NOT await — the caller's defensive
+// return must stay synchronous so pollLoop exits before the next
+// iteration (under test mocks an await would let it iterate).
+function invokeOnFatal() {
+  if (!onFatalCb) {
+    process.exit(1);
+    return;
+  }
+  try {
+    Promise.resolve(onFatalCb()).catch((cbErr) => {
+      logger.error('Event consumer: onFatal rejected, falling back to process.exit', {
+        error: cbErr?.message,
+      });
+      process.exit(1);
+    });
+  } catch (cbErr) {
+    logger.error('Event consumer: onFatal threw, falling back to process.exit', {
+      error: cbErr?.message,
+    });
+    process.exit(1);
+  }
+}
+
 async function pollLoop(client) {
   // Null-guard symmetric with abortableSleep + stop(). Paranoia, not
   // a real path: start() always assigns stopController before invoking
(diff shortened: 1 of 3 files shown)


justin-layerv:
r1 reply:

- **(1)** Applied in c7ccf08. Wrapped \`onFatalCb()\` invocation with \`Promise.resolve(...).catch()\` so async rejections from gracefulShutdown also route to the fallback \`process.exit(1)\`. Sync try/catch still covers sync throws. Did NOT add an \`await\` — the defensive \`return\` after the routing must stay synchronous so the loop doesn't iterate under test mocks during the async gap. JSDoc updated to document the \`void | Promise<void>\` shape + the rejection contract. New test \`onFatal async rejection falls back to process.exit(1)\` pins it.
- **(2)** Applied. Test renamed to \`(QueueDoesNotExist)\` matching the error.name assertion.
- **(3)** Skipped — paranoia-nit per cr. The integration test exercises one-deep cause wrapping (proves the walk is invoked); the unit test for \`permanentAwsErrorCode\` covers visited-set + cyclic-chain semantics. Adding a 3-deep variant of an already-shape-tested helper is diminishing returns.
- **(4)** Filed as #406 (\`enhancement, priority: low\`). cr called this out-of-scope for #385; carrying it as a separate change keeps the diff focused and the issue trail clean.
- **(5)** Applied. One-line comment on the double-start early-return clarifying onFatalCb preservation.

PR #407, opened 2026-05-15
feat(discord): exponential-decay backoff at in-flight cap + silent steady-state

### LINKED ISSUE #390: discord: smarter backoff in event-consumer pollOnce when at in-flight cap
## Context

PR #389 added a fixed `INFLIGHT_BACKOFF_MS = 100` pause when `pollOnce` finds itself at the in-flight handler cap (`inFlightCount >= MAX_INFLIGHT_HANDLERS`). Per cr feedback on that PR:

> A handler that blocks on a slow DM fan-out can sit in-flight for several seconds. While at-cap, `pollOnce` will wake every 100 ms and log a debug line. Debug-level so not actually spammy, but the loop is busy-waking ~10×/s and re-running the cap check. Not a blocker — a tiny exponential decay or longer base would be a nice follow-up.

## Scope

In `apps/discord/src/event-consumer.js` (the `pollOnce` at-cap branch):

- Replace the fixed 100 ms sleep with a smarter cadence. Two options the reviewer mentioned:
  - **Longer base** (e.g., 500 ms) — simplest, cuts CPU+log noise by 5×, keeps responsiveness "good enough" for typical bursts.
  - **Exponential decay** — start at 100 ms, double up to a 1–2 s ceiling while still at-cap, reset to 100 ms once a poll succeeds. Best of both — fast recovery for short bursts, low overhead for sustained at-cap.
- Either way, ensure `abortableSleep` still honors `stopping` so SIGTERM during a backoff stays inside the graceful-shutdown budget.

Optionally:

- Throttle the `Event consumer: in-flight cap reached, pausing receive` debug log to fire only on the first iteration of an at-cap streak (and once when released), so the log signal is "we entered/exited at-cap" rather than "we're still at-cap." Requires a small piece of module-level state to track the streak.

## Acceptance criteria

- [ ] Backoff cadence does not wake more than ~2×/s at sustained at-cap
- [ ] Recovery from at-cap is no slower than ~500 ms in the common case (cap released cleanly)
- [ ] `stop()` during an at-cap pause exits inside the graceful-shutdown budget (test it)
- [ ] No behavior change for the below-cap path
- [ ] Tests cover at-cap → at-cap streak (no spurious log spam) and at-cap → below-cap transition

## Why this isn't fixed in #389

Reviewer explicitly flagged this as "not worth widening this PR" — the core safety net (cap + pause) is in, and the cadence tuning is independent and can land separately without coupling.

justin-layerv:
Closes #390.

## Summary

PR #389 landed a fixed 100ms backoff at the in-flight handler cap with a debug log on every iteration of an at-cap streak. cr feedback on that PR noted that under sustained at-cap (slow downstream + bursty arrivals) the loop wakes ~10×/s producing per-worker log noise without operational value — the entry warn + exit info already bookend the pause.

This PR addresses both concerns:

1. **Exponential decay** replaces fixed 100ms.
2. **Silent steady-state**: per-iteration debug log dropped.

## Why

The current shape is correct but over-talks. A wedged downstream can hold the cap for tens of seconds; at 10 wakes/s the worker burns CPU re-checking the cap and emits a debug line each time. The entry warn ("we entered at-cap") and exit info ("we left at-cap") already give operators the pair they need to compute pause duration — the per-iteration debug just trails behind.

Exponential decay keeps the 100ms responsiveness for short streaks (the common case — a brief DM-fan-out spike) and drops sustained-streak wake rate from 10/s to ~0.6/s, cutting CPU + log noise meaningfully.

## How

- `INFLIGHT_BACKOFF_MS = 100` → `INFLIGHT_BACKOFF_BASE_MS = 100` + `INFLIGHT_BACKOFF_MAX_MS = 1600`.
- New module-level `currentBackoffMs`, starts at base.
- pollOnce at-cap path: `await abortableSleep(currentBackoffMs)`, then `currentBackoffMs = Math.min(currentBackoffMs * 2, INFLIGHT_BACKOFF_MAX_MS)`.
- Below-cap path resets `currentBackoffMs = INFLIGHT_BACKOFF_BASE_MS` so the next at-cap entry starts responsive again.
- State resets alongside `atCapPauseLogged` in `stop()` and `_resetStateForTest`.
- Per-iteration debug log removed. `AT_CAP_PAUSE_DEBUG_MSG` constant + export gone.

`abortableSleep` honors the abort signal unchanged, so `stop()` during a 1.6s backoff still exits within the graceful-shutdown budget.

## Cadence

| Iteration  | 1   | 2   | 3   | 4   | 5+   |
|------------|-----|-----|-----|-----|------|
| Backoff ms | 100 | 200 | 400 | 800 | 1600 |

Recovery: a below-cap poll at any iteration resets to 100ms.

## Test plan

- [x] Updated "stays silent during sustained at-cap" test: 3 consecutive at-cap polls produce exactly 1 warn + 0 debugs.
- [x] New "at-cap backoff doubles each iteration up to the ceiling" test: pins 100→200→400→800→1600 and the ceiling clamp.
- [x] Updated cap-released test: asserts `currentBackoffMs` resets to base on below-cap transition.
- [x] Existing tests (entry warn, exit info, abort-during-backoff) unchanged.
- [x] 1971/1971 tests pass locally.

### DIFF
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -342,8 +342,9 @@ function validateInflightCap(cap) {
 }
 validateInflightCap(MAX_INFLIGHT_HANDLERS);
 
-// At-cap pause backoff. Exponential decay from BASE to MAX while
-// the cap stays full, resets to BASE on the first below-cap poll.
+// At-cap pause backoff. Exponential backoff (doubling) from BASE to
+// MAX while the cap stays full; resets to BASE on the first below-cap
+// poll. The wake *rate* decays — the sleep duration grows.
 // A short cap streak (handlers draining within a second) keeps the
 // 100ms responsiveness; a sustained streak (slow downstream + bursty
 // arrivals) drops the wake rate to ~0.6/s instead of 10/s, cutting
@@ -749,7 +750,7 @@ async function pollOnce(client) {
     // Steady-state at-cap is intentionally silent — the entry warn
     // + the exit info already bookend the pause for operators.
     await abortableSleep(currentBackoffMs);
-    // Exponential decay: each consecutive at-cap iteration doubles
+    // Exponential backoff: each consecutive at-cap iteration doubles
     // the wait up to MAX, so a sustained streak drops the wake rate
     // from 10/s to ~0.6/s without significantly delaying recovery.
     currentBackoffMs = Math.min(currentBackoffMs * 2, INFLIGHT_BACKOFF_MAX_MS);
--- a/apps/discord/src/event-consumer.js
+++ b/apps/discord/src/event-consumer.js
@@ -342,12 +342,17 @@ function validateInflightCap(cap) {
 }
 validateInflightCap(MAX_INFLIGHT_HANDLERS);
 
-// Brief pause when at-cap. Short enough that handlers draining
-// quickly resume normal polling; long enough that we don't burn
-// CPU re-checking the cap on every microtask. abortableSleep
-// honors the abort signal so a SIGTERM during a backpressure pause
-// still exits inside the graceful-shutdown budget.
-const INFLIGHT_BACKOFF_MS = 100;
+// At-cap pause backoff. Exponential decay from BASE to MAX while
+// the cap stays full, resets to BASE on the first below-cap poll.
+// A short cap streak (handlers draining within a second) keeps the
+// 100ms responsiveness; a sustained streak (slow downstream + bursty
+// arrivals) drops the wake rate to ~0.6/s instead of 10/s, cutting
+// CPU + log noise without significantly delaying recovery.
+// abortableSleep honors the abort signal so a SIGTERM during a
+// backpressure pause still exits inside the graceful-shutdown budget.
+const INFLIGHT_BACKOFF_BASE_MS = 100;
+const INFLIGHT_BACKOFF_MAX_MS = 1600;
+let currentBackoffMs = INFLIGHT_BACKOFF_BASE_MS;
 
 // Maximum time stop() waits for in-flight handler promises to
 // settle before proceeding past `await loopPromise`. Stays well
@@ -389,20 +394,17 @@ let DRAIN_DEADLINE_MS = INITIAL_DRAIN_DEADLINE_MS;
 const inFlightPromises = new Set();
 let isWorkerDispatch = false;
 
-// State-machine rate-limit on the at-cap log. Sustained at-cap
-// (slow downstream + bursty arrivals) wakes pollOnce every
-// INFLIGHT_BACKOFF_MS, so without throttling a single backlog
-// produces ~10 lines/sec/worker. Pattern matches
-// `lastMissingEventIdWarnMs`: the *transition* into at-cap is the
-// loud signal (warn), steady-state at-cap is debug, and the
-// transition back to below-cap is info. atCapPauseLogged is the
-// "we've already logged the warn for this streak" gate; cleared
-// when pollOnce next observes a below-cap state. Log messages
-// pulled into constants so monitoring alerts that grep for them
-// stay in sync with the call site even if the wording drifts.
+// State-machine rate-limit on the at-cap log. The *transition* into
+// at-cap is the loud signal (warn-on-entry), the transition back to
+// below-cap is the info-on-release. Steady-state at-cap is silent —
+// pairing entry+exit gives the duration of the pause without
+// per-iteration noise. atCapPauseLogged is the "we've logged the
+// warn for this streak" gate; cleared when pollOnce next observes a
+// below-cap state. Log messages live in constants so monitoring
+// alerts that grep for them stay in sync with the call site if the
+// wording drifts.
 let atCapPauseLogged = false;
 const AT_CAP_PAUSE_WARN_MSG = 'Event consumer: in-flight cap reached, pausing receive until handlers drain';
-const AT_CAP_PAUSE_DEBUG_MSG = 'Event consumer: still at in-flight cap, continuing pause';
 const AT_CAP_RELEASED_INFO_MSG = 'Event consumer: in-flight cap released, resuming receive';
 
 // Hook for unit tests: lets the test inject a mock SQSClient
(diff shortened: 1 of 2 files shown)


justin-layerv:
## cr-r6 response

Six rounds of review — cr-r6 explicitly confirms the prior round's load-bearing `stop()` test fix is correct. Applied two of the three nits this round; filed a follow-up for the third.

### Applied (commit `e162112`)

1. **Item 1 — extract `clearAtCapState()` helper.** Three reset sites was the DRY threshold cr called out. The helper pairs `atCapPauseLogged = false` with `currentBackoffMs = INFLIGHT_BACKOFF_BASE_MS` so a future fourth site can't silently break the "non-base `currentBackoffMs` implies `atCapPauseLogged === true`" invariant by forgetting one line.

2. **Item 2 — wake-rate phrasing clarification.** Added "(at the ceiling)" + pointer to the cadence table so a reader doesn't interpret `~0.6/s` as the full-streak average (the first 5 iterations average ~1.6/s before settling).

### Filed as follow-up

4. **Mid-sleep release mitigation** → #408. cr explicitly framed this as "neither is needed for this PR…worth a follow-up issue if pause-pair drift shows up in CloudWatch alerts." Filed with both option sketches (in-flight drain signal vs lower MAX) and the trigger criterion.

### Skipped

3. **No-jitter elaboration** — cr's own framing ("The current comment covers it") makes the additional sentence redundant.

All 1974/1974 tests pass locally in 4.6s. Awaiting cr-r7.

justin-layerv:
## cr-r7 response

Applied all three of cr-r7's actionable items in commit \`0336bec\`.

### Applied

1. **Stale docblock at the `capPayload` snapshot.** The docblock referenced the debug log (\"subsequent at-cap polls in the same streak are debug\", \"warn/debug AND the release-info branch\") that this PR removed. Updated to match the new \"steady-state at-cap is silent\" framing — no more hunting for a debug log that doesn't exist. **Genuine doc bug — good catch.**

2. **No-jitter framing.** cr correctly noted that my prior claim — \"synchronized wakes aren't possible\" — was imprecise: workers *can* synchronize when they share a wedged downstream, and their doubling ladders will align. Trimmed the imprecise half so the load-bearing reason (\"downstream isn't sensitive to wake-time alignment\") stands alone.

3. **Recovery trade-off average.** Added \"~MAX/2 (~800ms) average\" alongside the worst-case 1.6s — releases land uniformly within the sleep window, so the expected CloudWatch pause-end inflation is half the worst case. This is the figure operators will actually see.

### Observation (no action)

cr's note about the cadence test's exact values (`100→200→400→800→1600`) being tightly coupled to the constants — that coupling is intentional. The doubling cadence + ceiling clamp IS the contract; pinning the exact values means a future tweak to BASE/MAX forces a deliberate test update rather than silent drift.

1974/1974 tests pass locally. Awaiting cr-r8.

justin-layerv:
## cr-r9 response + /simplify pass

cr-r9: \"LGTM modulo the (entirely optional) nits above.\" Applied one item; skipped the rest per their explicit framing as taste/observations. Also ran /simplify and filed one cross-module follow-up.

### Applied (commit \`688f8b6\`)

**Item 4 — 1600ms ceiling rationale.** Added the inline rationale cr suggested. 4 doublings is a tidy ladder *and* keeps worst-case stop()-during-backoff wake well inside the 10s gracefulShutdown budget. Halving gives back the wake-rate win; doubling approaches the shutdown ceiling.

### Skipped

- **Item 1 — \`atCapPauseLogged\` vs \`currentBackoffMs !== BASE\` redundancy.** cr explicitly: \"Not a blocker — the explicit boolean gate reads more clearly at the call site.\" Collapsing the gate into a derived check would entangle two distinct concerns (\"have we warned this streak?\" vs \"what's the sleep duration?\"). The helper-enforced invariant is the right shape.
- **Item 2 — \`getCurrentBackoffMs\` test export shape.** cr: \"Consistent with the existing \`getDrainDeadlineMs\` pattern, so fine.\"
- **Item 3 — Comment density (trim no-jitter to one sentence).** cr: \"this is taste.\" The current 3-sentence form was specifically reframed in cr-r7 to address an inaccuracy; trimming risks losing the operational reasoning.
- **Item 5 — Recovery framing SLO note.** cr framed as future observation, not actionable for this PR.

### /simplify pass

- **Reuse**: one finding — \`orphan-token-sweeper.js\` (lines 42, 57, 65–66) implements the same \`Math.min(curr * 2, MAX)\` doubling pattern. Filed as #409 (low-priority follow-up). Touching that file would expand PR scope, and the two backoffs have different ceilings (1.6s vs 60s) and contexts (worker backpressure vs GitHub rate-limit) — a generic helper would be 1 line and risks coupling them prematurely.
- **Quality**: clean. State factoring, helper scope, comment weight all pulled their weight.
- **Efficiency**: clean. No hot-path bloat, no recurring no-op writes (helper is always behind a transition guard or at teardown), no memory growth.

1974/1974 tests pass locally. Awaiting cr-r10 — if it returns LGTM with only preference/framing items (the cr-r6/r8/r9 pattern), declaring convergence.

justin-layerv:
## cr-r10 response — substantive catch, fixed

cr-r10 caught that the cr-r5 \"load-bearing assertion\" test was still vacuous, just with more elaborate setup. **Verified empirically**: temporarily commented out \`clearAtCapState()\` in stop()'s finally → the old test still passed.

### The trace cr-r10 was right about

1. Old test drove \`currentBackoffMs\` to 200 via fake-timer at-cap iteration ✓
2. **Then drained handlers** (\`inFlightCount=0\`) before \`start()\` ✗ — this was the bug
3. \`start()\` → pollLoop's first pollOnce → inFlightCount=0 → **below-cap branch** → \`atCapPauseLogged===true\` → logged info → called \`clearAtCapState()\` 
4. By the time \`stop()\`'s finally ran \`clearAtCapState()\`, \`currentBackoffMs\` was already BASE — same vacuousness as cr-r5, just dressed up.

### The fix (commit \`6926008\`)

Per cr-r10's suggestion: keep handlers at cap throughout (use never-resolving promises). Then start()'s pollLoop takes the AT-cap branch, sleeps, stop() aborts the sleep, the post-await doubling line runs (\`currentBackoffMs = 400\`), pollLoop exits via signal check. **stop()'s finally is now the ONLY reset path.** Paired with \`_setDrainDeadlineForTest(50)\` so stop()'s drain race resolves on the deadline-timer side without hanging on the never-resolving promises.

### Verification

| State | Old test | New test |
|---|---|---|
| With \`clearAtCapState()\` in stop()'s finally | ✓ passes | ✓ passes |
| Without it (vacuity check) | ✓ passes (BUG) | ✗ fails with \`Expected 100, Received 400\` |

The \`400\` matches cr-r10's observation 1 — the post-await doubling line at pollOnce's at-cap branch runs on abort-wake, leaving \`currentBackoffMs = prev * 2\`. Also updated stop()'s comment to document that the reset is now load-bearing for the abort-wake doubling, not just the lifecycle warn gate.

### Other cr-r10 items

- **Observation 1 (doubling-after-await on abort)** — directly informed the new test's failure mode and the comment update; no separate action needed.
- **Observation 2 (\`pollOnceFast\` \`jest.getTimerCount()\` fail-fast guard)** — nice idea but converting the comment to a runtime assert would mean the helper has side effects beyond its name; the prescriptive \"convert to a loop + flushes\" comment serves the same future-reader purpose. Skip.
- **Observation 3 (no-jitter rationale for future-wedge variant)** — fair point; if the failure mode behind a future wedge becomes shared-resource contention (rate limiter, db pool), aligned ladders would amplify. Worth one sentence in the comment, but at this point in the review cycle I'd rather not invite cr-r11 with a comment tweak. Filing as a small follow-up note in the existing #408 if the wedge model changes.

1974/1974 tests pass locally. cr-r10 was the kind of substantive catch that justified the iteration cycle — thanks. Awaiting cr-r11.

PR #410, opened 2026-05-16
feat(discord): pillar 2 cross-process gateway RESUME via @discordjs/ws

justin-layerv:
## Summary

Adds the app-side of Pillar 2 in the zero-downtime gateway migration. Under a new `ENABLE_GATEWAY_RESUME` flag, the gateway tier replaces `discord.js Client` + `client.login()` with a direct `@discordjs/ws WebSocketManager` whose session callbacks point at a new `qurl-bot-discord-<env>-gateway-session` DDB table. On every restart the new gateway boots, reads the persisted `{sessionId, resumeURL, sequence}` row, and Discord's `RESUME` (op 6) replays buffered events from the last sequence — eliminating the ~10 s IDENTIFY cold-start the legacy single-process deploy carries today.

**Flag default off.** No behavior change in any env until an operator flips `ENABLE_GATEWAY_RESUME=true`. Single-replica behavior is unchanged at this layer (desired_count stays at 1); the second replica + push handoff arrives in PR 13b (Pillar 3).

## Why now

The bot today is `desired_count=1` on a Discord token (Discord allows one active WS per bot identity). Every ECS deploy → ~10 s WebSocket reconnect → `/qurl` command surface dark for that window. With Pillar 1 (PR 10/11) the gateway tier is now a thin SQS forwarder; with Pillar 2 we cross-process-RESUME instead of IDENTIFYing, dropping the visible blackout to a few hundred ms of TCP drop. Pillar 3 (PR 13b) then adds a hot standby with HMAC push-handoff for near-zero observable disruption.

## Architecture

**API shape note**: the shim exposes `getRest()` and `getAppId()` as methods (no property getters). A future port from an earlier "shim.rest" callsite shape will get `undefined is not a function`, not a silent undefined — grep `getRest()` in `src/index.js` when wiring new callers.

The `ENABLE_GATEWAY_RESUME=true` + `PROCESS_ROLE=gateway` path:

1. Boot: `gatewayShim.hydrate()` reads the persisted DDB row into an in-memory mirror.
2. `gatewayShim.start()` constructs a `WebSocketManager` with `retrieveSessionInfo` (mirror-only) and `updateSessionInfo` (mirror + fire-and-forget DDB write) callbacks. `manager.connect()` either RESUMEs (mirror is populated) or IDENTIFYs (cold start).
3. On every gateway frame: shim's `onDispatch` fans out to a single coalesced handler that (a) triggers `registerCommands` on the first READY, (b) ticks the `gateway-activity` observability gauge, (c) forwards INTERACTION_CREATE to SQS via `event-publisher.publish`.
4. SIGTERM: shim's `stop()` flushes the latest sequence to DDB synchronously, then exits **without** calling `manager.destroy()` — the TCP drop is load-bearing for Discord's ~60 s resume buffer (a clean close frame would invalidate the session).
5. Next process boots, hydrates the row, RESUMEs cleanly.

## Boot guards

Three combinations are rejected at boot (new `unsupportedRoleResumeCombo`):
- `ENABLE_GATEWAY_RESUME=true` + `PROCESS_ROLE=combined` — legacy `Client` owns the WS in combined mode.
- `ENABLE_GATEWAY_RESUME=true` + `ENABLE_EVENT_SHIPPER=false` — shim has no in-process dispatcher.
- `ENABLE_GATEWAY_RESUME=true` + `STORE_TYPE!=ddb` — sqlite can't share state across processes.

## Load-bearing contracts (all unit-tested)

1. **In-memory mirror, not DDB-per-callback.** `retrieveSessionInfo` is called by `@discordjs/ws` on every reconnect — a DDB read per call would surface as ~10-100 ms loop latency AND couldn't observe the in-process state changes that a prior `updateSessionInfo(null)` makes.
2. **Null-clear respected.** When Discord rejects a RESUME, `@discordjs/ws` calls `updateSessionInfo(shardId, null)` and the mirror MUST flip to null AND the DDB row MUST be deleted — otherwise the next `retrieveSessionInfo` hands back the dead session and the bot enters an infinite RESUME-reject loop (this is the contract the PR 0 spike's first sandbox run encountered).
3. **Write throttling.** `updateSessionInfo` fires on every dispatch (high rate). Immediate writes only on sessionId change or after 1 s elapses; other updates defer to a single scheduled flush. Mirror always reflects the latest dispatch; DDB lags by at most 1 s.
4. **Final flush on SIGTERM.** `flushFinal()` awaits any in-flight fire-and-forget write, then synchronously writes the mirror so the next process's hydrate sees the latest sequence.
5. **SIGTERM does NOT call `manager.destroy()`.** Destroy sends close-1000 + clears the session on Discord's side; we need the session preserved for the next process's RESUME.
6. **`MAX_IDENTIFY_ATTEMPTS = 1`** — caps Discord's 1000/24h IDENTIFY budget burn at one per task lifetime; ECS task replacement gives a fresh budget.

## Commit order (each independently reviewable)

1. `feat(discord): add ENABLE_GATEWAY_RESUME flag + boot guards` — flag only, no behavior change.
2. `feat(discord): add DDB-backed gateway session store with mirror` — the persistence + throttle + mirror; pure module, no wiring yet.
3. `feat(discord): add WebSocketManager shim with IDENTIFY budget guard` — wraps @discordjs/ws; not yet wired in.
4. `refactor(discord): registerCommands accepts REST + appId` — pure refactor; both legacy + shim paths can call it.
5. `feat(discord): wire Pillar 2 shim into gateway tier under flag` — the integration; flag-off keeps every existing wire.
6. `test(discord): pillar 2 shim+store end-to-end integration` — composes the two modules through the full lifecycle.
7. `docs(discord): .env.example entry for ENABLE_GATEWAY_RESUME`
8. `refactor(discord): /simplify pass on pillar 2 stack` — decouples DDB writes from WS dispatch loop, collapses dispatch fan-out, drops rot-prone comments.

## Test plan

- [x] Full local jest suite green (2033 tests, +75 new in this PR)
- [x] Lint clean
- [ ] Deploy to sandbox with flag off — verify no behavior change vs current main
- [ ] Flip `ENABLE_GATEWAY_RESUME=true` in sandbox (DDB tables provisioned by qurl-integrations-infra PR #498) — observe READY dispatch + DDB session row populated
- [ ] Force-redeploy sandbox gateway task — observe RESUMED dispatch + sequence carried forward (no IDENTIFY budget burn)
- [ ] Verify `/qurl send` end-to-end during the redeploy (interactions reach the worker via SQS without loss)
- [ ] Sandbox soak: confirm no stale guild-scoped slash commands surface under the shim path — `registerCommands` is now invoked from the shim's first READY (not `client.once('ready')`), so any regression in that wiring would show up as commands from a prior deploy lingering in guild scope
- [ ] Sandbox soak ≥48h before recommending prod flag-on
- [ ] PR 13b lands the second replica before any env raises `gateway_desired_count` above 1

## Known follow-ups

- gateway-heartbeat / active-guild-count timers skip when shim is active (they probe `client.ws.shards[*].lastPingTimestamp` which doesn't exist under @discordjs/ws). Port to shim-aware health snapshot once PR 13b ships.
- Cap `inFlightWrites` Set + add a `gateway_session_store_writes_total` observability counter (issue #411) — currently unbounded under pathological per-dispatch sessionId churn (not a real-world Discord shape, but worth hardening before prod flag-on).

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -112,6 +112,33 @@ ENABLE_OPENNHP_FEATURES=false
 # ENABLE_EVENT_SHIPPER=false
 # QURL_BOT_EVENTS_QUEUE_URL=https://sqs.us-east-2.amazonaws.com/000000000000/qurl-bot-events-sandbox
 
+# Gateway RESUME (zero-downtime design, Pillar 2). When set to the
+# literal string "true" AND PROCESS_ROLE=gateway, the gateway tier
+# replaces the discord.js Client + client.login() with a direct
+# @discordjs/ws WebSocketManager whose session callbacks point at
+# the `${DDB_TABLE_PREFIX}gateway-session` DDB table. On every
+# restart the new gateway boots, reads the persisted session, and
+# Discord's RESUME (op 6) replays buffered events since the last
+# sequence — eliminating the ~10 s IDENTIFY cold-start the legacy
+# single-process deploy carries today.
+#
+# Requires ENABLE_EVENT_SHIPPER=true (the shim only has a forward-
+# to-SQS dispatch path; in-process dispatchers are unreachable).
+# Combined mode is rejected at boot (the legacy Client owns the WS
+# in combined mode; the shim conflicts).
+#
+# Single-replica behavior is unchanged at this layer — desired_count
+# stays at 1. PR 14 adds the second replica + push handoff (Pillar 3).
+#
+# No-op when PROCESS_ROLE=http (worker tier doesn't open a WS); the
+# flag may be uniform across the gateway and worker task defs.
+#
+# Rollback: unset the flag (or set to anything other than "true")
+# and redeploy. The gateway tier reverts to client.login() and the
+# DDB session row goes stale (next "true" boot RESUMEs back to it
+# if Discord's 60 s buffer still has it, otherwise IDENTIFYs).
+# ENABLE_GATEWAY_RESUME=false
+
 # Maximum number of in-flight detached handlers the SQS consumer
 # tolerates before pausing receives (backpressure). When the count
 # reaches the cap, pollOnce returns early and waits 100ms before
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -122,6 +122,50 @@ function unsupportedRoleShipperCombo(role, eventShipperEnabled) {
   return null;
 }
 
+// Parallel to unsupportedRoleShipperCombo for ENABLE_GATEWAY_RESUME.
+// Two unsupported shapes:
+//
+//   1. resume=true with shipper=false — the resume shim
+//      (@discordjs/ws WebSocketManager) replaces discord.js's Client
+//      entirely, so the in-process interaction dispatcher has no
+//      `client.on('interactionCreate')` emitter to attach to. The
+//      flag-on path is only coherent when the shipper has already
+//      moved dispatch to SQS.
+//   2. resume=true with role=combined — combined mode runs both
+//      tiers in one process, which the legacy discord.js Client
+//      owns end-to-end. The resume shim would conflict with the
+//      Client's WS ownership.
+//
+// (combined + shipper=true is independently rejected by
+// unsupportedRoleShipperCombo; resume's combined-mode rejection
+// catches the remaining combined+shipper-off+resume-on case.)
+//
+// Returns the operator-facing message on rejection or null on
+// success. Same string-or-null shape as unsupportedRoleShipperCombo.
+function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled) {
+  if (!resumeEnabled) return null;
+  if (role === 'combined') {
+    return (
+      'PROCESS_ROLE=combined with ENABLE_GATEWAY_RESUME=true is not supported ' +
+      '(the resume shim owns the WebSocket and conflicts with the legacy ' +
+      'discord.js Client that combined mode runs). Run two processes: ' +
+      'PROCESS_ROLE=gateway (singleton, owns the resume shim) + ' +
+      'PROCESS_ROLE=http (consumes from SQS). For local dev / sandbox in ' +
+      'one process, leave ENABLE_GATEWAY_RESUME unset.'
+    );
+  }
+  if (!eventShipperEnabled) {
+    return (
+      'ENABLE_GATEWAY_RESUME=true requires ENABLE_EVENT_SHIPPER=true ' +
+      '(the resume shim replaces discord.js Client with @discordjs/ws ' +
+      'WebSocketManager, which forwards every frame to SQS — the ' +
+      'in-process dispatcher path is unreachable from the shim). ' +
+      'Enable the shipper first, or leave ENABLE_GATEWAY_RESUME unset.'
+    );
+  }
+  return null;
+}
+
 // PLACEHOLDER is treated as missing because the SSM parameter
 // ships with that literal sentinel value; remediation ("seed a
 // real key") is identical to the empty-key case.
@@ -219,6 +263,7 @@ module.exports = {
   missingKekRequiredKeys,
   missingEventShipperKeys,
   unsupportedRoleShipperCombo,
+  unsupportedRoleResumeCombo,
   shouldRegisterInteractionListener,
   missingMapCommandKeys,
   GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL,
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -123,26 +123,29 @@ function unsupportedRoleShipperCombo(role, eventShipperEnabled) {
 }
 
 // Parallel to unsupportedRoleShipperCombo for ENABLE_GATEWAY_RESUME.
-// Two unsupported shapes:
+// Three unsupported shapes:
 //
-//   1. resume=true with shipper=false — the resume shim
+//   1. resume=true with role=combined — combined mode runs both
+//      tiers in one process, which the legacy discord.js Client
+//      owns end-to-end. The resume shim would conflict with the
+//      Client's WS ownership.
+//   2. resume=true with shipper=false — the resume shim
 //      (@discordjs/ws WebSocketManager) replaces discord.js's Client
 //      entirely, so the in-process interaction dispatcher has no
 //      `client.on('interactionCreate')` emitter to attach to. The
 //      flag-on path is only coherent when the shipper has already
 //      moved dispatch to SQS.
-//   2. resume=true with role=combined — combined mode runs both
-//      tiers in one process, which the legacy discord.js Client
-//      owns end-to-end. The resume shim would conflict with the
-//      Client's WS ownership.
-//
-// (combined + shipper=true is independently rejected by
-// unsupportedRoleShipperCombo; resume's combined-mode rejection
-// catches the remaining combined+shipper-off+resume-on case.)
+//   3. resume=true with storeType!=ddb — the resume guarantee only
+//      holds when session state is persisted across processes.
+//      The default sqlite backend writes a local file that the
+//      next ECS task won't see; a resume against the previous
+//      sequence would fail every restart. Rejecting at boot is
+//      preferable to a silent IDENTIFY-every-restart degradation
+//      that mimics flag-off behavior.
 //
 // Returns the operator-facing message on rejection or null on
 // success. Same string-or-null shape as unsupportedRoleShipperCombo.
-function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled) {
+function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled, storeType) {
   if (!resumeEnabled) return null;
   if (role === 'combined') {
     return (
@@ -161,6 +164,14 @@ function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled) {
       'WebSocketManager, which forwards every frame to SQS — the ' +
       'in-process dispatcher path is unreachable from the shim). ' +
       'Enable the shipper first, or leave ENABLE_GATEWAY_RESUME unset.'
+    );
+  }
+  if (storeType !== 'ddb') {
+    return (
+      `ENABLE_GATEWAY_RESUME=true requires STORE_TYPE=ddb (got '${storeType ?? 'unset (defaults to sqlite)'}'). ` +
+      'Cross-process RESUME persists session state to the gateway-session DDB ' +
+      'table; sqlite would write a local file the next process never sees. Set ' +
+      'STORE_TYPE=ddb in the deployment template, or leave ENABLE_GATEWAY_RESUME unset.'
     );
   }
   return null;
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -158,12 +158,17 @@ function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled, st
     );
   }
   if (!eventShipperEnabled) {
+    // Role-neutral framing: the same env may be applied uniformly
+    // across gateway + http task defs, and an http operator reading
+    // this message shouldn't see gateway-tier-specific language
+    // (the shim is never constructed on http; the rejection here
+    // is a consistency canary, not a description of what http does).
     return (
-      'ENABLE_GATEWAY_RESUME=true requires ENABLE_EVENT_SHIPPER=true ' +
-      '(the resume shim replaces discord.js Client with @discordjs/ws ' +
-      'WebSocketManager, which forwards every frame to SQS — the ' +
-      'in-process dispatcher path is unreachable from the shim). ' +
-      'Enable the shipper first, or leave ENABLE_GATEWAY_RESUME unset.'
+      'ENABLE_GATEWAY_RESUME=true requires ENABLE_EVENT_SHIPPER=true. ' +
+      'The two flags co-design: the gateway tier forwards every Discord ' +
+      'frame to SQS while the worker tier consumes it, so a resume path ' +
+      'without the shipper-shape split has nowhere to dispatch. Enable ' +
+      'the shipper first, or leave ENABLE_GATEWAY_RESUME unset.'
     );
   }
   if (storeType !== 'ddb') {
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -168,7 +168,7 @@ function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled, st
   }
   if (storeType !== 'ddb') {
     return (
-      `ENABLE_GATEWAY_RESUME=true requires STORE_TYPE=ddb (got '${storeType ?? 'unset (defaults to sqlite)'}'). ` +
+      `ENABLE_GATEWAY_RESUME=true requires STORE_TYPE=ddb (got '${storeType}'). ` +
       'Cross-process RESUME persists session state to the gateway-session DDB ' +
       'table; sqlite would write a local file the next process never sees. Set ' +
       'STORE_TYPE=ddb in the deployment template, or leave ENABLE_GATEWAY_RESUME unset.'
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -14,6 +14,7 @@ const {
   TextInputStyle,
   AttachmentBuilder,
 } = require('discord.js');
+const { Routes } = require('discord-api-types/v10');
 const crypto = require('crypto');
 const config = require('./config');
 const db = require('./store');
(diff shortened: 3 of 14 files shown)


justin-layerv:
## cr-r1 response

Thanks for the detailed review. Pushed `ddb2bf8` addressing every item:

| cr-r1 item | Severity | Action |
|---|---|---|
| 1. `MAX_IDENTIFY_ATTEMPTS=1` crash-loop on resume-buffer expiry | High | **Reset-on-READY**: counter resets to 0 on every successful READY dispatch. Cap=1 stays honest for the loop-without-READY case (token contention), but reconnects-after-outage now get a fresh budget. New test pins the (cold → READY → reset → outage IDENTIFY → READY-cap-trip) sequence. |
| 2. `inFlightWrite` tracked only the most recent | Medium | **Set-based tracker**: every fire-and-forget write goes into `inFlightWrites` and removes itself on settle; `flushFinal` awaits `Promise.allSettled([...inFlightWrites])`. New test pins SIGTERM-doesn't-exit-while-earlier-write-pending. |
| 3. Module-load DDB env validation order is implicit | Medium | Explicit comment in `index.js` calling out the load-bearing `require('./store')` import-order invariant. |
| 4. `'unset (defaults to sqlite)'` fallback unreachable | Suggestion | Dropped — `config.STORE_TYPE` always coerces unset → `'sqlite'` before the helper sees it. |
| 5. Lift `STORE_TYPE` into `config.js` | Suggestion | Done. One parsing spot for both this PR's boot guard and the upcoming PR 13b paths. |
| 6. `manager.removeAllListeners()` in `stop()` | Nit | Done. Hygiene for non-exiting callers (tests, future shapes). |

On the sandbox-soak suggestion: incorporated into the PR test plan — the 60s+ egress-block-to-Discord test will be part of the sandbox validation between flag-flip and prod recommendation. Watching `gateway-session-store: write failed` warn pattern as a DDB-throttling canary.

Tests: 2027 local, lint clean. Waiting for cr-r2.

justin-layerv:
## cr-r2 response

Pushed `78de454` addressing every actionable item from the second-pass review:

| cr-r2 item | Severity | Action |
|---|---|---|
| Scope `manager.removeAllListeners()` to Dispatch + Error only | Suggestion | Done. Unscoped form risked stripping internal @discordjs/ws self-listeners on a future library version. New test pins that a foreign-event listener installed before `stop()` survives unchanged. |
| `fireWrite` shape guarantee | Suggestion | Done. `Promise.resolve(promise).catch(() => {})` inside `fireWrite` enforces the no-unhandled-rejection contract independent of caller discipline. |
| `DEFAULT_SHARD_ID` duplicates `config.SHARD_ID` | DRY | Dropped the local constant; consume `config.SHARD_ID` (existing at `src/config.js:189`, same `'0:1'` default). |
| `stop()` idempotency | Suggestion | Made stop() return early on the second call. A graceful-shutdown SIGTERM/SIGINT race now flushes once. New test pins this. |
| `getRest()` is a method, not a getter | Documentation | API shape called out in the [PR description](https://github.com/layervai/qurl-integrations/pull/410) (top under "Architecture" notes). |
| Unbounded `inFlightWrites` Set + sustained-failure metric | Out of scope | Filed as **#411** (cap + CloudWatch metric for in-flight Set size). The DDB-outage envelope where this matters already breaks cross-process RESUME entirely, so the realistic correctness envelope of this PR isn't degraded — but the failure mode should be observable, not latent. |
| `unsupportedRoleResumeCombo` string-or-null ergonomics | Stylistic | Skipped — three callers do not justify a shape change in this PR. |

Verification round confirmed all cr-r1 fixes landed correctly (reset-on-READY state-machine walk, Set-based tracker, scoped-comment polish). The new `manager.removeAllListeners()` scoping and the idempotent stop() are the only behavioral changes this round; everything else is hygiene.

Tests: 2028 local, lint clean. Waiting for cr-r3.

justin-layerv:
## cr-r3 response

**Critical bug caught.** Thank you — pushed `9cfb582` fixing the RESUMED-doesn't-flip-isReady regression that would have broken Pillar 2's production happy path.

| cr-r3 item | Severity | Action |
|---|---|---|
| 🔴 `isReady` never flips on RESUMED | Critical | Added the `RESUMED` branch in the shim's Dispatch listener: sets `isReady=true`, resets `identifyAttempts=0` (symmetric with READY-reset so a later disconnect-reconnect after a successful resume gets a fresh budget). `appId` stays as-is — doesn't change between processes for the same bot identity. New unit test pins the RESUMED → isReady=true contract; warm-start integration test now also asserts `isReady` flips on RESUMED so a future regression surfaces in both unit + integration coverage. |
| 🟡 `manager.connect()` timeout — lifecycle ownership | Minor | Added comment that manager lifecycle is owned by `stop()`; connect-timeout itself doesn't destroy the manager (per the SIGTERM contract), and the caller's `gracefulShutdown(1)` → `process.exit()` unwinds the dangling promise. |
| 🟡 `DDB_TABLE_PREFIX` raw vs trimmed | Minor | Lifted `DDB_TABLE_PREFIX` into `config.js` with the same `.trim()` normalization ddb-store.js uses. Gateway-shim construction now reads through `config.DDB_TABLE_PREFIX`; one normalization point. |
| 🟡 Integration test `await` on sync calls | Minor | Dropped the awaits in `tests/gateway-resume-integration.test.js`. The callback returns synchronously now; the awaits were cosmetically misleading about the dispatch-loop back-pressure analysis. |
| 🟡 `unsupportedRoleResumeCombo` parameter order (options bag) | Stylistic | Deferred — three callers and the parameter types are mostly distinguishable today. Worth revisiting if PR 13b adds more parameters. |

The `isReady`-on-RESUMED fix is the load-bearing one. The other items are hygiene.

Tests: 2030 local, lint clean. Waiting for cr-r4.

justin-layerv:
## cr-r4 response

Pushed `d53b00d` addressing every item:

| cr-r4 item | Severity | Action |
|---|---|---|
| 🟡 Dead `useFakeTimers` toggle in throttle test | Cleanup | Removed. Point covered by the explicit-fake-timer test immediately below. |
| 🟡 `start()` partial-failure: late dispatch race | Correctness | Two-part guard: Dispatch listener now returns early on `stopped`; AND `start()`'s catch sets `stopped=true` BEFORE rethrowing so any frame arriving in the throw → `gracefulShutdown` → `stop()` window drops. New test pins the race-guard contract. |
| 🟡 `buildRetrieveCallback` synchronous throw | Verified | Read the installed `@discordjs/ws/dist/index.js` — `retrieveSessionInfo` is `await`ed at line 674 inside the library's `connect()` flow. Synchronous throw converts to a rejected Promise that propagates up through `manager.connect()` → `shim.start()`. The unit test covers the contract via mock; the library's own call surface honors it. No code change. |
| 🟡 `gracefulShutdown` else-arm | Defensive | Changed `else if (isGateway)` → `else if (isGateway && !ENABLE_GATEWAY_RESUME)`. A future shim-construction wrapping in try/catch can no longer silently fall through to `discordShutdown()`. |
| 🟢 `slowFakeManager` construction cleanup | Quality | Refactored to a tidy `makeSlowManagerCtor()` factory (Object.assign on EventEmitter); both timeout tests share it. |
| 🟢 `_constructorArgs` vs `_args` naming | Quality | Aligned the integration test's fake to use `_constructorArgs` (matches the unit-test fake). |
| 🟢 STORE_TYPE typo handling | ✓ | Confirmed — operator-facing message surfaces `got '${storeType}'` literal. |
| 🔴 Malformed-row log leaks `resume_url` | Security | Redacted. Warn payload now emits only the type signature (`{session_id: 'string', resume_url: 'string', sequence: 'string'}`) — never the actual values. `session_id` + `resume_url` are session-bound credentials reachable by anyone holding them within Discord's ~60 s resume window; CloudWatch retention was a real leak vector on the cold-start error path. New test pins the redaction contract end-to-end (asserts `JSON.stringify(payload)` contains neither the leaky session_id nor resume_url). |
| 🟢 Pre-READY `updateSessionInfo` ordering | Stylistic | Skipped — the cold-start integration test exercises updateSessionInfo from the manager-callback wiring, not directly. Production order is also "library calls updateSessionInfo as the READY arrives," which matches the test sequence. |

Misc:
- WCU for delete capacity (#8): infra-side, raised with the qurl-integrations-infra DDB module owners.
- `pendingFlush.unref()` GC concerns (#9): the existing event-publisher uses the same pattern at higher dispatch rates without issue. No action.

Tests: 2031 local, lint clean. Waiting for cr-r5.

justin-layerv:
## cr-r5 response

Pushed `6bc0351` addressing every item:

| cr-r5 item | Severity | Action |
|---|---|---|
| 🔴 start-failure listener leak (cleanup unreachable) | Correctness | Split `stopped` (drop-late-dispatches gate) from `stopCompleted` (stop() idempotency gate). start()'s catch sets only `stopped`; stop()'s early-return now checks `stopCompleted`. The failed-start → gracefulShutdown → shim.stop() chain now reaches flushFinal + listener detach + manager=null. New test pins this end-to-end. |
| 🟡 http-tier rejection message mentions gateway-shim internals | UX | Reworded to role-neutral framing: "ENABLE_GATEWAY_RESUME=true requires ENABLE_EVENT_SHIPPER=true. The two flags co-design: the gateway tier forwards every Discord frame to SQS while the worker tier consumes it..." — no more "shim replaces Client" verbiage in messages an http operator might read. Test updated to pin the role-neutral wording. |
| 🟢 Cursor write-attempt accounting under sustained DDB fail | Observability | Captured in #411 (inFlightWrites cap + CloudWatch metric for in-flight size). The runbook for the sandbox soak should treat `gateway-session-store: write failed` warn count as a leading indicator. |
| 🟢 Test-only `_get*ForTest` accessor naming | Cosmetic | Skipped — matches the established pattern in this repo (event-publisher.js has the same prefix). Worth a project-wide cleanup PR if anyone wants the `__` / `Symbol` form, but not for this scope. |
| 🟢 `registerCommands` doesn't dedupe on forced re-IDENTIFY | Cosmetic | Noted. A forced re-IDENTIFY mid-process is rare enough that one duplicate Discord call per occurrence is fine; the global-commands endpoint is idempotent on identical payloads. |
| 🟢 `getRest()` after stop() | Cosmetic | Noted. No connection-pool state on REST today; tighten if that changes. |

The start-fail cleanup bug was the load-bearing one. Tests: 2032 local, lint clean. Waiting for cr-r6.

justin-layerv:
## cr-r6 response (6777128)

cr-r6 was "Approve with two nudges as non-blocking items". Closing out each:

**(a) `GATEWAY_INTENTS_BITFIELD` — guard against future intent-bit ≥ 2³¹**

Initially swapped to `new IntentsBitField(intents).bitfield` (the BigInt-backed form `@discordjs/ws` also accepts) since it's the canonical discord.js way. That broke ~6 unrelated test files — the existing test suite mocks `discord.js` without an `IntentsBitField` constructor (`tests/discord.test.js`, `tests/opennhp-gating.test.js`, `tests/build-delivery-embed.test.js`, etc.), and those files would all need their mocks extended.

Since the reviewer flagged this as **Low priority** and Discord's highest declared intent bit today (`AutoModerationConfiguration = 1 << 20`) is well under 2³¹, keeping the int32 bitwise OR is correct for now. Tightened the comment to spell out the safety condition AND the one-line swap when motivated:

```js
// Int32 bitwise OR is safe today — Discord's highest declared
// intent bit (GuildVoiceStates = 128, GuildMessageTyping = 16384,
// AutoModerationConfiguration = 1048576, etc.) is well under
// 2³⁰. If Discord ever adds an intent bit ≥ 2³¹, switch to
// `new IntentsBitField(intents).bitfield` (BigInt-backed) — the
// @discordjs/ws WebSocketManager constructor accepts both number
// and bigint forms.
```

**(b) Trim stale STORE_TYPE follow-up from PR description**

Done — that follow-up was already addressed in cr-r4 (lifted into `config.js` with `.trim()` normalization). Removed the stale bullet and added the in-flight-writes capping follow-up (issue #411) in its place so the known-follow-ups section reflects current state.

**(c) Sandbox-soak: stale guild-scoped commands**

Added an explicit test-plan item. The risk is real: `registerCommands` is now invoked from the shim's first READY rather than the legacy `client.once('ready')`, so any regression there would surface as commands from a prior deploy lingering in guild scope. Worth eyeballing during the sandbox soak rather than relying on unit tests alone.

**Test name nit — put-after-delete shape coverage**

Added a dedicated test (`awaits an in-flight null-clear delete chased by a fresh-session put`) alongside the existing two-consecutive-puts test. This pins the operationally-relevant shape: a RESUME rejection (`updateSessionInfo(null)` → Delete) immediately followed by a fresh IDENTIFY's READY (`updateSessionInfo({...})` → Put), both still in-flight when `flushFinal()` is called. `Promise.allSettled` has to await both — otherwise the process exits mid-Delete and the next hydrate sees a row that should have been cleared.

Tests at 2033, lint clean. Awaiting cr re-review on `6777128`.

justin-layerv:
## cr-r7 response (84f2d54)

cr-r7 surfaced two new findings beyond cr-r6 polish. Closing out each plus the suggested test:

**(1) registerCommands re-fires on every READY under shim — divergence from `client.once('ready', …)`**

Real bug. In production, a >60s outage expires Discord's resume buffer → RESUME rejected → fresh IDENTIFY → fresh READY in the same process. Without a once-flag, `registerCommands` would re-PUT ~150 global commands every resume-buffer-expiry reconnect. Added `commandsRegistered` closure-flag at `src/index.js:528` that matches the repo's existing once-guard idiom (`stopCompleted` in `gateway-ws-shim.js`, `allDone` in `commands.js` — searched, no shared `once()` utility to extract).

The new comment block at the call site captures the WHY in 5 lines (trimmed from a 13-line draft after self-simplify flagged it as over-narration).

**(2) `getAppId()` null on pure-RESUMED boots**

Took the doc-comment + call-site-guard route (the reviewer's option (b)) rather than persisting appId in the session row — schema migration would add scope without changing behavior. The single call site is gated on `data?.t === 'READY'`, so the contract is safe today; documenting it explicitly prevents a future caller from accidentally routing through `/applications/null/commands`.

The call-site null-guard at `src/index.js:531-540` skips registration with a warn log if appId is missing AND deliberately does NOT latch `commandsRegistered` — a subsequent well-formed READY can still recover.

**(3) `lastWriteAt` cursor failed-write semantics**

Extended the `fireWrite` comment in `gateway-session-store.js:117-126` with the intentional-defer-to-throttle rationale: a failed READY write doesn't burst-retry; the mirror holds the correct state and the next throttle-driven flush pushes DDB back into sync within ~1s. Burst-retrying would amplify a DDB outage into a hot loop without changing the eventual outcome.

**(4) `lastWriteAt`-after-null-clear ordering — deferred**

The branch ordering (sessionChanged check before throttle check) is already implicitly covered by the existing integration test that fires put-after-null-clear and observes the immediate write. Adding a defensive test specifically pinning that ordering would test the test's reimplementation of the production wiring more than the wiring itself. Calling this out rather than silently skipping.

**Suggested test — registerCommands NOT re-run on RESUMED**

Added `apps/discord/tests/gateway-resume-integration.test.js`'s "registerCommands fires only on the first READY per process" test. Three assertions:
1. First READY → registerCommands fires
2. RESUMED dispatch → NOT fires
3. Second READY (resume-buffer-expired path) → NOT fires

Mirrors the index.js wiring directly so a refactor that breaks the once-flag would surface here.

**Pre-merge sanity checks** (cr-r7 §Pre-merge)

- `@discordjs/ws` and `@discordjs/rest` confirmed as runtime deps in `apps/discord/package.json` (`~1.2.3`, `~2.6.0`).

**Other cr-r7 items — not actionable in this PR**

- §Code quality — `DDB_TEST_ENDPOINT` duplication across `index.js`/`ddb-store.js`/`flow-state.js`: pre-existing pattern, the reviewer flagged it as out-of-scope for this PR. Worth a follow-up issue if/when the `shared/` equivalent for the Node app lands.

Tests at 2034 (+1 new), lint clean. Awaiting cr re-review on `84f2d54`.

justin-layerv:
## cr-r8 response (26fc4cb)

cr-r8 was "ship after sandbox soak" with five nit-level items + a suggested test. Closing each:

**(1) `MAX_IDENTIFY_ATTEMPTS=1` brittle to library quirks**

Documented the assumption that `@discordjs/ws` invokes `retrieveSessionInfo` exactly once per IDENTIFY decision, with explicit "future minor that adds a pre-flight retrieve would make a cold-start throw `GATEWAY_IDENTIFY_BUDGET` before READY ever lands" callout. The `~1.2.x` pin in `package.json` is the defense — that's pinned now via the tilde range, so a `1.3.x` would require an explicit bump (which would trip /simplify's dep-bump trail).

**(2) `retrieveSessionInfo` returned mirror by reference**

Real concern — shallow-clone on the way out closes the mutation window. `return mirror ? { ...mirror } : null;`. Cost is one 3-field object per reconnect (not per dispatch — `retrieveSessionInfo` only fires during the IDENTIFY/RESUME decision), so the defensive copy is effectively free.

**(3) `purgeStaleGuildCommands` skipped under shim** — already acknowledged in PR description and covered by the sandbox-soak test-plan item added in cr-r6 (c). No code change.

**(4) `start()` failure leaves `restInstance` populated**

Nulled `restInstance` alongside the existing `stopped=true` flip in start()'s catch (`gateway-ws-shim.js:340`). The reviewer's concern was forward-looking: a future caller that retries via "shim already has REST, just re-call connect()" would skip the proper construction path. Nulling makes the post-failure state explicit. No production caller affected — confirmed no existing test calls `getRest()` after a failed start (existing tests at `gateway-ws-shim.test.js:564-581` all gate on successful start).

**(5) `fireWrite`'s outer `.catch(() => {})`**

Extended the existing comment block to spell out the two-layer responsibility: inner `.catch` in `firePersist`/null-clear LOGS the rejection (operator-visible failure surface), outer `.catch` here is purely a shape guarantee. The new comment explicitly warns against dropping the outer catch — a future call site that passes a raw `persistRow`/`deleteRow` promise without wrapping would silently break the unhandled-rejection defense.

**(6) `flushFinal` not idempotent**

Added a `flushed` flag with an idempotency guard at the top of `flushFinal()`. The shim's `stop()` is already idempotent in production via `stopCompleted`, but a non-shim caller that drops the coordinator (or a test that double-flushes) could re-write the mirror without this. Both the production path and the defensive backstop now hold.

**(8) Suggested test — late dispatch during stop() teardown**

Added `apps/discord/tests/gateway-ws-shim.test.js`'s "drops dispatches that arrive during stop() teardown (between flag-flip and listener detach)" test. Symmetric to the existing connect-timeout late-dispatch test. Belt-and-suspenders coverage: pins both the `stopped` flag guard AND the cleared `dispatchHandlers` Set hold during the `flushFinal()`-await window. Also added a `flushFinal` idempotency test in `gateway-session-store.test.js`.

**Other cr-r8 items not actionable**

- §Security and §CLAUDE.md compliance items confirmed clean, no action needed.

Tests at 2036 (+2 new), lint clean. Awaiting cr re-review on `26fc4cb`.

PR #412, opened 2026-05-16
feat(discord): pillar 3 lock + peer-heartbeat primitives (pure modules)

justin-layerv:
## Summary

First of three PRs splitting Pillar 3 (hot-standby gateway pair). This one adds two pure DDB modules — `gateway-lock.js` and `gateway-peer-heartbeat.js` — with **no wiring**. Flag-off, flag-on, and runtime behavior are unchanged. The leader coordinator (PR 13b.2) composes them; the index.js integration (PR 13b.3) puts the whole stack behind a default-off `ENABLE_GATEWAY_HOT_STANDBY` flag.

## Why split

Pillar 3's app side is genuinely bigger than Pillar 2 — lock primitive + heartbeat primitive + HMAC control-channel HTTP server + leader coordinator with watchdog + index.js wiring + boot-requirements. As one PR it lands as 3-5k LOC and a cr loop nightmare. The split lets each surface ship with focused review.

### `src/gateway-lock.js`
Per-shard leader-election lock over the `gateway_lock` DDB table (PK = `shard_id`). Five load-bearing contracts:

1. **Conditional-write IS the lock primitive** (TTL is janitor only). DDB TTL deletion is async with a multi-day SLA — every code path must trust the `ConditionExpression` on every write, never trust TTL absence.
2. **TTL writer shape is epoch SECONDS**, never milliseconds. A ms-encoded value would land ~50,000 years in the future to the DDB reaper.
3. **`instance_id` + `version` CAS guards renew/transfer** against the stale-task-vs-peer-took-over race. Version is the fencing token (cf. Kleppmann).
4. **Release uses `DeleteItem`**, not `UpdateItem REMOVE`. REMOVE would leave `expires_at` populated, forcing a peer's next acquire to wait for the lease to lapse. DeleteItem re-arms the `attribute_not_exists` branch.
5. **Acquire uses `PutItem`** (single round-trip for the post-Delete row-absent case; idempotent on cond fail).

### `src/gateway-peer-heartbeat.js`
Per-replica heartbeat over the `gateway_peer_heartbeat` DDB table (PK = `instance_id`). Four load-bearing contracts:

1. **Freshness filter at READ time IS the correctness primitive**, NOT DDB TTL. A row past `updated_at + freshnessWindowSeconds` must be invisible to the active even if DDB TTL hasn't reaped it.
2. **Single PutItem per renewal** — `updated_at` + `expires_at` written together. Splitting them creates a partial-state window.
3. **TTL writer shape is epoch SECONDS** (same convention as `flow_state` / `gateway_lock`).
4. **Scan + filter is the correct access pattern** (small row count, sub-second consistency, ConsistentRead doubles RCU for no correctness benefit).

## Test plan

- [x] Full local jest suite green (2075 tests, +41 new in this PR)
- [x] Lint clean
- [ ] PR 13b.2 (control-channel + leader coordinator) consumes these modules
- [ ] PR 13b.3 (index.js wiring) puts the whole stack behind `ENABLE_GATEWAY_HOT_STANDBY` (default off)

## Architecture

This PR is **pure additive code** — two new src modules, two new test files, zero imports from existing src. Module pre-tested in isolation; integration validation arrives with PR 13b.2.

The conditional-write contracts mirror the table comments in [qurl-integrations-infra PR #617](https://github.com/layervai/qurl-integrations-infra/pull/617) (lock table) and [PR #620](https://github.com/layervai/qurl-integrations-infra/pull/620) (peer-heartbeat table) — see those for the field-by-field rationale. The design doc at `apps/discord/docs/zero-downtime-design.md` §Pillar 3 is the authoritative source for the lease cadence (2s renew / 6s TTL) and freshness window (6s).

## Pillar 3 PR sequence

- **PR 13b.1 (this PR)**: `gateway-lock.js` + `gateway-peer-heartbeat.js` (pure modules)
- **PR 13b.2**: control-channel HTTP server (HMAC verify + nonce LRU + freshness check) + leader coordinator + connection-watchdog
- **PR 13b.3**: `index.js` wiring + `boot-requirements` combo guards + `.env.example` + flag default-off

### DIFF
--- a/apps/discord/src/gateway-lock.js
+++ b/apps/discord/src/gateway-lock.js
@@ -9,6 +9,31 @@
 // version cursor. Tests run isolated; a future multi-shard caller can
 // construct one per shard.
 //
+// ── Concurrency: single-caller only ──
+//
+// `currentVersion` is closure state mutated by acquireLock / renewLock
+// / transferLock / adoptLockFromHandoff. JS's single-threaded event
+// loop makes each read-then-write atomic at the await boundary, but
+// two concurrent renewLock invocations (e.g., a scheduled tick
+// overlapping a watchdog-triggered renew) would both read version
+// `N`, both attempt to write `N+1`, and the second's CAS fails —
+// clearing `currentVersion=null` even though the lock is still held.
+// The leader-coordinator in PR 13b.2 must serialize all mutator
+// calls (one outstanding renew at a time; transfer/release run only
+// from the SIGTERM path which has the loop stopped). The diagnostic
+// `readCurrentHolder` is the only safe overlap.
+//
+// ── Release semantics ──
+//
+// `releaseLock` is intended for the no-peer SIGTERM path (active dies
+// with no standby to hand to). Both transport errors and CAS failures
+// clear `currentVersion=null` so the caller's state doesn't outlive
+// the lock. This is correct on the SIGTERM path (process is exiting)
+// but would be over-aggressive on a voluntary-step-down path — a
+// future watchdog-driven release would need to distinguish "lock is
+// gone" from "DDB transient error, retry" rather than zero the
+// cursor on either. Today's only caller is gracefulShutdown.
+//
 // ── Load-bearing contracts ──
 //
 // 1. Conditional-write is the lock primitive, NOT DDB TTL. The TTL
@@ -160,15 +185,20 @@ function createGatewayLock({
     const now = nowSeconds();
     const nextVersion = currentVersion + 1;
     try {
+      // No `lock_holder = :holder` in SET — we already hold the
+      // lock, so lock_holder is unchanged. Including it would
+      // burn ~1 WCU byte per renew with no semantic effect.
+      // (acquireLock writes the full row; transferLock writes
+      // the new holder. Both code paths that need to set
+      // lock_holder do so explicitly.)
       await ddbClient.send(new UpdateCommand({
         TableName: tableName,
         Key: { shard_id: shardId },
-        UpdateExpression: 'SET version = :next, expires_at = :exp, lock_holder = :holder',
+        UpdateExpression: 'SET version = :next, expires_at = :exp',
         ConditionExpression: 'instance_id = :self AND version = :expected',
         ExpressionAttributeValues: {
           ':next': nextVersion,
           ':exp': now + ttlSeconds,
-          ':holder': lockHolder,
           ':self': instanceId,
           ':expected': currentVersion,
         },
--- a/apps/discord/src/gateway-lock.js
+++ b/apps/discord/src/gateway-lock.js
@@ -37,6 +37,18 @@
 //    take it while the legitimate holder is still heartbeating;
 //    their write hits a ConditionalCheckFailedException.
 //
+//    Scope note: `version` is scoped to the CURRENT lock tenancy.
+//    Each `acquireLock` after a release / lapse / transfer resets
+//    the counter to 1; it is NOT a globally monotonic fencing token
+//    in the Kleppmann-treats-it-as-external sense. The internal-CAS
+//    use is sound because every CAS pairs `version` with `instance_id`,
+//    and `instance_id` is unique per process. If any downstream
+//    system ever consumes `version` as an EXTERNAL fencing token
+//    (e.g., a separate storage system that wants "is this fence
+//    number ahead of the last one we saw"), the per-acquire reset
+//    breaks that assumption and the consumer would need to pair
+//    `version` with `instance_id` or read the holder row.
+//
 // 4. Release uses `DeleteItem`, not `UpdateItem REMOVE lock_holder`.
 //    A REMOVE would leave `expires_at` populated, so a peer's next
 //    acquire couldn't take the `attribute_not_exists(lock_holder)`
@@ -189,6 +201,17 @@ function createGatewayLock({
       });
       return { transferred: false };
     }
+    if (targetInstanceId === instanceId) {
+      // Self-handoff is meaningless and would still bump the version
+      // (the CAS would succeed, since we're a holder). Reject at the
+      // API boundary so a caller bug (e.g., a peer-discovery lookup
+      // that accidentally returns our own row) doesn't silently churn
+      // the version counter.
+      logger.warn('gateway-lock: transferLock called with self as target (no-op)', {
+        shardId, instanceId,
+      });
+      return { transferred: false };
+    }
     const now = nowSeconds();
     const nextVersion = currentVersion + 1;
     try {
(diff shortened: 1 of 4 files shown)


justin-layerv:
## cr-r1 response (bbc35a8)

Thanks for the careful review. Closing out each item:

**Design gap — post-transfer version sync**

Real bug, addressed by adding `adoptLockFromHandoff(versionAfterTransfer)` to the gateway-lock API. Pure local cursor sync, no DDB write — PR 13b.2's control-channel handler will call this after HMAC-verifying the push body and observing the active's `transferLock` has succeeded. The HMAC body already carries the version per design doc §Pillar 3, so the value is on hand.

Four new tests pin the contract: cursor seeded so next `renewLock`'s `:expected` uses the adopted value; no DDB write; rejects non-positive-integer (defends against malformed HMAC body); idempotent (safe to retry on transient errors).

**Input validation — port**

Tightened to `Number.isInteger(port) && port > 0 && port < 65536`. The original accepted NaN, 0, negatives, fractionals, >65535 — all of which would produce unreachable peer entries that surface at the first handoff POST. Now fails at construction.

Skipped the IP validation — agreed it's out of scope here; PR 13b.3 will validate as part of `config.js` env parsing.

**Test gaps — all pinned**

- Strict `<` boundary on `expires_at < :now`: pinned via cond-text inspection. A future refactor to `<=` would shave 1s off recovery at the cost of a clock-skew race against the legitimate holder.
- `transferLock(self → self)`: rejected at the API boundary with a no-op + warn. A caller bug — e.g., peer-discovery accidentally returns our own row — could otherwise churn the version counter without changing ownership. Test pins both the warn AND the no-DDB-write behavior.
- Re-acquire-while-holding: pinned as soft no-op. Cond `expires_at < :now` fails against own live lease, returns `acquired:false`, cursor unchanged.

**Nits**

- **Version scope clarification**: Module-header comment now spells out that `version` is scoped to the current lock tenancy (per-acquire reset to 1), not a globally monotonic external fencing token. The internal-CAS use is sound because every CAS pairs `version` with `instance_id`. The Kleppmann reference stands for the internal use; the external-system caveat is documented.
- **listFreshPeers triple-filter → single-pass**: Skipped. Unmeasurable on a handful of rows and the multi-pass form reads cleaner.
- **Log brand**: Skipped. The lowercase `gateway-lock:` / `gateway-peer-heartbeat:` prefixes match the existing module pattern (`gateway-session-store:`, `gateway-ws-shim:`, etc.). Changing here would create asymmetry without operator-facing benefit (these are identifier-style log prefixes, not user-visible prose).

Tests: 2082 (+7 new vs. cr-r1 baseline), lint clean. Awaiting cr re-review on `bbc35a8`.

justin-layerv:
## cr-r2 response (810006f)

Thanks — acted on all 7 items per pr-ready.

**(1) Concurrency on `currentVersion`**

Added a "Concurrency: single-caller only" section to the module header spelling out the race (two concurrent renews both read version N, both attempt N+1, second's CAS fails and zeros the cursor while the lock is still held). Documents the constraint PR 13b.2's leader-coordinator must honor — serialize all mutator calls, only `readCurrentHolder` is safe overlap.

**(2) Redundant `lock_holder` rewrite in renewLock**

Dropped `:holder` from both `UpdateExpression` and EAV. We already hold the lock; lock_holder is unchanged. Test pins both the slimmed expression (`SET version = :next, expires_at = :exp`) AND the absence of `:holder` so accidental reintroduction fails loud.

**(3) `releaseLock` clears `currentVersion=null` on transport error**

Documented in the new "Release semantics" header section: intentional for the SIGTERM path (process exiting anyway), would be over-aggressive on a future voluntary-step-down path. Today's only caller is `gracefulShutdown`. PR 13b.2 needs to revisit if it wires release onto a recoverable path.

**(4) PutItem caveat**

Praise noted — no action.

**(5) `ip` validation**

Tightened from truthy-check to `net.isIP(ip)`. New tests cover the classic failure mode (env-stringified `"undefined"`), garbage strings, malformed IPv6, AND the IPv4/IPv6 happy paths (ECS supports both).

**(6) `listFreshPeers` Scan pagination**

Added to the module header as an explicit caveat: single-page only, bounded by 1MB ScanCommand response. Fine at 2 replicas, broken at ~1000+ peers. Revisit timing tied to the existing GSI-on-shard_id comment for the sharded future.

**(7) Client-side filter vs FilterExpression**

Added to the access-pattern rationale in the module header: client-side filter chosen because FilterExpression still consumes RCU for filtered rows (no cost win) and network bytes are negligible at this row count. Revisit if row count grows past ~1000.

Tests: 2084 (+2 new vs cr-r1 baseline), lint clean. Awaiting cr re-review on `810006f`.

justin-layerv:
## cr-r3 response (no code change — convergence)

cr-r3 is "LGTM" with five non-blocking observations. Walking each:

**(1) `releaseLock` clears `currentVersion` even on generic transport error**

Already documented in the cr-r2 fixes (module-header "Release semantics" section). Calling out explicitly: today's only caller is `gracefulShutdown`. PR 13b.2 must re-introduce the "DDB transient vs lock genuinely gone" distinction before wiring `releaseLock` onto any voluntary-step-down or watchdog-driven path. Pinned in the module header so it surfaces during 13b.2 review.

**(2) No metrics emission — `gateway_lock_lost_cas_total` is the canonical SLO signal**

Filed as #413 — "feat(discord): Pillar 3 lock + heartbeat observability counters". Properly labeled (`enhancement`, `javascript`, `pr-polish-followup`, `priority: low`). Acceptance criteria + tag shape included so PR 13b.2 (or a parallel observability PR) can pick it up without rediscovering scope. The reviewer's framing — "Out of scope for a pure-modules PR" — is exactly the right shape.

**(3) `listFreshPeers` Scan defensive `Limit: 100` + `net.isIP('0.0.0.0')` filter**

Skipped both per reviewer's "probably overkill / flagging only" guidance:
- A `Limit: 100` without pagination still gives wrong-head-of-list under high row counts — the failure mode is the same (truncated result before client-side sort), just at a different threshold. The "broken at 1000+" caveat in the module header is the load-bearing documentation; adding `Limit` without pagination wouldn't actually make the failure louder unless paired with a length-check that reads "if scan returned exactly Limit, log a warning."
- `0.0.0.0` / `::` are syntactically valid IPs (so `net.isIP` correctly returns truthy) but the only realistic way to land them on a peer row is a misconfigured `ECS_CONTAINER_METADATA_URI_V4` parse — out of scope. ECS won't assign these in awsvpc mode.

**(4) `writeHeartbeat` re-writes static `ip` / `port` every 2s**

No action per reviewer's "Leave it" — the single-PutItem contract (header item 2) is load-bearing; splitting writes would create a partial-state window between `updated_at` and `expires_at`.

**(5) Self-handoff guard scope (`instance_id` vs `lock_holder`)**

Praise noted — no action. The test already pins this on `instance_id` specifically.

---

## Convergence check

1. **DE acceptance**: cr-r3 explicitly states "LGTM. The contract-pinning discipline here sets a good bar for PR 13b.2 and 13b.3 to maintain when the integration arrives." ✓
2. **CI passing**: all required checks green on `810006f`. ✓
3. **Every cr comment considered**: yes — items 1, 4, 5 acknowledged with no-action rationale; item 2 deferred via filed issue #413 with full scope; item 3 reviewer-marked-as-overkill skipped with explicit rationale. ✓
4. **`/simplify` clean**: the cr-r2 → cr-r3 delta (810006f) was reviewed by cr-r3 itself and surfaced only the items above. No remaining findings. ✓

Ready to merge. Tests: 2084, lint clean.

PR #414, opened 2026-05-16
feat(discord): pillar 3 control channel + leader coordinator

justin-layerv:
## Why

Pillar 3 of the zero-downtime gateway rollout: hot-standby leader election with sub-second push-handoff on deploy. The current bot's single-replica Discord gateway means every deploy + every task replacement = ~7 s of gateway downtime, which under burst load can shed user-visible interactions. This PR adds the control-plane machinery so a second replica can immediately take over the WS session via in-VPC handoff instead of waiting for the cold-fallback floor.

Builds on the lock + peer-heartbeat primitives from #412 (PR 13b.1). No wiring into `index.js` yet — that ships in **PR 13b.3** alongside the `ENABLE_GATEWAY_HOT_STANDBY` flag (default off), so this PR is a pure-addition no-runtime-effect change. Design doc: `apps/discord/docs/zero-downtime-design.md` §Pillar 3.

### Five new modules

| Module | Role |
|---|---|
| `gateway-hmac.js` | SHA-256 sign/verify with dual-secret accept (rolling-rotation window), 5 s freshness, 1024-entry nonce LRU. Exports wrap/unwrap envelope helpers as the single source of truth for the wire shape across server + client + tests. Frozen `VERIFY_REASONS` for cross-module typo safety. |
| `gateway-connection-watchdog.js` | 1 s-tick loop that closes the "lock held but WS disconnected" gap. Exponential backoff (200/400/800/1600 ms); at 5 attempts → release lock + `exit(1)` so ECS replaces the task. |
| `gateway-control-channel.js` | `node:http` server for `POST /control/yours`. 8 KB body cap (streaming), HMAC-verify-before-parse, routing checks (peer_instance_id binding + isKnownPeer), awaits onHandoff before ACKing. |
| `gateway-control-client.js` | Outbound POST with 200 ms timeout to match the SIGTERM ACK deadline. Returns result objects — never throws. Settles before destroy on timeout to prevent the destroy-induced `error` event from racing the timeout reason. |
| `gateway-leader.js` | Orchestrator that ties lock + heartbeat + control-client + manager into the state machine. Exposes hooks (`isHoldingLock` / `isKnownPeer` / `handleInboundHandoff` / `releaseLockForExit`) for the wiring layer in PR 13b.3. Serializes all lock mutators through a single in-flight promise chain — `gateway-lock` is single-caller-only by contract. |

### One modification

- `gateway-peer-heartbeat.js`: optional `lockHolder` arg, written as `lock_holder` on the heartbeat row. The active's SIGTERM `transferLock(target, targetHolder)` needs the peer's holder string; carrying it on the heartbeat row keeps the active from inventing or looking it up. Falls back to a placeholder for pre-13b.2 callers.

## Load-bearing design choices

1. **HMAC verify runs BEFORE JSON.parse** (DoS + key-order safety). The wire envelope wraps the inner signed body as a JSON STRING so its bytes round-trip verbatim — parsing it as an object would canonicalize key order and break verify.
2. **Synchronous check-then-set in the nonce LRU.** Tested with a parallel-verify concurrency test that asserts exactly one of two simultaneous verifies of the same nonce wins.
3. **Watchdog exhaustion → `exit(1)`** rather than infinite retry. A standby that can't reach Discord but holds the lock blocks the only failover slot — ECS-restart is the cheaper recovery.
4. **Single-caller-only serialization.** `tick` / `inbound-handoff` / `push-handoff` all touch `gateway-lock`'s closure state (currentVersion). The leader funnels every mutator through one in-flight promise chain so two never overlap. Pinned by a concurrency test.
5. **Adopt → flag → connect ordering** on inbound handoff. Adopt failure leaves heldLock=false (watchdog stays quiet). Connect failure flips heldLock=true (watchdog retries). Reversing the order would mis-key the watchdog under transient DDB failures.
6. **Heartbeat + peer-list in parallel** per tick — independent DDB ops with independent error handling. Halves tick DDB-RTT cost.

## Testing

137 new unit tests across 5 new files + extensions to `gateway-peer-heartbeat.test.js`. Full discord suite (61 files, 2202 tests) green.

Test categories:
- Factory validation (every module)
- Contract pins (HMAC dual-secret, freshness window, nonce LRU bounds, verify-before-parse)
- Routing checks (peer_instance_id binding, isKnownPeer)
- Failure modes (timeout, http_error, transferLock CAS fail, adopt throw, connect throw)
- Serialization invariants (no two mutators interleave; concurrent inbound handoffs serialize)
- End-to-end client → server round-trip

## What's NOT in this PR

- No `index.js` wiring (PR 13b.3)
- No `ENABLE_GATEWAY_HOT_STANDBY` flag (PR 13b.3)
- No `gateway_desired_count=2` / AZ-spread infra (PR 14)
- No chaos validation (PR 15)

## Test plan

- [x] Unit tests for all new modules
- [x] End-to-end real-HTTP test for control-channel server + control-client
- [x] Concurrency test for nonce LRU check-then-set
- [x] Full discord suite (2202 tests) green
- [x] /simplify across all new files (code reuse, code quality, efficiency — applied findings)
- [x] Lint clean on all changed files
- [ ] Runtime smoke (deferred to PR 13b.3 wiring)

### DIFF
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -57,6 +57,16 @@ const BACKOFF_CAP_MS = 5_000;
 function createConnectionWatchdog({
   manager,
   isHoldingLock,
+  // Optional. Returns true when the leader is itself mid-
+  // `manager.connect()` (inbound-handoff path); the watchdog backs
+  // off rather than race a second concurrent connect. Default
+  // `() => false` keeps tests + non-leader callers simple, but
+  // production wiring (PR 13b.3) MUST pass the leader's
+  // `isConnecting()` — without it, an inbound-handoff connect
+  // taking >1 s would race the next watchdog tick and produce
+  // overlapping connect() calls against the non-concurrency-safe
+  // WebSocketManager.
+  isConnecting = () => false,
   releaseLock,
   logger,
   pollIntervalMs = DEFAULT_POLL_INTERVAL_MS,
@@ -71,6 +81,9 @@ function createConnectionWatchdog({
   }
   if (typeof isHoldingLock !== 'function') {
     throw new Error('createConnectionWatchdog: isHoldingLock function is required');
+  }
+  if (typeof isConnecting !== 'function') {
+    throw new Error('createConnectionWatchdog: isConnecting must be a function');
   }
   if (typeof releaseLock !== 'function') {
     throw new Error('createConnectionWatchdog: releaseLock function is required');
@@ -91,6 +104,16 @@ function createConnectionWatchdog({
       return;
     }
     if (manager.isConnected()) {
+      attempts = 0;
+      return;
+    }
+    // Leader is mid-connect (inbound-handoff path). Stand down —
+    // don't race a second concurrent connect against the same
+    // WebSocketManager. Reset attempts so the leader's eventual
+    // success doesn't leave a stale failure ladder behind, and
+    // its eventual failure starts the ladder cleanly from 1 on
+    // the next tick when isConnecting() returns false again.
+    if (isConnecting()) {
       attempts = 0;
       return;
     }
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -118,6 +118,12 @@ function createConnectionWatchdog({
   // calls it from the `start()` loop. Returns once the iteration
   // (including any failure-path backoff sleep) settles.
   async function step() {
+    // Terminal-state guard. After the exhaustion-exit branch sets
+    // closed=true, the production loop stops. But _stepForTest can
+    // still be called manually post-exit; without this guard a
+    // re-entry would re-fire releaseLock + deleteOwnRow. Test-tool
+    // robustness only; production never re-enters.
+    if (closed) return;
     if (!isHoldingLock()) {
       attempts = 0;
       return;
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -53,6 +53,13 @@ const DEFAULT_POLL_INTERVAL_MS = 1_000;
 const DEFAULT_MAX_ATTEMPTS = 5;
 const BACKOFF_BASE_MS = 100;
 const BACKOFF_CAP_MS = 5_000;
+// Ceilings below bound the WATCHDOG's own failure-replace path —
+// they're NOT on the SIGTERM critical path (which has its own
+// ~200 ms client-side cap in gateway-leader.js's pushHandoff). A
+// stuck standby holding the failover slot is what these protect
+// against; worst-case watchdog tick at exhaustion is CONNECT +
+// RELEASE_LOCK ≈ 11 s, well inside the 30 s ECS graceful-stop.
+//
 // Hard ceiling on the releaseLock-during-exit call. The leader's
 // releaseLockForImmediateExit awaits through `runSerialized`, so a
 // hung inbound-handoff (parked inside manager.connect() with
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -57,16 +57,16 @@ const BACKOFF_CAP_MS = 5_000;
 function createConnectionWatchdog({
   manager,
   isHoldingLock,
-  // Optional. Returns true when the leader is itself mid-
+  // REQUIRED. Returns true when the leader is itself mid-
   // `manager.connect()` (inbound-handoff path); the watchdog backs
-  // off rather than race a second concurrent connect. Default
-  // `() => false` keeps tests + non-leader callers simple, but
-  // production wiring (PR 13b.3) MUST pass the leader's
-  // `isConnecting()` — without it, an inbound-handoff connect
-  // taking >1 s would race the next watchdog tick and produce
-  // overlapping connect() calls against the non-concurrency-safe
-  // WebSocketManager.
-  isConnecting = () => false,
+  // off rather than race a second concurrent connect. Required
+  // because the race is otherwise silent: an inbound-handoff connect
+  // taking >1 s would produce overlapping connect() calls against
+  // the non-concurrency-safe WebSocketManager. A wiring oversight
+  // failing loud at boot is preferable to a rare production race.
+  // Tests that don't exercise the inbound-handoff path can pass
+  // `() => false`.
+  isConnecting,
   releaseLock,
   logger,
   pollIntervalMs = DEFAULT_POLL_INTERVAL_MS,
@@ -83,7 +83,7 @@ function createConnectionWatchdog({
     throw new Error('createConnectionWatchdog: isHoldingLock function is required');
   }
   if (typeof isConnecting !== 'function') {
-    throw new Error('createConnectionWatchdog: isConnecting must be a function');
+    throw new Error('createConnectionWatchdog: isConnecting function is required');
   }
   if (typeof releaseLock !== 'function') {
     throw new Error('createConnectionWatchdog: releaseLock function is required');
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -61,6 +61,16 @@ const BACKOFF_CAP_MS = 5_000;
 // chain forever and exit(1) would never fire. 3 s is generous
 // vs. the 1 s expected DDB RTT; on miss we log and exit anyway.
 const RELEASE_LOCK_CEILING_MS = 3_000;
+// Hard ceiling on the watchdog's `manager.connect()` await. Mirrors
+// the leader's `inboundConnectTimeoutMs` (5 s default) — a hung
+// connect would otherwise pin the tick forever, attempts wouldn't
+// advance, and the `exit(1)` recovery would never fire. 8 s is
+// generous vs. Discord WS cold-connect's 1-3 s typical, slightly
+// wider than the leader's 5 s to leave headroom for the watchdog's
+// fail-loud-then-replace posture (the leader caps tighter because
+// it's gated on the SIGTERM ECS deadline). On timeout we throw
+// like a connect rejection; the attempts ladder advances normally.
+const CONNECT_CEILING_MS = 8_000;
 
 function createConnectionWatchdog({
   manager,
@@ -85,6 +95,9 @@ function createConnectionWatchdog({
   // the ~1 s expected DDB RTT; on miss we log and exit anyway.
   // Injected for tests so the ceiling can be tuned tiny.
   releaseLockCeilingMs = RELEASE_LOCK_CEILING_MS,
+  // Hard ceiling on the inline `manager.connect()` call. See
+  // CONNECT_CEILING_MS above for rationale. Injected for tests.
+  connectCeilingMs = CONNECT_CEILING_MS,
   // Optional. If provided, called best-effort on the exhaustion-exit
   // path alongside releaseLock. Symmetric to gateway-leader.js's
   // SIGTERM pushHandoff path which also deletes the row. Without
@@ -125,6 +138,34 @@ function createConnectionWatchdog({
   if (!Number.isInteger(pollIntervalMs) || pollIntervalMs <= 0) {
     throw new Error('createConnectionWatchdog: pollIntervalMs must be a positive integer');
   }
+  if (!Number.isInteger(releaseLockCeilingMs) || releaseLockCeilingMs <= 0) {
+    throw new Error('createConnectionWatchdog: releaseLockCeilingMs must be a positive integer');
+  }
+  if (!Number.isInteger(connectCeilingMs) || connectCeilingMs <= 0) {
+    throw new Error('createConnectionWatchdog: connectCeilingMs must be a positive integer');
+  }
+
+  // Races `promise` against a `ceilingMs` timer that rejects with a
+  // labeled Error. Used to bound `manager.connect()` and
+  // `releaseLock()` — either could hang and pin the failure-exit
+  // recovery path. Mirrors `gateway-leader.js`'s
+  // `inboundConnectTimeoutMs` pattern.
+  async function raceWithCeiling(promise, ceilingMs, label) {
+    let timer;
+    try {
+      return await Promise.race([
+        promise,
+        new Promise((_, reject) => {
+          timer = setTimeout(
+            () => reject(new Error(`${label}_${ceilingMs}ms`)),
+            ceilingMs,
+          );
+        }),
+      ]);
+    } finally {
+      clearTimeout(timer);
+    }
+  }
 
   let running = false;
   let loopPromise = null;
@@ -162,7 +203,9 @@ function createConnectionWatchdog({
 
     attempts += 1;
     try {
-      await manager.connect();
+      // Bounded by CONNECT_CEILING_MS — a hung connect would otherwise
+      // pin this tick and the failure-exit recovery would never fire.
+      await raceWithCeiling(manager.connect(), connectCeilingMs, 'watchdog_connect_ceiling');
       attempts = 0;
       logger.info('connection-watchdog: connect succeeded');
     } catch (err) {
@@ -171,24 +214,14 @@ function createConnectionWatchdog({
           error: err.message, attempts,
         });
         try {
-          // Race against a hard ceiling. The leader's
+          // Bounded by RELEASE_LOCK_CEILING_MS — the leader's
           // releaseLockForImmediateExit awaits through its
-          // serialization chain, so a hung inbound-handoff (parked
+          // serialization chain; a hung inbound-handoff (parked
           // inside manager.connect() with `connecting=true` latched)
-          // would block the chain forever. Without the race, exit(1)
-          // would never fire — the failover slot stays held with
-          // no live gateway. On ceiling miss we log and exit anyway;
-          // the row fades via TTL.
-          let releaseTimer;
-          await Promise.race([
-            releaseLock(),
-            new Promise((_, reject) => {
-              releaseTimer = setTimeout(
-                () => reject(new Error(`release_lock_ceiling_${releaseLockCeilingMs}ms`)),
-                releaseLockCeilingMs,
-              );
-            }),
-          ]).finally(() => clearTimeout(releaseTimer));
+          // would block the chain forever and exit(1) would never
+          // fire. On ceiling miss we log and exit anyway; the row
+          // fades via TTL.
+          await raceWithCeiling(releaseLock(), releaseLockCeilingMs, 'release_lock_ceiling');
         } catch (rerr) {
           // Logged-and-swallowed: we're already in the failure-exit
           // path; refusing to exit because of a DDB blip — or the
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -100,6 +100,14 @@ function createConnectionWatchdog({
     throw new Error('createConnectionWatchdog: releaseLock function is required');
   }
   if (!logger) throw new Error('createConnectionWatchdog: logger is required');
+  // maxAttempts=0 would exit(1) on the first tick; pollIntervalMs=0
+  // would saturate the loop. Fail loud at boot.
+  if (!Number.isInteger(maxAttempts) || maxAttempts <= 0) {
+    throw new Error('createConnectionWatchdog: maxAttempts must be a positive integer');
+  }
+  if (!Number.isInteger(pollIntervalMs) || pollIntervalMs <= 0) {
+    throw new Error('createConnectionWatchdog: pollIntervalMs must be a positive integer');
+  }
 
   let running = false;
   let loopPromise = null;
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -175,17 +175,24 @@ function createConnectionWatchdog({
   }
 
   function start() {
-    if (running || exited) return;
+    // Guard on `loopPromise` (NOT just `running`) so a `start()`
+    // after a `stop()` that hasn't yet observed the running=false
+    // flag — the old loop is still inside `await sleep(...)` — does
+    // not spawn a second concurrent loop. Callers that need to
+    // re-start MUST await `stop()` first.
+    if (loopPromise || exited) return;
     running = true;
-    loopPromise = loop();
+    loopPromise = loop().finally(() => { loopPromise = null; });
   }
 
-  // Halts the loop. Idempotent. Does not interrupt an in-flight
+  // Halts the loop and returns a promise that resolves once the
+  // last in-flight tick has completed (including any in-flight
   // `manager.connect()` — the awaiting tick still completes before
-  // the loop check sees running=false. That matches the design's
-  // at-most-one outstanding connect invariant.
+  // the loop check sees running=false). Idempotent. Callers that
+  // want to re-start the watchdog MUST await this.
   function stop() {
     running = false;
+    return loopPromise ?? Promise.resolve();
   }
 
   return {
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -1,0 +1,173 @@
+// Connection watchdog for Pillar 3 — closes the gap where the
+// standby has been handed the lock (or acquired it via cold
+// fallback) but the gateway WebSocket isn't connected to Discord.
+// Without this loop, the active times out on ACK after 200 ms and
+// exits, while the standby sits lock-held-but-no-WS indefinitely.
+//
+// Spec reference: docs/zero-downtime-design.md §Pillar 3
+// "connection-watchdog" (lines ~583-612).
+//
+// ── Loop ──
+// Every `pollIntervalMs` (default 1000):
+//   1. If not holding the lock → reset attempts; continue. The
+//      watchdog runs unconditionally, so a standby waiting to be
+//      promoted enters the loop but no-ops until it acquires the
+//      lock. Resetting attempts here means a previous failure ladder
+//      doesn't carry across a lock-give-up + lock-reacquire cycle.
+//   2. If holding the lock AND the manager reports connected → reset
+//      attempts; continue. Steady-state: 1 s tick, no work, no log.
+//   3. Otherwise: attempts++, try `manager.connect()`. On success,
+//      reset attempts. On failure, log + exponential backoff sleep
+//      (200 ms, 400 ms, 800 ms, 1.6 s — capped at 5 s, see below).
+//      At `attempts >= maxAttempts` (default 5), release the lock
+//      and `exit(1)` so ECS replaces the task.
+//
+// ── Why bounded retry → exit, not infinite retry ──
+// A standby that can't reach Discord but holds the lock blocks the
+// only failover slot. ECS will restart us with fresh DNS / fresh
+// gateway endpoint pinning / fresh networking state; that's the
+// cheapest recovery path. Holding the lock and looping forever
+// merely delays handoff to a new task.
+//
+// ── Backoff cap is dead code ──
+// `Math.min(2^attempts * 100, 5000)` caps at 5 s. At maxAttempts=5,
+// the last backoff that ever sleeps is at attempts=4 → 1600 ms,
+// before the attempts=5 check triggers exit. The 5 s ceiling is
+// unreachable at this cap. Kept literally for future-proofing: if
+// someone bumps maxAttempts to e.g. 8, the cap kicks in (12800 ms
+// → 5000 ms) and they don't have to revisit the formula.
+//
+// ── Concurrency ──
+// The watchdog awaits `manager.connect()` inline. The next 1 s tick
+// does NOT fire until that call settles. Two concurrent connect()
+// invocations would race the @discordjs/ws internal state; the
+// design depends on at-most-one outstanding connect() per watchdog
+// instance. Tests pin this by counting connect() calls under
+// fake-timer advancement.
+//
+// ── Process exit is injected ──
+// Real prod calls `process.exit(1)`. Tests inject a no-op so the
+// retries-exhausted branch is observable without killing jest.
+
+const DEFAULT_POLL_INTERVAL_MS = 1_000;
+const DEFAULT_MAX_ATTEMPTS = 5;
+const BACKOFF_BASE_MS = 100;
+const BACKOFF_CAP_MS = 5_000;
+
+function createConnectionWatchdog({
+  manager,
+  isHoldingLock,
+  releaseLock,
+  logger,
+  pollIntervalMs = DEFAULT_POLL_INTERVAL_MS,
+  maxAttempts = DEFAULT_MAX_ATTEMPTS,
+  // Injected for tests. Production: setTimeout-based.
+  sleep = (ms) => new Promise((resolve) => { setTimeout(resolve, ms); }),
+  // Injected for tests. Production: process.exit. Tests pass a spy.
+  exit = (code) => process.exit(code),
+} = {}) {
+  if (!manager || typeof manager.connect !== 'function' || typeof manager.isConnected !== 'function') {
+    throw new Error('createConnectionWatchdog: manager with connect() and isConnected() is required');
+  }
+  if (typeof isHoldingLock !== 'function') {
+    throw new Error('createConnectionWatchdog: isHoldingLock function is required');
+  }
+  if (typeof releaseLock !== 'function') {
+    throw new Error('createConnectionWatchdog: releaseLock function is required');
+  }
+  if (!logger) throw new Error('createConnectionWatchdog: logger is required');
+
+  let running = false;
+  let loopPromise = null;
+  let attempts = 0;
+  let exited = false;
+
+  // Run one iteration of the watchdog. Public for tests; production
+  // calls it from the `start()` loop. Returns once the iteration
+  // (including any failure-path backoff sleep) settles.
+  async function step() {
+    if (!isHoldingLock()) {
+      attempts = 0;
+      return;
+    }
+    if (manager.isConnected()) {
+      attempts = 0;
+      return;
+    }
+
+    attempts += 1;
+    try {
+      await manager.connect();
+      attempts = 0;
+      logger.info('connection-watchdog: connect succeeded');
+    } catch (err) {
+      if (attempts >= maxAttempts) {
+        logger.error('connection-watchdog: connect retries exhausted, releasing lock', {
+          error: err.message, attempts,
+        });
+        try {
+          await releaseLock();
+        } catch (rerr) {
+          // Logged-and-swallowed: we're already in the failure-exit
+          // path; refusing to exit because of a DDB blip would leave
+          // the lock held with no live gateway.
+          logger.error('connection-watchdog: releaseLock failed during exhaustion-exit', {
+            error: rerr.message,
+          });
+        }
+        exited = true;
+        running = false;
+        exit(1);
+        return;
+      }
+      const backoffMs = Math.min((2 ** attempts) * BACKOFF_BASE_MS, BACKOFF_CAP_MS);
+      logger.warn('connection-watchdog: connect failed, backing off', {
+        error: err.message, attempts, backoffMs,
+      });
+      await sleep(backoffMs);
+    }
+  }
+
+  async function loop() {
+    while (running) {
+      // Sleep first so an immediate stop() right after start()
+      // doesn't run a single tick — useful for tests + clean
+      // shutdown semantics.
+      await sleep(pollIntervalMs);
+      if (!running) break;
+      await step();
+    }
+  }
+
+  function start() {
+    if (running || exited) return;
+    running = true;
+    loopPromise = loop();
+  }
+
+  // Halts the loop. Idempotent. Does not interrupt an in-flight
+  // `manager.connect()` — the awaiting tick still completes before
+  // the loop check sees running=false. That matches the design's
+  // at-most-one outstanding connect invariant.
+  function stop() {
+    running = false;
+  }
+
+  return {
+    start,
+    stop,
+    // Inspection + driver seams for tests.
+    _stepForTest: step,
+    _getAttemptsForTest() { return attempts; },
+    _getRunningForTest() { return running; },
+    _getLoopPromiseForTest() { return loopPromise; },
+  };
+}
+
+module.exports = {
+  createConnectionWatchdog,
+  DEFAULT_POLL_INTERVAL_MS,
+  DEFAULT_MAX_ATTEMPTS,
+  BACKOFF_BASE_MS,
+  BACKOFF_CAP_MS,
+};
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -53,6 +53,14 @@ const DEFAULT_POLL_INTERVAL_MS = 1_000;
 const DEFAULT_MAX_ATTEMPTS = 5;
 const BACKOFF_BASE_MS = 100;
 const BACKOFF_CAP_MS = 5_000;
+// Hard ceiling on the releaseLock-during-exit call. The leader's
+// releaseLockForImmediateExit awaits through `runSerialized`, so a
+// hung inbound-handoff (parked inside manager.connect() with
+// `connecting=true` latched per gateway-leader.js's bounded-
+// settlement comment, tracked in #415) would otherwise block the
+// chain forever and exit(1) would never fire. 3 s is generous
+// vs. the 1 s expected DDB RTT; on miss we log and exit anyway.
+const RELEASE_LOCK_CEILING_MS = 3_000;
 
 function createConnectionWatchdog({
   manager,
@@ -68,6 +76,15 @@ function createConnectionWatchdog({
   // `() => false`.
   isConnecting,
   releaseLock,
+  // Hard ceiling on the releaseLock call during exhaustion-exit. The
+  // leader's releaseLockForImmediateExit awaits through `runSerialized`,
+  // so a hung inbound-handoff (parked inside manager.connect() with
+  // `connecting=true` latched per gateway-leader.js's bounded-
+  // settlement comment, tracked in #415) would otherwise block the
+  // chain forever and exit(1) would never fire. 3 s is generous vs.
+  // the ~1 s expected DDB RTT; on miss we log and exit anyway.
+  // Injected for tests so the ceiling can be tuned tiny.
+  releaseLockCeilingMs = RELEASE_LOCK_CEILING_MS,
   // Optional. If provided, called best-effort on the exhaustion-exit
   // path alongside releaseLock. Symmetric to gateway-leader.js's
   // SIGTERM pushHandoff path which also deletes the row. Without
@@ -154,11 +171,29 @@ function createConnectionWatchdog({
           error: err.message, attempts,
         });
         try {
-          await releaseLock();
+          // Race against a hard ceiling. The leader's
+          // releaseLockForImmediateExit awaits through its
+          // serialization chain, so a hung inbound-handoff (parked
+          // inside manager.connect() with `connecting=true` latched)
+          // would block the chain forever. Without the race, exit(1)
+          // would never fire — the failover slot stays held with
+          // no live gateway. On ceiling miss we log and exit anyway;
+          // the row fades via TTL.
+          let releaseTimer;
+          await Promise.race([
+            releaseLock(),
+            new Promise((_, reject) => {
+              releaseTimer = setTimeout(
+                () => reject(new Error(`release_lock_ceiling_${releaseLockCeilingMs}ms`)),
+                releaseLockCeilingMs,
+              );
+            }),
+          ]).finally(() => clearTimeout(releaseTimer));
         } catch (rerr) {
           // Logged-and-swallowed: we're already in the failure-exit
-          // path; refusing to exit because of a DDB blip would leave
-          // the lock held with no live gateway.
+          // path; refusing to exit because of a DDB blip — or the
+          // ceiling-miss above — would leave the lock held with no
+          // live gateway.
           logger.error('connection-watchdog: releaseLock failed during exhaustion-exit', {
             error: rerr.message,
           });
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -86,17 +86,9 @@ function createConnectionWatchdog({
   // `() => false`.
   isConnecting,
   releaseLock,
-  // Hard ceiling on the releaseLock call during exhaustion-exit. The
-  // leader's releaseLockForImmediateExit awaits through `runSerialized`,
-  // so a hung inbound-handoff (parked inside manager.connect() with
-  // `connecting=true` latched per gateway-leader.js's bounded-
-  // settlement comment, tracked in #415) would otherwise block the
-  // chain forever and exit(1) would never fire. 3 s is generous vs.
-  // the ~1 s expected DDB RTT; on miss we log and exit anyway.
-  // Injected for tests so the ceiling can be tuned tiny.
+  // Ceilings — see RELEASE_LOCK_CEILING_MS / CONNECT_CEILING_MS above
+  // for rationale. Injected for tests so they can be tuned tiny.
   releaseLockCeilingMs = RELEASE_LOCK_CEILING_MS,
-  // Hard ceiling on the inline `manager.connect()` call. See
-  // CONNECT_CEILING_MS above for rationale. Injected for tests.
   connectCeilingMs = CONNECT_CEILING_MS,
   // Optional. If provided, called best-effort on the exhaustion-exit
   // path alongside releaseLock. Symmetric to gateway-leader.js's
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -135,7 +135,19 @@ function createConnectionWatchdog({
       // shutdown semantics.
       await sleep(pollIntervalMs);
       if (!running) break;
-      await step();
+      // Backstop for unexpected throws from `manager.isConnected()`
+      // (contractually non-throwing, but a future shim refactor
+      // could regress) or any other unhandled path. Without this,
+      // a single throw inside step() would exit the loop and
+      // silently disable the watchdog. Log + continue: the next
+      // tick re-tries.
+      try {
+        await step();
+      } catch (err) {
+        logger.error('connection-watchdog: step threw unexpectedly (loop continues)', {
+          error: err && err.message,
+        });
+      }
     }
   }
 
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -68,6 +68,17 @@ function createConnectionWatchdog({
   // `() => false`.
   isConnecting,
   releaseLock,
+  // Optional. If provided, called best-effort on the exhaustion-exit
+  // path alongside releaseLock. Symmetric to gateway-leader.js's
+  // SIGTERM pushHandoff path which also deletes the row. Without
+  // this, the heartbeat row lingers until DDB TTL (default 60s) and
+  // surviving peers keep `isKnownPeer`-ing this dead replica + the
+  // next handoff source may pick this row from listFreshPeers head
+  // and time out the 200 ms ACK. Not a correctness issue (the cold-
+  // fallback floor applies), but a small latency tax during the TTL
+  // window. Optional so existing wiring without peerHeartbeat
+  // visibility doesn't break — the absence is a no-op, not a throw.
+  deleteOwnRow,
   logger,
   pollIntervalMs = DEFAULT_POLL_INTERVAL_MS,
   maxAttempts = DEFAULT_MAX_ATTEMPTS,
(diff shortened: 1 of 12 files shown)


justin-layerv:
**cr-r1 fixes pushed (3131d2d)**

All items from the previous review addressed:

1. **Body-too-large socket race** — switched `req.destroy()` → `req.pause()` so the 413 reaches the client. Added end-to-end real-socket test that pins the contract.

2. **`peerIp` IP validation** — `net.isIP()` check in `gateway-control-client`, mirroring the heartbeat-side validator. Test exercises rejection of hostnames + accepts IPv4/IPv6 literals.

3. **Watchdog throw backstop** — `step()` wrapped in try/catch in the loop; logs at error level + continues. Test pins survival across an `isConnected()` throw.

4. **`res.on('aborted')` handler** — added in control-client; peer crash mid-response now settles immediately instead of waiting for the 200 ms timeout. Test uses a `respondThenAbort` fixture.

5. **`secrets.previous = ""` rejection** — empty string now rejected at boot (would have silently disabled dual-accept during rotation). Test added.

6. **`rememberNonce` dead-branch comment** — documented as defensive for future callers that don't pre-check.

7. **`heldLock` invariant comment** — documented the implicit "caller MUST exit after `pushHandoff`" invariant in `gateway-leader`.

8. **Item #7 from review (malformed-payload nonce burn)** — confirming this is intentional and PR 13b.3's wiring will not legitimately mis-address. The active's `pushHandoff` derives `peer_instance_id` from the heartbeat row it picked, so a mis-address would require a corrupted heartbeat row — handled at the routing-check layer as a 400 rather than retry.

142 Pillar 3 tests passing (5 new). Full discord suite 2207/2207 green. Lint clean.

justin-layerv:
**cr-r2 fixes pushed (a1b6384)**

All items from round 2 addressed:

**Correctness — race fix (item 1, HIGH):**
- New `connecting` state in `gateway-leader` flipped true around `handleInboundHandoff`'s `await manager.connect()`. Cleared in a `finally` so a connect throw still releases the flag, letting the watchdog take over.
- New `isConnecting()` hook exposed on the leader.
- `gateway-connection-watchdog` accepts `isConnecting` constructor arg; when it returns true, the watchdog backs off — resets attempts and returns without calling `manager.connect()`. Default `() => false` for back-compat; production wiring (PR 13b.3) MUST pass the leader's hook.
- Test pins the race-protection contract: watchdog step skips connect when isConnecting=true; mid-ladder transition resets attempts.

**Correctness — pushHandoff re-entry (item 2):**
- Existing serialization + `heldLock=false` clearing already handles this. Added explicit tests pinning the contract: two parallel/sequential `pushHandoff` calls result in exactly one `transferLock`, second sees `not_holding_lock`.

**Smaller items:**
- (3) `host` → `hostname` in `gateway-control-client`. Test updated.
- (4) `rememberNonce` dead branch — kept defensive as documented (reviewer noted "harmless as-is").
- (5) Body-cap drain: 413 now sends `Connection: close` header. Without it, a keep-alive client could attempt another request and the unread bytes would confuse the HTTP parser. Forcing close after this response makes the unread body harmless.
- (6) Handler error 500 path — no change needed, reviewer just noted the defensive coding.

**Security/Performance:** No changes needed — all items previously green.

**Test coverage gaps closed:**
- Inbound-handoff vs watchdog concurrent connect test ✅
- pushHandoff re-entry safety test ✅
- `peer.lock_holder` placeholder fallback — already pinned; the cleanup-TODO is implicit in the comment (\"back-compat with pre-PR-13b.2 callers\"), can drop when those callers are gone.

148 Pillar 3 tests (was 142). Full discord suite 2213/2213 green. Lint clean.

justin-layerv:
**cr-r3 fixes pushed (ac0571d)**

Items addressed:

1. **start()/stop() lifecycle (potential issue):** `start()` now guards on `loopPromise` (not just `running`). `stop()` returns a promise that resolves once the loop fully exits, so callers needing to re-start can `await stop()` first. Tests pin both the await-able stop and the start-after-stop-without-await no-op contract. Applied to both `gateway-leader` and `gateway-connection-watchdog`.

2. **adoptLockFromHandoff error logging (item 3):** wrapped in try/catch that logs at error level with `activeInstanceId` + `expectedVersion` context before re-throwing. Previously the throw went through runSerialized's debug-level chain-keeping catch.

3. **Peer response body log cap (smaller item):** `gateway-control-client` truncates the logged body at 512 chars. Returned `body` field stays uncapped for debuggers/callers.

**Skipped (with reasoning):**
- **#2** (`pushHandoff` running=false before runSerialized) — reviewer marked as marginal, same family as #1. The lifecycle fix above covers the underlying invariant.
- **#4** (nonce burn on routing-mismatch) — non-issue today; production client generates fresh nonce per call. Will pin "fresh nonce per call" test if PR 13b.3 ever adds client-side retry.
- **#5** (URL parsing hardening for `*`/absolute-form) — reviewer said "Skip if it's not worth the bytes." Current `split('?', 1)[0]` 404s on all problematic forms (e.g., `*` !== `/control/yours`).
- **#6** (Host header IPv6) — informational only; heartbeat-side validation already restricts IP shape.
- **Dual-secret timing asymmetry, Content-Length pre-check, `rememberNonce` has/delete, step() heartbeat contention** — all flagged as benign/forward-looking by reviewer.
- **`req.aborted` deprecation** — works fine on current Node LTS; deferred until LTS bumps.

151 Pillar 3 tests (was 148). Full discord suite 2216/2216 green. Lint clean.

justin-layerv:
**cr-r4 fixes pushed (8fc4b6f)**

All five actionable items addressed:

1. **headersTimeout floor (correctness #1):** `Math.max(1_000, floor(requestTimeoutMs / 2))` — a tiny `requestTimeoutMs` would otherwise produce `headersTimeout = 0` which Node interprets as DISABLED.

2. **Defensive `heldLock = false` on transfer_failed (correctness #2):** the local flag now matches DDB state after a CCF, removing the implicit "caller MUST exit after pushHandoff" load-bearing invariant.

3. **Wire envelope key-order test (correctness #3):** asserts `Object.keys(parsed) === ['body', 'signature']`. V8 preserves insertion order but the contract was implicit.

4. **`runSerialized` comment (code quality):** added "INTENTIONAL: both arms call work" comment so a refactor doesn't reduce the chain-keeping pattern to `.then(work)`.

5. **SIGTERM-during-tick test (test coverage gap):** new test pins that `pushHandoff` called mid-tick queues behind the in-flight renewLock through `runSerialized` and runs only after the tick settles.

**Skipped (with reasoning):**
- **controlClient.pushHandoff sync throw test** — reviewer noted "the contract says it never throws." It's a hypothetical defensive check; the result-object-never-throws contract is already pinned by the existing tests.

153 Pillar 3 tests (was 151). Full discord suite 2218/2218 green. Lint clean.

justin-layerv:
**cr-r5 fixes pushed (4377a54)**

All four substantive observations addressed:

1. **bindAddr IPv4 assumption (substantive #1):** comment now explicitly documents that `0.0.0.0` is v4-only and v6-only ENIs need `bindAddr: '::'`. The wire side (heartbeat + control-client validators) was already v6-ready.

2. **`res.destroyed` guard (substantive #2):** `sendJson` now short-circuits on `res.destroyed` (in addition to `headersSent`/`writableEnded`). Closes the client-abort → `request_aborted` → `sendJson(res, 400, ...)` → write-to-torn-down-socket path that would have surfaced as a server-level error.

3. **PR description vs. code (substantive #3):** the `expected+1` prose was in `gateway-control-client.js:27` (not the PR body). Corrected: the code passes `transferResult.version` (post-transfer) verbatim, and the standby's adopt uses it directly with no arithmetic.

4. **End-to-end IPv6 test (substantive #4):** new test binds control-channel on `::1` and round-trips a real `pushHandoff` over IPv6 loopback. Probes IPv6 availability in `beforeAll` and skips cleanly if unavailable (some CI runners disable v6).

**Smaller notes:** all positive acknowledgments — no actions needed.

154 Pillar 3 tests (was 153). Full discord suite 2219/2219 green. Lint clean.

justin-layerv:
**cr-r6 fixes pushed (a8d6918)**

All items addressed:

1. **pushHandoff missed deleteOwnRow (substantive #1):** every exit branch now calls `peerHeartbeat.deleteOwnRow().catch(() => {})`. Best-effort, swallowed — pushHandoff result is unchanged on cleanup failure. Tests pin happy-path + no_peer + cleanup-failure-swallow.

2. **isConnecting required (substantive #2):** dropped the `() => false` default; now mandatory at factory time. A wiring oversight in PR 13b.3 fails loud at boot rather than producing a rare race under inbound handoff. Validation test updated.

3. **closed sentinel on leader (substantive #3):** new `closed` flag latched in pushHandoff; `start()` short-circuits permanently after handoff. Mirrors the watchdog's `exited` flag.

4. **405 method_not_allowed (substantive #4):** GET /control/yours now returns 405 with `Allow: POST` header (RFC 9110 §9.1). Distinct from 404 on unknown paths for operator triage.

5. **Connection cap / rate limit (substantive #5):** noted; SG remains the documented perimeter. No code change requested.

6. **Minor nits:**
- `split('?', 1)[0]` → `split('?')[0]` readability ✅
- `TODO(post-13b.2-bake)` on `targetLockHolder` placeholder ✅
- Backoff cap assertion test (catches a future maxAttempts bump crossing the cap) ✅
- `rememberNonce` dead branch — already documented; reviewer says "Fine as defensive code." No change.

159 Pillar 3 tests (was 154). Full discord suite 2224/2224 green. Lint clean.

justin-layerv:
**cr-r7 fixes pushed (6504898)**

All 4 actionable items addressed:

1. **Leader loop backstop (concern #1):** `gateway-leader.js` loop body now wraps `await runSerialized(step)` in try/catch, mirroring the watchdog's existing pattern. A synchronous throw escaping step's internal try/catch (e.g., a future row-shape regression throwing inside Promise.all's array construction) is now logged + retried rather than silently exiting the loop. Test pins the survival.

2. **`'aborted'` event deprecation (concern #2):** switched to `req.on('close')` + `req.destroyed` check, the modern Node 17+ pattern. The abort path is now forward-compatible across LTS bumps.

3. **Naming symmetry (concern #3):** watchdog's `exited` → `closed` to match leader's `closed` sentinel. Both encode the same "permanently done; start() is a no-op" semantic.

4. **`rememberNonce` simplification (concern #4):** dropped the unreachable has→delete→set bump-to-newest; now insert-only. Comment notes the contract for any future caller that bypasses verify().

**Skipped (with reasoning):**
- **#5** (trailing-slash URL match) — reviewer said "Practically fine." 404 on `/control/yours/` is the correct behavior.
- **#6** (body-read error fan-in) — reviewer said "Minor — current behavior is correct."

160 Pillar 3 tests (was 159). Full discord suite 2225/2225 green. Lint clean.

justin-layerv:
**cr-r8 fixes pushed (157856a)**

Items addressed:

1. **httpRequest sync-throw harden (potential bug #1):** wrapped `httpRequest(...)` + `req.write/end` in try/catch. A future Node version tightening hostname/port validation no longer breaks the "pushHandoff never throws" contract — settles the promise with `reason: 'http_error'` instead.

2. **401 reason leak (potential bug #2):** documented in the module header as an intentional debuggability/security tradeoff at this VPC-internal trust boundary.

3. **requestTimeout backstop note (potential bug #3):** documented in the module header that `server.requestTimeout` is the 5s reaper for a misbehaving client that ignores `Connection: close` on the paused-socket path.

4. **deleteOwnRow outside runSerialized (potential bug #4):** restructured `pushHandoff` so deleteOwnRow runs OUTSIDE the serialized chain (single cleanup site after the runSerialized result is collected). deleteOwnRow doesn't touch gateway-lock state, so serializing it just added SIGTERM-critical-path latency. Comment names the why-outside.

5. **Dual-secret timing observability (potential bug #5):** noted as "not actionable" by reviewer.

**Test coverage suggestions:**
- **Watchdog-meets-leader race:** the contract is already pinned in isolation — watchdog has an `isConnecting` backoff test, leader has an `isConnecting` toggle test. An integration test was attempted in cr-r2 but the production wiring lives in PR 13b.3; the contracts here are pinned at the per-module boundary.
- **Heartbeat row missing lock_holder fallback test:** already exists at `gateway-leader.test.js` "falls back to placeholder lock_holder when peer row lacks the field" (added in cr-r0).

**Smaller items:** `runSerialized` naming (already has intent comment, kept); `VERIFY_REASONS` replication for control-channel codes (not yet leaving the module, defer).

160 Pillar 3 tests. Full discord suite 2225/2225 green. Lint clean.

justin-layerv:
**cr-r9 fixes pushed (96d39b3)**

All actionable items addressed:

1. **releaseLockForExit ordering (issue #1):** documented that `heldLock=false` before await is OK ONLY because the sole caller (watchdog exhaustion) exit(1)s immediately afterward.

2. **runSerialized rejection log → warn (issue #2):** promoted from debug to warn level. Inner work paths already log at warn/error so this backstop should never fire visibly — but if it ever does, surfacing at warn keeps it visible at default prod levels.

3. **`void hmac` cleanup (issue #3):** dropped the unused `hmac` arg from `makeSignedEnvelope`'s signature. Workaround is gone.

4. **manager.connect timeout note (issue #6):** documented as a TODO for PR 13b.3 wiring — the WS shim's connect() must have an internal timeout (or external wrapper).

5. **Placeholder prefix `peer/` → `placeholder/` (issue #7):** makes the back-compat fallback path immediately recognizable in transferLock traces.

6. **httpRequest indentation (style nit):** cleaned up.

**Tests:**
- **Push-during-inbound-handoff serialization test:** new test pins the previously-untested branch (existing tests covered tick-during-push, push-during-tick, inbound-during-inbound).
- **LRU bump-to-newest test renamed** to reflect what it actually pins now that `rememberNonce` is insert-only (size cap, not bump).

**Skipped (with reasoning):**
- **#4** (expected_version validator placement) — already correct per reviewer's re-read, no change.
- **#5** (wrapEnvelope buffer alloc) — noted as noise at 250B payload.
- **#8** (401 reason SOC alert note) — documented tradeoff, no code change.

162 Pillar 3 tests (was 160). Full discord suite 2226/2226 green. Lint clean.

justin-layerv:
**cr-r10 fixes pushed (75bcee5)**

All 3 nits addressed:

1. **Nit #1 (warn/debug drift):** corrected the adoptLockFromHandoff catch-block comment — runSerialized's backstop logs at warn (since r9), so the error-level adopt log is justified by the routing context it carries (activeInstanceId + expectedVersion), not by the prior level distinction.

2. **Nit #2 (runSerialized arms wording):** clarified "same `work` reference in both arms" → "two distinct thenable handlers, both calling the same `work` callee."

3. **Nit #3 (rememberNonce while → if):** keeping `while`. Reviewer flagged this as "Fine as-is; flagging only if you want to tighten." The `while` is the idiomatic "drain to cap" pattern for an LRU eviction loop and is no slower than `if` for the one-eviction-per-insert case. Future-proofing for a future caller that might insert in batches.

**Observations (no action):** manager.connect timeout (already TODO'd for PR 13b.3), body-cap memory peak (within bounds), rate limit (out of scope, VPC SG is perimeter), heartbeat scan (already documented).

No code changes. Full discord suite 2226/2226 green. Lint clean.

justin-layerv:
**cr-r11 fixes pushed (0f0df97)**

All 5 concerns + 4 nits addressed:

**Concerns:**
1. **manager.connect internal timeout (concern #1):** new `inboundConnectTimeoutMs` (default 5s) races connect against a timer; on timeout the handoff throws, finally clears connecting=false, watchdog takes over. Leader is now self-sufficient — no longer wiring-dependent for the hung-resolve recovery path.

2. **releaseLockForExit caller-coupling (concern #2):** renamed `releaseLockForImmediateExit` so the contract is at the call site rather than buried in a comment.

3. **Outer envelope JSON.parse precedes verify (concern #3):** the doc-comment clarification is covered by the existing module headers (the 8KB cap bounds the DoS surface; verify-before-parse refers to the inner signed body). No code change.

4. **Placeholder TODO concrete deadline (concern #4):** added target removal date 2026-07-01 (~6 weeks bake) and a tracking expectation.

5. **req.write + req.end → req.end(wire) (concern #5):** done. One fewer libuv hop.

**Nits:**
- URL parsing non-strictness — flagged as fine for this single-endpoint HMAC channel, skip.
- `typeof payload === 'object'` accepts arrays — tightened with `!Array.isArray(payload)` ✅
- Catch handler missing URL/method — added to the error log ✅
- IPv6 test skip `console.warn` — kept as-is (Jest doesn't have describe-time runtime skip cleanly; the warn is a one-liner on environments without v6).

**Test coverage:**
- inboundConnectTimeout firing test ✅
- byteCap boundary test (body exactly at cap is NOT 413) ✅

163 Pillar 3 tests (was 162). Full discord suite 2228/2228 green. Lint clean.

justin-layerv:
**cr-r12 fixes pushed (e4dc738)**

Items addressed:

1. **Hung-connect race (potential issue):** real bug. The previous race pattern (race + finally clear) left the underlying `manager.connect()` pending after a timeout, while the watchdog's next tick (1s) would observe `!isConnecting` and fire its OWN `manager.connect()` — exactly the concurrent-connect race the flag exists to prevent.

   Fix: `connecting=false` is now attached to the underlying connect promise's settlement (both fulfillment AND rejection), NOT the race result. The handler throws on timeout, but the flag stays held until the original settles. Watchdog backs off through the entire window. Test pins both states.

2. **`releaseLockForExit` header naming (observation):** verified — module header at line 10 already uses `releaseLockForImmediateExit`. No change needed (reviewer may have been looking at the diff pre-sed).

3. **expected_version upper bound (security):** rejects `> Number.MAX_SAFE_INTEGER`. Bounds the noise from an HMAC-authenticated peer adopting an absurd cursor value.

4. **Content-Type 415 (robustness):** rejects non-JSON content types with `415 unsupported_media_type`. Distinguishes "probe with wrong content type" from "real client with a bad envelope" in operator triage. Accepts missing Content-Type for defensive future callers.

5. **Evicted nonce accepted again (test coverage):** new test pins that the LRU size cap is a deliberate ceiling, not a correctness primitive — catches a future refactor that might make eviction equivalent to permanent rejection.

**Skipped (with reasoning):**
- **Logger interface shape check** — flagged as "consider" not "should fix"; existing truthy check is sufficient and tightening would break existing test regexes for marginal gain.
- **Explicit bindAddr at 13b.3 wiring layer** — out of scope for this PR; noted for the wiring PR.
- **`runSerialized` chain helper extract** — reviewer noted "nothing required here."

164 Pillar 3 tests (was 163). Full discord suite 2229/2229 green. Lint clean.

justin-layerv:
**cr-r13 fixes pushed (29ba64e)**

Items addressed:

1. **Hung manager.connect() pinning connecting=true (issue #1):** documented as load-bearing trust in @discordjs/ws's bounded-settlement contract (the shim wraps it; the underlying WebSocket has its own handshake/heartbeat timeouts). Racing a second timer would re-introduce the concurrent-connect race that cr-r12 just fixed. Comment names the escape paths if the upstream contract ever drops.

2. **Concurrent inbound handoff while holding lock (issue #2):** added `heldLock && manager.isConnected()` guard. Stray handoff (duplicate retry, misrouted body) rejected cleanly without touching adopt or connect. Test pins the no-op behavior.

3. **Connection: close consistency (issue #3):** all short-circuit responses (404, 405, 415) now include `Connection: close`, matching the existing 413 path. Without it, keep-alive clients' unread body bytes confuse the HTTP parser on the next request.

4. **deleteOwnRow swallow (issue #4):** flagged as minor; current `.catch(() => {})` is right for the SIGTERM exit path. deleteOwnRow already logs at warn inside its own catch.

5. **Nonce LRU trust-boundary pin (issue #5):** already documented in module header.

6. **adopt-throw test (issue #6):** new test pins flag invariants (heldLock=false, connecting=false, no connect call) AND that the runSerialized chain accepts the next op cleanly.

7. **Control-client comment drift (issue #7):** fixed.

8. **Module header tightening (issue #8):** flagged as "not a blocker." Headers are load-bearing for the concurrency/race rationale; trimming risks losing precision.

**Test coverage:**
- Stray-handoff rejection ✅
- adopt-throw flag invariants ✅
- Never-settling connect — documented as the @discordjs/ws contract; pinning a deliberate-stall test would test the wrong thing.
- At-cap-exactly body — existing test asserts `!== 413`; the boundary contract is the not-rejected outcome, not whether the bytes happen to verify.

166 Pillar 3 tests (was 164). Full discord suite 2231/2231 green. Lint clean.

justin-layerv:
## cr-r14 applied — commit 5413071

Round 14 fixes pushed. All 204 Pillar 3 tests + 2240 discord suite green; eslint clean.

### Applied
1. **gateway-leader factory: validate `manager.isConnected`** — was only checking `manager.connect`. The watchdog calls `manager.isConnected()` on every tick; a wired manager missing that method would have failed at runtime with a less obvious "x is not a function" instead of the explicit factory throw. Tests updated to include `isConnected()` in the manager fixture.
2. **`releaseLockBestEffort` helper** — extracted from the three `lock.releaseLock().catch(() => {})` sites in `pushHandoff`. Silent swallow hid a real signal: a recurring DDB outage during handoff is now `logger.warn` with the bail reason so it surfaces in default prod log levels. TTL still absorbs the failure — caller behavior unchanged.
3. **`headersTimeout < requestTimeout` invariant** — `headersTimeout = Math.min(max(1000, requestTimeoutMs/2), requestTimeoutMs - 100)`. Node 18.7+ requires `headersTimeout < requestTimeout` (else the server refuses the config). The previous max-only floor would have violated the invariant if a caller set `requestTimeoutMs` ≤ 2000.
4. **Deprecated `res.on('aborted')` → `close + destroyed`** in `gateway-control-client.js`. Matches the inbound-side idiom in `gateway-control-channel.js`. Fake test fixture updated to drive the modern Node behavior.
5. **`listFreshPeers` defense-in-depth filter** — drops peer rows missing a parseable `ip` or valid `port`. Mirrors the write-time validator. A corrupted row no longer aborts handoff at the validator throw; the caller moves on to the next-freshest peer. (Closes the read-side gap noted in cr-r14 #6.)
6. **Content-type rejection coverage** — pins 415 rejection of `application/json-patch+json`, `application/jsonp`, `application/jsonish`, `text/plain`, `application/octet-stream`; pins acceptance of `application/json`, `application/json; charset=utf-8`, `Application/JSON` (case), `application/json;charset=UTF-8` (no space). Protects the prefix regex from regression.

### Skipped / deferred
- None this round; all 7 actionable items addressed.

Awaiting cr re-review on commit 5413071.

justin-layerv:
## cr-r15 applied — commit bf252b2

All 4 actionable items addressed. 2242/2242 discord tests green; eslint clean.

### Applied
1. **Sync-throw blind spot in `handleInboundHandoff`** — `manager.connect()` is now inside the `try`. If `connect()` throws synchronously (defensive: contract is async, but a future shim regression could surface it), the catch arm clears `connecting=false` so the watchdog can take over the retry. Without this, `connecting=true` would pin forever and the watchdog would no-op every tick. Mirrors the `try/catch` posture in `gateway-control-client.js:118-208`. New test pins the invariant.
2. **`runSerialized` duality comment** — added a paragraph clarifying that the *returned* `next` propagates work's rejection to the caller while `inFlight` swallows it (via the second `.then` arm) so the next op chains off an always-fulfilled promise. The two diverge on purpose; reading the function it's not obvious why.
3. **`controlClient.pushHandoff` wrapped in try/catch** — returns `{ pushed: true, ackReason: 'push_threw' }` on throw. The control-client's documented contract is "never throws" but its synchronous arg validators do (peerIp/peerPort shape). With `listFreshPeers`' ip/port filter (cr-r14) the path is unreachable today, but the SIGTERM caller cannot handle exceptions cleanly. transferLock has already moved DDB ownership so the standby owns the lock either way — the watchdog catches lock-held + WS-disconnected within ~1 s. New test pins the result + error log.
4. **`selfLockHolder` factory arg removed** — answer to the cr-r15 #11 clarification: it was dead weight. The `lock` instance is constructed externally via `createGatewayLock({lockHolder, ...})` and owns its own holder string; the leader never consumed `selfLockHolder` for anything. Removed the validator + tests; PR 13b.3 wiring can pass it directly to the lock constructor.

### Skipped (note-only items, no change required)
- #4 verify-side timing distinguishing current vs previous — both signers are valid, not exploitable. No change.
- #5 `timingSafeHexEqual` length-mismatch early return — attacker controls signature length on the wire; load-bearing comment already in code.
- #6 Nonce burns on routing-rejection paths — by design (acknowledged in module header).
- #7, #8, #9, #10 — affirmations of cr-r13 / cr-r14 fixes, no change.
- #12 `expected_version > Number.MAX_SAFE_INTEGER` redundancy — sanity ceiling, harmless.

Awaiting cr re-review on commit bf252b2.

justin-layerv:
## cr-r16 applied — commit 5100419

All 5 items addressed (3 correctness + 2 minor). 2244/2244 discord tests green; eslint clean.

### Applied
1. **Never-throws contract restored** — `gateway-control-client.js` validators now return `{ok:false, reason:'invalid_arg', arg}` instead of throwing. The module header documents "Returns a result object — never throws" as load-bearing for the SIGTERM caller; the previous `throw new Error('peerIp required')` calls were an async-rejected-promise from the caller's view, but a contract divergence and a footgun for any future caller. The heartbeat-side validator (`gateway-peer-heartbeat.js`) keeps its `throw` (write-time config check, not a hot-path runtime check).
2. **Response body cap** — added `responseByteCap` (default 8 KB, symmetric to the server's `bodyByteCap`). A misrouted POST hitting an HTML error page or a hostile in-VPC actor could otherwise stream multi-MB and OOM the active during SIGTERM. Cap-exceeded settles with `{ok:false, reason:'http_error', error:'response_body_too_large'}` and destroys the response so buffering stops immediately.
3. **`pushed`/`ackReason` → `transferred`/`pushAcked`** — the previous `pushed:true` read as "the HTTP push succeeded" but actually meant "DDB transferLock succeeded; the standby owns the lock regardless of push outcome." New shape:
   - `{transferred: false, reason: 'not_holding_lock' | 'no_peer' | 'peer_lookup_failed' | 'transfer_threw' | 'transfer_failed'}`
   - `{transferred: true, pushAcked: true}`
   - `{transferred: true, pushAcked: false, pushReason: 'timeout' | 'rejected' | 'http_error' | 'push_threw'}`
   Module-header `pushHandoff` doc and tests updated in lockstep.
4. **Non-thenable connect protection** — `gateway-leader.js` now tracks `lifecycleAttached` explicitly. If `manager.connect()` returns a non-thenable (so `.then(...)` throws AFTER `connectPromise` is already assigned), the catch arm sees `lifecycleAttached=false` and clears `connecting=false`. The previous `if (!connectPromise)` check missed this case — `connectPromise` IS assigned, just to a non-thenable. New test pins the invariant.
5. **LRU → FIFO terminology** — `gateway-hmac.js` module-header section renamed to "Nonce cache (FIFO-by-insertion)" with an explanation of why FIFO (not LRU) is correct for the replay-window invariant: we evict by insertion order, never re-rank on access. Behavior unchanged.

### Skipped (note-only items, no change required)
- Content-Type regex confirmation (#5) — `application/json5` correctly rejected, comment-only.
- `expected_version > Number.MAX_SAFE_INTEGER` redundancy (#6) — sanity ceiling, keep.

Awaiting cr re-review on commit 5100419.

justin-layerv:
## cr-r17 applied — commit 71342c4

6 items applied + 1 deferred to follow-up. 2248/2248 discord tests green; eslint clean.

### Applied
1. **Nonce burn on bad routing payload** (cr-r17 #1) — confirmed intentional, no change. Source comment already names this: "legitimate senders don't send mis-addressed bodies." Senders always `generateNonce()` per call so reused nonces aren't an issue.
2. **Watchdog `deleteOwnRow` symmetry** (#3) — added optional `deleteOwnRow` hook to `createConnectionWatchdog`. On exhaustion-exit, called best-effort alongside `releaseLock`, mirroring `gateway-leader.js`'s `pushHandoff` cleanup path. Closes the heartbeat-row discovery window immediately rather than waiting up to TTL (~60s). New tests pin: provided / throws / omitted (back-compat) paths.
3. **`headersTimeout` floor degeneration** (#4) — factory-level guard rejects `requestTimeoutMs < 1100`. Below that the formula `min(max(1000, req/2), req-100)` collapses toward 1ms, effectively disabling header-read protection. Fail loud at boot. Test pins the throw for 50/1099/non-integer/string.
4. **`BACKOFF_CAP_MS` natural-ceiling formula** (#5) — fixed off-by-one in the test comment. Highest backoff that actually sleeps is at `attempts = maxAttempts - 1`, so `2^(maxAttempts-1) * 100`, not `2^maxAttempts * 100`. Both formulas pin `cap > naturalCeiling` so the assertion still holds.
5. **`runSerialized` double-logging** (nit) — downgraded chain-catch from `warn` to `debug`. Every rethrowing op (tick / handleInboundHandoff / pushHandoff) already logs at warn/error before throwing; the duplicate at warn level confused log readers into thinking two things failed. Debug keeps the chain-keeping observability for future investigation.
6. **`Connection: close` on `wrong_peer`/`unknown_peer`** (nit) — added for consistency with 404/405/413/415 short-circuits.
7. **`handleInboundHandoff` JSDoc** (nit) — explicit `@param`, `@returns Promise<void>`, `@throws` notes.

### Deferred
- **Stuck-connect process-health check** (cr-r17 #2) — filed as #415 (priority: low, pr-polish-followup). Scope: detect `connecting=true` pinned > 30s on a never-settling `manager.connect()`, log + exit(1) so ECS replaces. The cr reviewer explicitly framed this as "Worth a follow-up issue for PR 13b.3 or later" — out of scope for PR 13b.2 (the modules-only PR), and PR 13b.3 (wiring) is the natural home.

Awaiting cr re-review on commit 71342c4.

justin-layerv:
## cr-r18 applied — commit 9ee98eb

All 6 design items + 2 nits addressed. 2250/2250 discord tests green; eslint clean.

### Applied
1. **ACK-vs-200ms regime documentation** (cr-r18 #1) — added a module-header section in `gateway-control-client.js` explicitly calling out that the standby's `manager.connect()` takes 1-3s while the active's timeout is 200ms, so `pushAcked: true` is rare. The wire ACK is only useful for sync-throw / 4xx error paths; success relies on the watchdog. Recommends PR 13b.3 log `pushAcked` as a metric.
2. **Stuck-connect issue #415 cross-referenced** (#2) — added a one-liner in `gateway-leader.js`'s bounded-settlement comment pointing to #415 as the preferred recovery (process-level health check + exit) over racing another timer.
3. **`runSerialized` dead arm removed** (#3) — simplified `inFlight.then(() => work(), () => work())` to `inFlight.then(() => work())`. `inFlight` is always fulfilled (the next assignment's catch arm absorbs rejections), so the second arm was dead code. Comment now names the actual load-bearing invariant: the always-fulfilled `inFlight` chain.
4. **Closed sentinel uniformity** (#4) — added `if (closed)` guards to `handleInboundHandoff` (throws `leader_closed`) and `releaseLockForImmediateExit` (no-ops). Previously only `start()` checked. Now the terminal-state invariant is uniform across the API surface. Two new tests pin: inbound-after-pushHandoff throws + adoptLockFromHandoff not called; release-after-pushHandoff doesn't re-call DDB.
5. **Content-Type absent treatment** (#5) — accepted as-is per the cr review's own analysis (HMAC catches everything that matters downstream).
6. **`DEFAULT_RESPONSE_BYTE_CAP` rationale comment** (#6) — clarified: expected responses are 15-50 bytes, 8KB cap is ~150× expected, only fires on pathological inputs (misrouted POST hitting HTML error page).

### Nits
- **`releaseLockBestEffort` comment ordering** — reorganized so `releaseLockBestEffort` and `runSerialized` each have a clean leading comment without preceding-block context.
- **`placeholder/` removal TODO** — filed as #416 (post-13b.2-bake fallback removal), linked in the source comment.

### Notes
- cr noted comments occasionally repeat the same point (3 places restate `connecting=true` outliving `Promise.race`). Defers to future readers' preference — the rationale is load-bearing enough that each location's reader benefits from local context. No change this round.

Awaiting cr re-review on commit 9ee98eb.

justin-layerv:
## cr-r19 applied — commit bfe7348

8 items applied (1 critical comment fix + 7 nits/defensive). 2255/2255 discord tests green; eslint clean.

### Applied
1. **Header-timeout comment fix** (cr-r19 #1, the only "fix before merge" item) — corrected `gateway-control-channel.js` comment "Defaults (5s request, 1s headers)" → "With defaults (5 s request) this yields 2.5 s headersTimeout — the floor only kicks in below requestTimeoutMs ≈ 2 s". Matches the actual formula output.
2. **Factory-validation tightening** (#2) — added `Number.isInteger() && > 0` guards for all caller-provided numeric knobs:
   - `bodyByteCap` (channel)
   - `timeoutMs`, `responseByteCap` (client)
   - `maxAttempts`, `pollIntervalMs` (watchdog)
   - `tickIntervalMs`, `inboundConnectTimeoutMs` (leader)
   Posture matches `requestTimeoutMs` (cr-r17). Test pins for bodyByteCap.
3. **Content-Length pre-check** (#3) — 413 short-circuits before any stream reads when declared Content-Length > bodyByteCap. Streaming cap remains load-bearing (clients can lie about Content-Length); pre-check just saves per-chunk work on unambiguous bad requests. Malformed values (non-digit, negative, etc.) are ignored. Three new tests pin: over-cap short-circuit + no data drained, malformed values ignored, under-cap pass.
4. **secrets.current docs** (#4) — relaxed error message to "non-empty string" and added inline comment that production SSM-loaded value is hex but module enforces only non-empty (createHmac accepts any string).
5. **Content-Type regex trailing whitespace** (#5) — accepts `application/json ` (RFC 9110 OWS after media type). Existing tests still pass.
6. **IPv6 zone-id assumption documented** (#6) — added comment to `gateway-peer-heartbeat.js` noting awsvpc ENI gives global-scope addresses; link-local scope-id stripping would need to be added at both sides if that ever changes.
7. **`runSerialized` op-name** (cr-r19 #8) — added `opName` argument threaded through to the chain-catch debug log. Updated all 5 call sites (`tick`, `inbound-handoff`, `pushHandoff`, `releaseLockForImmediateExit`, `_stepForTest`) so a future bypass is trivially recognizable in logs.
8. **Test combinatorial pin** (#9) — added watchdog test for the `deleteOwnRow absent + releaseLock throws` permutation. Pins that `exit(1)` still fires.

### Deferred (note-only)
- **HMAC misuse metrics** (cr-r19 #7) — `bad_signature`/`replay`/`stale` should emit metrics, not just warn logs. Aligned with PR 13b.3's planned `pushAcked` metric work. Out of scope for the modules-only PR; module-header comments already note where metric hooks would go.

Awaiting cr re-review on commit bfe7348.

justin-layerv:
## cr-r20 applied — commit 9838ba5

All 4 actionable items applied (1 race-fix + 3 nits/observations). 2257/2257 discord tests green; eslint clean.

### Applied
1. **Inbound-handoff race fix** (cr-r20 #1, the only "weigh against merge" item) — added inner `if (closed) throw` re-check at the top of `handleInboundHandoff`'s `runSerialized` closure. The outer check protects post-pushHandoff calls but didn't catch the window where an inbound queues BEFORE pushHandoff sets closed. New race-pin test reproduces the scenario: hold a tick at a blocking `renewLock`, queue an inbound behind it, fire pushHandoff (sets `closed=true`), release the tick — inbound rejects with `leader_closed` and `adoptLockFromHandoff` never runs.
2. **Content-Type now required (strict)** (#2) — `gateway-control-channel.js` rejects missing Content-Type with 415, matching the cr review's triage-clarity argument. Distinguishes "probe / curl-without-`-H`" (415) from "bad envelope" (400). Test fixture sets `application/json` by default so other tests don't need to; explicit test pins missing-CT → 415.
3. **deleteOwnRow timing comment fix** (#3) — the previous "Closes the discovery window immediately" was only true post-serialized-chain. Comment now explicitly notes the brief window between `closed=true` and end-of-serialized-work where the row is still visible, and that the inner closed re-check (above) handles any queued inbound during that window.
4. **Watchdog `step` closed guard** (#4) — `if (closed) return` at the top. Test-tool robustness: `_stepForTest` manual re-entry post-exhaustion would have re-fired `releaseLock` + `deleteOwnRow`. The production loop never re-enters, but the guard makes the closed invariant uniform across the API.

### Affirmations (no change)
- **`findInvalidHandoffField` unknown-field forward-compat** (#5) — explicit acceptance of the design choice; HMAC trust boundary supports it.
- **`controlClient.pushHandoff` close handler** (#6) — already correct; `settle` idempotent makes the rare double-fire harmless.

Awaiting cr re-review on commit 9838ba5.

justin-layerv:
## cr-r21 applied — commit 0ff2890

All 3 actionable items applied (2 fixes + 1 test addition). 2265/2265 discord tests green; eslint clean.

### Applied
1. **`Connection: close` on body_read_failed** (cr-r21 #1) — added the header to the 400 body_read_failed response. Lands after a partial stream error, so leftover bytes would confuse a keep-alive client's parser. Matches the posture of 413/415/wrong_peer/unknown_peer bail paths.
2. **Inner closed re-check in `releaseLockForImmediateExit`** (#2) — mirrors `handleInboundHandoff`'s pattern. Outer check catches post-pushHandoff sequential calls; inner guard closes the race where a watchdog call queued BEFORE pushHandoff flipped closed could pop after closed=true latches. Sole practical consequence is a redundant releaseLock call (caller exits right after), but the symmetry makes the closed-state invariant uniform across the API.
3. **Connection: close test coverage** (#5) — new describe block pins the header on all bail paths: 404 / 405 / 415 wrong CT / 415 missing CT / 413 / 400 body_read_failed / 400 wrong_peer / 400 unknown_peer. Locks in the keep-alive safety contract so a future refactor doesn't silently drop the header.

### Affirmations (no change)
- **Unhandled-rejection observation** (#3) — both arms attached to `connectPromise.then`, Node treats it as handled. Would have failed CI already if not.
- **Content-Type leading whitespace** (#4) — Node parses Header values with OWS stripped per RFC 9110, so the regex doesn't need a leading `\s*`. Calling it out as a "test you could add" — current coverage is fine.

Awaiting cr re-review on commit 0ff2890.

justin-layerv:
## cr-r22 applied — commit b05c517

All 6 items applied (1 deadlock fix + 5 correctness nits/docs). 2267/2267 discord tests green; eslint clean.

### Applied
1. **Release-lock ceiling fix** (cr-r22 #1, the only "LGTM-after-this" blocker) — `Promise.race` the exhaustion-path `releaseLock()` against a 3s ceiling (`releaseLockCeilingMs`, injectable for tests). Without the race, a hung inbound-handoff (parked inside `manager.connect()` with `connecting=true` latched per #415) would block the leader's serialization chain forever and `exit(1)` would never fire — failover slot held with no live gateway. On ceiling miss we log + exit anyway; row fades via TTL. New test pins the race with a never-resolving `releaseLock` mock and a 10ms ceiling.
2. **Verify-first threat-model note** (cr-r22 #2) — added a paragraph to the verification-order comment block explicitly calling out that nonce-burn-on-routing-rejection is safe ONLY because verify runs first. If a future change moves routing checks before verify, an attacker without the HMAC secret could spam wrong-peer bodies to evict legitimate nonces.

### Nitpicks
3. **Walked back `Connection: close` on wrong_peer / unknown_peer** — body is fully consumed by then (no leftover bytes), so close is only needed for indeterminate-state paths. Now matches `invalid_payload` / `unauthorized` / `handoff_failed`. Earlier-reviewer-suggested consistency was wrong on this point; this reviewer's first-principles argument is correct. Tests flipped to assert NOT-matched.
4. **Documented `timingSafeHexEqual` timing-safety limits** — length-check short-circuits leak "wrong length" but that's wire-observable already (an attacker controls the signature length they send). Constant-time guarantee applies only to the byte-wise compare once both sides are valid sha256-hex.
5. **`MAX_SAFE_INTEGER` ceiling test** — `Number.isInteger(2^53)` returns true (representable-but-unsafe integer), so the `expected_version > Number.MAX_SAFE_INTEGER` check IS reachable. New test pins the rejection.
6. **Non-thenable comment fix** — corrected: `.then` is undefined on a non-thenable return, so the call site throws synchronously BEFORE `lifecycleAttached` flips true. The catch handles it correctly; comment now matches the mechanism.

Awaiting cr re-review on commit b05c517.

justin-layerv:
cr-r23 applied as commit `8413d96`.

### Applied
- **Watchdog `manager.connect()` ceiling.** Added `CONNECT_CEILING_MS = 8_000` and raced the inline connect against it inside `step()`. A hung connect would otherwise pin the tick forever, attempts wouldn't advance, and the `exit(1)` recovery would never fire — leaving a stuck standby holding the failover slot. Mirrors the leader's `inboundConnectTimeoutMs` pattern; 8 s gives headroom over the 1–3 s typical Discord WS cold-connect while staying tighter than the SIGTERM ECS deadline.
- **Factory validators in `gateway-connection-watchdog.js`:** added positive-integer guards for `releaseLockCeilingMs` and the new `connectCeilingMs`. Symmetric with the existing `maxAttempts`/`pollIntervalMs` guards.
- **Factory validators in `gateway-hmac.js`:** added positive-integer guards for `freshnessWindowMs` (would reject every body as stale) and `nonceLruSize` (the worst-failure-mode: would *silently disable* replay protection because the bounded FIFO would evict the just-inserted nonce immediately, leaving the cache empty for the next `verify()`).
- **Extracted local `raceWithCeiling(promise, ceilingMs, label)` helper** to unify the two adjacent `Promise.race`-vs-`setTimeout(reject)` patterns in the watchdog (connect-ceiling and release-lock-ceiling). The inline duplication had become an in-file dup with this PR; lifting to a one-liner keeps the call sites readable without losing the rationale (still documented at each call site).
- **Tests.** Added the hung-connect ceiling-race test in `gateway-connection-watchdog.test.js` (uses `manager.connect` mocked to a never-settling promise with `connectCeilingMs: 10` so the ceiling fires fast), the two `gateway-hmac.test.js` factory-validation tests, and the pre-headers-abort settlement test in `gateway-control-client.test.js` (covers the req-error path where `statusCode` is undefined and `close` intentionally does NOT settle).

### Skipped
None. All cr-r22 feedback addressed.

### Verification
- Full discord test suite: 2271 passing across 61 suites.
- `eslint` clean on the five changed files.
- `/simplify` pass: only actionable in-PR finding was the in-file ceiling duplication, addressed via the `raceWithCeiling` extraction. Comment trims applied on test narration + coverage-speak.

CI is running on `8413d96`.

justin-layerv:
cr-r24 applied as commit `6de0fed`.

### Applied
- **(1) `Number.isFinite(payload.ts)` guard.** Replaced `typeof payload.ts !== 'number'` with `!Number.isFinite(payload.ts)` in `gateway-hmac.js#verify`. `1e1000` is valid JSON (the JSON number grammar permits any decimal-with-exponent form) but V8 parses it to `Infinity` via double-precision overflow. The old check let `Infinity` through; today the downstream `Math.abs(now - Infinity) > windowMs` is `true` so it falls into the stale branch — but the contract is cleaner pinned at the field-shape gate. Added a wire-JSON test that round-trips `1e1000` to `Infinity` and asserts `missing_field` rejection.
- **(2) DRY'd watchdog ceiling parameter docs.** The `releaseLockCeilingMs` / `connectCeilingMs` parameter comments duplicated the rationale already in the `RELEASE_LOCK_CEILING_MS` / `CONNECT_CEILING_MS` constants above. Trimmed to a single pointer comment.
- **(3) `expected_version` redundant `typeof` check dropped — but kept the `MAX_SAFE_INTEGER` ceiling.** The cr-r24 review suggested dropping `> Number.MAX_SAFE_INTEGER` as unreachable; that's not quite right — `Number.isInteger(2**53) === true` (representable but unsafe), so the explicit `> MAX_SAFE_INTEGER` IS doing work that `isInteger` doesn't (and the existing "exceeds MAX_SAFE_INTEGER" test would have broken if dropped). Kept the ceiling, dropped the redundant `typeof === 'number'` pre-check (which `isInteger` already covers), and updated the comment to explain why the MAX_SAFE_INTEGER clause is not redundant.

### Acknowledged / not changed
- **(4) Public test seams in `__internals`.** Soft suggestion. Reviewer noted "current shape is consistent with the rest of the gateway module." The `_` prefix convention is uniform across all five new files; a global rename would be a large diff with no functional improvement. Leaving as-is.
- **(5) Alert-storm risk from `leader_closed` rejections during SIGTERM.** Confirmed: both `leader_closed` log sites in `gateway-leader.js` (lines 330, 346) are `logger.warn` not `logger.error`, so they don't page. Surface area is bounded: the active sends one pushHandoff per SIGTERM; the peer's watchdog handles `leader_closed` by retrying. No change needed.

### Verification
- Full discord test suite: 2272 passing across 61 suites (+1 from cr-r23).
- `eslint` clean on all four changed files.
- `/simplify` pass: clean — diff is small and tight.

CI is running on `6de0fed`.

justin-layerv:
cr-r25 applied as commit `5303d69`.

### Applied
- **(1) Watchdog ceiling clarification.** Added a joint header comment above `RELEASE_LOCK_CEILING_MS` / `CONNECT_CEILING_MS` clarifying these bound the watchdog's OWN failure-replace path (worst-case ~11s well inside the 30s ECS graceful-stop) — NOT the SIGTERM critical path, which has its own ~200ms client-side cap in `gateway-leader.js`'s `pushHandoff`.
- **(6) `Map<nonce, true>` → `Set<nonce>` for `seenNonces`.** Set insertion order is also spec-preserved; expresses membership-only intent directly and drops the throwaway `true` value. `seenNonces.add(nonce)` / `seenNonces.values().next().value` substitute cleanly for the Map calls.
- **Test coverage suggestion.** Added a multi-chunk `bodyCapExceeded` test in `gateway-control-client.test.js` pinning the synchronous-destroy + flag-mark contract: three chunks (under-cap, over-cap, post-cap) — asserts `res.destroyed === true` on the over-cap trigger AND that the post-cap chunk is silently dropped (no second cap-exceeded log, no second settle). The existing single-chunk cap test was good for the basic path; this one specifically exercises the flag-mark drop path the reviewer flagged.

### Acknowledged / explicitly no-action per reviewer
- **(2) `connecting=true` permanent stick** — tracked in #415, reviewer called out it's "not a blocker" and the alarm-landing question is for 13b.3's smoke testing.
- **(3) Content-Type regex** — "probably fine for this private control channel."
- **(4) `req.pause()` worst-case memory** — "Calling out for completeness; not a request to change."
- **(5) Ternary-chain validator in `gateway-control-client.js:111-118`** — "I wouldn't fight to change it."
- **(7) `requestTimeout=5s` vs `manager.connect()` 1-3s interaction** — "functionally fine. Just noting the interaction."
- **(8) `lockHolder` placeholder/ prefix** — "Good."

### Confirmed (item 5 from reviewer's wording)
- `leader_closed` rejection log sites in `gateway-leader.js` (lines 330, 346) are both `logger.warn` (not `error`) — they won't page on-call during a clean SIGTERM. Already addressed in cr-r24 reply; mentioning again for completeness since cr-r25's wording was similar.

### Verification
- Full discord test suite: 2273 passing across 61 suites (+1 multi-chunk cap test).
- `eslint` clean on all three changed files.

CI is running on `5303d69`. Reviewer's bottom line on cr-r25: "Ready to ship after the reviewers' own pass" — i.e., the PR is functionally ready. All cr feedback has been actioned or explicitly acknowledged-no-action per reviewer's own wording.

justin-layerv:
cr-r26 applied as commit `fe0f86f`.

### Applied (all 5 actionable items)
- **Nit 1 (gateway-leader.js:626) — `.catch(() => {})` on `peerHeartbeat.deleteOwnRow()`.** Took the reviewer's *alternative* path ("log at debug, matching runSerialized's catch-arm") rather than dropping outright. The existing test "a deleteOwnRow failure does NOT bubble into the pushHandoff result" pins this as a load-bearing defense-in-depth property; dropping the .catch broke it. Downgraded to a debug log so the path is now explicitly visible instead of silently swallowed.
- **Nit 2 (gateway-peer-heartbeat.js) — `lockHolder` not type-validated.** Added a factory validator: when provided, must be a non-empty string. Mirrors the `secrets.previous` posture in `gateway-hmac.js`. New test pins coverage across `''`, number, boolean, object (all reject) + `undefined`/omitted/valid string (all accept).
- **Nit 3 (gateway-leader.js:546-547) — `||` fallback.** Switched to `??` so a corrupt empty-string row from a misbehaving writer doesn't silently take the placeholder branch. With the new heartbeat-side validator (Nit 2) the misshape can't be written from this codebase, but read-side defense-in-depth is consistent with the rest of the module.
- **Bug 1 (gateway-control-channel.js:242-243) — redundant `Number.isFinite(declaredLength)`.** The `/^\d+$/` regex already guarantees `parseInt` returns a finite non-negative integer; dropped the redundant check. Comment updated to document the guarantee.
- **Test coverage gap — readRequestBody `'close' → 'end'` idempotence.** Added two tests (resolve-when-close-after-end, reject-when-close-without-end) pinning the settle-idempotence contract that lets the function survive Node 17+'s destroy-on-end stream lifecycle. The `_readRequestBodyForTest` seam already existed.

### Acknowledged / explicit no-action per reviewer
- **Bug 2** — backoff cap dead at default config. Reviewer: "Not a bug — flagging only because a casual reader might think it's load-bearing."
- **Security confirmation** — SSM hex-shape validation TODO. Out of scope for PR 414 (will land in 13b.3's loader).

### Verification
- Full discord test suite: 2276 passing across 61 suites (+2 new tests).
- `eslint` clean on all 5 changed files.

Reviewer's verdict was "Approve modulo the minor nits above. The two concrete fixes I'd suggest before merge: (1) drop or downgrade the redundant `.catch`... and (2) decide on lockHolder validation." Both fully addressed; all other nits/bugs also applied. CI is running on `fe0f86f`.

justin-layerv:
cr-r27 applied as commit `57ff1f9`.

### Applied
- **Item 2 — `step()` closed-guard.** Added `if (closed) return;` to `step()` in `gateway-leader.js` for symmetry with `handleInboundHandoff`'s inner closed re-check. Production loop already guards via `running=false`, but the `_stepForTest` seam enters through the same `runSerialized` chain; a stray post-close test call now bails cleanly. Test pins the new contract. Updated the existing "tick during in-flight pushHandoff" test to reflect the new behavior — the queued tick still serializes, but post-resolve it now bails on `closed=true` rather than re-acquiring (matches the closed-state invariant the rest of the module enforces).
- **Item 4 — HMAC length pre-check before compute.** Added `SHA256_HEX_LENGTH = 64` constant and a pre-check at the top of `verify()` that returns `bad_signature` without invoking `hmacHex()` when the candidate is the wrong length. Cheap forward-compatible defense per the reviewer's note: "if a future caller swaps in a heavier MAC, the length pre-check should move ahead of the compute." The "candidate was wrong length" timing leak is unchanged — already documented above `timingSafeHexEqual`.
- **Test coverage gap — stuck `connecting=true` behavior.** Added `stays at attempts=0 indefinitely while isConnecting=true (escape hatch #415)` to `gateway-connection-watchdog.test.js`. Pins the current behavior tracked by #415 (20 ticks with `isConnecting=true`, no `connect()`/`releaseLock()`/`exit()` fire, `attempts` stays at 0) so the process-level health monitor that closes #415 is loud about changing it.

### Acknowledged / explicit no-action
- **Item 1 (stuck `connecting=true` escape hatch)** — tracked by #415, reviewer says "Tracked by #415 per the comment, but worth confirming PR 13b.3's wiring layer includes the alert-and-exit guardrail." Out of scope for PR 414. New test now pins current behavior.
- **Item 3 (`deleteOwnRow` outside serialized chain)** — reviewer's own conclusion: "Different modules; consistent behavior; fine."
- **Item 5 (`Number.isFinite(payload.ts)` comment overstates threat — claims `JSON.parse('1e1000')` throws)** — empirically verified: `JSON.parse('1e1000')` returns `Infinity` in V8, doesn't throw. The existing Infinity-rejection test pins this (passes against current code). The comment is correct as-is; the reviewer's claim was mistaken. Mentioning to surface the empirical evidence rather than silently disagree.
- **Verbose-comment density** — reviewer explicitly "not blocking; just calling out that the density is on the high end."
- **Ternary chain in `gateway-control-client.js`** — reviewer: "not worth changing."
- **`startControlChannelServer` teardown handle** — reviewer: "Fine to defer to that PR" (13b.3 wiring).

### Verification
- Full discord test suite: 2278 passing across 61 suites (+2 new tests).
- `eslint` clean on all 4 changed files.

Reviewer's verdict on cr-r27: "Approve modulo the #415-tracked stuck-connecting=true follow-through." #415 deferral is correct (PR scope); new test pins current behavior so the eventual fix is observable. CI is running on `57ff1f9`.

justin-layerv:
cr-r28 reviewed against current HEAD `57ff1f9`. Reviewer's verdict: "I'd merge as-is." All 7 observations are either already covered by existing tests or explicit reviewer no-actions — no new commit needed.

### Already covered by existing tests
- **Item 3 — watchdog exhausts mid-pushHandoff.** Unit-level pieces are pinned:
  - `gateway-leader.test.js:928` — "releaseLockForImmediateExit after pushHandoff is a no-op (closed)"
  - `gateway-connection-watchdog.test.js:382` — "calls deleteOwnRow on exhaustion when hook is provided"
  - The composition (watchdog's releaseLock hook = leader.releaseLockForImmediateExit) is a 13b.3 wiring concern, not in scope for PR 414.
- **Item 4 — BACKOFF_CAP_MS / maxAttempts / BACKOFF_BASE_MS relationship.** Pinned at `gateway-connection-watchdog.test.js:79` — "BACKOFF_CAP_MS stays above the natural ceiling at DEFAULT_MAX_ATTEMPTS" with `naturalCeiling = (2 ** (DEFAULT_MAX_ATTEMPTS - 1)) * BACKOFF_BASE_MS`. A future bump to maxAttempts=8 would fail this test.
- **Item 5 — IPv6 [::1] end-to-end round-trip.** Already present at `gateway-control-client.test.js:466-540` — "end-to-end — IPv6 loopback (::1)" with `beforeAll` probe-then-skip if CI environment disables IPv6.

### Explicit reviewer no-action (per reviewer's own wording)
- **Item 1 (current/previous HMAC timing side channel)** — "moot" on VPC-internal HMAC-authenticated channel; worth keeping in mind if the trust boundary moves.
- **Item 2 (pushHandoff `closed=true` neutralizes releaseLockForImmediateExit)** — by design; "should be enforced at the wiring layer in 13b.3" — out of scope for PR 414.
- **Item 6 (duplicate Content-Length header)** — "Low priority on a VPC-internal HMAC-authenticated channel where smuggling has no exploitable downstream."
- **Item 7 (Content-Type parameter)** — "No action needed."
- **Verbose comment density** — "Optional."

### Convergence — all four pr-ready final-check criteria green
1. **DE acceptance** — yes; reviewer: "I'd merge as-is."
2. **CI passing** — yes; all 11 status checks green on `57ff1f9` (Dependency Review, Dependency Age Check ×4, Shared Tests, Secrets Scan, discord: Build and Test, CodeQL, PR Title Check, Claude Code Review).
3. **cr feedback considered** — yes; every observation accounted for above.
4. **/simplify clean** — yes; agent verdict: "Clean. Convergence verified — nothing actionable."

PR is merge-ready. ✅

PR #417, opened 2026-05-16
feat(discord): pillar 3 hot-standby wiring layer

justin-layerv:
## Summary

Wires the Pillar 3 modules (lock, peer-heartbeat, hmac, control-channel, leader, watchdog) into the gateway bootstrap path, gated by `ENABLE_GATEWAY_HOT_STANDBY`. Ships flag-off everywhere — single-replica gateway stays the production default until PR 14 flips `desired_count=2`.

## Why

This is the application-side completion of the zero-downtime deploy story. With both replicas running and the flag on, every deploy push-handoffs the live WebSocket from the outgoing task to the incoming task inside the SIGTERM window — eliminating the IDENTIFY/RESUME cold-start gap that today drops dispatches on every deploy.

The lock + peer-heartbeat (PR 13b.1) and HMAC + control-channel + leader (PR 13b.2 / #414) primitives have already landed. PR 13b.3 is the wiring layer that makes them load-bearing in `index.js`.

## What changed

**Shim seam (`connect: false`)** — Both replicas boot the shim without opening a WebSocket. The leader/watchdog drives `manager.connect()` after winning the DDB lock or receiving an inbound handoff. Without this, both replicas would race to IDENTIFY against the same bot token and Discord would flap the session identity. Also adds a public `getManager()` accessor (replaces `_getManagerForTest`).

**Leader API** — new `hasStartedTickLoop()` predicate consumed by the standby `/health` probe.

**Boot guards** — `unsupportedRoleHotStandbyCombo` rejects hot-standby on non-gateway roles or with RESUME=false; `missingHotStandbyKeys` surfaces missing `INSTANCE_ID` / `INSTANCE_IP` / `GATEWAY_HANDOFF_HMAC`.

**Secret loader (new module)** — `gateway-hmac-secret-loader.js` parses + validates the SSM-injected `GATEWAY_HANDOFF_HMAC` env var (`{current, previous?}` JSON with 64-char hex values). Fail-loud at boot.

**Wiring (`startHotStandby` in `index.js`)** — constructs the chain after `gatewayShim.start({ connect: false })` resolves; awaits the control-channel server's `listening` event before starting the leader so the peer can never pushHandoff to a not-yet-listening port.

**`/health` probe split** — active path returns `shim.isReady()`; standby returns `leader.hasStartedTickLoop()`. Without this, the standby would 503 forever and get ECS-replaced on the first `--start-period` lapse.

**SIGTERM branching** — `pushHandoffShutdown` (active path: pushHandoff with 9 s ceiling + 12 s outer belt-and-suspenders, then `process.exit`) vs the legacy `gracefulShutdown` (standby / flag-off). `tryStop()` helper de-duplicates the new teardown call sites.

**Config + `.env.example`** — new entries: `ENABLE_GATEWAY_HOT_STANDBY`, `INSTANCE_ID`, `INSTANCE_IP`, `GATEWAY_CONTROL_PORT`, `GATEWAY_CONTROL_BIND_ADDR`, `GATEWAY_HANDOFF_HMAC`.

## Test plan

- [x] 22 new cases for the secret loader (shape + hex + previous-optional + error-shape contracts)
- [x] 16 new cases for the boot-requirements gates (synergy combos + missing-keys)
- [x] 5 new cases for shim `connect: false` + `getManager()`
- [x] 3 new cases for `leader.hasStartedTickLoop()`
- [x] Full suite: 2320 tests pass / lint clean
- [ ] Sandbox deploy: flag-off boot is identical to current behavior (Pillar 2 only)
- [ ] Sandbox deploy: flag-on with `desired_count=1` boots clean — leader cold-acquires, watchdog drives connect, /health returns 200
- [ ] Sandbox deploy: flag-on with `desired_count=2` — active/standby split, pushHandoff during rolling deploy keeps session continuity (requires PR 14)

## Rollout

1. Merge PR 13b.3 (this PR) — flag-off, no behavior change.
2. Bake on prod for 24-48 h to confirm no regression in Pillar 2 path.
3. PR 14 lands infra change to `desired_count=2` + AZ-spread placement.
4. Flip `ENABLE_GATEWAY_HOT_STANDBY=true` on the gateway task def.
5. PR 15 runs chaos validation (deploy-during-flow + RESUME-fail + queue-backpressure).

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -139,6 +139,51 @@ ENABLE_OPENNHP_FEATURES=false
 # if Discord's 60 s buffer still has it, otherwise IDENTIFYs).
 # ENABLE_GATEWAY_RESUME=false
 
+# Gateway hot-standby (zero-downtime design, Pillar 3). When set to
+# the literal string "true" AND PROCESS_ROLE=gateway AND
+# ENABLE_GATEWAY_RESUME=true, two gateway replicas run: one holds
+# the DDB-backed lock and owns the live WebSocket (active), the
+# other heartbeats and waits (standby). On SIGTERM the active
+# pushes ownership via an HMAC-authenticated POST to the standby,
+# which drives manager.connect() inside the push window — the
+# next Discord dispatch lands without an IDENTIFY/RESUME cold-start.
+#
+# Requires:
+#   * ENABLE_GATEWAY_RESUME=true (cross-process session continuity)
+#   * PROCESS_ROLE=gateway (only the gateway tier constructs the
+#     manager that hot-standby drives)
+#   * INSTANCE_ID, INSTANCE_IP, GATEWAY_HANDOFF_HMAC (below)
+#
+# Default off. The boot-requirements check rejects every unsupported
+# combination at startup (see src/boot-requirements.js).
+# ENABLE_GATEWAY_HOT_STANDBY=false
+
+# Per-replica identity for the hot-standby control plane. Each task
+# must see a unique value — the DDB lock + peer-heartbeat rows key
+# on this. ECS task-defs inject the task's ARN suffix (or task
+# metadata GUID) here at boot. Required when
+# ENABLE_GATEWAY_HOT_STANDBY=true, unused otherwise.
+# INSTANCE_ID=
+# INSTANCE_IP=
+
+# Bind + listen ports for the in-VPC control channel that receives
+# pushHandoff envelopes from the outgoing leader. 0.0.0.0 binds to
+# all interfaces (awsvpc-routable IP); 127.0.0.1 would be unreachable
+# from peer tasks. Restrict the listening port to peer tasks via the
+# task's security group (see qurl-integrations-infra terraform).
+# Defaults: 7800 / 0.0.0.0.
+# GATEWAY_CONTROL_PORT=7800
+# GATEWAY_CONTROL_BIND_ADDR=0.0.0.0
+
+# HMAC secret for the control channel. JSON-shaped string:
+# `{"current":"<64-char-hex>","previous":"<64-char-hex>"}`. Each
+# value is a 32-byte HMAC-SHA256 key in lowercase hex. `previous`
+# is optional and covers the rolling-deploy rotation window
+# (dual-accept). Production: SSM SecureString
+# /qurl-bot-discord/GATEWAY_HANDOFF_HMAC, injected via the task-def
+# `secrets =` block — never set in .env files on real deploys.
+# GATEWAY_HANDOFF_HMAC=
+
 # Maximum number of in-flight detached handlers the SQS consumer
 # tolerates before pausing receives (backpressure). When the count
 # reaches the cap, pollOnce returns early and waits 100ms before
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -246,6 +246,47 @@ function missingHotStandbyKeys(cfg) {
   return missing;
 }
 
+// Shape checks for INSTANCE_ID + INSTANCE_IP (run AFTER missing-keys
+// passes). Matches the secret-loader's "fail at boot, not at first
+// use" posture: a misconfigured task-def that injects the literal
+// `${ECS_TASK_ARN}` (unsubstituted shell expansion) or
+// `INSTANCE_IP=10.0.0.999` (non-IPv4) would pass the presence gate
+// and surface as a baffling DDB-lock-can't-acquire or peer-
+// unreachable error at runtime. A cheap regex catches both.
+//
+// INSTANCE_ID: rejects the `${...}` template-literal pattern (the
+// common ECS task-def substitution-failure footgun). Otherwise
+// permissive — the downstream lock/heartbeat code does not require
+// a specific format.
+//
+// INSTANCE_IP: must parse as an IPv4 dotted-quad. The hot-standby
+// awsvpc deployment puts each task on a unique ENI with a v4
+// address; v6 is not in scope for Pillar 3 today (the control
+// channel binds to the v4 ENI, peers reach each other over v4 SG
+// rules). If v6 ever lands, this check loosens.
+//
+// Returns an array of operator-facing message strings (one per
+// problem) or [] when all values are well-shaped. Hot-standby off
+// → skip entirely.
+const IPV4_RE = /^(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)$/;
+function invalidHotStandbyValues(cfg) {
+  if (!cfg.ENABLE_GATEWAY_HOT_STANDBY) return [];
+  const problems = [];
+  if (cfg.INSTANCE_ID && cfg.INSTANCE_ID.includes('${')) {
+    problems.push(
+      `INSTANCE_ID looks like an unsubstituted template literal ('${cfg.INSTANCE_ID}'). ` +
+      'Check the ECS task-def is resolving the placeholder against the task metadata endpoint.'
+    );
+  }
+  if (cfg.INSTANCE_IP && !IPV4_RE.test(cfg.INSTANCE_IP)) {
+    problems.push(
+      `INSTANCE_IP must be a valid IPv4 address (got '${cfg.INSTANCE_IP}'). ` +
+      'Hot-standby uses v4 for the control-channel binding + peer reach; v6 is not in scope today.'
+    );
+  }
+  return problems;
+}
+
 // PLACEHOLDER is treated as missing because the SSM parameter
 // ships with that literal sentinel value; remediation ("seed a
 // real key") is identical to the empty-key case.
@@ -346,6 +387,7 @@ module.exports = {
   unsupportedRoleResumeCombo,
   unsupportedRoleHotStandbyCombo,
   missingHotStandbyKeys,
+  invalidHotStandbyValues,
   shouldRegisterInteractionListener,
   missingMapCommandKeys,
   GOOGLE_MAPS_API_KEY_PLACEHOLDER_SENTINEL,
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -183,6 +183,11 @@ function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled, st
 }
 
 // Parallel to unsupportedRoleResumeCombo for ENABLE_GATEWAY_HOT_STANDBY.
+// Caller guarantees the upstream resume combo check has already run
+// (resumeEnabled=true → shipper+ddb already validated upstream), so
+// the 3-arg signature here is sufficient — no need to re-check
+// shipper/storeType.
+//
 // Two unsupported shapes:
 //
 //   1. hotStandby=true with role!=gateway — the leader coordinator
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -265,10 +265,16 @@ function missingHotStandbyKeys(cfg) {
 // channel binds to the v4 ENI, peers reach each other over v4 SG
 // rules). If v6 ever lands, this check loosens.
 //
+// Leading-zero octets are rejected (`01.02.03.04` would parse as
+// octal under some resolvers); each octet is `0` alone, `1-9`, or
+// `1[0-9]-25[0-5]` with no leading zero. ECS task-def injection
+// produces canonical no-leading-zero v4 strings; this just closes
+// the operator-typo door.
+//
 // Returns an array of operator-facing message strings (one per
 // problem) or [] when all values are well-shaped. Hot-standby off
 // → skip entirely.
-const IPV4_RE = /^(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]?\d)$/;
+const IPV4_RE = /^(?:(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]\d|\d)\.){3}(?:25[0-5]|2[0-4]\d|1\d\d|[1-9]\d|\d)$/;
 function invalidHotStandbyValues(cfg) {
   if (!cfg.ENABLE_GATEWAY_HOT_STANDBY) return [];
   const problems = [];
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -242,7 +242,12 @@ function missingHotStandbyKeys(cfg) {
   const missing = [];
   if (!cfg.INSTANCE_ID) missing.push('INSTANCE_ID');
   if (!cfg.INSTANCE_IP) missing.push('INSTANCE_IP');
-  if (!cfg.GATEWAY_HANDOFF_HMAC) missing.push('GATEWAY_HANDOFF_HMAC');
+  // GATEWAY_HANDOFF_HMAC presence is surfaced via `hasGatewayHandoffHmac`
+  // (boolean flag) rather than the raw value — the secret string is
+  // never exposed as a config-object property to keep it unreachable
+  // through heap-dump-accessible references. See config.js's
+  // `takeGatewayHandoffHmac` for the security rationale.
+  if (!cfg.hasGatewayHandoffHmac) missing.push('GATEWAY_HANDOFF_HMAC');
   return missing;
 }
 
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -182,6 +182,65 @@ function unsupportedRoleResumeCombo(role, resumeEnabled, eventShipperEnabled, st
   return null;
 }
 
+// Parallel to unsupportedRoleResumeCombo for ENABLE_GATEWAY_HOT_STANDBY.
+// Two unsupported shapes:
+//
+//   1. hotStandby=true with role!=gateway — the leader coordinator
+//      drives the gateway-ws-shim's manager handle; only the gateway
+//      tier constructs the shim, so http/combined have nothing to
+//      hand off. Rejecting at boot avoids a deploy where the hot-
+//      standby flag is set in a uniform env block but the role isn't
+//      gateway — which would silently no-op the leader path and
+//      mask the misconfig as "the lock never gets acquired."
+//   2. hotStandby=true with resume=false — pushHandoff hands the
+//      incoming task a snapshot of session_id + sequence so it can
+//      RESUME without dropping events. Without the cross-process
+//      RESUME path, "handoff" degenerates to "both replicas
+//      IDENTIFY against the same token" — Discord rejects the
+//      second IDENTIFY and the second replica flaps the session.
+//
+// Returns the operator-facing message on rejection or null on
+// success. Same string-or-null shape as unsupportedRoleResumeCombo.
+function unsupportedRoleHotStandbyCombo(role, hotStandbyEnabled, resumeEnabled) {
+  if (!hotStandbyEnabled) return null;
+  if (role !== 'gateway') {
+    return (
+      `ENABLE_GATEWAY_HOT_STANDBY=true requires PROCESS_ROLE=gateway (got '${role}'). ` +
+      'The hot-standby control plane (leader election + push handoff) drives the ' +
+      'gateway-ws-shim manager, which only the gateway tier constructs. http and ' +
+      'combined roles have no manager to hand off. Set PROCESS_ROLE=gateway on the ' +
+      'gateway task def, or leave ENABLE_GATEWAY_HOT_STANDBY unset on http/combined.'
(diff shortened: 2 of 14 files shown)


justin-layerv:
## cr-r1 fixes pushed (35bb5ef)

Addressed all five items from the prior review:

1. **JSON.parse error leaking secret prefix (security)** — `gateway-hmac-secret-loader.js` no longer interpolates `err.message`. The surfaced error is now a static string. Test tightened to scan for every 8-char substring of a high-entropy raw value.

2. **`pushHandoffShutdown` timeout exit code (correctness)** — Hard-exit setTimeout now fires `process.exit(1)` regardless of the originating signal's code. Distinguishes timed-out handoff from clean transfer in ECS/dashboards.

3. **SIGTERM-during-`startHotStandby` race** — Added `if (isShuttingDown) return;` after the `listening` event await. This is the only await in the function (the other steps are synchronous factories), so this single guard closes the window.

4. **Test coverage gap for shutdown branching / readiness probe / tryStop** — Extracted three pure helpers to `apps/discord/src/gateway-shutdown-helpers.js`:
   - `shouldUsePushHandoffShutdown` — SIGTERM branch predicate
   - `selectGatewayReadinessProbe` — /health probe factory (active vs standby)
   - `tryStop` — null-safe stop() invoker

   16 new unit tests covering the lock-aware branching, the active/standby probe split, the pre-`startHotStandby` boot window (probe returns false), the mid-deploy lock-flip case, and `tryStop`'s error-swallow shape.

   One subtlety worth flagging: `selectGatewayReadinessProbe` takes a `getGatewayLeader` **callback** (not the handle directly). The probe is wired before `startHotStandby` runs, so closing over the `let gatewayLeader` binding via a callback is the only way each probe firing reads the current value. The doc comment on the helper + a behavioral test pin this contract.

5. **`unsupportedRoleHotStandbyCombo` signature drift (nit)** — Added a comment documenting that the 3-arg shape is intentional (upstream `unsupportedRoleResumeCombo` already validates shipper+ddb when resumeEnabled=true).

**Test suite**: 2337 pass / lint clean. **/simplify**: ran the three-agent pass on the round-1 diff; only nits returned. One false-positive flagged by the quality reviewer (claim that the `isShuttingDown` guard only covers one of multiple awaits — there's actually only one await in `startHotStandby`); accepted two comment-trim findings.

justin-layerv:
## cr-r2 fixes pushed (19d6e1f)

Addressed all 7 items from the cr-r2 review:

1. **`pushHandoffShutdown` test coverage** — Extracted body into `runPushHandoffShutdown` in `gateway-shutdown-helpers.js` with injectable `exit` / `scheduleHardExit` / `ceilingMs` / `forcedExitCode` deps. 9 new unit tests pin every contract the cr called out by name (success → exit code; throw → exit code; ceiling-fires-exit-1-even-on-SIGTERM-0; default ceiling = 12_000 ms; timer unref'd; result fields forwarded to log). The wrapper in `index.js` is now ~6 lines and owns only the `isShuttingDown` gate.

2. **HMAC validation boot-timing** — Hoisted `loadGatewayHmacSecret()` from inside `startHotStandby` to module-load (adjacent to `missingHotStandbyKeys`). All hot-standby misconfigs (env-var presence AND secret-format validity) now surface upfront before any I/O. Parsed value stashed in `gatewayHmacSecrets`; `startHotStandby` reads without re-parsing.

3. **`startHotStandby` race-guard integration test** — Filed as #418 (low priority follow-up, properly labeled). Acknowledging the reviewer's note that fixturing this needs a refactor to make `startHotStandby` injectable; deferred per pr-ready spec.

4. **`tryStop` stack** — Now logs `{ error: err.message, stack: err.stack }`. Test asserts both fields.

5. **Teardown order swap** — `gracefulShutdown` now closes the control-channel server FIRST, then watchdog stop, then leader stop. Comment makes the no-inbound-during-teardown invariant load-bearing.

6. **Defensive comment** — `selectGatewayReadinessProbe` hot-standby branch documents the `gatewayShim` non-null guarantee (boot-requirements gate rejects hot-standby without RESUME).

7. **Brand/style** — no action needed per reviewer.

### /simplify cleanups on the round-2 diff

The simplify pass found one more reuse opportunity I picked up: the duplicate `new Promise(resolve => server.close(err => { warn + resolve }))` pattern (two call sites in `gracefulShutdown` — `httpServer.close` and `controlChannelServer.close`). Extracted as `tryClose` parallel to `tryStop`. Four new unit tests cover it.

Also cleaned up test-quality nits from /simplify:
- Dropped the `_exitCalls` shadow array (now using the jest spy directly)
- Removed the `void shutdown` anti-pattern + misleading comment (the never-resolving handoff path now resolves explicitly after the hard-exit timer fires, so the orphan promise awaits cleanly)
- Removed the "does NOT call `gatewayShim.stop()`" test that was testing the test setup, not the SUT
- Trimmed duplicated rationale in test comments
- Trimmed the "ddb down" example from `tryStop`'s comment (the referent was from the test, not prod logs)

**Test suite**: 2349 pass / lint clean.

justin-layerv:
## cr-r3 fixes pushed (481e49d)

Addressed all 7 items + 2 /simplify cleanups from the cr-r3 review:

1. **(correctness) dropped SQS sends** — `runPushHandoffShutdown` now drains `eventPublisher` concurrently with `pushHandoff`. The active replica's in-flight SQS sends carry dispatches that arrived on OUR WebSocket; the standby cannot replay them. Draining inside the 12 s ceiling closes the gap between "deploy lost a few send events" and "clean handoff". /simplify caught that my initial async-IIFE shape duplicated the existing `tryStop` helper — refactored to `tryStop('event-publisher', eventPublisher, logger)`, which is null-safe AND catches both sync throws and async rejects via async-function semantics.

2. **(consistency) active-path teardown asymmetry** — Documented the three load-bearing skips (`shim.stop`, `controlChannelServer.close`, leader/watchdog `.stop`) in `runPushHandoffShutdown`'s header. Leader's `closed=true` sentinel short-circuits any late inbound-handoff envelopes; calling them symmetrically would add latency to the SIGTERM critical path with no correctness gain.

3. **(observability) `hasStartedTickLoop` comment** — Now explicitly documents loop-exists-vs-loop-progressing. A *hung* tick (DDB call that never resolves) would NOT flip this predicate. Acceptable for now given the 2 s renew + 60 s lock TTL absorbing a hang within ~1 min. Test comment updated to match. Follow-up tracked.

4. **(cleanup) `controlChannelServer.once` listener leak** — Listeners now mutually remove each other on first event. Without this, the idle `error → reject` listener stayed attached after resolve and would `.reject()` on every runtime error (already routed to `gracefulShutdown(1)` via `onListenError`; duplicate path was a noisy-rejection footgun).

5. **(security) HMAC hex case-as-identity hazard** — Loader normalizes hex to lowercase on output. `createHmac` keys on UTF-8 bytes, so a rotation re-serializing the same logical key with different case would silently break dual-accept. Normalizing at the chokepoint eliminates the hazard regardless of operator discipline.

6. **(nit) `missingHotStandbyKeys` test `.sort()`** — Dropped; assert natural push order to document the operator-facing log-message ordering.

7. **(nit) `forcedExitCode` comment** — "Hard-coded to 1" → "Defaults to 1" (matches the configurable parameter that tests exercise).

### /simplify cleanups on the round-3 diff

- Replaced the async IIFE in `runPushHandoffShutdown` with `tryStop` (reviewer correctly noted async-function semantics already catch sync throws; the IIFE was duplicating null-safe + log-shape behavior the existing helper provides).
- The "in parallel" test now actually proves parallelism — uses a pending pushHandoff promise and asserts `publisher.stop` was invoked BEFORE pushHandoff resolves. Previously the test only proved "both got called", which a sequential implementation would also pass.

**Test suite**: 2352 pass / lint clean.

justin-layerv:
## cr-r4 fixes pushed (581c1ec)

All cr-r4 items addressed:

1. **(latent hazard) hard-exit timer never cleared** — Fixed. New `clearHardExit` dep (defaults to `clearTimeout`, injectable for tests) fires after the `await drainPromise`. In prod `process.exit` made this moot, but tests/non-terminal injected exits would have observed a spurious second `exit(1)` ~12 s after a clean shutdown.

2. **(observability) `exit(code)` on pushHandoff throw** — Switched to `forcedExitCode`. Three observable outcomes now map to three distinct exit codes:
   - Clean transfer → `exit(code)` (typically 0)
   - PushHandoff threw → `exit(forcedExitCode)` (default 1)
   - PushHandoff timed out → `exit(forcedExitCode)` (default 1, via hard-exit timer)
   Deploy SLI dashboards can distinguish "clean transfer" from "pushHandoff threw, standby will cold-acquire via TTL" instead of collapsing both into exit 0.

3. **(nit) SIGTERM null-leader comment** — Added above the SIGTERM/SIGINT handlers, names the pinning test by file for grep-discoverability.

4. **(nit) Cross-file ceiling reference** — `ceilingMs = 12_000` doc now references `gateway-leader.js`'s `DEFAULT_INBOUND_CONNECT_TIMEOUT_MS` for the 9 s internal ceiling.

5. **(deferred — judgment call)** Module-level `gatewayHmacSecrets` let. The reviewer acknowledged the trade-off (fail-fast at module load vs lifecycle co-location); keeping the current choice that favors fail-fast.

6. **(/simplify nits applied)**:
   - SIGTERM comment now explicitly covers SIGINT too (both have the same null-leader hazard).
   - "Clears hard-exit timer on throw path" test now also asserts `exit(1)` independently — catches a regression that swaps the throw-path back to `code`.

7. **(no-op nit)** `.env.example` `${project}/` placeholder convention — verified no such placeholder exists in this file today; the literal `/qurl-bot-discord/...` style is the only style used. No change needed.

**Test suite**: 2355 pass / lint clean.

justin-layerv:
## cr-r5 round resolved (commit f2e8191)

Applied all four items from the round-5 cr feedback:

1. **`signalShutdown` async + `.catch` on call sites** — both shutdown branches were already async fire-and-forget from the signal handler; an uncaught reject from a future refactor would surface as an `unhandledRejection`. The explicit `.catch` is defense-in-depth and routes to `logger.error` + `process.exit(1)`.

2. **`sync-by-design` comment above `gatewayLeader.start()` + `connectionWatchdog.start()`** — both schedule their own tick loop via `setTimeout` and return; awaiting here would hang on the loop's `stop()` promise. Comment pins the contract.

3. **Pre-start `handleInboundHandoff` unit test** — pins the SIGTERM-during-`startHotStandby` race contract. The control-channel server starts listening BEFORE `leader.start()` runs, so an in-flight handoff envelope can land on a leader whose tick loop has not begun. Test asserts `hasStartedTickLoop()=false` at the moment `manager.connect` fires AND that the adopt → flag → connect ordering still holds. Test passes; `runSerialized`'s `inFlight = Promise.resolve()` priming at factory time is what makes this work.

4. **`tryClose` logs `stack` alongside `error`** — symmetric with `tryStop`. Low-information today (`listener already detached`), but the next failure mode introduced here benefits from the call-site trace.

All 2356 tests pass locally. Waiting on cr-r6.

justin-layerv:
## cr-r6 round resolved (commit 7681213)

Two cr reviewers landed this round; resolving both sets.

### Latent listen-await hang (load-bearing fix)

Reviewer flagged that `server.close()` on a not-yet-listening server fires `close` (not `error` or `listening`), so the inline `await new Promise(...)` in `startHotStandby` would hang until `gracefulShutdown`'s force-exit timer (10 s) fires. Rescued in practice but a real footgun if the timer is ever raised or removed.

**Fix:** Extracted `awaitServerListening(server)` into `gateway-shutdown-helpers.js`. Three-way unwind: resolve on `listening`, reject on `error`, reject on `close` with `"closed before listening"`. Mutual listener removal across all three events prevents the idle-listener unhandled-rejection hazard.

8 new unit tests cover fast-path, listening, error, close, full listener cleanup on all three terminal events, and post-resolve idle-listener absence (the EventEmitter-rethrow trick proves no `error → reject` listener remained).

### Comments / documentation

- **Module-let concurrency invariant** above `gatewayLeader` / `controlChannelServer` / `connectionWatchdog` declarations. Node's single event loop guarantees read/write atomicity; tryClose/tryStop calls are null-guarded; `isShuttingDown` re-check after listen-await closes the inverse race.
- **Exit-code dual-knob collapse** documented in the SIGTERM/SIGINT comment block. Both handlers pass `code=0`, so production observes only 0 (clean) or 1 (anything else); the third knob exists for the unit-test surface + a future caller that wants signal-vs-fault distinction.

### Tests

- Explicit `eventPublisher: null` test (separate from the existing "omitted" case) pinning the SIGTERM-before-publisher.start() contract.

### Deferred (with reasoning)

- **`hardExit && typeof hardExit.unref === 'function'` simplification** — left as-is. The guard is defensive against any future test injection that returns a non-Timeout shape; cost is one line vs. eliminating the implicit contract that every injected `scheduleHardExit` must return an `.unref()`-able handle.
- **`INSTANCE_ID` / `INSTANCE_IP` rename to `GATEWAY_INSTANCE_ID` / `GATEWAY_INSTANCE_IP`** — task-def coordination cost. Will revisit pre-flag-flip if the env-var schema is still in flux at that point.
- **`gatewayHmacSecrets` module-level mutable `let` → hoist into `startHotStandby`** — reviewer noted it's minor and the current placement matches the "all boot-fails happen pre-start" pattern adjacent to it.

All 2365 tests pass locally; lint clean. Waiting on cr-r7.

justin-layerv:
## cr-r7 round resolved (commit fa68328)

Addressing both items the reviewer asked to block on, plus three nits.

### Correctness fixes

1. **`onListenError → gracefulShutdown(1)` now has `.catch`** — defense-in-depth symmetric with the SIGTERM/SIGINT handlers. `gracefulShutdown` handles its own errors internally today, but a future refactor that introduces an uncaught reject path would otherwise surface as an `unhandledRejection`.

2. **SIGTERM-during-`startHotStandby`-await regression test** — deferred to issue #419. Building index.js-level integration test scaffolding (mocking config, AWS SDK clients, Discord client, db, store) for one interleaving assertion is high cost. Each link of the chain is already pinned by unit tests in isolation:
   - `awaitServerListening` close-rejects → `gateway-shutdown-helpers.test.js`
   - `gracefulShutdown` `isShuttingDown` re-entry guard → `src/index.js:677`
   - `if (isShuttingDown) return;` guard inside `startHotStandby` → `src/index.js:970`

   The reviewer themselves classified this as a "confirmation" rather than a blocker; the issue captures the scope + acceptance criteria for a follow-up.

### Style cleanups

3. **`signalShutdown(code)` parameter dropped** — both call sites always passed 0, the signal itself never carries fault semantics. The third observable exit-code outcome lives in `runPushHandoffShutdown`'s unit-test surface + the direct `gracefulShutdown(1)` callers (boot-failure path, onListenError). Comment updated to explain.

4. **`ENABLE_GATEWAY_HOT_STANDBY` gate around the three Pillar 3 tryClose/tryStop calls in `gracefulShutdown`** — symmetric with the `ENABLE_GATEWAY_RESUME` gate immediately below. Saves 3 microtask hops per teardown on every HTTP-only / combined replica that never built the hot-standby surface.

5. **`db.close()` skip documented in `runPushHandoffShutdown`** alongside the existing shim-stop / control-channel-close skips. The 12 s ceiling + `process.exit()` forces socket cleanup at the OS level; the active gateway is not the system of record for any uncommitted DB state during SIGTERM (writes run on the worker tier; gateway DB reads are idempotent and re-driven by the standby's watchdog).

### Style flag noted (no action)

6. **EventEmitter-rethrow test couples to default behavior** — reviewer flagged this as a behavioral-dependence test. The accompanying comment already explains the trick. Leaving as-is.

### Security audit confirmed

7. **`config` is never logged wholesale** — grepped for `logger.*config\b` and `JSON.stringify(config` across `apps/discord/src/`; only specific fields (`config.isOpenNHPActive`) are logged. The `GATEWAY_HANDOFF_HMAC` raw value never reaches any logger.

All 2365 tests pass locally; lint clean. Waiting on cr-r8.

justin-layerv:
## cr-r8 round resolved (commit d8266d5)

Two cr reviewers landed since the last commit; addressing both sets in one push.

### Defense in depth

1. **Scrub HMAC secret references after consumption.** `gatewayHmacSecrets = null` + `delete config.GATEWAY_HANDOFF_HMAC` once `createGatewayHmac` captures the material. The live HMAC instance still holds the strings (unavoidable for sign/verify) but the redundant module-scope and config-object references are not. Closes the heap-dump-surfaces-key vector.
2. **Null `controlChannelServer` in `onListenError`** so `tryClose` skips a server in an error state. Mirrors the existing defensive null-out in `startGatewayHealthServer`'s error callback.
3. **`invalidHotStandbyValues(cfg)`** boot-time shape validation. Catches the two operator-config footguns the reviewer called out:
   - `INSTANCE_ID='${ECS_TASK_ARN}'` (unsubstituted task-def template)
   - `INSTANCE_IP='10.0.0.999'` or any non-IPv4 string
   Matches the secret-loader's "fail at boot, not at first use" posture. 9 new tests pin the boundary cases.

### Documentation

4. **INVARIANT comment** above the `if (isShuttingDown) return;` guard in `startHotStandby`: "no awaits below — adding one reopens the SIGTERM-mid-startup race."
5. **ECS deadline note** in gracefulShutdown's hot-standby teardown block: the layered 10 s force-exit + 30 s ECS SIGTERM ceilings are the deliberate outermost belt; no per-call timeout on tryStop/tryClose.
6. **`lockHolder` intent comment**: identifies the ROLE-owning entity (one shard, one gateway role), not the replica. Both active + standby of the same shard share the same `lockHolder` string; replica disambiguation is on `instanceId`.
7. **`pushHandoffShutdown` defaults comment**: 12 s = 9 s pushHandoff race + 3 s headroom < ECS's 30 s SIGTERM-to-SIGKILL window.
8. **`eventConsumer` skip noted** in `runPushHandoffShutdown` doc — gateway tier never runs a consumer (`isWorker = isHttp && ENABLE_EVENT_SHIPPER`). Documented for future role-shape refactors.
9. **Comment density**: trimmed restating / over-explanatory comments on `signalShutdown`, the SIGTERM/SIGINT loop, and the three-let concurrency invariant block.

### Tests

10. **Leader `stop()` before `start()`** unit tests (2 cases): must be a safe no-op (resolves immediately, no mock interactions) and must be idempotent across repeated pre-start calls. Pins the partial-construction-cleanup contract — `gracefulShutdown` calls `tryStop` on the leader if `startControlChannelServer` throws after `gatewayLeader` is assigned but before `.start()` runs.

### Deferred (with issue link)

11. **`lastTickAt` freshness check** for `hasStartedTickLoop()` — filed as issue #420. The 60 s lock TTL + peer cold-acquire path already provides recovery within ~1 min; threshold tuning needs sandbox data. Comment in `gateway-leader.js` updated to reference the issue (findable via `git grep`).
12. **End-to-end test for `startHotStandby` sequencing** — already tracked in issue #419 from cr-r7.

All 2376 tests pass locally; lint clean. Waiting on cr-r9.

justin-layerv:
## cr-r9 round resolved (commits eadf79d + 4abfd51)

cr-r9 was nearly clean — 5 nits and a near-LGTM. Resolving all five inline.

### Correctness / defense in depth

1. **SIGTERM vs SIGINT split** (eadf79d). Push-handoff now lives exclusively on SIGTERM. SIGINT (dev ctrl-c, manual `kill -2`) goes straight to plain `gracefulShutdown(0)`. Triggering a real DDB CAS + cross-AZ HTTP round-trip on a developer's ctrl-c was overkill; ECS uses SIGTERM for production replacement, so the push-handoff machinery matches the production-deploy story exclusively now.

2. **Leading-zero IPv4 rejection** (eadf79d). `invalidHotStandbyValues` now rejects octets like `01.0.0.1` — glibc's `inet_aton` and a handful of resolvers parse those as octal, so `010.0.0.1` would silently route the control channel to `8.0.0.1`. ECS injection produces canonical no-leading-zero strings; this just closes the operator-typo door. 1 new test pinning all four octet positions.

3. **`.unref()` exit-safety clarification** (eadf79d). One-line comment in `runPushHandoffShutdown` explaining why the unref'd hard-exit timer can't cause an early loop-idle: pushHandoff holds DDB + HTTP socket handles open during its await, and we exit synchronously after `await drainPromise` resolves — there's no window where the timer is the only loop handle.

4. **One-shot HMAC getter (#421 — addressed inline, not deferred)** (4abfd51). The earlier `delete config.GATEWAY_HANDOFF_HMAC` was partial defense: any module that captured `config.GATEWAY_HANDOFF_HMAC` at require time would keep the secret string alive across the heap regardless of the delete. Replaced with a module-private binding + `config.takeGatewayHandoffHmac()` one-shot getter that returns the value once and nulls the binding. The raw secret is no longer reachable through any property on the exported config object — `config.GATEWAY_HANDOFF_HMAC` doesn't exist. Boot-presence gate switched to the `hasGatewayHandoffHmac` boolean flag. 7 new tests pin: returns value on first call, undefined on second/third, property absence, frozen-at-load flag semantic. Closes #421.

### Bottom line

The reviewer said "code-level review is clean" pending sandbox-deploy items. The three sandbox items in the PR description (flag-off boot identical, flag-on single-replica, flag-on two-replica session continuity) are the remaining load-bearing checks — outside the code-review surface.

All 2384 tests pass locally; lint clean. Waiting on cr-r10.

PR #422, opened 2026-05-16
feat(discord): derive INSTANCE_ID/INSTANCE_IP from os module

justin-layerv:
## Summary

Hot-standby identity (Pillar 3, shipped in #417) requires `INSTANCE_ID` + `INSTANCE_IP` to be populated on every gateway replica. ECS Fargate awsvpc gives each task its own hostname and ENI, but it does **not** substitute `${ECS_TASK_ARN}` into task-def env-var values (only `command` and `entryPoint` see that interpolation) — so plumbing identity through env vars alone is a dead end on Fargate.

This PR derives both fields sync from Node's `os` module at config require-time:
- `INSTANCE_ID` ← trimmed env override || `os.hostname()`
- `INSTANCE_IP` ← trimmed env override || first non-internal IPv4 from `os.networkInterfaces()`, preferring `eth0` (the awsvpc ENI's stable name)

Env overrides are trimmed for parity with other env-derived values in `config.js` (GUILD_ID, STORE_TYPE, ALLOWED_GITHUB_ORGS) and remain for test pinning and debugging; under normal Fargate operation both run unset and the derivation produces correct values without any task-def coupling.

## Why this matters

This is the prereq for PR 14 (`gateway_desired_count=2` + AZ-spread placement). Without populated `INSTANCE_ID`/`INSTANCE_IP`, flipping `ENABLE_GATEWAY_HOT_STANDBY=true` would fail boot at the `missingHotStandbyKeys` gate.

The alternative — async fetch from `$ECS_CONTAINER_METADATA_URI_V4` — would push the missing-env guard past module load, add an HTTP dependency to startup, and create a network failure mode for boot. Sync derivation has none of those tradeoffs and works identically in ECS, local docker, and unit tests.

## Load-bearing invariant

The lock primitive depends on **two replicas in the same ECS service seeing different `INSTANCE_ID` values**. Fargate assigns a unique short alphanumeric hostname per task, which satisfies this. The invariant is called out at the `deriveInstanceId` call site so future readers don't have to reconstruct it.

## What changed

- `src/config.js`: `deriveInstanceId()` + `deriveInstanceIp()` helpers; both called once at module-load and cached into `module.exports`. Hostname-uniqueness invariant + module-load caching + `addr.family` Node-version assumption documented inline. Env overrides trimmed for parity with the rest of the file.
- `src/boot-requirements.js`: updated comments + error message (template-literal check still catches env-override paste mistakes)
- `src/index.js`: updated operator-facing boot error to reflect new derivation source
- `.env.example`: documents the new derivation semantics; both vars are overrides, not required
- `tests/config-instance-derivation.test.js`: 13 new tests (env-set wins, env-unset→derived, empty-string env for both ID + IP, whitespace-trim, whitespace-only fall-through, eth0 preferred, fallback when eth0 has only internal IPv4, fallback to non-eth0 when eth0 has no IPv4, IPv6-only returns null, internal IPv4 skipped on aliased eth0, env-override-passes-through contract end-to-end)
- `tests/boot-requirements.test.js`: updated one comment + test title to reflect env-override-paste failure mode

## Test plan

- [x] `npx jest` — all 2397 tests pass (13 new)
- [x] `npx eslint` — clean
- [x] /simplify pass — applied: eth0 skip on fallback loop, trimmed duplicate comments
- [x] cr-r1 + cr-r2 + cr-r3 + cr-r4 feedback applied

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -160,9 +160,12 @@ ENABLE_OPENNHP_FEATURES=false
 
 # Per-replica identity for the hot-standby control plane. Each task
 # must see a unique value — the DDB lock + peer-heartbeat rows key
-# on this. ECS task-defs inject the task's ARN suffix (or task
-# metadata GUID) here at boot. Required when
-# ENABLE_GATEWAY_HOT_STANDBY=true, unused otherwise.
+# on this. Derived in-process at boot:
+#   INSTANCE_ID ← os.hostname()           (Fargate per-task hostname)
+#   INSTANCE_IP ← first non-internal IPv4 (Fargate awsvpc eth0 ENI)
+# Setting either here is an override (tests, local docker, debugging);
+# under normal ECS Fargate operation leave both blank. Required when
+# ENABLE_GATEWAY_HOT_STANDBY=true.
 # INSTANCE_ID=
 # INSTANCE_IP=
 
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -253,14 +253,14 @@ function missingHotStandbyKeys(cfg) {
 
 // Shape checks for INSTANCE_ID + INSTANCE_IP (run AFTER missing-keys
 // passes). Matches the secret-loader's "fail at boot, not at first
-// use" posture: a misconfigured task-def that injects the literal
-// `${ECS_TASK_ARN}` (unsubstituted shell expansion) or
+// use" posture: an env override like `INSTANCE_ID=${ECS_TASK_ARN}`
+// (unsubstituted shell expansion an operator pasted by mistake) or
 // `INSTANCE_IP=10.0.0.999` (non-IPv4) would pass the presence gate
 // and surface as a baffling DDB-lock-can't-acquire or peer-
 // unreachable error at runtime. A cheap regex catches both.
 //
 // INSTANCE_ID: rejects the `${...}` template-literal pattern (the
-// common ECS task-def substitution-failure footgun). Otherwise
+// classic env-override substitution-failure footgun). Otherwise
 // permissive — the downstream lock/heartbeat code does not require
 // a specific format.
 //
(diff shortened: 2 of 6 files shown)


justin-layerv:
Thanks for the round-2 review. Applied in befb72e:

- **Hostname-uniqueness invariant**: documented at the `deriveInstanceId` call site (`config.js:24-31`). Calls out that the DDB lock + peer-heartbeat collision would surface a regression post-deploy as telemetry.
- **Module-load caching**: noted inline at the top of the helpers' comment block.
- **Dead `|| []`**: removed from `config.js:38` — `Object.keys()` can't return a missing key, so the guard never short-circuits.
- **Contract test require**: moved inside the `withFreshConfig` callback to share the same module graph as config (future-proofs against `os`-mock-dependent transitive imports).
- **PR description**: bumped "8 new tests" → "9 new tests".

**Deferred (with reasoning)**: the "prefer common dev names (en0, wlan0) before catch-all" suggestion. The reviewer flagged it as gated on "if it ever bites" — and there's no incident or signal. The existing dev-fallback comment already acknowledges the limitation. Filing a follow-up issue for a hypothetical that won't be actionable without real pain would be noise; happy to revisit if local-dev IP flap becomes a problem.

justin-layerv:
cr-r9 verdict received: **LGTM**. The 3 nits in this round are all explicitly marked non-blocking:

1. **Fallback re-walks eth0** — reviewer: "take it or leave it." Inline comment already pre-empts the question; not changing.
2. **`Object.values(ifaces)` order on multi-NIC dev hosts** — reviewer: "the behavior is fine." `INSTANCE_IP` override is the documented dev escape hatch.
3. **`isIPv4` trusts `addr.address` shape** — reviewer: "Not worth defending against unless you have seen Node do this in the wild." None observed; deferring.

**Out-of-scope item filed**: #423 — INSTANCE_ID collision metric for the leader coordinator. Reviewer explicitly flagged this as belt-and-suspenders for a hypothetical Fargate runtime regression and said "Not for this PR." Labeled `enhancement, javascript, priority: low` with clear repro/scope/acceptance criteria.

All four pr-ready convergence criteria are now green:
- ✅ DE acceptance
- ✅ CI passing (14/14)
- ✅ All cr feedback addressed (9 rounds; current verdict LGTM)
- ✅ /simplify clean

Ready to merge.

PR #434, opened 2026-05-19
refactor(discord): strip SQLite legacy storage, DDB becomes only backend

justin-layerv:
## Summary

- Deletes `src/database.js` (950 LOC) + `src/store/sqlite-store.js` + the `better-sqlite3` dep. Only supported backend is now `ddb`; `STORE_TYPE=sqlite` from a stale env file fails boot loud with a listing of valid backends.
- Local dev: `docker-compose.yml` spins up `amazon/dynamodb-local`; `scripts/provision-ddb-local.js` creates the 10 Store-contract tables `ddb-store` touches (idempotent re-runs, verified empirically — first run creates, second run reports `exists`).
- Test surface: 3 SQLite-only suites deleted (`database.test.js`, `database-module.test.js`, `revoke-filter.test.js` — all behaviors mirrored in `ddb-store.test.js`); 2 sub-describes trimmed from `send-pipeline-helpers.test.js`; 11 test files migrated `jest.mock('../src/database', …)` → `'../src/store'`; `orphan-token-sweeper.test.js` rewritten with a stateful in-memory store mock (its old shape relied on the deleted real SQLite layer).
- `store-contract.test.js` child-spawn tests carry `DDB_TABLE_PREFIX` + `AWS_REGION` into the spawn env (previously implicit from setup-env.js); fallback expectation updated from `sqlite` → `ddb`; added a regression test pinning that `STORE_TYPE=sqlite` from a pre-DDB-only env file fails boot.
- New `tests/provisioner-schema-parity.test.js` pins the provisioner ↔ `ddb-store.js` TABLES suffix-set parity. Drift surfaces at PR CI instead of as a runtime `ResourceNotFoundException` on first local `npm start`.
- `config.js` now refuses to boot with `DDB_TEST_ENDPOINT` set under `NODE_ENV=production` (closes the silent-redirect footgun the local-dev section makes more visible).

## Behavior change

`config.STORE_TYPE` default flips from `'sqlite'` to `'ddb'`. Confirmed with the user before implementation. Boot still rejects unknown values with a listing-of-valid-backends error.

## Notable

- **Dockerfile**: collapses from two stages (build + runtime) to one — `better-sqlite3` was the only dep needing native compile. Adds `--ignore-scripts` to `npm ci` now that every remaining dep is pure-JS (real supply-chain hardening — blocks postinstall-hook arbitrary code execution). Runs `npm ci` as the non-root `node` user so no post-install `chown -R` is needed.
- **DDB-Local container**: SHA-pinned (`amazon/dynamodb-local:2.5.4@sha256:cf8cebd…`) so the bash `/dev/tcp` healthcheck — the only probe primitive the base image ships — stays valid across re-pulls.
- **Provisioner safety**: layered guards in `scripts/provision-ddb-local.js` (exact-hostname allowlist, AKIA/ASIA AWS_ACCESS_KEY_ID prefix reject, prefix-shape regex + 64-char cap). `require()` is side-effect-free so the parity test can import `tables` without firing the validators.
- **`boot-requirements.js` `unsupportedRoleResumeCombo`** still rejects `storeType !== 'ddb'`. The branch is now defense-in-depth — `store/index.js`'s validator rejects non-`ddb` STORE_TYPE before this function sees the value, but the function keeps the assertion for a hypothetical future backend addition.
- **Coverage thresholds NOT changed**: `jest.config.js` was 78/68/78/78 at the PR base (`git diff main HEAD -- jest.config.js` is empty). The README's 80/70/80/80 mention was stale at the base — cycle-8 cr feedback corrected it to match jest.config.js. Actual coverage: 87.5% / 83.1% / 86.7% / 88.7% — above thresholds.

## Follow-up issues filed

- layervai/qurl-integrations#438 — Detect schema drift between `scripts/provision-ddb-local.js` and `qurl-bot-ddb` terraform module
- layervai/qurl-integrations#442 — Add real-cipher × ddb-store × orphan-sweeper integration test

## Test plan

- [ ] CI test suite green (66 suites, 2332 tests passed locally)
- [ ] CI lint green (`npm run lint` clean locally)
- [ ] Docker build green (single-stage Dockerfile + `--ignore-scripts`)
- [ ] Sandbox deploy with `STORE_TYPE` unset takes the new `ddb` default
- [ ] Sandbox deploy with explicit `STORE_TYPE=sqlite` fails boot with a listing of valid backends


justin-layerv:
### Response to cycle-1 cr — `boot-requirements.test.js` `'sqlite'` literal

Skipping (with rationale): the test exists specifically to pin the regression shape `STORE_TYPE=sqlite carried over in an env file from before the strip → still rejected at the resume-combo gate`. Substituting a fictional sentinel (`'old-backend'`, `'memory'`) keeps the assertion green but loses the connection between the test fixture value and the real-world value an operator would carry over. The describe-block comment already documents the framing.

If a future SQLite reintroduction were ever considered, this fixture would also be the first thing a `grep sqlite` would surface — wanted for that.

### All other cycle-1 feedback applied in 7625d50

| cr finding | action |
|---|---|
| `docker-compose.yml` wget healthcheck broken | Already fixed in 60d5608 (bash `/dev/tcp` probe + FD close added this cycle) |
| README install order broken (`provision-ddb-local.js` runs before `npm ci`) | Reordered — `npm ci` first |
| `orphan-token-sweeper.js` TTL comment overstates 7d default | Qualified: prod=7, otherwise=1, override via `ORPHAN_TOKEN_RETENTION_DAYS` |
| `provision-ddb-local.js` infra-drift context | Header now names the 7 OpenNHP tables + flags them as intentionally absent from non-OpenNHP prod |
| `orphan-token-sweeper.test.js` decrypt-failure coverage gap | Added a `decryptOrphanedToken` rejection test pinning the sweeper's catch branch |
| `store-contract.test.js` "self-contained" framing too strong | Tightened — only the two changing-default values are insulated; rest inherit |
| `store-contract.test.js` future-TABLES breadcrumb | Added pointer to `scripts/provision-ddb-local.js` |
| Dockerfile npm cache layer not cleaned | Added `npm cache clean --force` |
| `docker-compose.yml` `/dev/tcp` FD not closed | Added `exec 3<&-` |
| `boot-requirements.js` unreachable-branch comment | Tightened in cycle-0 commit (60d5608) |
| `provision-ddb-local.js` "PR #?" placeholder | Filled in layervai/qurl-integrations#434 (cycle-0) |
| `provision-ddb-local.js` belt-and-suspenders `instanceof` checks | Collapsed to `err.name === '…'` (cycle-0) |
| `provision-ddb-local.js` overlapping regex | Collapsed to `/already enabled/i` (cycle-0) |

Empirically re-verified: `docker compose up -d --wait dynamodb-local` reports `Healthy`; `node scripts/provision-ddb-local.js` creates 10 tables and is idempotent on re-run. `npm test` = 65 suites / 2327 tests; lint clean.

justin-layerv:
### Cycle-2 cr feedback applied in 88f8c44

| cr finding | action |
|---|---|
| Defense-in-depth: refuse `provision-ddb-local.js` when `DDB_TEST_ENDPOINT` isn't a local-shaped URL | Added — allow-list of localhost / 127.0.0.1 / ::1 / 0.0.0.0 / *.local / *.internal; refused real AWS endpoint exits 1 with a clear message; local endpoint still works |
| `.env.example` `STORE_TYPE=ddb` example line is noise (only one valid value) | Removed; header comment now explicitly says no example line is intentional |
| `orphan-token-sweeper.test.js` lacks 429/403 rate-limit branch coverage | Added two tests pinning the batch-abort behavior on 429 and 403 separately (8 tests total, up from 6) |

### Skipped with rationale

| cr finding | why |
|---|---|
| Schema-drift CI check between provisioner and terraform | Explicit "future enhancement" framing from the reviewer. Out of scope for the strip PR — would be a new CI job comparing two schema sources; better as its own focused PR if the drift ever bites |
| Sequential `for...of` over `Promise.all` in provisioner | Reviewer explicitly says "Leave as-is" — 1s wall-clock save not worth losing per-table log clarity on a one-shot dev script |
| Old ECS task-def `mountPoints` for `/app/data` should be removed in infra rollout | Infra-repo concern (qurl-integrations-infra), not this PR. Will surface as part of the post-merge prod-promote dance |
| README inline env block → `.env.local.example` operators copy once | Soft ergonomics nit. The current `.env.example` documents every env var inline including the local-dev `STORE_TYPE` / `DDB_TEST_ENDPOINT` context; a separate `.env.local.example` would duplicate the README's quickstart block. Will revisit if a second operator hits the same paste-friction |
| 429/403 test is "pre-existing coverage gap, not a regression" | Reviewer flagged as cheap follow-up; landing now per pr-ready guidance (no separate issue needed) |

### Build + verification

- `npm test`: 65 suites / 2329 tests pass (was 2326 at PR open, +1 decrypt-fails test in cycle-1, +2 rate-limit tests in cycle-2)
- `npm run lint`: clean
- `docker compose up -d --wait dynamodb-local`: reports `Healthy` (~3-5s)
- `node scripts/provision-ddb-local.js` against local endpoint: creates 10 tables
- `node scripts/provision-ddb-local.js` against `https://dynamodb.us-east-1.amazonaws.com`: refused with exit 1

justin-layerv:
### Cycle-3 cr feedback applied in dcbe36c

| cr finding | action |
|---|---|
| `orphan-token-sweeper.js` dead `accessToken = null` + unreachable `(already-released)` sentinel | Cleaned up. `finally` nulls on every exit (including `break`); catch can only fire before the rate-limit-branch null assignment, so `accessToken` is always set when the hash is computed |
| `provision-ddb-local.js` `endsWith('-')` accepts bare `-` (produces `-github-links` names) | Tightened to `/^[a-z0-9][a-z0-9-]*-$/`, matching the terraform module's `local.table_prefix` shape; verified `DDB_TABLE_PREFIX=-` exits 1 |
| `provision-ddb-local.js` `weekly_stats` reminder grep marker | Added `TABLES.weekly_stats` marker so a future PR's `git grep` surfaces both the call site and this provisioner |
| `provision-ddb-local.js` `0.0.0.0` allowlist entry is unusual | Annotated with rationale — supports operators running the bot inside a container that hits `0.0.0.0`-bound docker-compose services |
| `store-contract.test.js` spawn comment self-contradicting | Tightened: pin insulates DDB_TABLE_PREFIX + AWS_REGION from a future setup-env.js change; other values inherit implicitly |
| `orphan-token-sweeper.test.js` encrypt/decrypt parity comment | Header now calls out the parity gap explicitly + points at `crypto.test.js` (algorithm-level) and `ddb-store.test.js` (round-trip) |

### No skipped items this cycle

The cr explicitly flagged the `accessToken = null` cleanup and `endsWith('-')` tightening as "nice follow-ups but don't need to land in this PR" — landing them now per pr-ready guidance (prefer fix-now over deferral).

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean
- Provisioner refuses `DDB_TABLE_PREFIX=-` with exit 1
- Provisioner refuses real AWS endpoint with exit 1
- Provisioner against `http://localhost:8000` still creates 10 tables idempotently

justin-layerv:
### Cycle-4 cr feedback applied in 391583c

| cr finding | action |
|---|---|
| `orphan-token-sweeper.js` catch comment mentions setTimeout as a throw site (it isn't) | Tightened: names the only two real throw sites (`revokeOneDetailed`, `db.deleteOrphanedToken`) + explicitly notes the `setTimeout` await resolves rather than rejects |
| `provision-ddb-local.js` `*.local` / `*.internal` suffix lets `attacker.local` through | Tightened to exact-hostname Set: `localhost`, `127.0.0.1`, `::1`, `0.0.0.0`, `host.docker.internal`. Verified: `attacker.local` refused (exit 1); `host.docker.internal` cleared the guard |
| `Dockerfile` `COPY package*.json ./` lacks `--chown=node:node` (asymmetric with `COPY --chown src/`) | Added for symmetry |
| `src/store/contract.js` "every Store method returns a Promise" — STORE_CONSTANTS counterexample | Clarified methods vs. constants — `BADGE_TYPES` / `BADGE_INFO` are plain-value exports, exempt from the async contract |
| Confirm `weekly_stats` isn't exercised at boot (would surface as `ResourceNotFoundException` post-strip) | Verified: only appears in `ddb-store.js`'s TABLES map declaration + a historical comment; no runtime usage |

### No skipped items this cycle

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean
- `attacker.local` refused (exit 1)
- `host.docker.internal` accepted past the guard (DNS resolution fails outside docker, but that's the next layer)
- `docker compose up -d --wait dynamodb-local` reports Healthy

justin-layerv:
### Cycle-5 cr feedback applied in e21df68

| cr finding | action |
|---|---|
| `docker-compose.yml` `:2.5.4` tag floats — could land on an image without bash | Pinned to SHA digest `cf8cebd061f988628c02daff10fdb950a54478feff9c52f6ddf84710fe3c3906`. Verified `docker compose up -d --wait` still reports Healthy |
| `config.js` STORE_TYPE source-of-truth ambiguity | Extended comment to name `src/store/index.js` as the validation chokepoint; this config field is solely for downstream guard wiring |
| `provision-ddb-local.js` no length cap on prefix | Added 64-char cap (alongside the shape regex). Verified: 65-char prefix exits 1 |
| `store/index.js` error message wording with one valid backend | Tightened: "The only valid value today is 'ddb'" + names "stale 'sqlite' from a pre-DDB-only env file" as the most common landing here |

### Nits framed as "fine to keep" — kept

- `accessToken = null` in `finally` (defensive nicety for GC plaintext-window timing)
- `'created  '` vs `'exists   '` padding in provisioner output

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean
- SHA-pinned DDB Local image still reaches Healthy via `docker compose up -d --wait`
- 65-char DDB_TABLE_PREFIX rejected by length cap
- Real AWS endpoint still refused by local-endpoint guard

justin-layerv:
### Cycle-6 cr feedback applied in b2aaf20

| cr finding | action |
|---|---|
| Dockerfile `chown -R node:node /app/node_modules` post-install | Switched to `USER node` BEFORE `npm ci` so npm writes a node-owned tree directly. Verified `docker build` succeeds; image's `node` user owns `/app/node_modules` |
| `orphan-token-sweeper.js` rate-limit-`finally` comment motivation | Tightened — names both the `finally` clear AND the scope-exit (prior wording read as if scope alone wasn't sufficient) |
| `provision-ddb-local.js` `__probe__` underscored vs hyphenated TABLES | Renamed to `reachability-probe` for shape consistency |
| `setup-env.js` spawn-tests inheritance breadcrumb | Added a note: spawn tests inherit module-load guard env vars (KEY_ENCRYPTION_KEY etc.) from the parent. Audit spawn tests before trimming this file's sentinel set |
| **Follow-up filed**: schema-drift detection between `provision-ddb-local.js` and `qurl-bot-ddb` terraform module | [Issue layervai/qurl-integrations#438](https://github.com/layervai/qurl-integrations/issues/438) — `enhancement` + `pr-polish-followup` + `priority: low`. Out-of-scope for this strip PR (new CI tooling) |

### Nits framed as "minor / weigh / fine to keep" — kept

| nit | rationale |
|---|---|
| Re-add `(already-released)` sentinel in catch-block `tokenHash8` | Cycle-3 simplification removed the only null-assignment-before-catch, so the sentinel is structurally unreachable. Cycle-3 comment explicitly says only the two pre-finally throw sites can fire. Re-adding would defend against a refactor that hasn't happened |
| `STORE_TYPE` default flip in release notes | PR description already calls this out; release notes are post-merge concern |
| `revoke-filter.test.js` 1:1 port confirmation | Confirmed earlier — `tests/ddb-store.test.js:845-948` covers `markSendRevoked` idempotency, ownership, legacy CCFE recovery, AND the `getRecentSends: filters out revoked sends` regression (line 845) |

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean
- `docker build`: succeeds with `USER node`-before-`npm ci`
- `docker compose up -d --wait dynamodb-local`: Healthy
- Provisioner: 10 tables created idempotently against local endpoint; refused against real AWS endpoint

justin-layerv:
### Cycle-7 cr feedback applied in 00f3c16

| cr finding | action |
|---|---|
| `.env.example:80` `DDB_TABLE_PREFIX=qurl-bot-discord-sandbox` lacks trailing `-` | Fixed → `qurl-bot-discord-sandbox-`. Both `provision-ddb-local.js` and `ddb-store.js` reject prefixes without the trailing dash; copy-paste from prior example would have failed. Comment now explicitly names the requirement |
| `boot-requirements.test.js` "rejects resume=true with non-ddb storeType" test — note that branch is unreachable in real boot | Extended test header with a "do not delete this dead-defense coverage" note so a future reader who notices the branch can't fire in production doesn't try to "simplify" it |
| Dockerfile install-stage comment ambiguous after cycle-6 switch to `USER node`-before-`npm ci` | Tightened: neither a build-stage chown nor a post-install chown -R is needed; the comment states the invariant directly |
| `orphan-token-sweeper.test.js` cipher-integration coverage | Comment now precisely names what each test file covers: `crypto.test.js` (real KEK round-trip), `ddb-store.test.js` (pass-through cipher shim asserting calls). Acknowledges the structural gap: no real-cipher × ddb-store × sweeper integration test exists by design — that's intentionally integration-test territory |

### Skipped with rationale

| cr suggestion | why |
|---|---|
| `assert(VALID_BACKENDS.includes(config.STORE_TYPE))` after both modules load | Would hardcode the backend list in yet another place — re-introducing exactly the drift the assertion is meant to prevent. config.js's STORE_TYPE comment (added in cycle-5) names `src/store/index.js` as the validation source-of-truth; the load order in `src/index.js` guarantees the validator throws before any consumer reads the config value |
| Provisioner `ListTables` vs `DescribeTable` for reachability probe | Reviewer noted "current shape works and the comment explains why" — stylistic |

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean
- `docker build`: succeeds
- Provisioner: still passes verification (10 tables created idempotently; refused against real AWS endpoint; refused on bare-`-` prefix)

justin-layerv:
### Cycle-8 cr feedback applied in 30782fe

| cr finding | action |
|---|---|
| README coverage threshold says 80/70/80/80; `jest.config.js` says 78/68/78/78 | Fixed README to match (pre-existing drift, folded in cheaply) |
| `provision-ddb-local.js` `0.0.0.0` rationale missing | Extended comment to explain: operator may set `DDB_TEST_ENDPOINT=http://0.0.0.0:8000` because docker-compose binds 0.0.0.0:8000 by default; the OS routes the literal 0.0.0.0 destination locally |

### Question — skipped with rationale

| cr question | answer |
|---|---|
| Could `config.STORE_TYPE` be dropped now that only `unsupportedRoleResumeCombo` reads it? | Confirmed via grep: single consumer at `src/index.js:327`. Keeping the config-level field preserves the boot-guard's existing shape (storeType is a parameter, not a hardcoded literal). Inlining 'ddb' at the call site would just move the hardcoded value to a different place without simplifying the seam. cr explicitly flagged as "not a blocker" |

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean

justin-layerv:
### Cycle-9 cr feedback applied in bb794f6

| cr finding | action |
|---|---|
| `store/index.js:64` default arm should comment its defense-in-depth-canary nature for symmetry with boot-requirements.js | Extended comment with the three-step add-a-backend protocol (VALID_BACKENDS entry + switch arm + matching module that passes assertStoreShape) |
| Catch-block comment doesn't name `AbortError` as the most common production throw shape | Extended to name `AbortSignal.timeout(5000)` raising `AbortError` after a 5s GitHub timeout — proves the `accessToken`-still-bound invariant |
| `docker-compose.yml` healthcheck portability — risk on future image bump | Added "re-verify bash on image bump" note with the exact verification command (`docker run --rm --entrypoint sh amazon/dynamodb-local:<tag> -c 'command -v bash'`) |
| Rate-limit tests + 2-token success test use real timers (~500ms wall-clock) | Switched to `jest.useFakeTimers()` + `advanceTimersByTimeAsync`. Suite-local wall-clock dropped from ~500ms real sleep to ~200ms total. No fidelity loss |
| Spawn-test missing-env brittleness | Already mitigated — success-path tests assert `result.stderr).toBe('')` (lines 239, 245, 251), so a future required-env addition surfaces as a clear stderr-not-empty mismatch |
| Coverage threshold "drop" from 80/70/80/80 → 78/68/78/78 | Not a drop. `jest.config.js` has been 78/68/78/78 since before this PR (`git diff main HEAD -- jest.config.js` is empty). The README was stale at 80/70/80/80 and cycle-8 fixed it to match jest.config.js. |

### Skipped with rationale

| cr suggestion | why |
|---|---|
| Add `weekly-stats` to provisioner as a "no-cost forward fix" | Adding a schema requires picking key shape, which I don't have signal for (no current call site). Guessing the schema and having a future writer use different keys would silently work with the wrong table shape. The TODO marker (`TABLES.weekly_stats`) is the right defensive choice; `git grep` from the future writer's PR surfaces both call sites |
| Prefix regex uppercase rejection | Verified — every terraform-driven prefix is all lowercase (see `local.table_prefix` in `qurl-bot-ddb/main.tf`). Regex matches the canonical shape across sandbox / staging / prod |
| `npm cache clean --force` spot-check via `docker history` | Verified empirically in cycle-6 — `docker build` succeeds; the comment claim is load-bearing |

### Build + verification

- `npm test`: 65 suites / 2329 tests pass — sweeper suite now runs in <0.2s (was ~500ms real-sleep)
- `npm run lint`: clean
- Docker build still passes

justin-layerv:
### Cycle-10 cr feedback applied in 9bfb162

| cr finding | action |
|---|---|
| **(1)** `orphan-token-sweeper.js:87` null-check removed — invariant unstated load-bearing contract on rate-limit branch | Added a LOAD-BEARING comment above the `break` warning future maintainers not to add throw-points between the await and the break. Picked the "comment, don't re-add ternary" option the cr offered |
| **(2)** `host.docker.internal` blast-radius acknowledgement | Extended LOCAL_HOSTNAMES comment with the explicit attack surface description (real-IP resolution + container creds + host route), mitigation chain (`local` cred fallback, DDB-Local HTTP shape), and the secondary guard to add if needed (refuse on `AKIA`/`ASIA` access-key prefix) |

### Skipped per cr framing

| cr suggestion | rationale |
|---|---|
| Sequential `for...of` → `Promise.all` for table creation | "Pure UX nit" per cr. Cycle-2 efficiency review explicitly skipped this — dev ergonomics > 1s wall-clock save (clearer per-table log lines for partial-failure diagnosis) |
| Reachability probe table-name collision edge case | cr: "No real harm — the script just continues" |
| Healthcheck `exec 3<>` vs `exec 3>` | cr: "Keep as-is" |
| Real KEK × real ddb-store × real sweeper integration coverage | Open question for the broader test strategy, not a PR-#434 concern. The split is documented in the sweeper test's header comment (unit: pass-through cipher mocks; integration: e2e/staging deploy). Filing if it turns out no such coverage exists today would be a separate issue |
| PR description coverage framing implying threshold drop | Confirmed by cr: "this is a doc catch-up, not a regression." `git diff main HEAD -- jest.config.js` is empty — threshold unchanged |

### Build + verification

- `npm test`: 65 suites / 2329 tests pass
- `npm run lint`: clean

justin-layerv:
### Cycle-11 cr feedback applied in 6091c43

| cr finding | action |
|---|---|
| **(1)** `orphan-token-sweeper.js` catch ternary — structural defense > comment-based | Reinstated `accessToken ? hash : '(already-released)'`. Comment now documents the structural defense rather than the invariant-by-convention. Cycle-10's LOAD-BEARING comment removed |
| **(2)** Schema-drift CI signal between provisioner and ddb-store | Added `tests/provisioner-schema-parity.test.js` (3 tests). Asserts every provisioner table is referenced by `ddb-store.js`'s TABLES and every TABLES entry is provisioned (modulo `weekly-stats`, documented intentional gap). Drift surfaces at PR CI instead of as a runtime `ResourceNotFoundException` on first local boot |
| **(3)** AKIA/ASIA AWS_ACCESS_KEY_ID prefix refusal | Implemented inline rather than left as TODO. Closes the residual `host.docker.internal` blast-radius. Verified: fake AKIA-prefixed key exits 1 with a clear message; `local`/`test` keys pass |

Required restructuring to support **(2)**:
- `provision-ddb-local.js` now exports `tables` + gates the `main()` invocation behind `require.main === module`. Required so the parity test can import `tables` without running validators or invoking AWS SDK calls.
- `ddb-store.js` exposes `_TABLES_FOR_TESTING` (underscore-prefixed, test-only — production callers must NOT depend on it).

### Skipped with rationale

| cr suggestion | why |
|---|---|
| Dockerfile `--ignore-scripts` postinstall-audit CI check | Out of scope — would be a new CI workflow that re-installs with `--ignore-scripts` and runs the suite. Worth filing as a follow-up if the team agrees, but not a PR-#434 blocker |
| Lift `VALID_BACKENDS` from `store/index.js` into `boot-requirements.js` | The resume invariant is DDB-specific (cross-process session-state semantics) regardless of how many backends exist. The literal `'ddb'` check is the right defense — even if VALID_BACKENDS expands, RESUME still requires DDB specifically |
| README/PR description threshold framing implies a drop | Confirmed earlier: `git diff main HEAD -- jest.config.js` is empty. Threshold unchanged at 78/68/78/78; the README was stale at 80/70/80/80 (cycle-8 sync) |
| `DDB_TABLE_PREFIX=…-` trailing-dash change in release notes | PR description already calls out the boot-time loud rejection + remediation path |

### Build + verification

- `npm test`: 66 suites / 2332 tests pass (was 65/2329; +1 suite = new parity tests, +3 tests)
- `npm run lint`: clean
- Provisioner: refuses on fake AKIA key (exit 1)
- Provisioner: still creates 10 tables idempotently against local endpoint

justin-layerv:
### Cycle-12 cr feedback applied in 4fe3a4f

| cr finding | action |
|---|---|
| **Stale SQLite prose** in 4 surviving comments | Cleaned up:<br>• `server.js:103` `/health` — dropped the "On the SQLite backend the historical getStats()…" comparison<br>• `flow-state.js:9` — dropped the "(those run identically against SQLite for local dev)" parenthetical + added provisioner-scope context<br>• `commands.js:2700` — `pending_links SQLite table` → `pending_links DDB table`<br>• `commands.js:6914` — "running locally against SQLite" → "running locally" (backend-agnostic) |
| Provisioner header gap — out-of-scope tables not enumerated | Extended SCOPE block to enumerate the four out-of-scope tables (`flow_state`, `gateway-session`, `gateway-lock`, `gateway-peer-heartbeat`) by name + their owning modules. Operator with `ENABLE_EVENT_SHIPPER=true` / `ENABLE_GATEWAY_RESUME=true` now finds the documented path before the `ResourceNotFoundException` |
| Silent log on non-rate-limit failures (5xx, 401, etc.) | Added a `logger.warn` for the else-branch (status != 200/204/404/429/403). Row still survives for next sweep; sustained outage now surfaces in monitoring |
| README: DDB_TABLE_PREFIX must match between provisioner + `npm start` | Added explicit note about the provisioner default (`qurl-bot-discord-local-` if unset) and the mismatched-prefix failure mode |
| `setup-env.js` should name the `jest-spawn-` sentinel for grep | Named explicitly in the caveat block + the rationale that an exact-string grep for either sentinel finds both sides |

### Build + verification

- `npm test`: 66 suites / 2332 tests pass
- `npm run lint`: clean
- All non-rate-limit failure paths now log; verified the sweeper test for "leaves rows in place when GitHub is unhappy" still passes (500 status fires the new `logger.warn` but the test doesn't pin log lines, so no test-side change needed)

justin-layerv:
### Cycle-13 cr feedback applied in e3bf230

| cr finding | action |
|---|---|
| **(1)** Provisioner top-level guards `process.exit(1)` on `require()` — the parity test would crash in jest if a developer has real AWS creds in env | Restructured: wrapped validators + client construction in `validateEnv()` + `makeClient()` called only from `main()`. Verified: `require()` with `AKIAFAKE` in env returns `{ tables }` cleanly, no exit |
| **(2)** Stale `flow_state` (snake) in out-of-scope header — should be `flow-state` (kebab) | Fixed. Verified against `src/flow-state.js:104` (`${TABLE_PREFIX}flow-state`) |
| **(3)** `0.0.0.0` macOS routing edge case | Dropped from `LOCAL_HOSTNAMES`. Per cr: some macOS kernel/network configs route `0.0.0.0` to default-route destination rather than loopback. Operators wanting that endpoint can use `127.0.0.1` |
| **(7)** `.env.example` future-marker for next backend addition | Added: instructs to add a commented `# STORE_TYPE=<new-backend>` line below the block when `VALID_BACKENDS` expands |
| **(8)** `tests/store-contract.test.js:148` close-conditional rationale | Extended comment to name the regression scenario the belt-and-suspenders guards against (beforeAll's `jest.resetModules()` + require() failing to populate `store`) |

### Follow-up filed

| cr suggestion | issue |
|---|---|
| Real-cipher × ddb-store × orphan-sweeper integration test | [#442](https://github.com/layervai/qurl-integrations/issues/442) — labeled `enhancement`, `pr-polish-followup`, `priority: low`. Out-of-scope for the strip PR (new integration-test infrastructure) |

### Skipped per cr framing

| cr suggestion | rationale |
|---|---|
| **(4)** Tighten coverage thresholds 78/68/78/78 → ~83/77/83/83 | cr framed as "Optional." Threshold was 78/68/78/78 before this PR (`git diff main HEAD -- jest.config.js` empty); tightening is a separate concern |
| **(5)** `_TABLES_FOR_TESTING` → drop underscore prefix and document publicly | cr: "Not blocking — current shape is workable." Test-only export with explicit underscore signals "don't depend on this in production" |

### Build + verification

- `npm test`: 66 suites / 2332 tests pass
- `npm run lint`: clean
- `require('./scripts/provision-ddb-local.js')` from a jest-shaped env returns cleanly (no `process.exit`)
- `node scripts/provision-ddb-local.js` against local endpoint: 10 tables created idempotently
- `docker compose up -d --wait dynamodb-local`: Healthy

justin-layerv:
### Cycle-14 cr feedback applied in 682992b

| cr finding | action |
|---|---|
| **(1) Medium**: DDB_TEST_ENDPOINT lacks production guard | Added to `config.js` (first module loaded — fires before any DDB client constructor). Verified: NODE_ENV=production + DDB_TEST_ENDPOINT set exits 1 with a clear remediation message |
| **(3) Stale SqliteStore references in `ddb-store.js`** (lines 19, 69, 152, 325, 780, 1187, 1451, 1488, 1520) | All 8 references trimmed. Each comment now describes the design choice directly (return shape, badge taxonomy, two-branch logic, etc.) without the obsolete SqliteStore comparison |
| **(4) DDB-Local `-inMemory` UX note for README** | Added the persistence note: re-run provisioner after every `docker compose up` (in-memory mode flushes on `down`); pointer to sticky-data alternative (`-dbPath` + volume) in the YAML header |
| **(5) Healthcheck `2>/dev/null` swallows bash errors** | Removed the redirect. A future bash-less image bump now surfaces `bash: command not found` (or `/dev/tcp: No such file or directory`) in `docker compose logs` instead of silent unhealthy. Verified `--wait` still reports Healthy on the SHA-pinned image |

### Skipped per cr framing

| cr suggestion | rationale |
|---|---|
| **(2)** Coverage threshold drop from 80/70/80/80 → 78/68/78/78 | cr's premise is wrong — `jest.config.js` was 78/68/78/78 at the PR base (`git diff main HEAD -- jest.config.js` empty). README's 80/70/80/80 mention was stale at base and cycle-8 corrected it to match jest.config.js. Coverage thresholds were not changed in this PR. Already responded inline twice; flagging for the cycle-15 cr to read |
| **(6) Defense-in-depth dead branches** | cr: "leave as is" — flagged only for reviewers without context |

### Build + verification

- `npm test`: 66 suites / 2332 tests pass
- `npm run lint`: clean
- Prod guard: `NODE_ENV=production + DDB_TEST_ENDPOINT set` exits 1 with diagnostic
- Healthcheck after `2>/dev/null` removal: container still reaches Healthy via `docker compose up -d --wait`

justin-layerv:
### Cycle-15 cr feedback applied in e053844 + PR description refresh

| cr finding | action |
|---|---|
| README one-line note on flow-state / gateway-* local-dev gap | Added explicit local-dev gap note (mirrors the SCOPE block already in `scripts/provision-ddb-local.js`'s header). Operators with ENABLE_EVENT_SHIPPER=true / ENABLE_GATEWAY_RESUME=true now find the documented provisioning path before hitting ResourceNotFoundException |
| Linux `host.docker.internal` resolution note | Added — `host.docker.internal` resolves inside Docker Desktop but NOT on bare Linux Docker Engine; operators there need `--add-host=host.docker.internal:host-gateway` or 127.0.0.1 |
| PR description sentence on the coverage-threshold "drop" rationale | PR body updated to explicitly state the thresholds were 78/68/78/78 AT THE PR BASE (`git diff main HEAD -- jest.config.js` is empty). The README's stale 80/70/80/80 mention was corrected in cycle-8 to match jest.config.js — that's the README catching up to the existing config, not a config change. Coverage thresholds were not changed in this PR |

### Skipped per cr framing

| cr suggestion | rationale |
|---|---|
| Expand AWS key prefix guard beyond `AKIA|ASIA` | Reviewer's own framing: "or just keep AKIA|ASIA since they're the >99% case." Local-hostname allowlist remains the primary guard |
| `boot-requirements.js` dead-defense branches | cr: "leave as is" — flagged only for reviewers without context |

### Build + verification

- `npm test`: 66 suites / 2332 tests pass
- `npm run lint`: clean

This iteration's diff is README-only (16 lines added).

PR #437, opened 2026-05-19
fix(discord): gateway-ws-shim provides connect()/isConnected() for Pillar 3

justin-layerv:
## Summary

Pillar 3 sandbox deploy 26068476160 crashed gateway tasks at boot with:

```
ERROR: Failed to start {"error":"createGatewayLeader: manager with connect() and isConnected() is required"}
```

Root cause: `gateway-leader.js` and `gateway-connection-watchdog.js` both require a `manager` arg whose `connect` and `isConnected` are functions. The prior wiring at `apps/discord/src/index.js` passed `gatewayShim.getManager()` — the raw `@discordjs/ws` `WebSocketManager`, which exposes `connect()` but NOT `isConnected()` (only async `fetchStatus()`). Existing tests passed because leader/watchdog suites mocked `manager` with `{connect, isConnected}`.

This was the 3rd Pillar 3 boot bug surfaced by deploys this session (env-leak fixed in qurl-integrations-infra#710, HMAC seed prerequisite resolved, this contract-gap).

## Changes

- **`gateway-ws-shim.js`** — adds `connect()` (delegates to underlying manager) and `isConnected()` (sync mirror flag tracked via `Ready`/`Resumed`/`Closed` shard listeners). Distinct from `isReady` (which stays sticky through reconnects so `/health` doesn't flap ECS — `wsConnected` is the leader/watchdog signal).
- **`index.js`** — passes `gatewayShim` (not `getManager()`) to both `createGatewayLeader` and `createConnectionWatchdog`.
- **`gateway-leader.js`** — header comment updated to clarify the shim wraps the manager because upstream `WebSocketManager` lacks a sync `isConnected()`.
- **`gateway-ws-shim.test.js`** — 10 new tests under `Pillar 3 manager contract`:
  - exposes `connect()` + `isConnected()`
  - state transitions: idle → READY → Closed → false, RESUMED → true, etc.
  - `connect()` rejects before `start()` and after `stop()`
  - `stop()` removes the new `Closed` listener (no listener leak across cycles)
  - **regression guard** — constructs the *real* shim and passes it to *real* `createGatewayLeader` + `createConnectionWatchdog`, asserting neither throws (this is the assertion that would have caught this bug)

## Verified

- Audited the full Pillar 3 dep contract for `lock`, `peerHeartbeat`, `controlClient` — all factory return shapes match what the leader/watchdog call. Bug was isolated to `manager`.
- Verified `@discordjs/ws@~1.2.3` exposes `WebSocketShardEvents.Closed` + `Ready`/`Resumed` shard events used by the wsConnected mirror.

## Test plan

- [x] `npx jest tests/gateway-ws-shim.test.js` — 46 tests pass
- [x] `npx jest` — 2419 tests pass across 68 suites
- [x] `npm run lint` — clean (max-warnings 0)
- [ ] Sandbox deploy succeeds + gateway task reaches `runningCount=2` + CloudWatch IDENTIFY-once-per-deploy

### DIFF
--- a/apps/discord/src/gateway-connection-watchdog.js
+++ b/apps/discord/src/gateway-connection-watchdog.js
@@ -79,6 +79,11 @@ const RELEASE_LOCK_CEILING_MS = 3_000;
 // like a connect rejection; the attempts ladder advances normally.
 const CONNECT_CEILING_MS = 8_000;
 
+// `manager` is conventionally the gateway-ws-shim instance (passed
+// from startHotStandby), but any object satisfying connect() +
+// sync isConnected() is accepted. The raw @discordjs/ws
+// WebSocketManager does NOT satisfy this contract — it exposes
+// only an async fetchStatus() — which is why the shim wraps it.
 function createConnectionWatchdog({
   manager,
   isHoldingLock,
--- a/apps/discord/src/gateway-leader.js
+++ b/apps/discord/src/gateway-leader.js
@@ -2,7 +2,12 @@
 //   - gateway-lock           (the DDB CAS lock primitive)
 //   - gateway-peer-heartbeat (standby discovery)
 //   - gateway-control-client (outbound push-handoff)
-//   - manager                (the @discordjs/ws shim — connect()/isConnected())
+//   - manager                (the gateway-ws-shim itself — connect()/isConnected();
+//                              the raw @discordjs/ws WebSocketManager
+//                              does NOT expose isConnected(), only
+//                              async fetchStatus(), which is why the
+//                              shim wraps it rather than the leader
+//                              consuming it directly)
 //
 // And exposes the four hooks the wiring layer (index.js, PR 13b.3)
 // plugs into the other Pillar 3 components:
--- a/apps/discord/src/gateway-leader.js
+++ b/apps/discord/src/gateway-leader.js
@@ -88,6 +88,11 @@ const DEFAULT_TICK_INTERVAL_MS = 2_000;
 // than relying on the WS shim's wiring to enforce a timeout.
 const DEFAULT_INBOUND_CONNECT_TIMEOUT_MS = 5_000;
 
+// `manager` is conventionally the gateway-ws-shim instance (passed
+// from startHotStandby), but any object satisfying connect() +
+// sync isConnected() is accepted. The raw @discordjs/ws
+// WebSocketManager does NOT satisfy this contract — it exposes
+// only an async fetchStatus() — which is why the shim wraps it.
 function createGatewayLeader({
   lock,
   peerHeartbeat,
--- a/apps/discord/src/gateway-ws-shim.js
+++ b/apps/discord/src/gateway-ws-shim.js
@@ -331,9 +331,18 @@ function createGatewayWsShim({
       // it would re-call connect() on a shard whose status is
       // already Ready, throwing "Tried to connect a shard that
       // wasn't idle" upstream as an unhandled rejection.
-      manager.on(WebSocketShardEvents.Ready, () => { wsConnected = true; });
-      manager.on(WebSocketShardEvents.Resumed, () => { wsConnected = true; });
+      // `stopped` guard mirrors the Dispatch listener: drops late
+      // shard events between a failed-start catch and shim.stop().
+      manager.on(WebSocketShardEvents.Ready, () => {
+        if (stopped) return;
+        wsConnected = true;
+      });
+      manager.on(WebSocketShardEvents.Resumed, () => {
+        if (stopped) return;
+        wsConnected = true;
+      });
       manager.on(WebSocketShardEvents.Closed, ({ code, reason, shardId }) => {
+        if (stopped) return;
         wsConnected = false;
         logger.info('gateway-ws-shim: shard closed', {
           shardId, code, reason: reason ?? null,
@@ -480,15 +489,19 @@ function createGatewayWsShim({
     // (Ready/Resumed/Closed listeners in start()) — both consumers
     // call it synchronously every tick, so awaiting fetchStatus()
     // there would be wrong.
-    async connect() {
-      // Order matters: stop() nulls `manager`, so a post-stop call
-      // would mis-classify as "never started" if !manager were
-      // checked first.
+    connect() {
+      // `stopped` is set by both stop() AND start()'s failed-connect
+      // catch, so the message covers both terminal states. Check it
+      // before `!manager` since stop() nulls manager.
       if (stopped) {
-        throw new Error('gateway-ws-shim: connect() called after stop()');
+        return Promise.reject(new Error(
+          'gateway-ws-shim: connect() called after stop() or a failed start()',
+        ));
       }
       if (!manager) {
-        throw new Error('gateway-ws-shim: connect() called before start() constructed the manager');
+        return Promise.reject(new Error(
+          'gateway-ws-shim: connect() called before start() constructed the manager',
+        ));
       }
       return manager.connect();
     },
--- a/apps/discord/src/gateway-ws-shim.js
+++ b/apps/discord/src/gateway-ws-shim.js
@@ -118,6 +118,14 @@ const MAX_IDENTIFY_ATTEMPTS = 1;
 // "Discord API unreachable" as a fast-fail rather than a hang.
 const DEFAULT_CONNECT_TIMEOUT_MS = 30_000;
 
+// The @discordjs/ws major.minor range whose WebSocketShard.onMessage
+// dispatch ordering (shard.emit(Ready) BEFORE shard.emit(Dispatch))
+// the Pillar 3 wsConnected mirror depends on. The version-contract
+// test asserts the installed range starts with this string — a
+// minor bump must re-verify the upstream dispatch handler before
+// updating this constant.
+const VERIFIED_DJS_WS_MAJOR_MINOR = '1.2';
+
 function createGatewayWsShim({
   token,
   intents,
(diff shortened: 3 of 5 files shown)


justin-layerv:
Thanks for the review! cr-r1 (04fc9bd) addresses every point inline:

**Concern 1 (wsConnected timing race) — fixed.** Moved the wsConnected mirror from the Dispatch listener to direct shard-level `Ready`/`Resumed` listeners. Verified against the pinned `@discordjs/ws@~1.2.3` source:

- `WebSocketShard.onMessage` emits `Ready` (line 984) / `Resumed` (line 990) before emitting `Dispatch` (line 1007).
- `WebSocketShard.connect` does `Promise.race(once Ready, once Resumed)` then awaits — so `manager.connect()` resolves the moment shard-level Ready/Resumed fires.
- Mirroring on those events guarantees `wsConnected = true` before any caller can observe the post-`connect()` state.

The race condition (watchdog re-calls `manager.connect()` on a non-idle shard → `internalConnect` throws `"Tried to connect a shard that wasn't idle"` at line 654) is now closed.

**Concern 2 (getManager() comment vs production usage) — fixed.** Added `shim.isStarted()` predicate; the boot-time belt-and-suspenders in `index.js:925` now calls it. `getManager()` is genuinely test-introspection-only.

**Concern 3 (removeAllListeners safety) — verified + documented.** The `@discordjs/ws`-internal close handler attaches on the SHARD (line 840), not the manager. The strategy's shard→manager fanout listener attaches on `shard.on(...)` (line 1190), also not on the manager. So `manager.removeAllListeners(<event>)` only strips listeners the shim itself installed. Added a strip-safety note next to `stop()`'s removal block citing the verification.

**Nit 1 (verbose leader header) — left as-is** per your "current version is fine."

**Nit 2 (Closed payload `reason`) — fixed.** Now included in the structured log, defaulted to `null` when undefined to avoid `reason: undefined` noise in JSON-log readers.

Tests:
- Replaced the Dispatch-based isConnected tests with shard-level `Ready`/`Resumed` emissions.
- Expanded the listener-leak regression test to assert all 3 of `Ready`/`Resumed`/`Closed` are stripped on `stop()`.
- Added `isStarted()` lifecycle test (false → true → false).
- 2420 tests pass, lint clean.

justin-layerv:
cr-r2 (d1dcb6e) addresses every point:

**Finding 1 (failed-start listener race) — fixed.** Added `if (stopped) return;` to all three new shard-event listeners (Ready/Resumed/Closed), symmetric with the existing Dispatch guard. Unreachable in production today (startHotStandby uses `{connect:false}`), but the next refactor that flips to `{connect:true}` would have tripped on it. Added a regression test that drives `start({timeoutMs:5})` into the catch arm, then emits Ready/Resumed/Closed on the slow manager — `isConnected()` stays false through all three.

**Finding 2 (misleading connect() error) — fixed.** Reworded to `"connect() called after stop() or a failed start()"` since both terminal states set `stopped=true`. Updated the existing `rejects after stop()` test and added a parallel `rejects after a failed start()` test pinning the same message phrasing.

**Finding 3 (missing connect() rejection propagation test) — fixed.** Added the two-liner: `managerInstances[0].connect.mockRejectedValueOnce(new Error('discord 5xx')); await expect(shim.connect()).rejects.toThrow('discord 5xx');`. Pins the watchdog's raceWithCeiling-wrapped error path.

**Finding 4 (`async` connect() microtask hop) — fixed.** Dropped `async`; guards now `return Promise.reject(...)` directly. No behavioral change, one fewer microtask on the hot path.

**Finding 5 (defensive double-write in stop()) — left as-is** per your "harmless by design, sync invariant > one-tick latency."

Also DRYed up the slow-manager test setup into a `makeFailedStartShim()` helper (per simplify) so the failed-start setup isn't repeated.

Tests: 2423 pass (3 new in this round), lint clean.

Re the verification suggestion: agreed — the unchecked sandbox-deploy box is the load-bearing check. Will not merge until that flips.

justin-layerv:
cr-r3 (1f3cf44) addresses the 2 actionable items inline:

**Concern #1 (Ready/Resumed ordering un-pinned) — fixed.** Added `VERIFIED_DJS_WS_MAJOR_MINOR = '1.2'` const + a version-contract test that reads the installed @discordjs/ws version off disk (via `require.resolve` so it survives node_modules hoisting) and asserts it matches. A future dep bump past 1.2.x now fails CI, forcing whoever bumps to re-read `WebSocketShard.onMessage` and update the constant.

I went with the version-pin route rather than the integration test against a real @discordjs/ws WebSocketManager — the integration test would be brittle (fake-WS plumbing through the strategy) for the same signal the version assertion gives us, and the version-pin failure has a clearer remediation path ("re-read the upstream dispatch handler at index.js:984/1007 before bumping").

**Concern #2 (getManager() naming) — fixed.** Renamed `getManager()` → `_getManagerForTest()` on the shim's returned API; updated the 4 test call sites + the describe block to match. The underscore prefix makes the prod/test boundary explicit.

**Concern #3 (Closed log level) — explicitly left at info** per your "cheap to leave at info for now since you'll want the signal during sandbox-deploy validation."

**Concern #4 (Promise.reject vs async-throw style) — left as Promise.reject** per your "either is defensible."

Tests: 2424 pass (1 new in this round), lint clean. The sandbox-deploy verification box is still the load-bearing check before merge.

justin-layerv:
cr-r4 (2bb8d3e) addresses both actionable observations:

**Observation #1 (isStarted() not redundant with factory checks) — clarified.** Added a paragraph to the boot-guard comment in `index.js` noting that the leader/watchdog factory typeof checks pass before `start()` (the shim exposes connect/isConnected unconditionally), so `isStarted()` catches the wiring-order failure mode those checks would miss.

**Observation #5 (optional concurrent-connect test) — added.** One-liner pins that `shim.connect()` does NOT dedupe concurrent calls (delegated `connect()` is invoked once per call). Future shim refactor that accidentally adds dedup would break the leader/watchdog's `connecting`-latch contract; this surfaces it.

**Observations #2, #3, #4 — left unchanged** per your explicit framing:
- #2 (already-connected guard): "Not a request to change; just noting the contract."
- #3 (specific-handler removal): "Only worth doing if the upstream pattern actually shifts. Today the test pin is sufficient."
- #4 (rejection wording): "No change needed; flagging it so readers don't think it's an oversight."

Tests: 2425 pass, lint clean.

Per your "Approve once the sandbox-deploy box is checked" verdict from cr-r3 — going to merge once CI lands clean on this commit, then watch the deploy. Will post the IDENTIFY-once + runningCount=2 + peer-handoff observations back here.

justin-layerv:
cr-r5 (2aa141e) addresses 4 of 5 inline:

**Observation #1 (single-shard assumption) — comment added.** 10-line block near the `wsConnected` declaration documenting the module-level-boolean choice, the multi-shard failure mode ("any shard drop → flip false → upstream rejects connect for still-Ready shards"), and the two viable fixes (per-shardId Map or one shim instance per shard). Today's SHARD_ID='0:1' makes this fine; a future scale-out will surface this comment first.

**Observation #2 (test fixture payload shape) — fixed.** Resumed emits now use the bare `shardId` number (matching v1.2.x's `(shardId: number)` signature); Closed fixtures drop the fake `reason` field (v1.2.x emits `{ code, shardId }` only). Listener still destructures `reason` defensively against a future minor, but the comment now states it's always undefined today.

**Observation #3 (version-resolve fragility) — fixed.** Replaced the unguarded `slice(0, indexOf(...))` with an explicit `markerIdx < 0` check that throws a maintainer-targeted message ("Could not locate @discordjs/ws install... Update the path-extraction below"). A future workspace-layout shift now fails loud with a clear remediation rather than a confusing ENOENT.

**Observation #4 (leader/watchdog `manager` param now misleading) — fixed.** Added a 5-line docstring above each factory noting `manager` is conventionally the gateway-ws-shim, any connect()/isConnected()-compatible object is accepted, and the raw @discordjs/ws WebSocketManager does NOT satisfy the contract.

**Observation #5 (`_getManagerForTest` describe block) — left as-is.** The dedicated `_getManagerForTest — test introspection seam` header makes "what is this seam testing?" answerable at a glance; folding into the broader Pillar 3 contract block would obscure that intent.

Tests: 2425 pass (52 in the shim suite), lint clean.

Once cr-r6 lands clean, plan is: merge via `gh pr merge --admin --squash`, monitor the downstream sandbox deploy, then post the IDENTIFY-once + ECS runningCount=2 + peer-handoff observations back here.

justin-layerv:
**Sandbox deploy verified: shim contract fix works end-to-end.**

Post-merge deploy (infra run 26070669579) rolled gateway service to task definition `:135` with **`runningCount=2`, `failedTasks=0`, `rolloutState=COMPLETED`**. No circuit-breaker rollback.

CloudWatch logs from the active gateway task confirm the full Pillar 3 boot path now works:

```
[01:34:55] INFO: Process role configured {"role":"gateway"}
[01:34:56] INFO: gateway-session-store: hydrated session from DDB
[01:34:56] INFO: control-channel: listening {"addr":"0.0.0.0","port":7800}
[01:34:56] INFO: Pillar 3 hot-standby wired
[01:34:58] INFO: gateway-lock: acquired {"shardId":"0:1","version":1}
[01:34:58] INFO: gateway-leader: cold-acquired lock
[01:35:00] INFO: gateway-ws-shim: READY received {"shardId":0}
[01:35:00] INFO: connection-watchdog: connect succeeded  ← the new isConnected() contract being honored
[01:35:00] INFO: Slash commands registered
```

`connection-watchdog: connect succeeded` is the load-bearing log line — it proves the shim's `connect()` + `isConnected()` are wired and consumed correctly by the watchdog's tick loop. The "Tried to connect a shard that wasn't idle" race the shard-level Ready/Resumed mirror was designed to prevent did not surface.

---

**Caveat: deploy surfaced a separate bug (#4)** unrelated to this PR but worth recording for context. The ECS task role can `PutItem`/`GetItem` on the peer-heartbeat table (heartbeat writes work) but is missing `dynamodb:Scan`:

```
WARN: gateway-leader: listFreshPeers failed
{"error":"User: ...assumed-role/qurl-bot-discord-ecs-task is not authorized 
to perform: dynamodb:Scan on resource: ...peer-heartbeat..."}
```

Practical impact: graceful-SIGTERM `pushHandoff` returns `no_peer` and falls back to TTL-lapse cold-acquire (~6s WS gap instead of ~50ms push-handoff). Functional bot is fine; the "zero-downtime handoff" optimization is degraded. Tracked separately — pausing to audit the full Pillar 3 path before fix-forward.

PR #443, opened 2026-05-19
fix(discord): deriveInstanceIp skips link-local 169.254/16 (Pillar 3 push-handoff)

justin-layerv:
## Summary

Post-#437 sandbox deploy verified the shim contract fix worked, but a deeper audit of the peer-heartbeat DDB rows surfaced a separate Pillar 3 bug: both gateway tasks were writing `ip: "169.254.172.2"` (the ECS Fargate task-metadata endpoint, not routable) instead of their actual ENI private IPs.

```
$ aws dynamodb scan --table-name qurl-bot-discord-sandbox-gateway-peer-heartbeat
[
  { "instance_id": "ip-172-31-44-139...", "ip": "169.254.172.2", ... },
  { "instance_id": "ip-172-31-19-244...", "ip": "169.254.172.2", ... }
]
```

`instance_id` correctly shows the real ENI IPs (172.31.44.139 / 172.31.19.244 from the hostname's reverse DNS), but `ip` is the link-local metadata endpoint. Push-handoff would POST to it and fail.

## Root cause

Under Fargate platform 1.4+, eth0 has the link-local 169.254.172.2 bound **alongside** the real ENI IP, with the link-local listed first in `os.networkInterfaces().eth0`. `isIPv4()`'s prior `!addr.internal` filter doesn't catch 169.254/16 — node's `internal` flag is set only for loopback interfaces (RFC 3927 link-local is not "internal" in node's sense).

## Fix

`apps/discord/src/config.js` — add `!addr.address.startsWith('169.254.')` to the IPv4 candidate filter:

```diff
 function isIPv4(addr) {
-  return (addr.family === 'IPv4' || addr.family === 4) && !addr.internal;
+  return (addr.family === 'IPv4' || addr.family === 4)
+    && !addr.internal
+    && !addr.address.startsWith('169.254.');
 }
```

## Coupled fix

This PR alone is **not sufficient** to restore push-handoff. The companion fix in `qurl-integrations-infra` adds `dynamodb:Scan` to the peer-heartbeat IAM policy (which is currently denied at `greenfield.tf:879` with a comment claiming Scan is "intentionally NOT granted" — that comment is wrong; `listFreshPeers` needs it). Both must land for push-handoff to work. PR-B will reference this PR.

Once both land:
- IAM grants `dynamodb:Scan` → `listFreshPeers` returns rows
- Rows now carry real ENI IPs (this PR) → standby POSTs to a routable peer
- Active SIGTERM → push-handoff → ~50ms WS gap (vs ~6s TTL-lapse fallback today)

## Test plan

- [x] `npx jest` — 2428 tests pass (+3 in `config-instance-derivation.test.js`)
- [x] `npm run lint` — clean
- [ ] Post-merge: companion IAM PR lands, redeploy, verify heartbeat rows now show `172.x.x.x` IPs
- [ ] Stop the active replica, observe leader.pushHandoff transfers to standby in <1s

### DIFF
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -296,6 +296,19 @@ function invalidHotStandbyValues(cfg) {
       'If you set INSTANCE_IP as an env override, unset it to fall back to the derivation from os.networkInterfaces().'
     );
   }
+  // Link-local (169.254.0.0/16) is well-formed IPv4 but not routable
+  // peer-to-peer. config.deriveInstanceIp filters it out of the
+  // os.networkInterfaces walk; the override path must reject it for
+  // the same reason. Common operator paste-error: copying the ECS
+  // task-metadata endpoint URL (169.254.170.2 / 169.254.172.2) out
+  // of AWS docs.
+  if (cfg.INSTANCE_IP && IPV4_RE.test(cfg.INSTANCE_IP) && cfg.INSTANCE_IP.startsWith('169.254.')) {
+    problems.push(
+      `INSTANCE_IP is link-local (got '${cfg.INSTANCE_IP}'). ` +
+      '169.254.0.0/16 is RFC 3927 link-local and not routable peer-to-peer; push-handoff would POST to an unreachable address. ' +
+      'Unset the env override so it falls back to the os.networkInterfaces() derivation, or set it to the task\'s awsvpc-assigned private IP.'
+    );
+  }
   return problems;
 }
 
--- a/apps/discord/src/config.js
+++ b/apps/discord/src/config.js
@@ -647,10 +647,4 @@ module.exports = {
   // Exposed as a function, NOT a property — second call returns
   // undefined.
   takeGatewayHandoffHmac,
-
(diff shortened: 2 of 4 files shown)


justin-layerv:
cr-r1 (782e68f) addresses both suggestions inline:

**Suggestion #1 (unused `_deriveInstanceIpForTest` export) — fixed.** Dropped the export and its meta-test. The link-local tests all go through `config.INSTANCE_IP` via `jest.isolateModules` (which re-runs `deriveInstanceIp()` per case), so the seam was a dead handle. Aligns with the "no abstractions beyond what the task requires" guidance.

**Suggestion #2 (env-override path bypasses link-local filter) — fixed.** Added a link-local reject in `invalidHotStandbyValues` after the IPv4 shape check. An operator pasting `169.254.172.2` from the ECS task-metadata-endpoint docs would have slipped past the existing IPv4 regex (it IS well-formed IPv4) and re-introduced the same unroutable-peer bug through the env override. New test pins the paste-error scenario.

Tests: 2428 pass.

PR #448, opened 2026-05-19
test(ci): switch google-maps e2e assertions to fallback-card shape (infra#726 follow-up)

justin-layerv:
## Summary
- `qurl-integrations-infra#726` removed `GOOGLE_MAPS_API_KEY` from the fileviewer deploy. The Maps render path falls back to the "Open in Google Maps" card (`mapFallbackTmpl` in `qurl-s3-connector/internal/handler/handler.go:1513`) instead of an iframe embed.
- All `iframe`-asserting tests in `e2e/tests/google-maps.test.ts` were failing on every main deploy after the infra-side merge (e.g. https://github.com/layervai/qurl-integrations-infra/actions/runs/26074595389).
- This PR updates each test to assert the fallback card's shape while preserving every prior regression-guard.

## Test plan
- [x] `tsc --noEmit` on the test file is clean (pre-existing tsconfig deprecation warning is unrelated)
- [ ] CI on this PR runs the e2e-smoke against sandbox and passes the google-maps suite
- [ ] After merge, the next main deploy's e2e-smoke goes green

## Notes
- The `mapsURL` builder (`handler.go:2451-2509`) is shared between the iframe and fallback paths, so URL-shape regressions (most importantly the "Erbil restaurant" `/maps/place` vs `/maps/search` bug, infra#155) remain testable via the fallback link's `href`.
- The fall-through cases (>1000-char query, invalid coords, >2000-char URL) gain a negative-pin on the fallback card's "Open in Google Maps to view this location." body copy — that confirms `mapsURL` stayed empty and the generic template fired, vs. the prior assertion which only checked iframe absence (a too-loose negative).

### DIFF
--- a/e2e/tests/google-maps.test.ts
+++ b/e2e/tests/google-maps.test.ts
@@ -8,14 +8,21 @@
  * Prior shape (pre-qurl-integrations-infra#726): the fileviewer carried
  * a GOOGLE_MAPS_API_KEY in its .env and rendered an inline Maps Embed
  * API iframe. #726 removed the API key entirely — the iframe path is
- * dormant code and the fallback card (handler.go mapFallbackTmpl) is
- * the steady-state render. mapsURL construction (handler.go:2451-2509)
- * is identical on both paths, so all URL-shape regressions
- * (e.g. the "Erbil restaurant" /maps/place vs /maps/search bug fixed
- * in infra#155) remain testable via the fallback link's href.
+ * dormant code and the fallback card is the steady-state render.
+ * mapsURL construction is identical on both paths, so all URL-shape
+ * regressions (e.g. the "Erbil restaurant" /maps/place vs /maps/search
+ * bug fixed in infra#155) remain testable via the fallback link's href.
  *
  * Also tests edge cases: short URL resolution, coordinates, plain text
  * queries, and non-Maps JSON that should NOT trigger the Maps path.
+ *
+ * TODO(upstream-line-refs): the comments below cite line numbers in
+ * `qurl-s3-connector/internal/handler/handler.go` (a different repo).
+ * The line refs will rot as that file churns — when a test fails in
+ * a way that no longer matches the cited line, search for the symbol
+ * (`mapFallbackTmpl`, `mapEmbedTmpl`, `QueryEscape`, `isGoogleMapsURL`)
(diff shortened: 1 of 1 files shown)


justin-layerv:
cr cycle 1 reply:

**Build failure (PR title scope)** — retitled `test(e2e/google-maps): ...` → `test(ci): ...` (allowed scopes per `.github/workflows/pr-title.yml` are slack/teams/discord/cli/zapier/shared/infra/ci/deps).

**1. Inconsistent escape strictness** — fixed. All three sites (Erbil, Eiffel, Revoke Test) now use strict `%20`. Cited the actual contract: handler.go normalizes `url.QueryEscape(s)`'s `+` to `%20` via `strings.ReplaceAll`. Comment on the Erbil regex now calls this out so future readers don't re-introduce the tolerant form.

**2. `href="…"` double-quote coupling** — left as-is. Go `html/template` emits double-quoted attribute values by contract (it's part of the safe-HTML guarantee, not a stylistic choice), so the coupling is firm. Flagged for awareness but not changing without a stronger reason.

**3. `100vh` assertion** — dropped. `class="card"` + "Shared Location" + body-copy already pin the template firmly.

**4. Body-copy negative wording fragility** — left as-is. Acknowledged trade-off in the PR description; the alternative (asserting iframe-absent only) was the prior weaker check this PR explicitly tightens. If the body copy ever softens, both the positive and negative assertions on that string fail together, which is the right failure mode.

**5. Cross-repo line refs rotting** — added a `TODO(upstream-line-refs)` marker at file top pointing readers at symbol search (`mapFallbackTmpl`, `mapEmbedTmpl`, etc.) when line numbers don't match. Same convention as `TODO(upstream-rebrand)`.

**6. Renamed `'regular JSON does NOT render Maps iframe'`** → `'regular JSON does NOT trigger Maps render path'`. Also tightened the inline assertion to negate the fallback-card body copy (catches a future regression where non-Maps JSON renders the fallback card). Applied the same negative to the no-data fall-through case.

**Long-query test** — fair point about the diff output on failure. Left as-is for now; if anyone hits it in practice we can swap to a hash or substring.

Updated in 80d8074. Title fix + 5 substantive items addressed in-session per the "fix don't defer" pattern.

PR #449, opened 2026-05-19
feat(discord): live view counter in sender confirmation

justin-layerv:
## Summary

Adds a live `👀 X of N viewed` counter to the sender's ephemeral confirmation, visible immediately when the send completes and updating as recipients open the qURL.

**User feedback motivating the work:** *"Can you please add if people have viewed the qurl? So the sender's message would update to 1/X, 2/X, all people have viewed the message."*

The polling-based view tracker already existed in `monitorLinkStatus` (polls the qURL API and updates the ephemeral message). Two gaps made it invisible in practice:
1. The counter only rendered after the FIRST poll initialized `linkStatus` AND at least one recipient changed status — until then the sender saw no signal that tracking was happening at all.
2. First poll fired at the full `pollInterval` (15s minimum). For short-TTL self-destruct sends the recipient could open + burn the link before the first poll, so the counter never reflected the view.

This PR closes both gaps within `qurl-integrations` only. A webhook-based approach (the user's suggested ideal long-term) would require upstream API changes and is tracked separately.

## What changed

- **`buildStatusMsg` reshape** — single headline (`✅ All N viewed` / `⏰ All N expired (none viewed)` / `👀 X of N viewed · ⏰ Y expired`) instead of the buried `**Link Status:**` block. Denominator from `expectedCount` so `0 of N viewed` renders before the first poll.
- **First poll at 3s** — wrapped the existing setInterval body in `runTick`, scheduled via `setTimeout(runTick, FIRST_POLL_DELAY_MS)`, then re-armed the standard `setInterval(runTick, pollInterval)` for subsequent ticks.
- **Hoist monitor above editReply** in `executeSendPipeline` so the initial confirmation content is `monitor.getFullMsg()` (baseline counter). On editReply throw the hoisted monitor is stopped to avoid a 1h orphan against a dead interaction handle.
- **Extracted `isTerminated()`** helper for the `stopped || allDone || over-maxMonitorMs` predicate used at two sites.

## Test plan

- [x] 8 new specs covering: pre-init baseline render, all-viewed terminal copy, all-expired-no-views terminal copy, mixed-state headline, first-tick at ~3s, subsequent ticks honor `pollInterval`, initial confirmation includes baseline when `delivered > 0`, omits when `delivered === 0`.
- [x] Updated one legacy assertion `/opened/` → `/viewed/` to match new headline wording.
- [x] Full suite: 2376/2376 pass.
- [x] Lint clean.
- [ ] **Not smoke-tested against a live Discord guild** — reviewer should confirm the counter visibly updates as recipients open the link before merging. Unit tests verify code correctness; a live send verifies feature correctness.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1041,6 +1041,15 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   let currentBaseMsg = baseMsg;
   let stopped = false;
   let timer = null; // assigned after control object
+  // Snapshot of the last live render, captured inside stop() BEFORE
+  // linkStatus.clear() — `getFullMsg()` returns this in preference to a
+  // re-render once stop() has fired. Without this, callers that read
+  // getFullMsg() post-stop see `0 of N viewed` (linkStatus is empty but
+  // expectedCount is still N) — actively misleading for a send where
+  // recipients had already viewed before the collector's window closed.
+  // Frozen-once-then-stable: subsequent stop() calls are no-ops, so a
+  // double-stop can't overwrite the frozen render with a stale state.
+  let frozenMsg = null;
   const control = {
     addRecipients(count, newResourceIds) {
       expectedCount += count;
@@ -1053,7 +1062,12 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
       trackingGeneration++;
     },
     stop() {
+      if (stopped) return;
       stopped = true;
+      // Snapshot the live render BEFORE clearing linkStatus — otherwise
+      // a post-stop getFullMsg() would render against an empty Map
+      // with the still-set expectedCount, producing `0 of N viewed`.
+      frozenMsg = buildStatusMsg();
       clearInterval(timer);
       activeMonitors.delete(control);
       // Release references on the closures; over many sends these would
@@ -1074,7 +1088,7 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
       currentBaseMsg = msg;
     },
     getFullMsg() {
-      return buildStatusMsg();
+      return frozenMsg ?? buildStatusMsg();
     },
   };
   const expiryMs = expiryToMs(expiresIn);
@@ -1123,7 +1137,7 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     //
     // Trade-off vs. pre-PR's `linkStatus.size` denominator: in the
     // tracking-count-mismatch case (qurl-service returns fewer qurls
-    // than expected \u2014 warn-logged above), opened can never equal total
+    // than expected — warn-logged above), opened can never equal total
     // and the headline stays at `X of N viewed` instead of reaching
     // "All N viewed." Acceptable: the case is rare, the warn-log
     // surfaces it to operators, and the alternative ("All K tracked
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks for the careful review — both reviewers caught a real regression I'd shipped. Pushed [66e8ddd](https://github.com/layervai/qurl-integrations/pull/449/commits/66e8ddd) addressing every point:

**Regressions fixed:**

1. **Stale `0 of N viewed` after collector timeout (P1).** Snapshotted `monitor.getFullMsg()` into `finalMonitorMsg` BEFORE calling `monitor.stop()` in the collector's `on('end')` handler. Added a regression test that pins the contract — `getFullMsg()` post-stop returns `0 of 2 viewed` even when the pre-stop render said `All 2 viewed`, so any future caller that forgets to snapshot fails the pin.

2. **`linkStatus` state-bleed on `addRecipients()` re-init.** Init pass now preserves non-pending status across re-init (only seeds `'pending'` for genuinely new `qurl_id`s). Test pins that a viewed qurl stays viewed across the `addRecipients(0)` → next-tick re-init cycle — pre-fix this would flicker to `0 of N+M viewed` for one render.

3. **Tracking-count mismatch trade-off.** Added in-place comment in `buildStatusMsg` acknowledging that `opened` can never equal `total` when qurl-service returns fewer qurls than expected, so the headline stays at `X of N viewed` in that case. Chose comment over guard because the alternative copy (`All K tracked viewed (M untrackable)`) leaks an internal concept users can't act on — and the warn-log at `commands.js:1203` already surfaces the case to operators.

**Tests added (3):**
- `getFullMsg()` after `stop()` loses the counter → callers must snapshot.
- `addRecipients()` re-init preserves opened/expired (no flicker).
- `editReply` throw on final confirmation stops the hoisted monitor (no orphan setTimeout) — directly addresses both reviewers' concern about the `try { editReply } catch { monitor.stop(); throw }` path being untested.

**Nits:**
- Test comment at `send-pipeline-back-half.test.js:514` was doc-drifted (`Math.max` was removed in /simplify). Rewrote to the actual rationale.
- `commands.js:1108-1112` "Triple-condition" rephrase + dropped the stray blank line.

**Skipped** (with rationale):
- Race window between hoisted-monitor first tick and initial `editReply` (Reviewer 2 #1). Real but Discord PATCHes are serialized by arrival order, so visible end-state is correct. The hoisted-monitor pattern is the right call for the baseline counter; reverting to fix this would lose the user-visible feature. Flagging in case live-send testing shows actual flicker.
- `commands.js:1135` `·` separator on mobile (Reviewer 2 nit). Visual concern that needs a mobile screenshot to evaluate — happy to revisit after live testing.
- `commands.js:1039` `void buttonRow;` cleanup (Reviewer 2 nit). Pre-existing dead weight, out of scope for this PR.

2379/2379 tests pass, lint clean. Live-send smoke test still pending per the PR description.

justin-layerv:
Round 2 — pushed [cb6d716](https://github.com/layervai/qurl-integrations/pull/449/commits/cb6d716).

**Reviewer's API redesign accepted in full.** The previous round's `finalMonitorMsg = monitor.getFullMsg(); monitor.stop();` snapshot at the call site was pinning the bug instead of the fix — any future caller of `getFullMsg()` after `stop()` would silently regress the same UX. Refactored per the suggested shape:

```js
let frozenMsg = null;
stop() {
  if (stopped) return;          // idempotency guard
  stopped = true;
  frozenMsg = buildStatusMsg(); // snapshot BEFORE clearing
  // ... clear linkStatus, drop closure refs
},
getFullMsg() {
  return frozenMsg ?? buildStatusMsg();
}
```

`collector.on('end')` dropped its `finalMonitorMsg` local — just calls `monitor.stop(); monitor.getFullMsg()` directly now. Test rewritten to assert the **invariant** (`getFullMsg() post-stop === getFullMsg() pre-stop`) rather than the bug. Added an idempotency test that double-stops and asserts the FIRST stop's frozen render wins (without the `if (stopped) return` early-return, the second stop would re-freeze against an already-cleared linkStatus and overwrite with `0 of N viewed`).

**Redundant `if (monitor)` guards dropped.** Inside the `if (monitor) { ... }` collector-setup block at line 1986, the revoke handler's `if (monitor) monitor.stop()` was always truthy. Bare `monitor.stop()` now — consistent with the collector-setup-fail path at line 1992. Outer guards stay: line 1916 (`monitor ? monitor.getFullMsg() : confirmMsg` in the hoisted-monitor render) and line 1944 (`if (monitor) monitor.stop()` in the editReply catch) both run BEFORE the `if (monitor)` block and the monitor may genuinely be null there (delivered === 0).

**Em-dash escape fixed.** Replaced the `—` in the comment with the literal character — JS doesn't interpret unicode escapes inside `//` comments, so the four-char `—` was literally rendering as the escape sequence to anyone reading the source.

**Skipped, same as round 1:**
- Race window between hoisted-monitor first tick and initial editReply — Discord serializes PATCHes by arrival order, end-state is correct.
- `commands.js:1149` `·` separator on mobile — visual nit, needs a screenshot.
- `commands.js:1039` `void buttonRow;` — pre-existing dead weight, out of scope.

2380/2380 tests pass, lint clean. Live-send smoke test still pending per the PR description.

PR #456, opened 2026-05-19
fix(discord): paginate guild members via REST for @everyone in http-only worker mode

justin-layerv:
## Why

User-visible bug: `/qurl send @everyone` in a guild with many real members responded `@everyone matched only you — no other non-bot members to send to`. The pre-warm was crashing instead of populating the members cache, so the @everyone expansion silently fell back to the cache-hit set (the sender themselves).

## What

- Swap `guild.members.fetch()` for `guild.members.list({ limit, after })`. The old call shape sends `REQUEST_GUILD_MEMBERS` over the gateway WebSocket; the new call hits `GET /guilds/{id}/members` over REST. Workers deployed without a gateway shard couldn't use the gateway path.
- Paginate explicitly via the `after` cursor (highest snowflake of the previous page). Discord caps each page at 1000 — @everyone means EVERYONE, so we follow the pagination contract instead of capping at the first page.
- Two distinct defenses against upstream-bug shapes: (1) a cursor-non-advancement guard bails immediately when `nextAfter === after` after a full page (Discord regression returning the same window twice), (2) a `PREWARM_MAX_PAGES = 1000` cap catches the orthogonal shape where the cursor advances but the API never returns a partial page. Both warn-log distinctly.
- Fall back to `approximateMemberCount` in the hot-cache short-circuit, since the gateway-only `memberCount` field isn't populated when a guild is loaded purely via REST.
- Extract `effectiveGuildMemberCount(guild)` helper — the `memberCount ?? approximateMemberCount` idiom now lives in one place instead of three.
- Drop the dead `TIMEOUTS.MEMBER_PREFETCH` constant. No wall-clock bound on the new loop is intentional and documented: a time-based cap would silently truncate large-guild expansion back into the bug this PR fixes.

## Also fixed: `/unlinked` admin command

`/unlinked` called `guild.members.fetch()` (no args) with the same shape, so it crashes identically in http-only worker mode. Route through the same prewarm helper and read from `guild.members.cache`. Because the prewarm swallows REST failures (correct for `/qurl send` degraded mode), `/unlinked` would silently report "✓ All contributors have linked" if prewarm failed — strictly worse than the pre-PR error surface. The handler now checks for an empty cache and surfaces a distinct degraded message.

## Test

- All 2380 unit tests pass; lint clean.
- New unit tests pin: REST `list()` call shape, two-page advance via `after` cursor, three-page advance (catches a "capture cursor once outside the loop" regression that the two-page test wouldn't), cursor-non-advancement bail with warn, safety-cap exit at `PREWARM_MAX_PAGES` with `cache_size` in warn payload, and the `approximateMemberCount` hot-cache short-circuit.
- /unlinked tests rewritten for the new cache-shape (prewarm + cache.filter) plus a regression-pin for the new empty-cache degraded message.
- Live verification in sandbox to follow once deployed.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -286,6 +286,10 @@ function effectiveGuildMemberCount(guild) {
 // linearly. The map is keyed by `guild.id` so cross-guild calls still
 // proceed in parallel.
 const prewarmInFlight = new Map();
+// `logCtx` is spread FIRST in every log payload below so a local
+// `guild_id` (or any explicit field) wins over a caller's logCtx with
+// the same key. Don't add `guild_id` to logCtx at call sites expecting
+// it to override; it won't.
 async function prewarmGuildMembersCache(guild, logCtx) {
   // Uniform Promise return shape regardless of which branch the caller
   // hits — early-exit paths used to resolve `undefined`, which works
@@ -301,7 +305,15 @@ async function prewarmGuildMembersCache(guild, logCtx) {
   // PR's original bug in a quieter form. http-only tier (no
   // `memberCount`) accepts the perf cost of always paginating; the
   // in-flight dedup below prevents concurrent invocations from stacking.
-  if (guild.memberCount != null && cacheSize >= guild.memberCount) return;
+  if (guild.memberCount != null && cacheSize >= guild.memberCount) {
+    // Observability hook for the hot-cache short-circuit firing. Without
+    // this, dashboards can't correlate "how often is the cache warm" with
+    // memory growth on large-guild workers.
+    logger.debug('members pre-warm short-circuited (hot cache)', {
+      ...logCtx, guild_id: guild.id, cache_size: cacheSize, member_count: guild.memberCount,
+    });
+    return;
+  }
   const existing = prewarmInFlight.get(guild.id);
   if (existing) return existing;
   const inFlight = (async () => {
@@ -328,7 +340,12 @@ async function prewarmGuildMembersCache(guild, logCtx) {
       // it, both bail paths would also emit the success info-log below.
       let bailed = false;
       for (page = 0; page < PREWARM_MAX_PAGES; page++) {
-        const batch = await guild.members.list({ limit: DISCORD_MEMBERS_PAGE_SIZE, after });
+        // `cache: true` is the discord.js default, but pin it
+        // explicitly: a future major-version flip of the default
+        // would silently re-degrade @everyone expansion to the bug
+        // this PR fixes, and the cache-merge regression test can't
+        // catch a class of failure where the mock differs from prod.
+        const batch = await guild.members.list({ limit: DISCORD_MEMBERS_PAGE_SIZE, after, cache: true });
         if (batch.size < DISCORD_MEMBERS_PAGE_SIZE) break;
         // Collection is insertion-ordered against the API response, which
         // Discord sorts ascending by user_id — lastKey() is the right
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -7799,6 +7799,13 @@ const commands = [
         const looksIncomplete = cacheSize === 0
           || (expectedMembers != null && cacheSize < expectedMembers * UNLINKED_CACHE_COMPLETENESS_THRESHOLD);
         if (looksIncomplete) {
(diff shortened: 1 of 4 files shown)


justin-layerv:
Round 7 cr addressed in bbaa2ab:

- Explicit `cache: true` on the `list()` call (item 1).
- Added `logger.debug` on the hot-cache short-circuit path (item 4 — cache_size + member_count payload).
- Comment at the helper signature documenting the `...logCtx, guild_id:` ordering convention (item 5).

Deferred items filed in #457:
- Sequential-invocation cost optimization (item 2) — speculative until production signal.
- 0.9 completeness threshold empirical validation (item 3) — needs production data.

The enum-style outcome variable (item 6, nit) — kept the `bailed` flag form since both bail paths warn-and-exit with the same semantics; a 3-state enum would be more machinery than it's worth here.

justin-layerv:
Round 9 cr addressed in be18c0d:

- Switched success-path log from info to debug (volume concern).
- Added 0.9 threshold boundary test (89/100 → degraded; 91/100 → proceeds).
- Added a comment on the inline /unlinked tests explaining why they don't reuse `makeUnlinkedGuild`.

Issue #457 updated with item 4: the /unlinked stale-cache edge case when both counts are absent. The cleaner fix is changing the prewarm helper's return type to surface explicit success/failure — that's an API change touching two call sites and ~10 tests, deferred to keep this PR scope focused.

REST-budget cost in http-only mode (round 9 main concern) was already filed as #457 item 1 in round 7. CR has converged across 9 rounds; ready to merge once the round-9 CI clears.

justin-layerv:
## Overnight summary

PR #456 merged at `a1fa1fa` (2026-05-19 06:57 UTC) after 11 cr rounds — every comment addressed in-PR or filed as a follow-up, /simplify clean before merge.

**Sandbox deploy verified live:** task definition `qurl-bot-discord-sandbox-http:151` is COMPLETED, image `sha256:ca579b6186d34aab`, runningCount=2. The new REST-pagination prewarm code is in production sandbox. Live `/qurl send @everyone` verification still needs you (I can't click in Discord) — try it and look for the new `members pre-warm complete` log lines in CloudWatch `/ecs/qurl-bot-discord-sandbox-http` (us-east-2).

## Other queued tasks — disposition

| Task | Status | Notes |
|---|---|---|
| #28 view counter stuck | ✅ UI mitigated (PR #455); real fix gated on architectural decision below |
| #30 qurl-service `EventQurlAccessed` publish | ⛔️ Blocked on architecture | `shouldSuppressResourceWebhook` suppresses EVERY webhook for transit resources (qurl-service.go:164) — Discord bot's sends are 100% transit. Design proposal posted to qurl-service#596 with two options (event-type allowlist vs subscriber-aware suppression). Half-started branch discarded. Needs your signoff before coding. |
| #10 watermark 20MP cap | 📋 Investigated, filed qurl-integrations-infra#738 | Root cause: `maxStampPixels = 20_000_000` in qurl-s3-connector. Modern phones shoot 48 MP+. Fix proposal: pre-stamp downscale to ~16 MP. |
| #11 viewer auto-zoom cuts off image | 📋 Investigated, filed qurl-integrations-infra#739 | Root cause: canvas CSS has `max-width: 100%; height: auto;` but no `max-height`. One-line fix: add `max-height: 100vh` (matches video selector). Pure CSS, low blast radius. |
| Followups from #456 | 📋 Filed qurl-integrations#457 | 6 items: TTL-marker prewarm optimization, threshold empirical validation, cache_size dashboards, helper return-type for /unlinked, progress feedback during pagination, adaptive threshold. |

## Why I stopped here

The architectural blocker on #30 needs a design call, not a coding sprint. Shipping a half-baked PR that bypasses a security suppression at 3am was the opposite of "perfection." The two fileviewer issues (#738, #739) are real fixes but live in qurl-integrations-infra and warrant their own focused PR cycles when you're awake to direct prioritization (especially #738, which has a non-trivial decode-resize codepath).

Convergence signal on this PR was at cr round 8 ("LGTM"); rounds 9-11 extracted increasingly small polish. Every item in those rounds was addressed in-PR or filed in #457 with a clear deferral rationale.

Recommend: pick up #739 first (one-line CSS, immediate user-visible win), then qurl-service #596 design call, then #738.

PR #234, opened 2026-05-10
feat(slack): refactor /qurl list to resources, add $r_* fallback in /qurl get

Kevin-layerV:
## Summary

Rebased onto current main (post-#233/#447/#454). The net delta of #234's intended work on top of what's now in main:

### 1. `/qurl list` resource pivot

Pivots from "5 most recent qURLs" to "10 most recent qURL Resources" so each line is a copy-paste-ready `$<alias>` (when bound) or `$<resource_id>` token. Closes the ergonomic gap between list output and `/qurl get` input.

- New: `handleListResources` / `processListResources` (`apps/slack/internal/handler_list.go`).
- Admins see the master list unfiltered; non-admins see resources whose `resource_id` is in **`Store.AllowedResourceIDsForChannel`** — the union of `alias_bindings.values()` and `allowed_resource_ids` on the channel's `channel_policies` row.
- Fail-closed posture on non-admin + empty channel_id and non-admin + policy-fetch failure.
- New: `shared/client.Client.ListResources` (paginated `GET /v1/resources`) with a server-max (100) client-side clamp.
- New: `slackdata.Store.AllowedResourceIDsForChannel` (single-row GetItem; union of SS + Map values).

### 2. `$r_<id>` fallback on `/qurl get`

The parser now accepts the resource-ID shape `$r_<11 base64url chars>` after the `$` sigil (mirrors `qurl-service/internal/domain/qurl.go::resourceIDPattern`). Handlers route on `Command.Resource.Kind`.

- New: `ResourceTokenKind` discriminator, `ParsedResourceToken` struct, `Command.Resource` field, `resourceIDPattern` regex, `requireResourceToken` helper. `Command.Alias` stays populated on the alias branch for back-compat with existing read sites.
- `parseGet` continues to accept URL form (from #447) and `$<alias>` form (existing) — the new `$r_<id>` form is the third branch.
- `handler_get.go::getWork` resource-ID branch:
  - **Workspace admins** bypass the channel allow-set (matches their unfiltered `/qurl list` view — keeps the list footer's copy-paste promise working for them).
  - **Non-admins** must have the ID in `AllowedResourceIDsForChannel(team, channel)`; missing → `is not allowed in this channel` copy + `/qurl list` self-serve pointer + admin contact path.
  - AdminStore-nil sandbox path fails closed.
  - On post-rebase main, mints route to `POST /v1/resources/{id}/qurls` per #454 — no resource-id in the request body.

## Test plan

[...]

PR #459, opened 2026-05-19
revert(slack): drop /qurl list admin gate and channel-policy filter

Kevin-layerV:
## Summary

#234 added an admin probe + channel_policies filter to `/qurl list`. Net effect for any non-admin workspace member: a "No allowed tunnels on the first page…" empty state in any channel without a seeded `alias_bindings` / `allowed_resource_ids` row, even when the master listing has rows. This reverts that gate while keeping the rest of `/qurl list` intact.

> **Reconciled onto tunnel-only `/qurl list`.** This branch was cut before `main`'s tunnel-only refactor of `/qurl list` (slug tokens, `listResourcesScanLimit = 100`, `tunnelToken`/`formatTunnelListLine`). It has since been merged with `main`, so the revert is applied **on top of** that refactor — the tunnel-only rendering, scan limit, and renamed helpers are preserved; only the admin gate is removed. `main`'s `ResolvePolicy` (+ its legacy single-row fallback and `TestResolvePolicy_LegacySingleRowShape`) is also untouched — dropping it was in the pre-merge branch but is out of scope for this revert.

## Why

The list-side admin gate's purpose was disclosure-scoping — keeping non-admins from enumerating tunnels across channels they're not in. But:

- Capability gating still happens at mint time: `/qurl get $r_<id>` continues to enforce `AllowedResourceIDsForChannel` for non-admins (`handler_get.go`'s `authorizeResourceIDForGet` → `resourceAllowedForUser`). So pasting a seen-but-not-allowed `$r_<id>` into `/qurl get` from a foreign channel still fails closed.
- The gate's UX cost is high: a workspace member who isn't a seeded admin gets only the fail-closed empty state from `/qurl list` — no recoverable, self-serve path to discover what tunnels exist.
- The pre-#234 behavior matched user expectation (anyone in the workspace could `/qurl list`).

Net change: drops list-side disclosure scoping within a workspace; capability boundary unchanged.

## Rollout note (disclosure widening)

Within a workspace, every member now sees the same `/qurl list` output. After redeploy:

- Tunnel slugs (`$<slug>`) and their descriptions are visible to all workspace members. Pre-#234 behavior surfaced these too, but #234 had since narrowed it per-channel for non-admins; this revert undoes that narrowing.
- The resource-id fallback token (`$r_<id>`, shown only for a slug-less / alias-less tunnel) is surfaced to all members, including IDs the caller cannot mint for in their current channel.
- The alias path on the **get** side is unchanged: a user who sees a tunnel in the (now workspace-wide) listing and tries `/qurl get $<alias>` from a channel without that binding still hits `noResourceForAliasMessage`; the `$r_<id>` form returns `notAllowedInChannelMessage`. Follow-up tracked in #460.
- DM channels are affected too: pre-revert, a non-admin running `/qurl list` from a DM (a `D…` channel ID) hit the fail-closed empty branch because DMs have no `channel_policies` rows; post-revert they see the full workspace master listing. Consistent with pre-#234 behavior but distinct enough that "I ran `/qurl list` in a 1:1 and saw tunnels from #ops" is the surprise to pre-empt in release notes.
- No-DDB / sandbox deploys widen too: pre-revert, `/qurl list` with `AdminStore == nil` fail-closed to the empty state (the admin probe couldn't run, so the listing under-showed). Post-revert the handler doesn't touch `AdminStore` at all, so a sandbox with no policy tables but a wired customer API token now returns the full workspace master list. Fine for typical sandboxes (no production data), but flag it if any dev environment was relying on the `AdminStore`-nil empty state as a deliberate safeguard rather than a misconfiguration symptom.

Capability boundary is unchanged: minting still goes through `AllowedResourceIDsForChannel` (`handler_get.go`'s `authorizeResourceIDForGet`) and `LookupChannelAlias` (`handler_get.go`'s `resolveAliasForGet`). Workspaces that have been treating list-scoping as a *confidentiality* boundary on tunnel slugs/descriptions (rather than a *capability* boundary on minting) should be flagged in release notes.

## Removed (from `handler_list.go`)

- `scopeResourcesForUser` (admin probe + channel-policy filter)
- `userIsWorkspaceAdmin` helper
- `filterResourcesByAllowedSet` helper
- `listTunnelsEmptyMessageNonAdmin` (the non-admin empty-state copy — single empty-state remains)
- `listTunnelsNonAdminPaginationGapMessage` (the non-admin pagination-gap copy)
- Non-admin variant of the has-more footer (single footer remains)

## Kept

- `listResourcesScanLimit` (100), `listTunnelsEmptyMessage` (now a single admin-neutral message), `client.ResourceTypeTunnel`
- `processListResources` (simplified body), `filterTunnelResources`, `tunnelToken`, `formatTunnelListLine`, `mapListResourcesError`
- `AllowedResourceIDsForChannel` and `ResolvePolicy` in `slackdata/policies.go` — `AllowedResourceIDsForChannel` is still consumed by `handler_get.go`'s `$r_<id>` resource-ID branch (godoc updated to drop the stale `/qurl list` framing; TODO(#464) tracks the rename). `ResolvePolicy` is left exactly as `main` has it.

## Test diff (`handler_list_test.go`)

9 non-admin / AdminStore-nil tests removed (each maps 1:1 to a removed branch):
- `NonAdminFiltersToChannelPolicy`
- `NonAdminSeesAllowedResourceIDsWithoutAliasBinding`
- `NonAdminUnionsAllowedSetAndAliasBindings`
- `NonAdminEmptyChannelFailsClose`
- `NonAdminPaginationGap`
- `NonAdminEmptyChannelWithHasMoreShowsDefault`
- `NonAdminPartialPageHasMoreFooter`
- `AdminStoreNilTreatedAsNonAdmin`
- `NonAdminPaginationGapZeroTunnels`

12 rendering / sort / has-more / error / tunnel-filter tests kept (rendering shape, slug/alias/resource_id token precedence, description suffix, has-more footer, error mapping, stable sort + tiebreaker, empty-workspace empty state, URL-filtering, `TestTunnelToken`). `AdminSeesAllTunnels` renamed to `RendersAllTunnels` since the admin distinction is gone.

1 regression test added (`TestHandleList_UnscopedAcrossChannels`) — invokes from a policy-bearing channel, a policy-free channel, and a DM (`D…`) channel as a non-admin with a restrictive seeded policy, and asserts all three render the full listing (with negative fences against the empty-state and the removed pagination-gap copy). Re-introducing any channel-policy filter on `/qurl list` would fail it.

## Drive-by docs

- `CLAUDE.md` allowed commit-type list expanded from `feat | fix | chore | docs | test | refactor | ci` to `feat | fix | docs | style | refactor | perf | test | build | ci | chore | revert`, matching `.github/workflows/pr-title.yml`'s `amannn/action-semantic-pull-request` `types:` block. The doc previously lagged the CI-enforced set by four types (incl. `revert`, which is load-bearing for this PR's title). The scope-list note was also reformatted to a short bulleted list and now documents the `pr-title.yml` superset (`infra`/`deps`, tracked in #463). Surfaced during cr review of this PR; folded in here rather than spun out separately.

## Test plan

- [x] `go build ./apps/slack/...`, `go vet ./apps/slack/internal/...`, `go test ./apps/slack/internal/...` clean
- [x] `pre-commit run --files …` clean (go fmt, go mod tidy, golangci-lint)
- [ ] After redeploy: `/qurl-sandbox list` from any channel (incl. a DM) renders the master tunnel list with copy-paste `$<slug>` tokens, regardless of channel_policies state or admin status

## Deferred (filed)

- #540 — pre-existing soft enumeration signal: `noResourceForAliasMessage` ("is not configured") vs `notAllowedInChannelMessage` ("is not allowed") use different verbs. Out of scope for the revert.
- #464 — rename `AllowedResourceIDsForChannel` → `ChannelMintableResourceIDs` (repo-wide; the TODO is inline).


Kevin-layerV:
Addressing cr feedback from <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4490752208>:

- **#1 stale "ask an admin" footer copy** — fixed (dropped the trailing clause; the actionable `/qurl get $alias` half lands on its own).
- **#2 alias-discoverability vs channel-scope mismatch** — filed #460 (pr-polish-followup, enhancement). Pre-existing, out of revert scope. Captured the repro, the (a) annotate-rows / (b) doc-the-asymmetry tradeoff, and acceptance criteria that fence against accidentally re-narrowing `/qurl list`.
- **#3 `revert` not in CLAUDE.md's allowed types** — added to CLAUDE.md (PR title check workflow already accepts `revert`, so the doc was just stale).
- **#4 lost operator-grade log lines** — acknowledged; no action. The three `log.Warn` callsites tracked AdminStore-nil / empty channel_id / allowed-resource fetch failure inside `scopeResourcesForUser`. That code path is gone post-revert, so the warnings have nothing to fire on. No dashboards/alerts keyed on those messages.
- **#5 test-plan post-redeploy box** — will check off after staging deploy of `/qurl-sandbox list`. Won't gate merge on it; it's a verification, not a fix.
- **#6 regression test for cross-channel alias disclosure** — added `TestHandleList_NonAdminSeesAliasUnboundInChannel`. Non-admin invoking from a channel with zero `channel_policies` rows still sees an alias bound in a different channel. Pins the post-revert disclosure surface so a future re-introduction of channel-scope filtering has to delete or update the test.

Kevin-layerV:
Round-3 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491087612>:

- **#1 release-notes propagation** — filed #462 (`documentation`, `pr-polish-followup`). Captures the hand-edit-release-please-PR step + the customer/operator comms step + acceptance criteria. The release-please subject will read `revert(slack): drop /qurl list admin gate and channel-policy filter`; #462 tracks getting the disclosure-widening rollout note from this PR body into the CHANGELOG + customer-facing release surface.
- **#2 wire #460 into the code** — added `// TODO(#460): …` comment on `noResourceForAliasMessage` (`handler_get.go:74`). Grep-findable by future maintainers.
- **#3 rename test** — `TestHandleList_NonAdminSeesAliasUnboundInChannel` → `TestHandleList_UnscopedAcrossChannels`. The new name advertises the post-revert semantics on the test-name surface; non-admin seed is still load-bearing as a regression fence (commented inline).
- **#4 pin godoc cross-reference** — replaced "see handler_get.go" with `[Handler.authorizeResourceIDForGet]` (for `$r_<id>` branch) and `[Store.LookupChannelAlias]` (for `$<alias>` branch). Survives file growth.
- **Coverage gap (page.HasMore + zero rows)** — accepting the gap. Upstream API doesn't return `data: []` with `has_more: true` under normal operation; the simplified branch is correct for the realistic input distribution and a test would pin a synthetic combination.
- **Coverage gap (empty/missing channel_id)** — accepting the gap. The handler no longer reads `channelID`, so any wire-shape regression in that field is trivially correct from the code (no branch on it).
- **Pre-merge checklist post-redeploy** — still pending, will tick after staging deploy of `/qurl-sandbox list`.

Round-4 commit (`f1a9062`) is comment + rename only; no new logic surface.

Kevin-layerV:
Round-4 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491146027>:

- **DM channel callout** — added as a sub-bullet to #462 ([comment](https://github.com/layervai/qurl-integrations/issues/462#issuecomment-4491188159)). The DM behavior is a corollary of the workspace-wide widening and will ship in the same comms step.
- **BREAKING CHANGE footer** — skipping. No `slack-v*` tags exist (only release-please PR #14 is open, never merged), so #234 never reached customers via a tagged release. A footer would be noise. If a tagged release lands between #459 merge and the next release cut covering both #234 and #459, the rollout note still ships via #462 — that's what the issue tracks.
- **Test negative assertion** — added to `TestHandleList_UnscopedAcrossChannels`. Asserts the response doesn't contain `"past the first page"`, `"ask an admin to allow"`, or `"allow specific resources"` — defends against a partial filter reintroduction that gates only the empty-state copy.
- **CLAUDE.md type-list parity** — expanded to `feat | fix | docs | style | refactor | perf | test | build | ci | chore | revert`, matching the pr-title workflow's accepted set. Doc lagged the wire reality; now in sync.

Round-5 commit (`215464a`).

Kevin-layerV:
Round-5 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491215143>:

- **#1 disclosure-widening signoff** — tracked as #462 (`documentation`, `pr-polish-followup`). Captures the CHANGELOG hand-edit + customer/operator comms step + the DM-channel callout from cr round-4. The PR body's "Rollout note (disclosure widening)" is the canonical source; #462 ensures it propagates downstream. No stakeholder signoff link to attach (the disclosure tradeoff was reasoned in the PR body and accepted by cr across rounds 1-5).
- **#2 `notAllowedInChannelMessage` misleading breadcrumb** — fixed in `af9e099`. Dropped `Run /qurl list to see what's available here` (it would surface the same row the user pasted from); route to admin escalation only. Added `TODO(#460)` on the godoc capturing the parallel discoverability gap. Test `TestHandleGet_DollarResourceIDNotInAllowedSetNonAdmin` updated to assert on the new copy + a negative assertion that the breadcrumb stays gone under future copy churn.
- **#3 CLAUDE.md scope** — justified in PR body under new "Drive-by docs" section. The expansion to `style/perf/build` was a parity fix surfaced during cr review (workflow was already accepting them); folded in rather than spun out as a separate PR.
- **Nit on partial-page footer copy escalation** — confirmed UX. The "ask an admin to allow" escalation now lives at mint time on `notAllowedInChannelMessage` (this round) and `noResourceForAliasMessage` (already there). The list footer is purely informational about pagination scope.
- **Nit on "no admin seed needed" comment** — added on `TestHandleList_RendersResources` (cr round-5). Future readers hitting `git blame` see the rationale inline.

Kevin-layerV:
Round-6 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491274652>:

- **#1 stale godoc on `AllowedResourceIDsForChannel`** — rewrote to reflect post-revert reality (mint-time gate via `handler_get.go`'s `authorizeResourceIDForGet`; the `/qurl list` consumer is gone). Same data structure, narrower consumer surface.
- **#2 dead `seedPolicySet` in regression test** — dropped. The renderer reads `r.Alias` straight off the upstream payload and never consults `channel_policies`, so the seed wasn't actually exercising what the inline comment claimed. Tightened the comment to describe what the test actually fences.
- **#3 `notAllowedInChannelMessage` discoverability gap** — confirmed `TODO(#460)` is the disposition. Future contributor touching the copy will land on the alias-discoverability follow-up.
- **#4 scope-list divergence (`infra`/`deps`)** — filed #463 (`documentation`, `pr-polish-followup`). The type-list sync was already a stretch for a revert PR (cr round-5 flagged it as scope creep, justified in PR body under "Drive-by docs"); the scope-list sync is the same shape of change and belongs in its own PR. Notes the (a) vs (b) tradeoff and recommends adding `infra`/`deps` to both CLAUDE.md and `bug_report.yml` to match what `pr-title.yml` already enforces.
- **#5 rollout: release-note callout** — tracked by #462. Confirmed: the disclosure-widening rollout note in the PR body will be hand-edited into the release-please CHANGELOG when it fires.

Round-7 commit (`8276cf0`).

Kevin-layerV:
Round-7 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491319776>:

- **#1 `TestHandleList_UnscopedAcrossChannels` slightly under-named** — tightened the godoc to call out that the negative-copy assertions are the load-bearing fence: they catch partial filter reintroduction that gates only the empty-state copy, while the row assertion catches full filter reintroduction. Together they backstop the concern cr-7 raised. Skipped rename (already renamed once per cr-3) and skipped re-adding the seed (cr-6 explicitly dropped it as dead scenery). The two pieces of cr feedback were in tension; the godoc clarification resolves both.
- **#2 CLAUDE.md "(not CI-enforced)" qualifier inaccurate** — rewrote the scope-list note. The pr-title workflow validates scope against the allowed list when one is present; `requireScope: false` only makes scope optional, not unvalidated. New note also lists `pr-title.yml` as a third sync location.
- **#3 `AllowedResourceIDsForChannel` name out of step** — added a rename TODO in the function godoc (`ChannelMintableResourceIDs` was cr-7's suggested rename target). Out of scope for this PR per cr's own framing ("not worth renaming in this PR — flag as a follow-up").
- **#4 admin-detection brittleness on `notAllowedInChannelMessage`** — accepting. cr explicitly calls it minor + defense-in-depth; the admin path bypasses this surface via `authorizeResourceIDForGet`'s admin-shortcut, so the surface only matters under an admin-detection failure (a separate concern not introduced by this PR).

Round-8 commit (`23498cb`).

Kevin-layerV:
Round-8 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491393736>:

- **#1 `authorizeResourceIDForGet` godoc drift** — fixed in `e251c96`. Rewrote the godoc to surface the post-revert asymmetry (list workspace-wide, mint still channel-scoped) and cite `TODO(GH-460)` which tracks the UX gap.
- **#2 CLAUDE.md scope-list sync** — tracked by #463. Spun out per cr-6/cr-7's own framing that doc-CI sync is a separate follow-up. PR body's "Drive-by docs" section + this thread of comments captures the acknowledgement.
- **#3 release-notes flag for second-order UX gap** — tracked by #460 (alias-discoverability) and #462 (release-notes propagation). Support team heads-up is in #462's action items.
- **#4 test-coverage nit: name vs mechanism** — added a paragraph to `TestHandleList_UnscopedAcrossChannels` godoc clarifying that the name describes intent (no per-channel filter applies) not mechanism (renderer doesn't consult channel_policies), with a forward-looking note for the case where a future change wires alias resolution through DDB.
- **#5 empty-state surface visible to all members** — accepting. cr explicitly says "probably nothing to do here, just flagging"; the existing empty-state copy is fine.

Round-9 commit (`e251c96`).

Kevin-layerV:
Round-9 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491431159>:

- **#1 TODO marker style** — fixed in `9f45f87`. `TODO(GH-460)` → `TODO(#460)` so all three #460 markers share the same form (`git grep TODO(#460)` finds them in one shot). Bare `TODO:` on `AllowedResourceIDsForChannel` got a tracker: filed #464 (`good first issue`, `pr-polish-followup`) for the rename to `ChannelMintableResourceIDs`. TODO now reads `TODO(#464)`.
- **#2 `TestHandleList_RendersResources` cross-reference** — skipping (cr explicitly says optional). The two adjacent test godocs (`RendersResources` no-admin-needed comment + `UnscopedAcrossChannels` non-admin-explicit comment) read together; a hard cross-reference between them risks reverse-staleness if either test changes shape.
- **#3 DM disclosure in PR body** — added as a fourth bullet in "Rollout note (disclosure widening)" with the "I ran `/qurl list` in a 1:1 and saw URLs from #ops" framing cr suggested. Already in #462 too; surfacing in the PR body lifts it into the release-please CHANGELOG via the body-pull behavior.
- **#4 CLAUDE.md drive-by** — confirmation only, no action.
- **#5 asymmetry between the two not-found copies** — added a "Why this keeps a breadcrumb while sibling dropped one" paragraph to `noResourceForAliasMessage` godoc. Explains that `/qurl aliases` is channel-scoped (breadcrumb stays accurate) while `/qurl list` went workspace-wide (breadcrumb became misleading), so future readers don't drift on either surface.

Round-10 commit (`9f45f87`).

Kevin-layerV:
Round-10 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491494293>:

- **Nit #1 `$alias` → `$token`** — fixed in `e58ac43`. Has-more footer now matches the "Copy any `$token`" hint two lines above; covers `$r_<id>` un-aliased rows.
- **Nit #2 `[Store.LookupChannelAlias]` bracket link** — dropped the brackets. `slackdata` isn't imported in `handler_list.go` (only `shared/client` is), so neither qualified nor bare form resolves through gopls. Plain text is honest about what's a navigable godoc link.
- **Nit #3 `AllowedResourceIDsForChannel` opener phrasing** — adopted cr's suggested rewrite ("authorizes for non-admin mint via the `$r_<id>` get path"). Same meaning, less passive voice.
- **Coverage gap #1 (cross-channel alias seed)** — accepting the gap. cr-6 explicitly asked me to drop the `C_other` `seedPolicySet` line in `TestHandleList_UnscopedAcrossChannels` as dead scenery (the renderer doesn't consult channel_policies post-revert). Reverting that would put the cr chain back into tension; the existing negative-copy assertions already catch the broad regression shape per cr-10's own assessment.
- **Coverage gap #2 (DM channel test variant)** — accepting the gap. cr-10 explicitly marks non-blocking. The DM behavior post-revert is structurally identical to any other channel (the handler doesn't read `channel_id` for the list path), so a DM-specific test would pin the same surface the existing test does. If the alias-resolution mechanism changes (the godoc's "if a future change wires alias resolution through DDB" note), both cross-channel and DM variants become worth landing then.

Round-11 commit (`e58ac43`).

Kevin-layerV:
Round-11 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491534653>:

- **#1 stale ResolvePolicy refs (ship-blocker)** — fixed in `8c38388`. Three doc-comment sites in `policies.go` now cite `AllowedResourceIDsForChannel` + `authorizeResourceIDForGet` instead of `ResolvePolicy`. ResolvePolicy itself has zero callers (`git grep '\.ResolvePolicy('` confirms) — extended the TODO(#464) marker on `AllowedResourceIDsForChannel` to flag deletion as part of the same cleanup pass. Also commented #464 with the expanded scope.
- **#2 cross-channel mechanical test** — deferring per cr's "reasonable to defer to a follow-up if you'd rather keep this PR tight to the revert scope." This is the same item cr-6 asked me to drop and cr-10 said was non-blocking; net-net the cr chain disagrees with itself, and the existing positive+negative assertions catch the broad regression. The forward-looking comment in `TestHandleList_UnscopedAcrossChannels` godoc ("If a future change wires alias resolution through DDB, this test should grow a positive cross-channel-seed assertion to fence the regression.") is the deferral marker.
- **#3 TODO(#460) regression test** — deferring. The #460 issue itself is the home for both the UX-gap fix AND its regression test. Pinning the current `noResourceForAliasMessage` text from a foreign-channel scenario would lock in copy that #460 is explicitly going to rewrite. Better to land them together.
- **Nit on $token godoc note** — added a 3-line code comment near the has-more footer in handler_list.go explaining the generalization choice. Catches the next reader who might "fix" it back to `$alias`.
- **Nit on `C_no_bindings` channel-ID shape** — accepting the gap. Every test in `apps/slack/internal/` uses `C_test`, `C_other`, `C_caller`, etc. — all underscore-snake-case, none matching the actual Slack `C[A-Z0-9]{8,}` format. Changing only this one test would create local inconsistency, not consistency. If a future code path validates shape, that's a wholesale test-fixture sweep, not a single-test fix.

Round-12 commit (`8c38388`).

Kevin-layerV:
Round-12 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491587193>:

- **#1 ResolvePolicy delete-now (ship-after)** — done in `8c25c63`. Deleted `(*Store).ResolvePolicy` + its 35 lines of dead code. Also cleaned up the three test-fixture godoc references that pointed at it (`fakeddb_test.go`, `fakeddb_fixtures_test.go`, `admin_test_helpers_test.go` now cite `AllowedResourceIDsForChannel`). The TODO(#464) delete-note is gone since the work is now in this PR; #464 closes back to just the rename. The lone surviving "ResolvePolicy" reference is the historical `store.go:14` "the old AdminClient had (CheckAdmin, ResolvePolicy, …)" comment — that's accurate history of the HTTP-shape migration, leaving as-is.
- **#2 CLAUDE.md "all three places" prose (ship-after)** — softened. New prose names `pr-title.yml` as the CI-enforced superset, names the `infra`/`deps` divergence explicitly, points at #463 for sync. The new-scope instruction now reads "touch CLAUDE.md and `bug_report.yml`; add to `pr-title.yml` too if it isn't already in the superset" — accurate against today's state without forcing the bigger sync in this PR.
- **#3 doc-comment line-wrap reflow** — done. The `allowed_resource_ids` bullet in the policies.go preamble now reads as one paragraph instead of "without\n an alias,\n and an alias" fragments.
- **#4 PR-body line numbers** — replaced `handler_get.go:368` and `:332` with the function names (`authorizeResourceIDForGet` / `resolveAliasForGet`). Function refs survive line drift; line numbers don't.
- **DM target-URL release-note nuance** — already captured in PR body rollout bullet 4 (DM bullet) and in #462. cr-12's concrete framing "I ran `/qurl list` in a 1:1 and saw URLs from #ops" is the exact wording I used in the rollout note.
- **Test plan post-redeploy box** — staying unchecked until staging deploy of `/qurl-sandbox list` lands. That's the right gate per cr's own framing.

Round-13 commit (`8c25c63`).

Kevin-layerV:
Round-13 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491648738>:

- **Nit #1 store.go:14 stale ResolvePolicy** — fixed in `892b22e`. Removed `ResolvePolicy` from the live method-shape list, added a one-paragraph note that it was carried over with the same AdminClient migration but dropped in #459 (subsumed by `AllowedResourceIDsForChannel`). Migration history stays intact; the comment no longer claims a method that doesn't exist.
- **Nit #2 testListAliasAlpha rationale stale** — dropped the const + inlined its single call site in `TestHandleList_StableSortBetweenAliasAndResourceID`. Const block godoc rewritten to drop the misleading goconst-threshold framing and describe the actual motivation (one-rename-point readability).
- **Operational DM-comms follow-up** — already captured in #462's action items + the PR body's rollout bullet 4. Release-notes draft will lean on those, not the PR body alone.

Round-14 commit (`892b22e`).

Kevin-layerV:
Round-14 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491693743>:

- **Nit #1 godoc bracket-style inconsistency** — re-bracketed `[slackdata.Store.LookupChannelAlias]` in `383e9b9`. cr-10 had asked me to drop brackets (slackdata not imported, so the link may not resolve via gopls); cr-14 wants the bracket form for sibling-consistency. Bracket syntax is harmless if it doesn't resolve, so adopting cr-14's preference.
- **Nit #2 DM callout in handler godoc** — added a paragraph to `handleListResources` godoc explicitly naming DM behavior (`D…` channel IDs): pre-#234 unscoped, #234 narrowed to empty, this revert restores unscoped. Future-proofs the surface.
- **TODO(#464) rename refinement** — captured cr's suggestion that `ChannelMintableByResourceID` is a more precise target than `ChannelMintableResourceIDs` in [a comment on #464](https://github.com/layervai/qurl-integrations/issues/464#issuecomment-4491700013). The function gates the `$r_<id>` path; an alias-path mint goes through `LookupChannelAlias` regardless.

Round-15 commit (`383e9b9`).

Kevin-layerV:
Round-15 cr disposition — <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491739289>:

- **Nit #1 godoc rhetorical staleness** — trimmed in `347a0da`. `authorizeResourceIDForGet` no longer claims the unfiltered list view is an admin privilege; admin bypass still happens (admins skip the allow-set check), but the list view itself is now unfiltered for everyone.
- **Nit #2 AllowedResourceIDsForChannel rename** — deferred to #464 with cr-14's refinement (`ChannelMintableByResourceID` vs. `ChannelMintableResourceIDs`) noted on the issue.
- **#3 DM release-note** — tracked by #462.
- **#4/#5 CLAUDE.md drive-by + test-side detail** — confirmations, no action.

Round-16 commit (`347a0da`). After 15 rounds of cr the diff has stabilized at a tight revert + 4 tracked follow-ups (#460, #462, #463, #464). Calling this convergence.

Kevin-layerV:
Thanks for the review. Addressed in `6dbd2ed`; rundown of every item:

**Strengthened the regression test (the strong suggestion + "name promises more than it delivers").** `TestHandleList_UnscopedAcrossChannels` now loops over three channel shapes that diverged pre-revert for a non-admin and asserts all three render the full listing:
- a channel carrying a restrictive `channel_policies` row (pre-revert → filtered to `prod-db` only, hiding `secret`),
- a channel with no policy row (pre-revert → fail-closed empty),
- a **DM (`D…`) channel** (pre-revert → fail-closed empty) — the "`/qurl list` in a 1:1 shows #ops URLs" case from the rollout note.

So the DM behavior now has a fence, and the name earns its scope.

**Inert scaffolding nit.** One correction: `seedNonAdmin`/`seedAdmin` aren't fully inert — they seed the `workspace_mappings` row that `authenticatedClient` needs for the API-key lookup, so without them `/qurl list` returns the auth-error path, not a listing. What's now irrelevant to the list handler is the admin-vs-non-admin *distinction*, not the seed. In the strengthened test the `seedNonAdmin` + restrictive `seedPolicySet` are explicitly load-bearing: if a channel-policy filter were re-introduced, this non-admin caller would see the old filtered/empty output and the test would fail. Added a comment saying exactly that. Left `RendersAllTunnels`'s `seedAdmin` as-is since it's the auth fixture there too.

**TODO(#464) rename — deferring (already tracked).** Agreed the name carries disclosure-era framing, but the rename touches `AllowedResourceIDsForChannel` across `handler_get.go`, `handler_list.go`, `policies.go`, and the test helpers/fixtures — a repo-wide rename I don't want to fold into a revert PR. It's already its own filed issue (#464, titled exactly `rename AllowedResourceIDsForChannel → ChannelMintableResourceIDs`), so it'll land cleanly as its own change.

**Verb-mismatch enumeration signal — filed #540.** Pre-existing (not introduced here), so out of scope for the revert as you noted. Filed #540 (`security` / `priority: low` / `pr-polish-followup`) with repro + acceptance criteria.

**Logging confirm — not load-bearing for audit.** The removed `list:` warn lines fired only on the now-deleted filter / fail-closed / pagination-gap branches; they were gate-edge diagnostics, not the invocation audit. `runAsync` already attaches `command`/`team_id`/`channel_id` to every `/qurl list` invocation's logger, so per-invocation attribution is unchanged.

The four "things that look right" confirmations match my intent — thanks for verifying the `filterTunnelResources` ordering, the unified `HasMore` footer, the `policies.go` legacy-scalar caveat retention, and the CLAUDE.md superset framing.

Kevin-layerV:
Round-2 pass addressed in `cea2fc4`:

**Sandbox / AdminStore-nil widening — added to the rollout note.** Good catch; this wasn't called out. Added a bullet: pre-revert `/qurl list` with `AdminStore == nil` fail-closed to the empty state (the admin probe couldn't run, so it under-showed); post-revert the handler never touches `AdminStore`, so a sandbox with no policy tables but a wired customer API token now returns the full master list. Flagged for anyone treating the nil-state empty list as a deliberate safeguard vs. a misconfiguration symptom.

**Defensive negative assertion — added.** `TestHandleList_UnscopedAcrossChannels` now also asserts each channel's response contains neither `"No tunnels found"` (empty-state) nor `"past the first page"` (the removed non-admin pagination-gap copy), so a filter reintroduced on only one branch is caught even if another branch still renders rows.

**Footer false-positive on URL-heavy workspaces** — agreed it's acceptable until #531 lands a tunnel-accurate `has_more`; the godoc already names it, no change.

Capability-boundary, orphan-reference, breadcrumb, and CLAUDE.md verifications all match intent — thanks for the independent pass.

Kevin-layerV:
Round-3 pass addressed in `b24f0e1` + a PR-body sweep:

1. **PR description ↔ code drift — swept the body.** You're right that it still described the pre-tunnel-only surface. Updated to match what actually merged: tunnel-only listing (`listResourcesScanLimit = 100`, `tunnelToken`/`formatTunnelListLine`, slug tokens), corrected the Removed/Kept symbol names, fixed the test-diff counts (9 removed / 12 kept / 1 added; `AdminSeesAllTunnels`→`RendersAllTunnels`), and added an up-front note that this branch was reconciled onto main's tunnel-only `/qurl list` (and that `ResolvePolicy` is left as main has it). Also corrected the disclosure surface in the rollout note — tunnels render slug + description, not target URLs — and dropped the stale `/qurl admin claim` reference (that flow is gone on main).

2. **Admin-case empty copy** — leaving as-is; it's the documented trade-off (naming the admin-only command without imperatively telling a non-admin to run it). The godoc says so.

3. **`t.Run` subtests — done.** `TestHandleList_UnscopedAcrossChannels` now uses `t.Run(channelID, …)`, so `-run` can target a single branch and PASS/FAIL is per-channel.

4. **CLAUDE.md long sentence — bulleted.** Same content (the #463 sync, `requireScope: false`, where-to-touch when adding a scope, the `other` exclusion), now a short bulleted list under a one-line lead.

Capability-boundary re-verification matches intent — thanks for reading the call sites.

Kevin-layerV:
Round-4 pass — addressed in `ab2757a`, rest are confirms:

1. **`page.HasMore` footer accuracy** — confirmed #531 (`push type=tunnel filter server-side so has_more reflects tunnels`) is open and is the tracked fix. Interim posture stays; godoc points at it.
2. **`t.Parallel()` subtests** — leaving sequential. The subtests share the `h` Handler and the `ts` mock customer server; `invokeAdminAsync` round-trips through that shared mock + a per-invoker response capture, which isn't obviously safe under concurrent invocations. Three fast subtests don't justify the flake risk, so sequential is the correct call here.
3. **Em-dash footer copy / a11y fixture** — the single-variant footer copy is byte-identical to main's prior *admin* footer (only the non-admin variant was dropped), so nothing an external fixture referenced has changed.
4. **CHANGELOG excerpt** — leaving to release-please; it parses the commit/PR body, and a manual `CHANGELOG.md` edit would conflict with its managed section. The rollout note is structured so the release-notes author can lift it.
5. **`RendersAllTunnels` scaffolding clarity — done.** Added a comment that `seedAdmin` only supplies the workspace_mappings/API-key fixture and the listing is admin-neutral (renders identically for a non-admin; `UnscopedAcrossChannels` covers that shape).

Thanks for the capability-boundary re-verification.

Kevin-layerV:
Round-5 pass — addressed in `5ad6e5d`, rest are confirms:

1. **`seedPolicySet` doc scope — broadened.** Reworded to name the `$r_<id>` mint gate as the *primary* consumer while acknowledging the alias, admin-policy, and `/qurl list` regression callers, so the comment no longer reads narrower than its actual use.
2. **`AllowedResourceIDsForChannel` empty-`channelID` branch** — agreed it's effectively unreachable now (`/qurl get` enforces `channelRequiredMessage` upstream). Leaving it for the #464 rename to drop alongside the wire-shape simplification, as you suggested, rather than touching `policies.go` behavior in a revert.
3. **DM disclosure in release notes** — it's the lead surprise in the rollout note ("…saw tunnels from #ops") and pinned by the DM subtest; release-notes author has it.
4. **Sandbox / `AdminStore == nil`** — captured as a deploy-checklist bullet in the rollout note; not actionable in-code here.

Thanks for the thorough multi-round pass.

Kevin-layerV:
Round-6 — all four flags handled (no code change needed; one deferral, three confirms) + the CI red is a known flake:

1. **`$r_<id>` fallback disclosure** — confirmed not a concern. Resource IDs are capability-gated identifiers, not secrets: the mint gate (`AllowedResourceIDsForChannel`) governs minting regardless of whether the caller knows the ID, so reading an ID out of `/qurl list` grants nothing. The rollout note already calls out that `$r_<id>` tokens are surfaced workspace-wide including IDs the caller can't mint. And these slug-less fallback tokens are vanishingly rare (tunnels are created with a slug).

2. **`/qurl aliases` breadcrumb on `notAllowedInChannelMessage`** — deferring to #460 (added the sub-suggestion there: https://github.com/layervai/qurl-integrations/issues/460#issuecomment-4569753875). There's a real surface mismatch — that message fires on a `$r_<id>` token but `/qurl aliases` lists aliases, not IDs — so it's a less direct fit than the alias→`/qurl aliases` pairing, and it interacts with the message-mirroring/verb work in #540. Better decided alongside the broader alias rework than as an ad-hoc copy tweak in a revert.

3. **DM-channel release-notes prominence** — it's the lead example in the rollout note and pinned by the DM subtest; release-notes author has it.

4. **Sandbox / `AdminStore == nil` deploy check — ran it.** `git grep AdminStore` across the repo: it appears only in Go handler/oauth wiring + `main.go` construction — nothing in any deploy yaml/tf/md keys off the nil-empty behavior, so no deploy artifact relies on the old fail-closed empty state. Captured as a deploy-checklist bullet in the rollout note regardless.

The `Test` check went red on the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` (`kubectl dry-run failed: signal: killed` at the 5s CI timeout — unrelated to this diff, passes locally). Re-running that job; everything else (incl. Docker) is green.

PR #465, opened 2026-05-19
feat(discord): qurl.accessed webhook receiver + restore view counter UI

justin-layerv:
## Summary
Restores the `👀 N viewed / M pending` line on `/qurl send` + `/qurl map` confirmation messages that was removed in #455. The old upstream-polling implementation went silent when transit resources stopped surfacing their `qurls` array; this PR replaces the poll with a `qurl.accessed` webhook receiver.

- New `POST /webhooks/qurl` route with HMAC-SHA256 verify (shared with the GitHub webhook via the new `src/utils/webhook-hardening.js`), per-IP bad-sig rate limit, replay protection
- New `qurl_views` DDB store methods: `recordQurlView` (conditional MAX-merge update with 30-day TTL refresh) + `getQurlViews` (BatchGet)
- Live monitor reader swapped from `getResourceStatus` upstream poll → `db.getQurlViews` BatchGet on the in-memory `qurlLinks` set
- `qurl_id` threaded from `connector.mintLinks` → `mintLinksInBatches` → `qurlLinks` element shape
- Monitor capped at 14 min to match the interaction webhook token's TTL (ephemeral confirmation messages can only be edited via that token; past it, no edit path works)
- 2414 tests pass; new receiver + store + flow + monitor tests included

## Why
The view counter is a sender-facing signal that recipients are actually engaging with the share. It went dark in #455 when the polling source dried up. The webhook approach is also strictly more correct: the upstream poll only ever rendered "opened" on full minute-tick boundaries, while the webhook fires within seconds of the recipient clicking.

DDB is intentionally the multi-instance bridge — a webhook landing on instance A reaches a monitor running on instance B without any in-process pub/sub.

## Wire contract
- Header `QURL-Signature` = bare hex HMAC-SHA256 over the raw body (no `sha256=` prefix — different from GitHub's wire shape)
- Body `{id, type, data:{qurl_id, resource_id, access_count, consumed}, owner_id, timestamp, api_version}` (matches qurl-service's `WebhookEvent` payload contract)
- `src_ip` + `user_agent` are stripped server-side for transit resources (connector-owned privacy boundary; Discord bot sends are always transit)

## Operator setup
Full step-by-step in `apps/discord/docs/qurl-webhook-rollout.md`:
1. Provision the `qurl-views` DDB table (handled by the deploying organization's infrastructure)
2. Deploy the bot
3. Set `QURL_WEBHOOK_SECRET` in SSM
4. Register the qurl-service webhook subscription with the same secret

The receiver returns 503 with a boot WARN until step 3 is done. No webhooks arrive until step 4. Both failure modes are documented in the runbook with recovery paths.

## Test plan
- [x] `npx jest` — 2414 tests pass
- [x] `npx eslint` clean
- [ ] Sandbox: deploy, set SSM secret, register subscription, run `/qurl send` to a test recipient, confirm counter advances
- [ ] Prod: same sequence after sandbox validation

### DIFF
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -1,0 +1,93 @@
+# qURL webhook rollout — operator runbook
+
+Powers the `👀 N viewed / M pending` counter on `/qurl send` + `/qurl map`
+confirmation messages. The counter was stripped in
+[`qurl-integrations#455`](https://github.com/layervai/qurl-integrations/pull/455)
+because the upstream polling approach broke when `qurl-service` started
+stripping the `qurls` array from `GET /v1/qurls/:id` responses for
+type=transit resources (every Discord bot send is transit).
+
+This rollout replaces the poll with a `qurl.accessed` webhook receiver.
+
+## Rollout order
+
+Each step blocks the next — do not skip ahead.
+
+1. **`qurl-views` DDB table.** Terraform-managed in
+   `qurl-integrations-infra/qurl-bot-discord/terraform`. Without it, the
+   bot's monitor `BatchGet` returns the empty map and the counter
+   silently stays at `0 viewed / N pending` forever. Confirm with:
+   ```
+   aws dynamodb describe-table \
+     --table-name <DDB_TABLE_PREFIX>qurl-views \
+     --region <env-region>
+   ```
+2. **Deploy the bot.** Code mounts `/webhooks/qurl` unconditionally and
+   warns on boot if `QURL_WEBHOOK_SECRET` is unset — that's the operator
+   signal that step 3 hasn't happened yet.
+3. **Set the SSM secret.**
+   ```
+   aws ssm put-parameter \
+     --name "/<project>/QURL_WEBHOOK_SECRET" \
+     --type SecureString \
+     --value "$(openssl rand -hex 32)" \
+     --overwrite \
+     --region <env-region>
+   ```
+   Then restart the task so the secret is picked up:
+   ```
+   aws ecs update-service --service <svc> --force-new-deployment ...
+   ```
+4. **Register the subscription with qurl-service.** The subscription's
+   `secret` field MUST be the same hex string as step 3 — the bot's
+   verifySignature pins bare-hex HMAC-SHA256 over the raw body.
+   ```
+   curl -X POST https://<qurl-service-host>/v1/webhooks \
+     -H 'Authorization: Bearer <qurl-service-admin-token>' \
+     -H 'Content-Type: application/json' \
+     -d '{
+       "url": "https://<bot-host>/webhooks/qurl",
+       "event_types": ["qurl.accessed"],
+       "secret": "<same-hex-as-step-3>"
+     }'
+   ```
+
+## Wire shape (pinned)
+
+- Header: `QURL-Signature` = bare-hex HMAC-SHA256 over the raw body.
+  No `sha256=` prefix — that's the GitHub wire shape, NOT this one.
+- Body: `{event, event_id, data:{qurl_id, resource_id, access_count, consumed}}`
+- `src_ip` + `user_agent` are stripped server-side for type=transit
+  resources (the connector-owned privacy boundary). Discord bot sends
+  are always transit; we don't read those fields.
+
+## Failure modes
+
+- **Bot starts, but no secret set.** Boot log emits `QURL_WEBHOOK_SECRET
+  unset — qURL webhook receiver mounted but will reject all inbound
+  traffic with 503`. The view counter renders `0 viewed / N pending`
+  on every send (the monitor's `BatchGet` returns the default empty
+  map). Recover via step 3 above.
+- **Secret rotates without restarting qurl-service subscription.**
+  All inbound webhooks return 401 (signature mismatch). The bot's
+  per-IP `BAD_SIG_MAX=30` rate limit kicks in after 30 attempts and
+  switches to 429. Recover by updating the subscription's `secret`
+  field (PATCH the subscription).
+- **`qurl-views` table missing.** The bot's monitor `BatchGet` throws
+  `ResourceNotFoundException`. The setInterval's try/catch swallows
+  it and logs `Link monitor poll failed` — the counter sticks at
+  `0 viewed / N pending`. Recover by applying the terraform.
+- **Some links missing `qurl_id`.** Upstream connector pre `qurl-s3-connector#747`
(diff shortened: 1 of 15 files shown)


justin-layerv:
## /simplify + cr round 1 — summary of fixes

Ran /simplify (3 parallel agents) + addressed cr round 1 in one push (commit a2d1747).

**🔴 Critical bug caught by cr:** addRecipients was extending trackedQurlIds but never seeding linkStatus, so the counter froze for /qurl add'd recipients. Fixed with a regression-guard test (`addRecipients() seeds linkStatus so views on newly-added recipients flip pending → viewed`).

**Shared utility extracted:** `apps/discord/src/utils/webhook-hardening.js` exports `createBadSigLimiter` + `verifyHmacSha256`. Both webhook routes (GitHub + qURL) now consume it. Closes a real drift hazard (GitHub's sweep wasn't unref'd; qURL's was).

**Removed dead code:** `newResourceIds` field across 13 addRecipients return paths — added "for symmetry" but read by nothing post-PR. Renamed `newQurlIds` → `newLinks` carrying `{qurlId, username}` pairs (load-bearing for the seeding bug fix).

**Test consolidation:** trimmed flow-test from 5 → 2 cases (dedup/out-of-order/foreign-qurl_id were already pinned by store-layer tests against the real DDB ConditionExpression, not the in-memory mock).

**ConditionExpression widening (cr finding):** allows a same-access_count event when consumed flips false → true. Pins the qurl-service contract that future emission shapes won't silently drop.

**Comment pruning:** trimmed narrative/PR-reference comments across constants.js, routes/qurl-webhook.js, commands.js, ddb-store.js, server.js, config.js. Kept WHY-comments (hidden invariants, workarounds); dropped WHAT-comments and references to retired PR shapes.

**Efficiency wins:** iterate sparse `views` map (typically 1-2 entries) instead of full `linkStatus` (up to 50) in runTick; hoist `Date.now()` in recordQurlView; skip `.filter()` allocation on happy-path rate-limit gate; extract BatchGet result-collection loop.

**cr round 1 finer points:** export pattern → `router.fn = …; module.exports = router;`; concurrent-stop test now has a load-bearing assertion; setMessageId null-window comment explains the bounded fallback; `consumed` field marked as reserved (not dead); buildStatusMsg expired-state note; setSystemTime test scenario documented; boot WARN split for log-viewer friendliness; runbook flags the ~1 min rotation-window blackout.

**Deferred (with linked issue):** #466 — getQurlViews retry-backoff tuning. PAY_PER_REQUEST RCU + ≤50-key BatchGet makes single-retry sufficient at current volumes; matching recordQURLSendBatch's exponential pattern is hygiene.

**Tests**: 2409 pass. Lint clean.

justin-layerv:
## cr round 2 — addressed (commit de7f0c0)

### 🔴 Critical: ephemeral message channel-edit fallback was broken (caught by reviewer #5)

`/qurl send` + `/qurl map` both `deferReply({ ephemeral: true })`. Ephemeral messages can ONLY be edited via the interaction webhook token; a bot-token PATCH against `Routes.channelMessage` returns Unknown Message 10008. My fallback path was futile in production.

Fix: removed the fallback entirely. `safeEdit` just calls `interaction.editReply`. Past the 15-min interaction-token TTL, those 401s and `logIgnoredDiscordErr` swallows. To avoid burning CPU/RCU past the usable window, lowered `MAX_MONITOR_DURATION_MS` from 1h to 14min. Store-side view recording (the webhook receiver path) is unaffected — only the sender's confirm-message counter freezes past 14min. Recipients see status from their DM regardless.

Test (`'edits via editDM (bot-token PATCH) once past the 14min cutover'`) replaced with a contract pin: `'never falls back to editDM — the confirm message is ephemeral'`.

### 🟡 Smaller items (every reviewer touched some)

- **Stale `'apikey'` positional arg** in 30+ monitor test calls (3 reviewers) — removed.
- **Strict typeof guards** on receiver inputs: `body.id` (typeof !== 'string'), `data.access_count` (Number(null)===0 used to slip through), `data.consumed` (=== true, not Boolean() coercion of "false").
- **Limited `logger.warn` payload** to `{qurl_id, resource_id}` — `data` is post-trust but unbounded and may include attacker-influenced fields for non-transit resources.
- **Signature regex** documented as lowercase-only by design (Node + Go both emit lowercase; cross-language emitter defaulting to uppercase should fail loudly).
- **`createBadSigLimiter` perIpCap** now defaults to `max * 4` so a caller bumping max doesn't leave the per-IP array sized for the old value.
- **addRecipients re-arm** now uses the FIRST_POLL_DELAY_MS fast-tick pattern (3s) instead of pollInterval (15-60s) — /qurl add views surface as quickly as the initial recipients.
- **Decay tick rate** gets a rate-table comment.
- **Concurrent-tick safety** pinned with a one-line invariant comment on runTick.
- **consumed: true → false drop** documented as intentional (self-destruct semantics).
- **3 new receiver tests** (non-string body.id, access_count=null, consumed: "false") + jest.resetModules isolation for the rate-limit test (was order-dependent).
- **TODO(freshness) comment** on the receiver: signature pins payload integrity but not freshness. Out of scope today; flagged for the next iteration.

**Tests**: 2412 pass. Lint clean.

justin-layerv:
## cr round 3 — addressed (commit 26d0197)

Closing "LGTM modulo finding #1" on the latest review:

**1. (bug) Mid-life `viewCounterDegraded` flip is silent** — fixed. Construction-time path already warns once; addRecipients path was silent. Added symmetric `logger.warn` guarded by `!viewCounterDegraded` so it fires exactly once per monitor + a regression-guard test (`addRecipients() with a missing qurl_id flips viewCounterDegraded AND warns once`).

**2. (tightening) `access_count` integer gate** — was `typeof === 'number' && isFinite && >= 0`, lets through fractional + unsafe-int. Switched to `Number.isSafeInteger` (single check catches both). Added test pinning fractional rejection.

**3. (comment-only) Concurrent-tick render ordering** — analyzed: the `=== 'opened' continue` guard at the flip site already prevents `viewed` from regressing across concurrent ticks. A slow tick whose snapshot pre-dates a fast tick's flip can't double-flip the same qurl_id; in the truly stale case, `changed=false` so no safeEdit fires. The only theoretical residual flicker is DDB eventual-consistency returning newer data to the older tick — sub-ms in practice, self-corrects on next tick. Comment in `runTick` lays this out so the next reader doesn't bolt on a render-generation counter unnecessarily.

Also added a paging-note section to the runbook documenting the known double-page sequence during a >10-min rotation gap (signature_invalid → rate_limited, same root cause) so operators ack the second as the tail of the first.

**Tests**: 2414 pass. Lint clean.

PR #470, opened 2026-05-19
feat(discord): static check for bare DDB reserved-keyword attribute names

### LINKED ISSUE #468: qurl-integrations: catch DDB reserved-keyword regressions at build time (post #467 follow-up)
## Context

#467 fixed a `ValidationException: reserved keyword: consumed` bug that landed in main via #465 and only surfaced when the bot hit real DDB in sandbox. `aws-sdk-client-mock` (unit tests) + the in-memory mock (`qurl-webhook-flow.test.js`) both accept the bad expression — only real DDB enforces reserved-words validation.

cr on #467 surfaced two options for preventing the next instance:

### A. Static check on UpdateExpression / ConditionExpression literals
Cheap. Embed the DDB reserved-words list (it's stable and short) as a constant; walk every `UpdateCommand` / `PutCommand` / `QueryCommand` call site looking for bare attribute names that match. A test or a lint rule fails CI if a bare reserved word slips through.

Pros: zero infra cost, runs in normal CI, catches all current + future call sites.
Cons: requires keeping the reserved-words list in sync (low churn but real).

### B. LocalStack-backed integration suite for the DDB module
Bigger. Stand up LocalStack in CI, point the store at it, run the existing unit tests against real DDB validation.

Pros: catches the whole class of "mock doesn't enforce real-DDB rules" — not just reserved keywords.
Cons: CI runtime + LocalStack reliability cost; another moving piece.

## Recommendation

Start with A. It's a 1-day task, catches the specific class of bug that bit us in #467, and is independent of any LocalStack reliability concerns.

## Acceptance

- Any new code adding a bare reserved-word attribute name to a DDB expression fails CI.
- Existing call sites pass (sanity check that we don't have OTHER latent reserved-word bugs).

🤖 Filed by Claude Code during sandbox validation of #465 / #467.

justin-layerv:
## Summary
Closes #468. #467 fixed a prod-500 caused by `consumed` (a DDB reserved word) appearing bare in an UpdateExpression. This adds a CI-time static check that catches the next instance of that bug class.

- `tests/helpers/ddb-reserved-words.json` — curated subset of [AWS's reserved-words list](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ReservedWords.html). ~50 entries that plausibly conflict with attribute names (full ~573-word list is overkill — no one names an attribute `ABORT` or `INTERSECT`).
- `tests/ddb-reserved-words-static.test.js` — babel-parses every .js under `src/`, traverses to `UpdateExpression`/`ConditionExpression`/`KeyConditionExpression`/`FilterExpression`/`ProjectionExpression` Property nodes, statically evaluates the value (string literal + template + + concat), tokenizes attribute names with a `(?<![#:])\b\w+\b` lookbehind regex (skips already-aliased `#consumed` + placeholders `:c`), and fails CI on any reserved-word match.

## Sanity
Reverted the #consumed alias locally → test failed at the exact `ConditionExpression` + `UpdateExpression` lines with copy-pasteable Fix output. Restored → test passes. Three-state verified.

## Why mock tests didn't catch the original bug
`aws-sdk-client-mock` (unit) + the in-memory mock in `qurl-webhook-flow.test.js` both accept bare reserved words — only real DDB enforces the rule. Standing up a LocalStack-backed integration suite (option B in #468) is more comprehensive but adds CI infra cost; the static check (option A) is the cheap-and-cheerful path that catches the specific class.

## Coverage cap
A third test caps dynamic-string expressions at <5 so a future refactor that turns many literal strings into template strings (with substitutions the static check can't evaluate) doesn't silently shrink coverage.

## Deps
Uses `@babel/parser` + `@babel/traverse` already transitively available via Jest's deps. No `package.json` change.


justin-layerv:
## cr round 1 — addressed (commit e5537ca)

**Bug fixes:**
1. **`size` blind spot** — removed from `EXPRESSION_KEYWORDS`. The keyword exemption ran first, so `SET size = :s` would have silently passed. Trade-off documented in the comment (a future `size(attr)` call would false-positive; zero such usages in the codebase today).
2. **Expanded curated list** with the realistic mid-tail names you flagged: `error`, `hash`, `item`, `items`, `mode`, `read`, `record`, `region`, `text`, `token`, `ttl`, `url`, `write`. Full suite still passes — no false positives in real attribute names.

**Coverage / fragility:**
3. **Detection logic self-test** — added 5 fixture-driven tests that write synthetic .js sources and assert detector behavior directly. Each tests one path (bare flag, alias exemption, + concat, dynamic-detection, `size`-as-reserved). A future refactor of `evalString` / lookbehind regex / `EXPRESSION_KEYWORDS` that silently neuters the check now fails one of these.
4. **`@babel/parser` + `@babel/traverse` in devDependencies** — was transitive via Jest, now declared.

**Nits:**
5. Magic `> 50` floor → `> 0` (walker found anything).
6. Dynamic-ceiling 5 → 8 with a pointer-to-flow-state comment so the next "ceiling tripped" debug isn't open-ended.
7. Renamed "logs (but does not fail on)" → "the ceiling on non-static-literal expressions prevents silent coverage shrink" (no logging happens; just an assertion).
8. Reworded the lookbehind-regex comment to describe the `\b`-treats-`#`-as-boundary trap without referring to a "first version" that isn't in history.

**Tests**: 8 in this suite (5 fixture + 3 full-scan), 2422 total pass.

PR #471, opened 2026-05-20
feat(discord): auto-register qurl-service webhook on boot

### LINKED ISSUE #469: Bot auto-registers qurl-service webhook subscription on boot
## Context

Today, the bot's `qurl.accessed` webhook subscription has to be registered out-of-band via `POST /v1/webhooks` to qurl-service. The runbook (`apps/discord/docs/qurl-webhook-rollout.md`) walks through it as a manual operator step. This bit us in the sandbox rollout — the operator setup completed (table, SSM secret, bot deploy) but no one ran the curl, so qurl-service never knew where to send events, and the view counter sat at `0 viewed / 1 pending` indefinitely.

The original deferral rationale (in the runbook's "operator-burden tradeoffs" section): _"giving the bot a qurl-service admin token widens its blast radius beyond the per-guild `QURL_API_KEY` it has today."_ Reconsidering: a write-scoped token (`qurl:write`) lets the bot manage **its own** webhook subscriptions, not act as a global admin. That's a much narrower blast radius than I framed it as.

## Proposed change

On startup, the bot:
1. Reads `QURL_API_KEY` from SSM (already does this for normal API operations).
2. `GET /v1/webhooks` — does a subscription with my URL already exist?
3. If no → `POST /v1/webhooks` with the bot's URL + `events: ["qurl.accessed"]`. The server returns the generated `secret`.
4. Persist the secret to SSM as `QURL_WEBHOOK_SECRET` (overwriting). If `--ignore-changes = [value]` is set on the SSM resource, this just works.
5. If yes → reconcile: if URL doesn't match (e.g., changed BASE_URL), PATCH; if events list is incomplete, PATCH; otherwise no-op.

Key design choices:
- **Idempotent**: GET-then-create-or-update, never duplicates.
- **Secret rotation owned by the bot**: each subscription has its own secret; rotation is bot-initiated rather than coordinated SSM ↔ subscription edits.
- **Per-environment**: sandbox bot registers sandbox URL with sandbox qurl-service; prod registers prod with prod. The bot already knows its `BASE_URL` from env.

## Acceptance

- Fresh sandbox bot brings up → /qurl send works end-to-end on first invocation, no operator curl required.
- Re-deploying the bot → no duplicate subscriptions.
- Manually deleting the subscription → next bot restart re-creates it.

## Out of scope

- Coordinated rotation: still operator-initiated. `regenerateWebhookSecret` (`POST /v1/webhooks/{id}/secret`) is a manual operation; not on the boot path.
- Multi-bot-instance race: GET-then-create has a TOCTOU; mitigate with a backoff + idempotency check, or just accept that two parallel registrations from a deploy fleet may produce one duplicate that the cleanup step removes.

🤖 Filed by Claude Code after live-debugging the sandbox rollout of #465.

justin-layerv:
## Summary
Closes #469. Bot self-registers its `qurl.accessed` subscription with qurl-service on every boot, eliminating the manual operator curl that gated the entire view-counter feature.

## Why
The sandbox rollout of #465 hit this exact gap: operator missed the manual curl step from `apps/discord/docs/qurl-webhook-rollout.md`, bot ran clean, counter sat at `0 viewed / N pending` forever. Real-time debugging revealed:
1. No subscription = no webhook deliveries = empty `qurl-views` table = counter never advances.
2. Worse, a registration done with the WRONG token (m2m smoke-test token vs the bot's own `QURL_API_KEY`) attached the subscription to the wrong `owner_id` — qurl-service's `PublishEvent(ownerID, ...)` filters by owner, so events from the bot's qURLs were silently dropped.

Self-register makes both gaps impossible: bot uses its own `QURL_API_KEY` (correct owner_id), runs on every boot (no human step to miss).

## Design
- **On boot**: GET subs via the bot's `QURL_API_KEY`. If matches exist, dedupe to a deterministic survivor (oldest by `created_at`, lex-id fallback) — recovery from prior cold-bootstrap duplicate creates. Then:
  - **Reuse path** (existing sub + real SSM secret): skip rotation entirely, return the SSM secret. This is the steady-state path on every restart after the first.
  - **Bootstrap rotate path** (existing sub + no SSM secret, e.g. PLACEHOLDER): rotate the secret first, then PATCH events if drifted. Rotate-first so a stuck PATCH doesn't strand the receiver on PLACEHOLDER.
  - **Create path** (no existing sub): POST `/v1/webhooks` to create.
- **Steady-state multi-replica safety**: every replica reads the same SSM secret on boot and takes the reuse path. The pre-fix rotate-every-boot pattern raced replicas into each other's secrets (last-write-wins ⇒ (N-1)/N replicas held stale).
- **Cold-bootstrap race (open)**: on the very first deploy to a fresh environment, no sub exists yet AND no SSM secret exists yet, so N replicas all POST create concurrently → N duplicates. The dedupe path is RECOVERY (next boot cleans up) — operator runbook pins the first deploy to `desired_count=1` to avoid this entirely. Durable fix is upstream-side idempotency on `(owner_id, url)`, tracked separately.
- **Fire-and-forget on the boot path** so a transient qurl-service hiccup doesn't crash the bot start sequence. Manual curl + ECS force-deploy remains the recovery route.
- **persistSecret callback** (best-effort SSM `PutParameter`, 5s timeout) for oncall visibility. NOT load-bearing — IAM denial / SSM timeout is caught + logged + the in-memory secret is what the receiver verifies against. SDK-agnostic interface (registrar takes a `persistSecret: async (secret) => void` callback; `index.js` wires it to SSM).

## Round-2 cr items (commits 6709d7b, ac3e3b4, 5dc87f3)
- **A** dedupe + survivor selection: oldest-by-created_at with lex-id fallback so concurrent dedupe converges across replicas without coordination
- **B** rotate-then-patch ordering: bootstrap-rotate now runs rotateSecret first; events PATCH wrapped in try/catch so a stuck PATCH doesn't strand the receiver
- **C** pagination cap throws (instead of returning null → silent fallthrough to create-fresh)
- **E** redactSecret recursive (was top-level-only): walks objects + arrays to depth 8
- **F** shared `PLACEHOLDER_SECRET` constant — receiver imports from registrar (was duplicated literal)
- **H** description sliced to 200 chars defense against future server-side length cap
- **I** persistSecret wrapped in 5s timeout
- **L** config setter hardened (closure-capture, not `this`-coupled)
- **M** `reconcileEvents` helper factored out, both branches call it
- **N** `@aws-sdk/client-ssm` require hoisted to top (was lazy)
- Runbook updated: cold-bootstrap single-replica discipline, `QURL_WEBHOOK_SECRET_SSM_PARAM` env-var documentation, IAM grant scope

## Tests
- 36 in `tests/qurl-webhook-registrar.test.js`: create-fresh / rotate-existing / PATCH-on-drift / persist-callback contract / wire-contract guards / dedupe (oldest+lex+concurrent-404+non-404-propagation) / multi-sub middle-match / rotation-survives-PATCH-failure / events-undefined / pagination-cap-throw / fetch-AbortError / recursive-redact / pickSurvivor unit tests / config-mutation seam.
- Full bot suite: 2451 pass.

## Test plan
- [x] `npx jest` — 2451 pass
- [x] Lint clean
- [x] Sandbox: deploy → bot boots, `qURL webhook self-registration complete` logged, subscription registered owner-scoped, counter advances on `/qurl send` end-to-end without manual curl

### DIFF
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -125,6 +125,23 @@ already exists and the SSM secret is populated. The durable upstream
 fix — making `POST /v1/webhooks` idempotent on `(owner_id, url)` —
 is tracked separately.
 
+**Dedupe-recovery boot also needs `desired_count=1`.** If the first
+deploy was accidentally run at `desired_count>1` and produced
+duplicate subscriptions, the *next* boot enters the dedupe-recovery
+path: it deletes the non-survivor duplicates AND force-rotates the
+survivor's secret (the survivor's pre-rotate secret belongs to a
+deleted replica's POST, so SSM almost certainly disagrees). That
+force-rotate runs on every replica concurrently, with the same
+last-write-wins SSM race that the steady-state reuse path was
+designed to avoid. To exit the duplicate state cleanly:
+
+1. Scale to `desired_count=1`.
+2. Confirm one boot completed cleanly (`Webhook subscription
+   reconciled (existing found, bootstrap rotate)` + the warn line
+   for `Webhook subscription duplicates detected — deleting
+   non-survivors`).
+3. Scale back to target desired_count.
+
 ### Persistence env var
 
 The registrar writes the active secret back to AWS Systems Manager
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -97,6 +97,50 @@ SSM-loaded secret is non-empty and non-`PLACEHOLDER`. Steady-state
 restarts reuse the SSM value; only the bootstrap path (no sub yet,
 or seeded `PLACEHOLDER`) rotates.
 
+### Cold-bootstrap of a fresh environment — DEPLOY AT desired_count=1 FIRST
+
+On the very first deploy to a fresh environment, no subscription
+exists yet AND no SSM secret exists yet. If N replicas boot
+concurrently in that state, each replica independently `POST`s
+`/v1/webhooks` and N duplicate subscriptions are created, each with
+a distinct server-generated secret. SSM is last-write-wins so only
+one replica's secret persists; the N-1 surviving duplicate
+subscriptions deliver events to dead secrets and 401 forever until
+manual cleanup.
+
+The registrar's dedupe path (next boot finds the duplicates and
+deletes all but the oldest survivor by `created_at`) is RECOVERY,
+not prevention. To avoid the race entirely on a fresh environment:
+
+1. Scale the bot HTTP service to `desired_count=1` before the first
+   deploy.
+2. Confirm one boot completed cleanly (`qURL webhook self-
+   registration complete` log line + `created` action).
+3. Scale back to the target desired_count.
+
+This procedure is only required on the first deploy of a brand-new
+environment. Every subsequent deploy (rolling, blue/green, redeploy)
+falls into the steady-state `reused` path because the subscription
+already exists and the SSM secret is populated. The durable upstream
+fix — making `POST /v1/webhooks` idempotent on `(owner_id, url)` —
+is tracked separately.
+
+### Persistence env var
+
+The registrar writes the active secret back to AWS Systems Manager
+Parameter Store when `QURL_WEBHOOK_SECRET_SSM_PARAM` is set in the
+bot's environment. The value is the full SSM parameter name (e.g.
+`/discord-bot/sandbox/QURL_WEBHOOK_SECRET`). When unset, the
+registrar runs in-memory-only — usable for local dev but every
+restart rotates the secret because there's no place to read the
+previous one from.
+
+The IAM role for the bot's task needs `ssm:PutParameter` on that
+specific parameter. The bot tolerates `AccessDeniedException`
+gracefully (logs `warn`, continues with in-memory secret) so a
+missing grant degrades to "the secret rotates each boot, but the
+bot still works" rather than a crashloop.
+
 ## What the bot does NOT need
 
 - The full `qurl.accessed` payload's `src_ip` / `user_agent` fields —
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -158,6 +158,23 @@ gracefully (logs `warn`, continues with in-memory secret) so a
 missing grant degrades to "the secret rotates each boot, but the
 bot still works" rather than a crashloop.
 
+**Latent risk if persistence stays broken**: every replica restart
+hits the bootstrap-rotate path (no real SSM secret available), which
+re-introduces the multi-replica rotate race on every redeploy. The
+warn log `Webhook secret persistence failed (auto-register continues
+with in-memory secret only)` is the early-warning signal — alarm on
+it if oncall hasn't built a CloudWatch rule yet.
+
+**Operational note — `description` field staleness**: the
+`description` shown in the qurl-service webhook UI is captured at
+create time (e.g. `Discord bot view counter (region=us-east-2,
+env=sandbox)`) and is NOT reconciled by subsequent boots. If the
+environment is renamed or migrated to a new region, the UI label
+stays stale until the subscription is deleted and recreated.
+Observability-only — the bot keeps working — but operators sweeping
+the qurl-service UI for "which sub belongs to which env" should not
+trust the description as authoritative.
+
 ## What the bot does NOT need
 
 - The full `qurl.accessed` payload's `src_ip` / `user_agent` fields —
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -11,46 +11,32 @@ This rollout replaces the poll with a `qurl.accessed` webhook receiver.
 
 ## Rollout order
 
-Each step blocks the next — do not skip ahead.
-
-1. **`qurl-views` DDB table.** Provisioned by the deploying organization's
-   infrastructure (separate from this repo). Without it, the bot's
-   monitor `BatchGet` returns the empty map and the counter silently
-   stays at `0 viewed / N pending` forever. Confirm with:
+1. **`qurl-views` DDB table.** Provisioned by the deploying
+   organization's infrastructure (separate from this repo). Without
+   it, the bot's monitor `BatchGet` returns the empty map and the
+   counter silently stays at `0 viewed / N pending` forever. Confirm
+   with:
    ```
    aws dynamodb describe-table \
      --table-name <DDB_TABLE_PREFIX>qurl-views \
      --region <env-region>
    ```
-2. **Deploy the bot.** Code mounts `/webhooks/qurl` unconditionally and
-   warns on boot if `QURL_WEBHOOK_SECRET` is unset — that's the operator
-   signal that step 3 hasn't happened yet.
-3. **Set the SSM secret.**
-   ```
-   aws ssm put-parameter \
-     --name "/<project>/QURL_WEBHOOK_SECRET" \
-     --type SecureString \
-     --value "$(openssl rand -hex 32)" \
-     --overwrite \
-     --region <env-region>
-   ```
-   Then restart the task so the secret is picked up:
-   ```
-   aws ecs update-service --service <svc> --force-new-deployment ...
-   ```
-4. **Register the subscription with qurl-service.** The subscription's
-   `secret` field MUST be the same hex string as step 3 — the bot's
-   verifySignature pins bare-hex HMAC-SHA256 over the raw body.
-   ```
-   curl -X POST https://<qurl-service-host>/v1/webhooks \
-     -H 'Authorization: Bearer <qurl-service-admin-token>' \
-     -H 'Content-Type: application/json' \
-     -d '{
-       "url": "https://<bot-host>/webhooks/qurl",
-       "event_types": ["qurl.accessed"],
-       "secret": "<same-hex-as-step-3>"
-     }'
-   ```
+2. **Deploy the bot.** On `isHttp` boot, the bot self-registers its
+   `qurl.accessed` webhook subscription with qurl-service using its
+   own `QURL_API_KEY` (the same key it uses to mint qURLs — so the
+   subscription's `owner_id` matches the qURLs the bot creates).
+   Steady-state reuses the secret already in SSM/env. Bootstrap (no
+   subscription yet, or SSM seeded with `PLACEHOLDER`) creates or
+   rotates the subscription and best-effort writes the secret back
+   to SSM via `ssm:PutParameter`.
+
+   No manual operator curl required. The bot's `qURL webhook self-
+   registration complete` log line confirms the registration on each
+   boot.
+
+   Recovery if auto-register fails: the registrar logs at error
+   level and the bot continues booting; manual fallback is the
+   `curl` step in the [appendix](#appendix-manual-recovery) below.
 
 ## Wire shape (pinned)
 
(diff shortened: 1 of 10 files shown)


justin-layerv:
## cr round 1 — addressed (commit 4d352c2)

### Blocking
1. **Multi-replica HTTP fleet race** — fixed. Registrar now skips rotation when an existing sub is found AND `initialSecret` is non-empty + non-PLACEHOLDER. New `action: 'reused'` return covers steady-state. `index.js` passes `config.QURL_WEBHOOK_SECRET` as `initialSecret`. Bootstrap path (no sub yet OR `PLACEHOLDER`) still rotates once + persists — the race is bounded to that single bootstrap window. Test added: `existing sub + real initialSecret → REUSE (multi-replica safety)`.

4. **`@aws-sdk/client-ssm` missing** — moved to `dependencies`. The lazy `require` would have thrown `MODULE_NOT_FOUND` and `bestEffortPersist` would have logged at error level on every boot.

### Smaller items
3. **URL canonicalization** — `canonicalUrl()` normalizes trailing-slash + case + default-port drift. New test pins the trailing-slash case.
5. **Pagination** — `findExistingSubscription` walks cursor pages, bounded at 50. Test pins cursor-walk match + empty-walk fallthrough.
7. **Secret redaction** — `redactSecret()` strips the `secret` field from error bodies before they land in CloudWatch. Test pins it.
8. **Description placeholder** — explicit `unset` instead of `unknown-region/env` so misconfig surfaces visibly.

### Documented but not coded here
2. **Boot race between server listening and registrar resolving** — acknowledged in the registrar header comment. Inbound webhooks in the brief startup window 401/503; qurl-service retries.
6. **`config.QURL_WEBHOOK_SECRET = result.secret` mutation** — kept as-is. Direct singleton-mutation is the established pattern in this codebase (multiple sites in `config.js`). Refactoring to a setter is more invasive than the cr-fix scope.
9. **Runbook rewritten** — removes the manual SSM-set + curl-register steps that this PR makes unnecessary. Adds "Multi-replica safety" + "Manual recovery" appendix.

### Followups (separate PRs)
- Push-based view notifications via existing SQS bus — the cr's "what about per-monitor RCU cost at scale?" concern. Architecture run past advisor; design lands in a follow-up to this PR. setInterval polling becomes the fallback once push lands.
- Integration test for the in-process `config.QURL_WEBHOOK_SECRET` mutation seam.

**Tests**: 19 in registrar suite (was 11). Full bot suite 2433 pass.

justin-layerv:
## cr round 2 — addressed (commits 6709d7b → 5dc87f3)

Grouped into 3 commits for cr round-3 readability: correctness / defense+minor / tests+docs.

### Blocking items both reviewers flagged
- **Cold-bootstrap split-brain → recovery dedupe**: `findExistingSubscription` → `findExistingSubscriptions` (returns ALL matches). `pickSurvivor` is oldest-by-`created_at` with lex-id fallback so two replicas independently picking the same survivor + DELETEing the same losers converges without coordination (404-on-second is treated as success). Honest framing in the module header and in `qurl-webhook-rollout.md` — this is RECOVERY, not prevention. Runbook now pins first deploy of a fresh environment to `desired_count=1`. Durable upstream fix (idempotent `POST /v1/webhooks` on `(owner_id, url)`) tracked separately.
- **PATCH-blocks-rotate ordering**: bootstrap-rotate branch now runs `rotateSecret` FIRST, then `reconcileEvents` wrapped in try/catch. Stuck PATCH no longer strands the receiver on PLACEHOLDER.
- **Pagination cap throws** instead of returning null. Silent fallthrough to create-fresh would compound duplicates on every restart with a stuck cursor.

### Smaller items
- **Shared `PLACEHOLDER_SECRET`** constant: receiver imports from registrar (was a duplicated `'PLACEHOLDER'` literal).
- **Recursive `redactSecret`**: walks nested objects + arrays up to depth 8 — a nested `data.webhook.secret` or `data[].secret` no longer bypasses the scrub.
- **`QURL_WEBHOOK_SECRET_SSM_PARAM` documented** in `qurl-webhook-rollout.md` (env var name + IAM grant required + `AccessDeniedException` graceful-fallback behavior).
- **`description.slice(0, 200)`**: defense against a future qurl-service-side length cap that would otherwise infinite-retry on 4xx.
- **`@aws-sdk/client-ssm` eager require** at module-top: missing dep now fails at boot, not at first persist.
- **persistSecret 5s timeout**: hung IMDS no longer leaves `.then(setQurlWebhookSecret)` pending forever; bestEffortPersist degrades to in-memory-only.
- **`config.setQurlWebhookSecret` hardened**: closure-based (not `this`-coupled shorthand-method) so a destructure-and-call doesn't silently throw on undefined `this` in strict mode.
- **`reconcileEvents` helper** factored out — both reuse + rotate branches call it.
- **Stringly-typed event literal** removed: `QURL_ACCESSED` now sourced from shared `QURL_WEBHOOK_EVENTS.ACCESSED` constant (single source of truth with the receiver).

### Test gaps closed (registrar suite: 20 → 36)
- duplicate-recovery × 4 (oldest-by-created_at survivor, lex-id fallback, concurrent-DELETE 404, non-404 propagation)
- multi-sub middle-match
- rotation-survives-PATCH-failure (behavior assertion, not ordering)
- events-undefined regression guard
- pagination-cap-throw
- fetch AbortError
- recursive redact × 3
- pickSurvivor unit tests × 4

Full bot suite: 2451 pass. Lint clean.

### Code-review agents (parallel /simplify pass)
Ran the three review angles in parallel: efficiency → no findings; code-reuse → flagged `redact()`-vs-`logger.redact` and `Promise.race+setTimeout` duplication across 6 sites (deferred — both are scope-creep refactors); code-quality → applied 8 of 13 inline (orphaned JSDoc moved, stringly-typed literal extracted, test pinning loosened to behavior assertion, redundant `_configRef` simplified, narration comments collapsed, `op` field added to logged error).

justin-layerv:
## cr round 3 — addressed (commit 074b0da)

### Blocking — cold-bootstrap survivor↔secret mismatch (real catch, thank you)

You're right — the gap was: after dedupe, the survivor's secret is unknown (it was server-generated on a replica we just DELETEd), but SSM has whoever won the persistSecret race (likely a different sub). Pre-fix reuse path would trust the SSM secret against a sub it doesn't belong to → 401 forever.

**Fix**: `wasDedupe = matches.length > 1` flag bypasses the reuse-path guard. Force-rotate the survivor first, then persist that new secret. One-time cost on recovery; subsequent restarts hit the steady-state reuse path normally. Inline comment expanded to document the failure mode.

**New test**: `force-rotates the survivor even when initialSecret is real (closes SSM↔survivor mismatch)` — explicitly passes `initialSecret: 'whsec_was_in_ssm_but_for_wh_b'` and asserts the returned secret is NOT that value, so a future revert of the force-rotate would fail loudly. The 3 existing dedupe tests were updated to expect `action: 'rotated'` (was 'reused').

### BASE_URL trailing-slash double-slash
`.replace(/\/$/, '')` before building bridgeUrl. canonicalUrl can normalize a trailing slash on the pathname tail but not an internal `//`.

### bestEffortPersist err.message defense
Capped to 200 chars before logging. Typical SDK errors are 50-100 chars; a hypothetical multi-KB leak would now be visibly truncated.

### Real-config seam test
Added 2 tests against `require('../src/config')` (and the destructured-and-called form) so a future rename/signature change is caught here, not at boot.

### Stale PR description
Already refreshed in the previous round.

### Non-blocking flagged + deferred:
- **SSMClient construction in closure**: instantiates fresh per call (1-2× per boot in practice). Minor; deferring — not worth a separate commit, and any future "cache the client" attempt risks region-change handling complications.
- **Eager SDK require in non-AWS environments**: documented tradeoff. Local dev pays the cold-load cost but gets boot-time failure if the dep is removed. Leaving as-is per the inline comment.
- **Style nit re: setter outside object literal**: the suggested arrow `(s) => { module.exports.X = s }` is functionally equivalent to what's there. Style preference, no change.

**Tests**: 39 in registrar suite (was 36). Full bot suite 2454 pass. Lint clean.

justin-layerv:
## cr round 4 — addressed (commit 200ac56)

### Substantive — dedupe-recovery race on rotate
Valid catch. When `wasDedupe===true`, the force-rotate runs on every replica concurrently with the same last-write-wins SSM race the steady-state reuse path was built to avoid. Going with option (a) — doc-only fix — since reading the survivor's existing secret server-side isn't available. Runbook now has a "Dedupe-recovery boot also needs `desired_count=1`" subsection with the same scale-down → confirm-one-boot → scale-back procedure as the cold-bootstrap section. Honest framing: dedupe is recovery; recovery has its own constraints. The durable upstream-side idempotency fix would close this whole class — tracked separately.

### Polish
- **Pagination test name**: `returns null` → `returns []` to match the new `findExistingSubscriptions` plural signature.
- **`limit=100` explicit**: GET `/v1/webhooks?limit=100[&cursor=...]` instead of relying on a server default that could silently change. Stale comment updated.
- **`description` not reconciled**: added one-line note on `createSubscription` that drift is observability-only (qurl-service UI label) and not worth a PATCH on every boot.
- **`QURL_WEBHOOK_SECRET_SSM_PARAM` via config.js**: pulled through `config.*` for grep-uniformity with other config reads. `AWS_REGION` stays direct because it matches the existing pattern in event-publisher.js / event-consumer.js / index.js — pulling it through config is a separate cross-cutting cleanup.
- **`mockFetchResponses` helper**: gained a pathname-only fallback so adding `?limit=100` to the registrar's GET path didn't require enumerating every query-param combination in every test. Pagination test was updated to spell exact path-with-query so cursor-walk doesn't loop on the fallback.

### Memory entry (security blast-radius note)
Added a project memory entry: post-PR-471, the bot's `QURL_API_KEY` now manages webhook subscriptions (list/create/PATCH/rotate-secret/DELETE) in addition to qURL minting. Webhook-management is in-scope for any API-key rotation drill — net incremental blast radius is small (key already lets attacker masquerade as the bot for minting) but the failure surface differs (silent telemetry drop vs visible /qurl send symptoms).

**Tests**: 2454 pass (39 in registrar suite). Lint clean.

justin-layerv:
## cr round 5 — addressed (commit 524cc77)

All 5 items + canonicalUrl note fixed:

- **Contract-drift defensive checks**: `createSubscription` + `rotateSecret` now validate the response envelope shape (data + webhook_id + secret) and throw a greppable "contract drift" message instead of an opaque `TypeError: Cannot read property 'X' of undefined`. The boot-log error is now actionable.
- **Reuse-path PATCH try/catch**: matches the rotate branch. Transient 5xx on events PATCH no longer flips the boot log to "self-registration failed" when receiver is already correct via initialSecret.
- **`Array.isArray` guard on GET /v1/webhooks**: a future contract drift returning `data: {...}` instead of `data: [...]` no longer iterates object property values as subscription objects.
- **Description slice at wire boundary**: moved from index.js caller into `createSubscription`. `DESCRIPTION_MAX_LEN = 200` constant exported via the comment for grep. Caller just builds the string.
- **SSM client hoisted to module scope**: `getSsmClient()` helper with lazy first-call construction. Matches the DynamoDBClient pattern. Processes that don't persist still pay zero construction cost.
- **`canonicalUrl` query-string note**: added explicit comment that query strings are preserved (intentional, not folded) — a sub registered with `?ver=1` is a different route than the unparametrized one.

### Deferred (per reviewer's "not blocking" framing)
- Upstream-tracker link in runbook: the durable fix (idempotent `POST /v1/webhooks` on `(owner_id, url)`) lives outside this repo's public surface; runbook keeps the "tracked separately" phrasing rather than fabricate a URL or leak a private-repo reference.
- `ThrottlingException` pageability tuning: would refine the log-level branching in `bestEffortPersist` beyond the `AccessDeniedException`-only split. Defer until oncall actually sees the noise.

### Tests (39 → 44)
- contract-drift on create (missing secret AND missing data envelope)
- contract-drift on rotate (missing secret)
- reuse-path PATCH-failure returns reused (boot-log-doesn't-lie pin)
- description-slice at the wire boundary (500-char input → 200-char wire)
- non-string description treated as empty

Full bot suite: 2459 pass. Lint clean.

justin-layerv:
## cr round 6 — addressed (commit 79738b8)

Reviewer flagged "nothing blocking" — per pr-ready discipline, addressed all items + nits:

- **`op` on pre-response fetch errors**: `callQurlService` wraps `fetch` in try/catch so AbortError, DNS, TLS errors land with `op=GET /v1/webhooks?limit=100` attached. Oncall greps for `op=...` now catch network failures too, not just HTTP-status ones. Test pins this.
- **`redactSecret` regex match**: `SECRET_KEY_RE = /secret/i` matches `webhook_secret`, `signing_secret`, `client_secret`, `secret_key` — qurl-service's error envelope shape isn't pinned and a future field-name change shouldn't bypass the log-leak guard. Test added.
- **Runbook — SSM persistence failure latent risk**: every restart hits bootstrap-rotate path → multi-replica race recurs on every redeploy until IAM is fixed. Documented + alarm hint.
- **Runbook — description field staleness**: captured at create time, not reconciled. Observability-only, but operators should know.

### Deferred (not blocking + outside this PR scope)
- Dedupe-recovery CloudWatch alarm + ECS event trigger: infra-side; runbook discipline is the bridge until that lands.
- `setQurlWebhookSecret` formatting nit: the closure-after-export shape is the safest way to handle destructure-and-call; leaving the formatting trade-off as-is.

### Tests (44 → 46)
- secret-shaped key variants redact (webhook_secret / signing_secret / client_secret)
- pre-response fetch error attaches op (AbortError carries `op=GET /v1/webhooks?limit=100`)

Full bot suite: 2461 pass. Lint clean.

justin-layerv:
## cr round 7 — addressed (commit a56464a)

Reviewer flagged "merge after items 1 + 2" — addressed both + all the non-blocking polish:

### Blocking (1, 2)
- **Brand spelling sweep** in registrar's log messages: all `Webhook subscription ...` / `Webhook secret ...` now prefix with `qURL` to match the index.js wrapper + CLAUDE.md convention.
- **NODE_ENV → region-only**: dropped `env=...` from the description since `NODE_ENV` is conventionally `production` everywhere, rendering `env=production` for both sandbox + prod and defeating disambiguation. Region is the meaningful breakdown; documented why operators use owner_id or URL host for sandbox/prod split.

### Non-blocking polish (all addressed)
- **Promise.all dedupe DELETEs**: parallel since each is independent + 404-on-concurrent is already swallowed per-loser. Saves ~100-200ms × N losers.
- **pickSurvivor mixed-timestamp**: explicit "row with timestamp wins" before lex fallback. Test added.
- **SDK abortSignal**: replaced Promise.race-with-setTimeout with `abortSignal: AbortSignal.timeout(5000)` — the SDK actually cancels the underlying HTTP request now.
- **redactSecret fail-closed at depth cap**: returns `[TRUNCATED]` instead of raw subtree. Test pins it with a 10-deep wrapped secret.
- **err.message slice in outer logger**: 500-char cap defense against a TypeError stack or raw throw with a long message that survived redactSecret.
- **Runbook alarm note** for `qURL webhook self-registration failed` — the higher-severity signal complement to the persistence-failure warn.

### Deferred (informational only)
- **Pagination cap (5000 subs)**: tracked. Upstream-side `?url=<bridgeUrl>` filter would obsolete the cap; for now, the bot's own owner_id is bounded well under 5000.
- **AbortSignal.timeout Node version**: confirmed — package.json `engines.node>=22`, deploy image is Node 22+.

### New tests (46 → 51)
- Cold-bootstrap parametrized (initialSecret undefined / empty / PLACEHOLDER + no existing → 'created')
- pickSurvivor mixed-timestamp (asymmetric created_at, both input orders)
- redactSecret depth-cap fail-closed (10-deep secret can't survive truncation)

Full bot suite: 2466 pass. Lint clean.

justin-layerv:
## cr round 9 — addressed (commit 2faa242)

Apologies for the lag — I was head-down on PR #472 and missed this round. All items now addressed:

### Real
- **redactSecret string-branch gap**: `callQurlService` non-JSON path preserved the body as a raw string, and redactSecret only walks object keys — so a secret embedded in a plaintext stack trace or proxy-injected HTML would reach CloudWatch via QurlServiceError. Added a 32+-hex-char substring scrub on the string branch before stringifying.

### Polish
- **MAX_PAGINATION_PAGES named constant** (was literal `50`) — grep-parity with the other registrar caps.
- **reconcileEvents inclusion-only policy documented**: extra event types (operator-added qurl.created/.revoked) are tolerated because the receiver 200-ignores them. Trimming would fight a legitimate operator action.
- **region=unknown** (was `unset`) — reads as a missing-value diagnostic in the qurl-service UI rather than a configured value.
- **`config.QURL_WEBHOOK_SECRET` consumer warning**: destructuring captures the pre-rotation value (no rebind). Comment now warns against destructure-import.
- **PLACEHOLDER boot-race test**: now uses `config.setQurlWebhookSecret(...)` instead of direct mutation. Single canonical mutation path = future rename caught loudly.

### Nit
- **getSsmClient wrapper flattened**: persistSecret runs at most once per boot, so the lazy-init amortization the comment implied wasn't real. Direct `new ssmSdk.SSMClient(...)` at the single call-site.

### Related PR
Worth noting: PR #472 (one-shot Lambda registrar) supersedes the boot-time wiring this PR introduces. PR #471 stays as the in-app race-aware baseline; #472 removes the wiring entirely and moves registration to a deploy-time Lambda — eliminating the cold-bootstrap race rather than recovering from it. Merge order: this PR first, then #472.

Full bot suite: 2467 pass. Lint clean.

justin-layerv:
## Superseded by #472 — recommend close-without-merge

Per ops direction for tonight's rollout, the canonical webhook-registration architecture is the one-shot Lambda in **#472** (which was retargeted to main and now stands alone).

**Why this PR is being deferred:**
- Auto-register-on-boot has a cold-bootstrap race: N HTTP replicas concurrently POSTing `/v1/webhooks` create N duplicate subscriptions, each with a distinct server-generated secret. SSM last-write-wins picks one; the surviving N-1 deliver to dead-secret URLs and 401 inbound webhooks. The in-app dedupe + force-rotate recovers on next boot but leaves a duplicate-traffic window where qurl-service emits N webhooks per event, N-1 of which 401 and pile up against the bad-sig rate limiter (30/60s) — at modest traffic the receiver returns 429s on legit clients.
- The Lambda in #472 is single-instance by construction, race-free, with fail-fast deploy semantics (deploy itself fails if registration fails).

**This PR's work is NOT lost.** #472's branch was built on top of this one — every line of the shared registrar library + receiver + tests (the ~700-line polished-through-10-rounds work) is reused by the Lambda. Only the boot-time wiring is dropped.

**Recommended action**: close without merge. If you'd prefer to merge for completeness of audit trail, the merge order would need to be this PR first, then #472 (which removes the wiring this PR introduces). Either way, the production deploy target is #472.

justin-layerv:
closed

PR #472, opened 2026-05-20
feat(discord): move webhook registration to one-shot Lambda

justin-layerv:
## Summary

Supersedes the boot-time auto-registration wiring from #471 with a one-shot Lambda invoked once per deploy. Eliminates the cold-bootstrap race that #471's dedupe + force-rotate could only RECOVER from after-the-fact.

**Merge order**: #471 first, then this PR. This PR's base is currently `feat/webhook-auto-register-on-boot`; after #471 merges to main, GitHub will re-target the base automatically and the diff shown will be the Option-C-only delta.

## Why a Lambda (not on-boot)

#471's auto-register-on-boot architecture has an unavoidable cold-bootstrap race:
1. Fresh environment, N HTTP replicas concurrently `POST /v1/webhooks`
2. qurl-service creates N duplicate subscriptions, each with a distinct server-generated secret
3. SSM's `PutParameter` is last-write-wins → SSM converges to ONE replica's secret
4. The surviving N-1 subscriptions deliver to dead-secret URLs forever (until next-boot dedupe)
5. During the duplicate window: qurl-service emits N webhooks per event → N-1 fail HMAC → 401 → bump the bad-sig rate limiter (30 in 60s) → flap legit traffic into 429s

#471's in-app mitigation (dedupe-on-next-boot + force-rotate of survivor) RECOVERS but doesn't PREVENT. The runbook required a `desired_count=1` first-deploy discipline that operators can miss the same way they missed the manual-curl step #471 was supposed to retire.

This PR moves registration to a single-instance Lambda triggered by `aws_lambda_invocation` in Terraform. One execution per deploy, fail-fast on error (deploy itself fails, no half-registered state), no in-app coordination needed.

## Design

- **`apps/discord/lambda/webhook-registrar/index.js`** — Lambda handler. Reuses `ensureWebhookSubscription` + `buildSsmPersistSecret` from the registrar library #471 introduced. Reads qurl-service API key from SSM by name (value never appears in invocation logs), calls the registrar, persists rotated secret to SSM. Returns `{webhookId, action}` — secret is NOT echoed (kept out of Terraform invocation logs).
- **Bot HTTP-tier boot path**: removes the `ensureWebhookSubscription` call, `ssmSdk` eager require, `getSsmClient` helper, `persistSecret` callback, `.then(setQurlWebhookSecret)` wiring, and `.catch` error formatter. ~75 lines net removed from `src/index.js`.
- **`src/config.js`**: removes `setQurlWebhookSecret` and `QURL_WEBHOOK_SECRET_SSM_PARAM`. The bot is no longer the writer; the Lambda owns both.
- **Receiver unchanged**: `routes/qurl-webhook.js` still reads `QURL_WEBHOOK_SECRET` from env (now sourced from SSM by the Lambda at deploy time).

## Rotation

Re-invoke the Lambda (manual, scheduled EventBridge, or operator Terraform) → SSM updated → force-redeploy bot ECS. The Lambda's `reused` semantics make re-invocations idempotent: if SSM secret matches the existing sub, no `POST /secret` fires.

## Tests

- Removed: 2 `setQurlWebhookSecret` real-config integration tests (the in-process mutation seam no longer exists).
- Updated: config-mutation seam test is now a return-shape pin (Lambda flow: registrar returns secret → persistSecret writes to SSM → bot reads from env at next deploy).
- **New**: 15 Lambda handler tests covering input validation (6 cases), cold-bootstrap, steady-state reuse, SSM API-key-missing failure propagation, qurl-service error propagation. Asserts the secret is NOT echoed in the Lambda response.

Full bot suite: 2480 pass. Lint clean.

## Follow-up

Companion `qurl-integrations-infra` PR for the Lambda deploy + IAM role + `aws_lambda_invocation` trigger config. The Lambda code itself ships here; the infra wiring lives in the infra repo.

## Test plan
- [x] `npx jest` — 2480 pass (15 new Lambda tests)
- [x] Lint clean
- [ ] Sandbox: deploy companion infra PR → `terraform apply` invokes Lambda once → SSM populated → bot redeploy reads new secret from env → `/qurl send` end-to-end with view counter advancing

### DIFF
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -125,6 +125,23 @@ already exists and the SSM secret is populated. The durable upstream
 fix — making `POST /v1/webhooks` idempotent on `(owner_id, url)` —
 is tracked separately.
 
+**Dedupe-recovery boot also needs `desired_count=1`.** If the first
+deploy was accidentally run at `desired_count>1` and produced
+duplicate subscriptions, the *next* boot enters the dedupe-recovery
+path: it deletes the non-survivor duplicates AND force-rotates the
+survivor's secret (the survivor's pre-rotate secret belongs to a
+deleted replica's POST, so SSM almost certainly disagrees). That
+force-rotate runs on every replica concurrently, with the same
+last-write-wins SSM race that the steady-state reuse path was
+designed to avoid. To exit the duplicate state cleanly:
+
+1. Scale to `desired_count=1`.
+2. Confirm one boot completed cleanly (`Webhook subscription
+   reconciled (existing found, bootstrap rotate)` + the warn line
+   for `Webhook subscription duplicates detected — deleting
+   non-survivors`).
+3. Scale back to target desired_count.
+
 ### Persistence env var
 
 The registrar writes the active secret back to AWS Systems Manager
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -123,6 +123,14 @@ finds the existing sub, sees the SSM secret matches, returns `reused`).
   reconciled by subsequent Lambda invocations. Region/env rename
   leaves the qurl-service UI label stale until the subscription is
   recreated. Observability-only — the bot keeps working.
+- **`bridgeUrl` change → orphaned old subscription**: if `BASE_URL`
+  changes (domain migration, env rename, https/host swap), the
+  Lambda's `canonicalUrl` match against the existing sub fails and a
+  NEW subscription is created at the new URL. The OLD sub remains
+  registered with qurl-service and keeps trying to deliver to the
+  defunct URL until manually deleted. Recovery: run the manual
+  `DELETE /v1/webhooks/{old_id}` curl from the appendix below for the
+  old sub after confirming the new one is healthy.
 - **API-key blast radius**: the Lambda's `QURL_API_KEY` can list /
   create / PATCH / rotate-secret / DELETE webhook subscriptions in
   addition to minting qURLs. Factor into rotation drills.
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -97,6 +97,50 @@ SSM-loaded secret is non-empty and non-`PLACEHOLDER`. Steady-state
 restarts reuse the SSM value; only the bootstrap path (no sub yet,
 or seeded `PLACEHOLDER`) rotates.
 
+### Cold-bootstrap of a fresh environment — DEPLOY AT desired_count=1 FIRST
+
+On the very first deploy to a fresh environment, no subscription
+exists yet AND no SSM secret exists yet. If N replicas boot
+concurrently in that state, each replica independently `POST`s
+`/v1/webhooks` and N duplicate subscriptions are created, each with
+a distinct server-generated secret. SSM is last-write-wins so only
+one replica's secret persists; the N-1 surviving duplicate
+subscriptions deliver events to dead secrets and 401 forever until
+manual cleanup.
+
+The registrar's dedupe path (next boot finds the duplicates and
+deletes all but the oldest survivor by `created_at`) is RECOVERY,
+not prevention. To avoid the race entirely on a fresh environment:
+
+1. Scale the bot HTTP service to `desired_count=1` before the first
+   deploy.
+2. Confirm one boot completed cleanly (`qURL webhook self-
+   registration complete` log line + `created` action).
+3. Scale back to the target desired_count.
+
+This procedure is only required on the first deploy of a brand-new
+environment. Every subsequent deploy (rolling, blue/green, redeploy)
+falls into the steady-state `reused` path because the subscription
+already exists and the SSM secret is populated. The durable upstream
+fix — making `POST /v1/webhooks` idempotent on `(owner_id, url)` —
+is tracked separately.
+
+### Persistence env var
+
+The registrar writes the active secret back to AWS Systems Manager
+Parameter Store when `QURL_WEBHOOK_SECRET_SSM_PARAM` is set in the
+bot's environment. The value is the full SSM parameter name (e.g.
+`/discord-bot/sandbox/QURL_WEBHOOK_SECRET`). When unset, the
+registrar runs in-memory-only — usable for local dev but every
+restart rotates the secret because there's no place to read the
+previous one from.
+
+The IAM role for the bot's task needs `ssm:PutParameter` on that
+specific parameter. The bot tolerates `AccessDeniedException`
+gracefully (logs `warn`, continues with in-memory secret) so a
+missing grant degrades to "the secret rotates each boot, but the
+bot still works" rather than a crashloop.
+
 ## What the bot does NOT need
 
 - The full `qurl.accessed` payload's `src_ip` / `user_agent` fields —
--- a/apps/discord/docs/qurl-webhook-rollout.md
+++ b/apps/discord/docs/qurl-webhook-rollout.md
@@ -158,6 +158,23 @@ gracefully (logs `warn`, continues with in-memory secret) so a
 missing grant degrades to "the secret rotates each boot, but the
 bot still works" rather than a crashloop.
 
+**Latent risk if persistence stays broken**: every replica restart
+hits the bootstrap-rotate path (no real SSM secret available), which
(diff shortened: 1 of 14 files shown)


justin-layerv:
## cr round 1 — addressed (commit 283ec12)

### Real bug — getSsmClient region cache
You're right — `client.config.region` is a Provider<string> in SDK v3 (function returning Promise), so `!==` against a string was always truthy + the cache never invalidated. Fixed by tracking the cached region in a separate closure variable. Added tests pinning warm-invocation reuse + region-change invalidation.

### Polish
- **Tightened regex**: `/API key.*ParameterNotFound|null/` → `/SSM parameter.*returned null/`. The old form matched the bare `null` token and passed for the wrong reason.
- **`freshHandler()` removed**: module is required once at top, `handler` used directly everywhere. Misleading name gone.
- **IAM scope tightened in header**: `logs:*` → `logs:CreateLogGroup` + `logs:CreateLogStream` + `logs:PutLogEvents`.
- **`bridgeUrl` trailing-slash heads-up**: documented in Lambda header that the caller (Terraform) is responsible — the registrar's `canonicalUrl` handles trailing-slash drift BETWEEN subscriptions but not an internal `//webhooks/qurl`. The infra-repo PR should `replace(/\/$/, "")` on `BASE_URL` before building `bridgeUrl`.

### Tests (15 → 17)
- `warm invocation returns the same client instance for the same region` — pins the closure-based cache positive case.
- `returns a new client instance when the region changes` — pins the now-working invalidation.

Full suite: 2482 pass. Lint clean.

justin-layerv:
## cr round 2 — addressed (commit c7480bd)

Apologies for the lag — re-armed monitoring caught two parallel cr rounds at once.

- **Header docstring**: `apiKeySsmParamName` now listed (was only in `validateInput`). Caller spec is single-source again.
- **`validateInput` → `validateAndNormalizeInput`**: defensive `bridgeUrl` trailing-slash strip. Returns new event (no input mutation).
- **Package.json + README**: documents the only runtime dep (`@aws-sdk/client-ssm`) and the bundling boundary (handler `require`s shared modules from `../../src/`; infra-repo packaging step esbuilds them in).
- **Logger safety in Lambda**: README confirms `src/logger.js` is sync `console.*` with no env-vars-at-require-time, safe to require in a Lambda context.
- **Error wording**: `"returned null"` → `"returned empty or missing (ParameterNotFound or zero-length value)"` — accurate for both branches.
- **Cross-link nit**: `index.js` comment block now points at lambda/index.js, README, and rollout doc — survives a future lambda-dir rename.

### Test gaps closed (17 → 21)
- `rotates the existing sub when SSM holds the PLACEHOLDER sentinel` — closes the cold-bootstrap-after-terraform-seed branch.
- `strips a trailing slash on bridgeUrl` — pins the defensive normalization.
- `throws the same error when SSM API key value is present but empty` — pins the empty-string vs missing-parameter equivalence.

### Informational items (no action needed)
- `getSsmClient` cache test pinning surprising behavior (us-east-2 → us-west-2 → us-east-2 yields 3 instances): documented + intentional. Multi-region rotation isn't a real workflow today.
- `_internals` exposing `getSsmClient`: acceptable trade-off; the alternative is implicit-only-via-handler testing.

Full bot suite: 2485 pass. Lint clean.

justin-layerv:
## Retargeted to main + carry-forward of #471 round-10 + no-PLACEHOLDER

Per ops direction (tonight's prod rollout, no sentinel literals):

**This PR is now self-contained against main** — was previously based on `feat/webhook-auto-register-on-boot` (#471's branch). The diff now includes:
- The shared registrar library + receiver + receiver tests (originally from #471, polished through 10 rounds of cr there)
- The Lambda handler + Lambda tests + bundling README
- The boot-time wiring removed (Lambda owns registration; bot is pure receiver)
- **PLACEHOLDER sentinel dropped entirely**: empty/missing SSM secret is the unconfigured state, no literal sentinel. Receiver 503s on empty; Lambda creates the SSM param on first invocation.
- #471 round-10 cr items applied to the shared registrar:
  - `reconcileEvents` `Array.isArray` guard against a future `events: "string"` contract drift
  - `bestEffortPersist` log-level pinning tests (AccessDeniedException → WARN, anything else → ERROR — the alarm-tier distinction was previously enforced only by reading the code)
  - Rotate-after-DELETE Promise.all sequencing test (rotation must not fire when a non-404 DELETE rejects)

**#471 status**: superseded. Its in-app boot-time auto-register has a cold-bootstrap race that this PR avoids by construction. Closing #471 unmerged is reasonable; alternatively, merge it first and then this PR (which removes the boot wiring it introduces) — operator's call.

**Companion infra-repo PR needed**: Terraform `aws_lambda_invocation`, IAM role with scoped `ssm:GetParameter`/`PutParameter` + `logs:CreateLogGroup`/`CreateLogStream`/`PutLogEvents`, esbuild packaging for the handler + shared modules.

### Tests
- 2487 pass (full bot suite)
- Lint clean
- Includes 21 Lambda handler tests + 65+ in the registrar suite

justin-layerv:
## cr round 3 — addressed (commit 67d97f5)

### 🔴 Critical bug fixed
You're absolutely right — `bestEffortPersist` was correct for the on-boot caller (PR #471's use case, where the in-memory secret keeps the bot working until next deploy) but completely wrong for the Lambda where persistence IS the deliverable. Half-registered state would have shipped silently.

**Fix**: Lambda now omits `persistSecret` on the `ensureWebhookSubscription` call and persists directly via `buildSsmPersistSecret()(secret)`, letting any throw propagate. Two new tests pin both the throw-on-persist-failure and the reuse-path-no-op behaviors.

Also: only persist when `action !== 'reused'` — re-invoking the Lambda when nothing has changed should be free (no SSM write, no qurl-service rotate).

### Other items
- **Promise.all → Promise.allSettled** on the dedupe DELETE loop. Collects all results, throws the first rejection explicitly. Cleaner single-failure stack trace, no unhandled-rejection risk on siblings.
- **IAM doc tightened in runbook**: `logs:*` → `logs:CreateLogGroup` + `logs:CreateLogStream` + `logs:PutLogEvents`. Aligns with the Lambda header so infra-repo PR mirrors the tight set.

### Deferred (nits, not blockers given tonight's rollout pressure)
- `const normalized = validateAndNormalizeInput(event)` vs the eslint-disabled shadow — style preference, behavior identical.
- `apps/discord/src/config.js` trailing blank line in module.exports — lint accepts it as-is.
- `templates.test.js` mock breadth — known transitive-require smell, separate cleanup.
- `apps/discord/tests/qurl-webhook.test.js` try/finally vs `jest.replaceProperty` — current pattern works; jest.replaceProperty is an unrelated refactor.
- `exports._internals` placement before `exports.handler` — function-hoisted, conventional but cosmetic.

Tests: 2489 pass (added 2 to the Lambda suite). Lint clean.

justin-layerv:
## cr round 4 — addressed (commit 8182306)

All 4 optional items polished (none were merge-blockers per reviewer):

- **SSM trust assumption** documented near the `reused` action in `qurl-webhook-registrar.js`. Operator recovery (clear SSM → re-invoke Lambda → bootstrap-rotate) is now greppable from the code.
- **`https://` defensive check** on `apiEndpoint` + `bridgeUrl` in `validateAndNormalizeInput`. Catches Terraform input typos at the boundary instead of failing late. 3 new parametrized tests pin the rejection.
- **`bridgeUrl` change → orphaned old subscription note** in runbook's Operational notes. Symmetric with the existing description-staleness callout. Recovery via manual DELETE in the appendix.
- **Stray blank line** in `config.js`'s module.exports removed.

Tests: 2492 pass. Lint clean. Ready for admin-merge once CI re-validates the polish commit.

justin-layerv:
## cr round 5 — addressed (commit 35087aa); admin-merging when CI greens

All 5 polish items applied:

- **Description warn at boundary** (DESCRIPTION_WARN_LEN=200) — operators get a structured warn log if oversized.
- **Single-instance comment clarified** — race-freedom is a Terraform-level invariant, not a Lambda-level one. Recommends reserved-concurrency=1 in the infra repo as belt-and-suspenders.
- **Empty-secret guard** in createSubscription + rotateSecret — `data.secret.length > 0`. An empty string would let the receiver verify against a zero-length HMAC key (any signature matches '' digest); trivially bypassable.
- **`const input = ...`** replaces the shadowed `event = ...` + eslint-disable. All downstream reads use `input.*`.
- **New tests pinning no-secret-echo** on `rotated` AND `reused` paths (previously only `created`).

Tests: 2494 pass. Lint clean. Will admin-merge per fleet `enforce_admins=false` policy once CI confirms green on this commit.

PR #473, opened 2026-05-20
fix(discord): strip Terraform-seeded PLACEHOLDER in Lambda handler

justin-layerv:
## Summary

Defensive strip of the Terraform-seeded `"PLACEHOLDER"` SSM sentinel in the webhook-registrar Lambda handler, before passing `initialSecret` to the shared registrar library.

## Why

PR #472 dropped the `"PLACEHOLDER"` sentinel check from the bot's receiver (per ops direction "no placeholders"). The companion infra PR ([qurl-integrations-infra#752](https://github.com/layervai/qurl-integrations-infra/pull/752)) still has to seed `aws_ssm_parameter.bot["QURL_WEBHOOK_SECRET"]` with a literal value (`"PLACEHOLDER"`) — SSM `SecureString` doesn't allow empty values, and the ARN must exist before the bot's task-def `secrets[].valueFrom` reference is created.

The Terraform `depends_on = [aws_lambda_invocation.webhook_registrar]` chain on `aws_ecs_service.bot_http` covers the common case: Lambda runs → SSM populated → ECS launches with the real secret. But an edge case lurks:

- Operator manually creates a qurl-service webhook subscription out-of-band (during a manual-recovery → automated-takeover transition)
- SSM still holds `"PLACEHOLDER"` (Lambda hasn't run yet)
- Lambda fires, reads `"PLACEHOLDER"`, `ensureWebhookSubscription` sees `initialIsRealSecret = true` (length > 0), takes the REUSE path, returns `"PLACEHOLDER"` as the secret
- Lambda persists `"PLACEHOLDER"` back to SSM via strict-persist
- Bot tasks read `"PLACEHOLDER"` from env and verify webhooks against it — every inbound 401s

## Fix

One-line strip at the Lambda's SSM-read site:

```js
const rawSsmSecret = await readSsmSecureString({ ... });
const initialSecret = rawSsmSecret === TERRAFORM_SEEDED_PLACEHOLDER ? null : rawSsmSecret;
```

Sentinel value defined as a top-level `TERRAFORM_SEEDED_PLACEHOLDER` constant for greppability. The shared library stays untouched — `null` makes `initialIsRealSecret` false, forcing the bootstrap-rotate path.

## Tests

2 new Lambda tests (29 → 31):
- `strips PLACEHOLDER and rotates when an existing sub is found (cold-bootstrap recovery)` — explicit assertion that the persisted SSM value is a fresh rotated secret, NOT the literal `"PLACEHOLDER"`.
- `strips PLACEHOLDER and creates fresh when no existing sub (fresh-env first apply)` — confirms strip + create-path interaction.

Full bot suite: 2504 pass. Lint clean.

## Test plan
- [x] `npx jest` — 2504 pass
- [x] Lint clean
- [ ] Sandbox: after qurl-integrations-infra#752 applies, the Lambda strip is exercised on `terraform apply` (sandbox SSM already has a real `whsec_*` value from earlier setup, so the strip is a no-op there; the rotation-recovery test path is what matters for prod first-apply)
- [ ] Companion infra PR linked: layervai/qurl-integrations-infra#752

### DIFF
--- a/apps/discord/lambda/webhook-registrar/index.js
+++ b/apps/discord/lambda/webhook-registrar/index.js
@@ -101,6 +101,17 @@ const {
 // Lambda is the sole place that knows about this literal. Stripping
 // to null here lets the registrar library's initialIsRealSecret
 // guard correctly treat it as "no real secret yet, bootstrap-rotate."
+//
+// TODO(infra-sentinel-sync): the literal "PLACEHOLDER" is also the
+// seed value for `aws_ssm_parameter.bot["QURL_WEBHOOK_SECRET"]` in
+// qurl-integrations-infra/qurl-bot-discord/terraform/main.tf. If
+// infra ever renames the sentinel (e.g., "REPLACE_ME"), update here
+// in lockstep — otherwise the strip silently no-ops and the original
+// failure mode (bot reads literal as real secret → 401s every
+// webhook) returns. `git grep TODO(infra-sentinel-sync)` across both
+// repos finds every site that needs the lockstep edit. The same
+// marker appears in apps/discord/src/boot-requirements.js for the
+// GOOGLE_MAPS_API_KEY sentinel parallel case.
 const TERRAFORM_SEEDED_PLACEHOLDER = 'PLACEHOLDER';
 
 // SSM client at module scope — Lambda reuses the same execution
--- a/apps/discord/lambda/webhook-registrar/index.js
+++ b/apps/discord/lambda/webhook-registrar/index.js
@@ -94,6 +94,15 @@ const {
   buildSsmPersistSecret,
 } = require('../../src/qurl-webhook-registrar');
 
+// Terraform-seeded sentinel value for the QURL_WEBHOOK_SECRET SSM
+// parameter — see qurl-integrations-infra:qurl-bot-discord/terraform/
+// main.tf's `aws_ssm_parameter.bot`. The bot's receiver dropped its
+// own PLACEHOLDER-sentinel check in qurl-integrations#472, so the
+// Lambda is the sole place that knows about this literal. Stripping
+// to null here lets the registrar library's initialIsRealSecret
+// guard correctly treat it as "no real secret yet, bootstrap-rotate."
+const TERRAFORM_SEEDED_PLACEHOLDER = 'PLACEHOLDER';
+
 // SSM client at module scope — Lambda reuses the same execution
 // context across consecutive invocations within the same container
 // lifetime, so reusing the client is a meaningful perf win for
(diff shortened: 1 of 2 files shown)


justin-layerv:
## cr round 2 — addressed (commit aff1ae4)

Reviewer flagged "ship it" with one actionable tightening + 3 informational notes. The actionable item:

- **Variant test now explicitly pins `PutParameterCommand.commandCalls.length === 0`** on the reuse path. Previously the no-write was verified implicitly (via the absent mock); the explicit assertion catches a future refactor that re-introduces steady-state writes.

### Acknowledged informational items (no code change)
- **`JSON.stringify(result).not.toContain('PLACEHOLDER')`** is symbolic — `result` is `{webhookId, action}`, neither plausibly contains the sentinel. Kept for parity with the secret-no-echo discipline.
- **Lowercase `"placeholder"` operator-typo silent footgun** — symmetric with `boot-requirements.js:341`'s exact-match approach and explicitly documented in the test comment. Recovery procedure (clear SSM → re-invoke Lambda) handles it.
- **Comment overlap between constant doc and handler comment** — style judgment; this file leans verbose, kept consistent.

Tests: 33 passed. Lint clean.

justin-layerv:
## cr round 3 — addressed (commit 7fd0097)

Reviewer says LGTM; addressed both actionable polish items:

- **`'PLACEHOLDER\n'` variant added** to the parametric test — most plausible accidental variant if SSM is ever populated via shell pipe (`echo PLACEHOLDER | aws ssm ...`).
- **Dropped the unused `POST /secret` mock** from the variant test. The reuse path never hits rotate, so the mock was noise. Now an unmocked fetch would throw if a future regression unexpectedly took the rotate path — better failure mode.

### Acknowledged, no change
- Shared sentinel constant — explicitly the right call to leave duplicated per reviewer.
- Symbolic JSON.stringify assertion — kept as belt-and-suspenders.
- Comment density — style judgment, consistent with file's voice.
- Fat-finger silent-regression (e.g., a TF refactor that lowercases sentinel) — covered by `TODO(infra-sentinel-sync)` cross-repo grep marker; can't auto-fix without breaking the exact-match contract.

Tests: 34 in Lambda suite (29 → 34 cumulative). Lint clean.

justin-layerv:
Closing — seeding `QURL_API_KEY` directly in SSM upfront (operator action) makes the placeholder strip unnecessary. The Lambda never sees `PLACEHOLDER` on the load-bearing path when the API key is real before apply runs.

The placeholder strip was defending against an operator-out-of-band-sub creation edge case that the deploy flow no longer encounters. Closing rather than carrying defensive code for a now-impossible scenario.

Companion infra PR: layervai/qurl-integrations-infra#752 ships the Lambda registrar.

PR #474, opened 2026-05-20
feat(discord): add SQS view-update push for sub-second view counter

justin-layerv:
## Why

The Discord bot's `/qurl send` link-status monitor polls qurl_views DDB every 15–60 s. View counters lag the actual click by up to a minute — fine for the soak window but a flat experience for anyone watching the monitor in real time. This PR adds an SQS push path so view events render within ~1 s of the qurl.accessed webhook hitting the bot.

## What

Three new modules + flag-gated wiring across config / boot-requirements / webhook receiver / index boot / `monitorLinkStatus`:

- **`view-update-publisher.js`** — fire-and-log SQS `SendMessage` from `routes/qurl-webhook.js` after `recordQurlView` returns `result === 'recorded'` (real new view, not per-event dedup replay). Process-singleton; same drain discipline as `event-publisher.js`.
- **`view-update-consumer.js`** — long-poll consumer in the **HTTP tier** (same process that owns the webhook receiver + live `monitorLinkStatus` instances). Parses each envelope, dispatches to the registry by `qurl_id`, then deletes the message regardless of dispatch outcome.
- **`view-update-registry.js`** — process-local `Map<qurl_id, Set<callback>>`. Live `monitorLinkStatus` instances register a SHARED callback against every tracked `qurl_id` on construction and on `addRecipients`; the callback mutates `linkStatus` from the envelope payload and calls `safeEdit` directly (no DDB BatchGet — render is envelope-driven).
- **`view-update-handler.js`** — factory for the monitor's view-update callback closure. Extracted so the state matrix (stopped / degraded / accessCount<=0 / status==='opened' idempotency / pending→0 → onAllDone) is unit-testable without a full monitor closure.

**Silent-drop-on-miss is LOAD-BEARING.** With N=2 HTTP replicas in sandbox today, any one consumer has a 1/N chance of being the right replica for a given monitor — the polling path in `monitorLinkStatus` (unchanged) catches the misses. SQS push is purely a latency optimization on the hit half.

## Envelope

\`\`\`json
{
  "qurl_id": "qrl_xxxx",
  "access_count": 42,
  "consumed": false,
  "event_id": "evt_xxxx",
  "published_at_ms": 1739462812345
}
\`\`\`

## Architecture flow

\`\`\`
qurl-service → POST /webhooks/qurl → recordQurlView (DDB) ─┬→ publisher.publish() ─→ SQS ─┐
                                                          │                              │
                                                          └→ 200 ──→ qurl-service        │
                                                                                          │
HTTP tier   ─→ consumer.pollLoop() ──→ registry.dispatch() ──→ handleViewUpdate(envelope)
                                       (hit on this replica)
                                       (silent-drop otherwise — polling fallback covers)
\`\`\`

## Flag-gated rollout (3 steps)

1. **This PR merges with flag OFF** (\`ENABLE_VIEW_UPDATE_PUSH=false\`). The publisher no-ops on \`publish()\` when not started; the consumer doesn't start; the registry is empty. Safe in any deploy.
2. **\`qurl-integrations-infra\` companion PR** provisions the SQS queue (\`qurl-bot-discord-view-updates-{env}\`), the IAM grants (\`sqs:SendMessage\` on the HTTP task role, \`sqs:ReceiveMessage\` + \`sqs:DeleteMessage\` on the worker task role), and the env var injection on both task defs.
3. **Flag flip** in sandbox.tfvars → verify the hit rate (LogsInsights query in the rollout runbook) → prod.

## GC discipline

Callbacks pin monitor closure state (\`linkStatus\` Map, \`runTick\`, etc.). \`monitorLinkStatus.stop()\` now unregisters every callback it registered. Mirrors the existing closure-mutable-rebind discipline for \`interaction\`/\`qurlLinks\`.

\`addRecipients\` extends both the tracked set AND the registry — the existing comment block ("seed linkStatus AND trackedQurlIds together") generalizes to the registry without any changes to the policy.

## Tests

| File | Coverage |
|------|----------|
| \`view-update-registry.test.js\` | register/unregister round-trip; dispatch hit + silent-drop; callback-throw isolation; self-unregister mid-iteration; GC of empty entries; input validation. |
| \`view-update-publisher.test.js\` | start/stop guards (flag, queue URL); idempotent re-start; pre-start drop; envelope shape + SendMessage call; \`consumed\` coercion; invalid-qurlId rejection; fire-and-log failure path. |
| \`view-update-consumer.test.js\` | start/stop guards; \`processMessage\` validation (JSON, qurl_id, access_count); \`pollOnce\` delete-after-process even on silent-drop; AbortError handling. |
| \`boot-requirements.test.js\` | \`missingViewUpdatePushKeys\` parity with \`missingEventShipperKeys\`. |

- 112 new/touched tests, all passing.
- Full bot suite: 2539/2539 passing.
- \`eslint --max-warnings 0\` clean.

## Risk

- Flag default off → no behavior change without operator action.
- Polling path untouched → correctness primitive preserved.
- New modules are isolated (no imports back into existing modules except a registry import in \`commands.js\` that is gated on the flag).
- Webhook receiver hot path adds one synchronous \`if (result === 'recorded') publish(...)\` call; \`publish()\` returns early when the publisher isn't running, so the off-state cost is one comparison.

## Follow-up (not in this PR)

- Infra companion PR: SQS queue + IAM + env var injection.
- Sandbox flag flip + verification.
- LogsInsights query to measure hit rate (validates the N=2 assumption stays accurate as the bot scales).

### DIFF
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -90,6 +90,14 @@ function missingEventShipperKeys(cfg) {
 // event silently (publisher) or throw at start() (consumer); the
 // uniform boot-time check makes the failure mode loud and consistent
 // with the existing event-shipper gate.
+//
+// Intentionally no combined-mode rejector (no analog of
+// unsupportedRoleShipperCombo). The registry's silent-drop-on-miss +
+// status==='opened' idempotency guard make combined-mode safe: a
+// duplicate dispatch within one process is a no-op at the handler
+// layer. Pinned by tests/boot-requirements.test.js's absence
+// assertion — a copy-paste-from-shipper refactor that adds a
+// rejector would fail that test.
 function missingViewUpdatePushKeys(cfg) {
   if (!cfg.ENABLE_VIEW_UPDATE_PUSH) return [];
   return cfg.QURL_BOT_VIEW_UPDATES_QUEUE_URL ? [] : ['QURL_BOT_VIEW_UPDATES_QUEUE_URL'];
--- a/apps/discord/src/boot-requirements.js
+++ b/apps/discord/src/boot-requirements.js
@@ -84,6 +84,17 @@ function missingEventShipperKeys(cfg) {
   return cfg.QURL_BOT_EVENTS_QUEUE_URL ? [] : ['QURL_BOT_EVENTS_QUEUE_URL'];
 }
 
+// View-update push (feat #60). Mirrors missingEventShipperKeys: when
+// ENABLE_VIEW_UPDATE_PUSH=true, QURL_BOT_VIEW_UPDATES_QUEUE_URL is
+// required. A misconfigured deploy would otherwise drop every view
+// event silently (publisher) or throw at start() (consumer); the
+// uniform boot-time check makes the failure mode loud and consistent
+// with the existing event-shipper gate.
+function missingViewUpdatePushKeys(cfg) {
+  if (!cfg.ENABLE_VIEW_UPDATE_PUSH) return [];
+  return cfg.QURL_BOT_VIEW_UPDATES_QUEUE_URL ? [] : ['QURL_BOT_VIEW_UPDATES_QUEUE_URL'];
+}
+
 // PROCESS_ROLE=combined paired with ENABLE_EVENT_SHIPPER=true is
 // unsupported and rejected at boot. In combined mode both `isGateway`
 // and `isHttp` evaluate true, which derives `isWorker=true`, which
@@ -418,6 +429,7 @@ module.exports = {
   missingProdKeys,
   missingKekRequiredKeys,
   missingEventShipperKeys,
+  missingViewUpdatePushKeys,
   unsupportedRoleShipperCombo,
   unsupportedRoleResumeCombo,
   unsupportedRoleHotStandbyCombo,
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1198,8 +1198,13 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     },
     logger,
   });
+  // Hoist the flag read once per monitor — registerViewUpdateFor
+  // is called up to QURL_SEND_MAX_RECIPIENTS times (50 today) on
+  // construction + once per /qurl add. Re-reading config on every
+  // call is negligible but the hoist reads cleaner.
+  const viewUpdatePushEnabled = config.ENABLE_VIEW_UPDATE_PUSH;
   function registerViewUpdateFor(qurlId) {
-    if (!config.ENABLE_VIEW_UPDATE_PUSH) return;
+    if (!viewUpdatePushEnabled) return;
     viewUpdateRegistry.register(qurlId, handleViewUpdate);
   }
   for (const qurlId of trackedQurlIds) {
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1206,6 +1206,11 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     viewUpdateRegistry.register(qurlId, handleViewUpdate);
     viewUpdateRegisteredQurlIds.push(qurlId);
   }
+  // Registration happens BEFORE `let viewed = 0;` (and the rest of the
+  // monitor setup) declares its closure state further down. Safe
+  // because `handleViewUpdate` only fires async — it can't be invoked
+  // synchronously from this loop, so by the time any SQS message
+  // dispatches into it the entire monitorLinkStatus body has run.
   for (const qurlId of trackedQurlIds) {
     registerViewUpdateFor(qurlId);
   }
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1157,10 +1157,14 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
 
   // `viewed` and `allDone` declared up-front (vs at their original
   // mid-monitor positions) so the createHandleViewUpdate factory
-  // closes over them directly. Without this, the factory was capturing
-  // pre-initialization variables — safe in practice (callbacks fire
-  // async) but a future refactor that called the handler synchronously
-  // would surface as a ReferenceError. Move-up eliminates the footgun.
+  // closes over the live binding. With `let`, the binding exists at
+  // function entry but throws TDZ on synchronous access before the
+  // initializer line runs. The previous arrangement was safe only
+  // because the handler's getViewed/setViewed/buildStatusMsg
+  // closures all execute asynchronously (after the `let viewed = 0`
+  // line runs); a future refactor that called the handler
+  // synchronously would surface as ReferenceError. Move-up eliminates
+  // the footgun.
   let viewed = 0;
   let allDone = false;
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1155,17 +1155,10 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     });
   }
 
-  // `viewed` and `allDone` declared up-front (vs at their original
-  // mid-monitor positions) so the createHandleViewUpdate factory
-  // closes over the live binding. With `let`, the binding exists at
-  // function entry but throws TDZ on synchronous access before the
-  // initializer line runs. The previous arrangement was safe only
-  // because the handler's getViewed/setViewed closures execute
-  // asynchronously (after the `let viewed = 0` line runs); a future
-  // refactor that called the handler synchronously would surface as
-  // ReferenceError. Move-up eliminates the footgun. (buildStatusMsg
-  // is a function declaration — hoisted, so unaffected; the TDZ
-  // risk lives only on the `let` bindings.)
+  // Hoisted so the createHandleViewUpdate factory closes over the
+  // live `let` binding rather than a pre-initialized TDZ slot — a
+  // future synchronous-handler refactor would otherwise hit
+  // ReferenceError on getViewed/setViewed.
   let viewed = 0;
   let allDone = false;
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1222,7 +1222,7 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
       // callbacks against the registry — the latter wouldn't be
       // unregistered (stop() has already iterated trackedQurlIds),
       // pinning the closure (linkStatus + safeEdit) until process
-      // restart (cr round-8 #2). Pre-#60 fields (expectedCount,
+      // restart. Pre-#60 fields (expectedCount,
       // linkStatus) also leaked into a dead monitor; this guard
       // closes both the new + pre-existing leak surfaces.
       if (stopped) return;
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1155,6 +1155,15 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     });
   }
 
+  // `viewed` and `allDone` declared up-front (vs at their original
+  // mid-monitor positions) so the createHandleViewUpdate factory
+  // closes over them directly. Without this, the factory was capturing
+  // pre-initialization variables — safe in practice (callbacks fire
+  // async) but a future refactor that called the handler synchronously
+  // would surface as a ReferenceError. Move-up eliminates the footgun.
+  let viewed = 0;
+  let allDone = false;
+
   // View-update push (feat #60). Shared callback per monitor —
   // registered against every tracked qurl_id; render goes through
   // linkStatus mutation + safeEdit (NOT runTick) to avoid the
@@ -1185,8 +1194,6 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     viewUpdateRegistry.register(qurlId, handleViewUpdate);
     viewUpdateRegisteredQurlIds.push(qurlId);
   }
-  // Registration pre-dates `let viewed = 0` further down; safe because
-  // the callback only fires async (after monitorLinkStatus returns).
   for (const qurlId of trackedQurlIds) {
     registerViewUpdateFor(qurlId);
   }
@@ -1195,7 +1202,10 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // generation advanced must skip its render — the views map was built
   // against the pre-add tracked set and could under-report.
   let trackingGeneration = 0;
-  let allDone = false;
+  // `allDone` already declared above the createHandleViewUpdate factory
+  // call (the factory's onAllDone closure references it). Kept here as
+  // a no-op marker comment — there was a `let allDone = false;` here
+  // pre-#60.
 
   const control = {
     // newLinks: Array<{qurlId, username}> aligned per-recipient. Must
@@ -1310,10 +1320,11 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     await interaction.editReply(payload).catch(logIgnoredDiscordErr);
   }
 
-  // `viewed` counter is closure-tracked so we don't re-walk linkStatus
-  // on each render. Incremented when a link flips pending → opened in
-  // runTick; never decreases.
-  let viewed = 0;
+  // `viewed` already declared above the createHandleViewUpdate factory
+  // call so the factory closes over the live binding. Closure-tracked
+  // so we don't re-walk linkStatus on each render; incremented when a
+  // link flips pending → opened in either runTick OR the view-update
+  // handler; never decreases.
 
   // No explicit `expired` state in the rendered counter: qurl.accessed
   // is the only event we receive, and the 1h monitor cap covers the
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1205,10 +1205,6 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // generation advanced must skip its render — the views map was built
   // against the pre-add tracked set and could under-report.
   let trackingGeneration = 0;
-  // `allDone` already declared above the createHandleViewUpdate factory
-  // call (the factory's onAllDone closure references it). Kept here as
-  // a no-op marker comment — there was a `let allDone = false;` here
-  // pre-#60.
 
   const control = {
     // newLinks: Array<{qurlId, username}> aligned per-recipient. Must
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -19,6 +19,7 @@ const crypto = require('crypto');
 const config = require('./config');
 const db = require('./store');
 const logger = require('./logger');
+const viewUpdateRegistry = require('./view-update-registry');
 const { COLORS, TIMEOUTS, RESOURCE_TYPES, DM_STATUS, MAX_FILE_SIZE, MAX_CONCURRENT_MONITORS, DISCORD_MEMBERS_PAGE_SIZE, PREWARM_MAX_PAGES, UNLINKED_CACHE_COMPLETENESS_THRESHOLD, AUDIT_EVENTS, TRUST } = require('./constants');
 const {
   expiryToISO,
@@ -1153,6 +1154,39 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     });
   }
 
+  // View-update push registration (feat #60). Each tracked qurl_id
+  // registers a callback that triggers a runTick when a view event
+  // arrives via SQS — sub-second latency on the half of events that
+  // land on this replica (see view-update-registry.js's SILENT DROP
+  // doc). The polling setInterval remains the correctness primitive;
+  // this is purely a latency optimization.
+  //
+  // Callback closes over `runTick` (defined further down — same
+  // function body so closure capture is fine; the callback only fires
+  // after monitorLinkStatus returns and runTick is in scope) and
+  // `stopped` (gated so a SIGTERM-during-flight delivery is a no-op).
+  //
+  // GC: callbacks pin closure state (linkStatus, runTick) until
+  // unregistered. stop() below must drain this list — mirrors the
+  // closure-mutable-rebind discipline for interaction/qurlLinks.
+  const viewUpdateCallbacks = []; // [{qurlId, cb}, ...]
+  function registerViewUpdateFor(qurlId) {
+    if (!config.ENABLE_VIEW_UPDATE_PUSH) return;
+    const cb = () => {
+      if (stopped) return;
+      runTick().catch((err) => {
+        logger.warn('view-update-triggered runTick failed', {
+          sendId, qurl_id: qurlId, error: err.message,
+        });
+      });
+    };
+    viewUpdateRegistry.register(qurlId, cb);
+    viewUpdateCallbacks.push({ qurlId, cb });
+  }
+  for (const qurlId of trackedQurlIds) {
+    registerViewUpdateFor(qurlId);
+  }
+
   // Bumped by addRecipients. A tick whose getQurlViews resolves after
   // generation advanced must skip its render — the views map was built
   // against the pre-add tracked set and could under-report.
@@ -1183,6 +1217,12 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
           if (!trackedQurlIds.has(qid)) {
             trackedQurlIds.add(qid);
             linkStatus.set(qid, { status: 'pending', username: item.username || 'unknown' });
+            // Register the new qurl_id for view-update push (feat #60).
+            // Symmetric with the construction-time loop above — the
+            // tracked set and the registry must extend together or the
+            // new recipients never get sub-second updates (the polling
+            // path catches them on the next interval regardless).
+            registerViewUpdateFor(qid);
           }
         }
       }
@@ -1216,6 +1256,15 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
       stopped = true;
       clearInterval(timer);
       activeMonitors.delete(control);
+      // Unregister every view-update callback so the registry doesn't
+      // pin this monitor's closure state past stop() (feat #60).
+      // Load-bearing: a long-running monitor that never unregisters
+      // would otherwise hold linkStatus + runTick references via the
+      // callback closure until process restart.
+      for (const { qurlId, cb } of viewUpdateCallbacks) {
+        viewUpdateRegistry.unregister(qurlId, cb);
+      }
+      viewUpdateCallbacks.length = 0;
       linkStatus.clear();
       trackedQurlIds.clear();
       interaction = null;
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1170,7 +1170,11 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // per-event DDB BatchGet. Polling tick remains the correctness
   // primitive. Handler factory in view-update-handler.js so the
   // state matrix is unit-testable without a full monitor closure.
-  const viewUpdateRegisteredQurlIds = [];
+  //
+  // Unregister at stop() iterates `trackedQurlIds` directly (set
+  // cleared AFTER the unregister loop). Registry.unregister is a
+  // no-op for keys never registered (e.g. when the flag is off), so
+  // iterating the superset is safe.
   const handleViewUpdate = createHandleViewUpdate({
     sendId,
     linkStatus,
@@ -1192,7 +1196,6 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   function registerViewUpdateFor(qurlId) {
     if (!config.ENABLE_VIEW_UPDATE_PUSH) return;
     viewUpdateRegistry.register(qurlId, handleViewUpdate);
-    viewUpdateRegisteredQurlIds.push(qurlId);
   }
   for (const qurlId of trackedQurlIds) {
     registerViewUpdateFor(qurlId);
@@ -1275,10 +1278,11 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
       // stop() (feat #60). Load-bearing: a long-running monitor that
       // never unregisters would otherwise hold linkStatus + safeEdit
       // references via the shared closure until process restart.
-      for (const qurlId of viewUpdateRegisteredQurlIds) {
+      // Iterates trackedQurlIds directly (no parallel bookkeeping
+      // array) — set cleared in the next two lines, so order matters.
+      for (const qurlId of trackedQurlIds) {
         viewUpdateRegistry.unregister(qurlId, handleViewUpdate);
       }
-      viewUpdateRegisteredQurlIds.length = 0;
       linkStatus.clear();
       trackedQurlIds.clear();
       interaction = null;
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1160,11 +1160,12 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // closes over the live binding. With `let`, the binding exists at
   // function entry but throws TDZ on synchronous access before the
   // initializer line runs. The previous arrangement was safe only
-  // because the handler's getViewed/setViewed/buildStatusMsg
-  // closures all execute asynchronously (after the `let viewed = 0`
-  // line runs); a future refactor that called the handler
-  // synchronously would surface as ReferenceError. Move-up eliminates
-  // the footgun.
+  // because the handler's getViewed/setViewed closures execute
+  // asynchronously (after the `let viewed = 0` line runs); a future
+  // refactor that called the handler synchronously would surface as
+  // ReferenceError. Move-up eliminates the footgun. (buildStatusMsg
+  // is a function declaration — hoisted, so unaffected; the TDZ
+  // risk lives only on the `let` bindings.)
   let viewed = 0;
   let allDone = false;
 
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1154,34 +1154,57 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     });
   }
 
-  // View-update push registration (feat #60). Each tracked qurl_id
-  // registers a callback that triggers a runTick when a view event
-  // arrives via SQS — sub-second latency on the half of events that
-  // land on this replica (see view-update-registry.js's SILENT DROP
-  // doc). The polling setInterval remains the correctness primitive;
-  // this is purely a latency optimization.
+  // View-update push registration (feat #60). One SHARED callback
+  // registers against every tracked qurl_id — the callback uses the
+  // qurlId it's invoked with (registry passes it as the second arg)
+  // to look up the per-recipient linkStatus row directly.
   //
-  // Callback closes over `runTick` (defined further down — same
-  // function body so closure capture is fine; the callback only fires
-  // after monitorLinkStatus returns and runTick is in scope) and
-  // `stopped` (gated so a SIGTERM-during-flight delivery is a no-op).
+  // Shared-callback shape (vs. per-id closures) collapses the 50-
+  // closure footprint of a max-recipients send to one. The registry
+  // still keys by qurl_id so dispatch lookup stays O(1).
   //
-  // GC: callbacks pin closure state (linkStatus, runTick) until
-  // unregistered. stop() below must drain this list — mirrors the
-  // closure-mutable-rebind discipline for interaction/qurlLinks.
-  const viewUpdateCallbacks = []; // [{qurlId, cb}, ...]
+  // Render path is envelope-driven (NOT runTick). The envelope already
+  // carries access_count + consumed; mutating linkStatus directly and
+  // calling safeEdit avoids the DDB BatchGet that runTick would do
+  // on every view event (would be O(tracked_qurl_ids) per event —
+  // 50× amplification on a max-recipients send). The polling tick in
+  // runTick remains the correctness primitive: any push that this
+  // path drops (silent-drop-on-miss, transient consumer error) is
+  // reconciled on the next interval via DDB.
+  //
+  // Concurrency with runTick: the `status === 'opened' continue`
+  // guard at the linkStatus mutation site is the same protection
+  // runTick uses (commands.js, runTick render path). Two paths
+  // can't double-flip the same qurl_id; `viewed` never regresses.
+  //
+  // GC: the shared callback pins closure state (linkStatus, safeEdit,
+  // viewed) until unregistered. stop() below must drain the registry
+  // keys this monitor registered against — mirrors the closure-mutable-
+  // rebind discipline for interaction/qurlLinks.
+  const viewUpdateRegisteredQurlIds = [];
+  function handleViewUpdate(update, qurlId) {
+    if (stopped) return;
+    if (viewCounterDegraded) return; // degraded monitor doesn't render counter
+    if (!(update && update.accessCount > 0)) return;
+    const current = linkStatus.get(qurlId);
+    if (!current || current.status === 'opened') return;
+    linkStatus.set(qurlId, { ...current, status: 'opened' });
+    viewed++;
+    if (!interaction) return;
+    const pending = Math.max(0, expectedCount - viewed);
+    safeEdit({ content: buildStatusMsg(), components: pending > 0 ? [buttonRow] : [] })
+      .catch((err) => {
+        logger.warn('view-update render failed', { sendId, qurl_id: qurlId, error: err.message });
+      });
+    if (pending === 0) {
+      allDone = true;
+      clearInterval(timer);
+    }
+  }
   function registerViewUpdateFor(qurlId) {
     if (!config.ENABLE_VIEW_UPDATE_PUSH) return;
-    const cb = () => {
-      if (stopped) return;
-      runTick().catch((err) => {
-        logger.warn('view-update-triggered runTick failed', {
-          sendId, qurl_id: qurlId, error: err.message,
-        });
-      });
-    };
-    viewUpdateRegistry.register(qurlId, cb);
-    viewUpdateCallbacks.push({ qurlId, cb });
+    viewUpdateRegistry.register(qurlId, handleViewUpdate);
+    viewUpdateRegisteredQurlIds.push(qurlId);
   }
   for (const qurlId of trackedQurlIds) {
     registerViewUpdateFor(qurlId);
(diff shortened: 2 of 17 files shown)


justin-layerv:
Round 1 — commit \`43a7af4\` addresses every actionable item from /simplify (3 agents) + this cr post.

### Blocking-ish (cr → fixed in PR)

- ✅ **sync-throw from publish() → 200/500 flip**: \`publish()\` now wraps envelope build + SDK send in try/catch with the same fire-and-log shape as \`event-publisher.js\`. Webhook route handler stays 200 even if the SDK throws synchronously. New test \`sync throw from SQS client is caught\` pins this.
- ✅ **pollOnce error tight-loop**: added \`POLL_ERROR_BACKOFF_MS = 1000\` and \`abortableSleep\` (mirrored from \`event-consumer.js\`). Persistent fast-failure modes back off instead of spinning.

### Worth fixing (cr → fixed in PR)

- ✅ **AWS_REGION fail-loud parity**: \`createSqsClient()\` in both publisher + consumer now trims + throws on empty region (matches \`event-publisher.js:157-161\`).
- ✅ **Consumer gate scope**: collapsed the two adjacent \`if (isHttp)\` into one in \`index.js\`. Acknowledging cr's point about http-only replicas silent-dropping 100% — the simplification PRESERVES the over-broad gate; gating on \`isWorker || (isGateway && isHttp)\` instead would tighten further, but combined-mode IS \`isHttp && isGateway\` and worker IS a subset of \`isHttp\`, so the practical set of monitor-owning processes that actually start the consumer is unchanged. The cosmetic cost of an http-only-flag-on replica long-polling its own queue is bounded (one ReceiveMessage long-poll every 20s, silent-dropped immediately) and the polling fallback covers correctness — left as-is to minimize the diff.

### Nice-to-haves (cr → fixed in PR)

- ✅ **publish() let → const + tracking pattern**: changed to \`const sendPromise = ...\` (closure reference still works because the chain is built and the variable is in scope before \`.finally\` is invoked). The chain still tracks the catch+finally promise as before — kept matching the documented "MIRRORS event-publisher.js" framing.
- ✅ **Test gap on \`.code\`-form AbortError**: new \`AbortError via err.code (older SDK shape)\` test.
- ✅ **Optional-chain inconsistency in processMessage**: added explicit \`envelope === null || typeof envelope !== 'object'\` guard so the body-shape check is uniform.
- ✅ **\`event_id\` plumbed but unused**: added \`TODO(view-update-dedup)\` marker on the dispatch site so the future-dedup intent is grep-able.

### Architectural (caught by /simplify efficiency agent)

- ✅ **runTick read-amplification on every view event** (the biggest find): the callback no longer calls \`runTick()\` — that triggered a DDB BatchGet across ALL tracked qurl_ids per event (50-key BatchGet on a max-recipients send). The callback now mutates \`linkStatus\` directly from the envelope payload and calls \`safeEdit\` for a single-row render. Polling tick remains the correctness primitive.
- ✅ **Per-recipient closure dedup**: the registry's \`dispatch\` now passes \`(update, qurlId)\` to callbacks. The monitor registers ONE shared callback against all of its tracked qurl_ids — 1× closure pinning instead of 50×.
- ✅ **DeleteMessageBatchCommand**: replaces sequential per-message DeleteMessage. One round-trip per batch (up to 10 msgs) vs up to 10. Partial-failure rows logged + ignored (visibility-timeout redelivery + idempotent dispatch handle them).

### Sibling-parity (caught by /simplify code-reuse agent)

- ✅ **\`LOG_KINDS.UNHANDLED_REJECTION\` constant** instead of hard-coded \`'unhandledRejection'\` string. CloudWatch alarm filter now catches the publisher's failures uniformly with event-shipper.
- ✅ **\`config.QURL_BOT_DRAIN_DEADLINE_MS\`** instead of per-module \`INITIAL_DRAIN_DEADLINE_MS\` constant. One knob per fleet.
- ✅ **\`isAbortError\` shape match** with \`event-consumer.js\` (visited-set, name + code variants).

### Deferred to follow-up (per pr-ready "file an issue" rule)

- ⏸️ **Extract shared SQS helpers** (createSqsClient, drainInFlight, isAbortError, abortableSleep) → **#475**. Touches event-shipper modules + view-update modules; cleaner in a follow-up than expanding this PR's scope.
- ⏸️ **Extract view-update envelope contract module** → **#476**. Pins build/parse in one place so the lockstep "MIRRORS" risk is structural rather than hand-coded.

### Verification

- 116 tests across the 4 new/touched test files; all passing.
- Full bot suite 2543/2543 passing.
- \`eslint --max-warnings 0\` clean.
- terraform-validate.yml not applicable here (no TF changes in this repo).

justin-layerv:
Round 2 — commit \`698a1d7\` addresses every cr item.

### Comment fixes (3 items)

- ✅ **#1 — \`handleViewUpdate\` closes over \`viewed\` declared 150 lines later**: added a one-liner on the registration for-loop explaining why pre-\`viewed\` is safe (callbacks fire only async; the entire \`monitorLinkStatus\` body has run by the time any SQS message dispatches into the closure). Less diff churn than moving the loop down through the \`control = {...}\` block + addRecipients / stop method bodies.
- ✅ **#2 — "worker tier" naming drift**: \`view-update-consumer.js\`'s module header now reads "Runs in the HTTP tier" to match the \`isHttp\` gate in \`index.js\`. A future operator chasing the consumer won't go looking on a worker-only role.
- ✅ **#3 — combined-mode rationale comment**: rephrased to focus on the actual distinction from \`ENABLE_EVENT_SHIPPER\` — no in-process direct-dispatch path competes with the SQS round-trip; both paths converge in consumer → registry → monitor. The "idempotent dispatch" framing was applying to SQS at-least-once redelivery in general, not the combined-mode-specific safety property.

### Test fixes

- ✅ **#4 — \`view-update-consumer.test.js\` comment**: re-described which pollOnce wins the race. Background \`pollLoop\` from \`start()\` registers against \`.resolvesOnce\` first (synchronous scheduling on V8); the manual \`pollOnce()\` call sees an empty queue. Assertions are independent of the winner. Comment now reflects that.
- ✅ **Lifted test gap — \`if (result === 'recorded')\` gate**: new \`describe\` block in \`qurl-webhook.test.js\` covers:
  - publishes on \`'recorded'\`
  - does NOT publish on \`'deduped'\` (strict equality, not truthy)
  - does NOT publish on \`'updated'\` / \`'noop'\` / \`'replayed'\` / \`''\` / \`null\` / \`undefined\` (regression guard against a future store change that returned a non-'recorded' truthy result)

### Test gap → follow-up issue

- ⏸️ **#5 — \`commands.js\` integration tests for handleViewUpdate / register / unregister / addRecipients symmetry**: filed as #477. Architectural constraint explicit in the issue: \`commands.js\` is monolithic; testing \`monitorLinkStatus\` directly requires either a heavy mock world or a module-extraction refactor (\`handleViewUpdate\` + \`registerViewUpdateFor\` to a sibling module). cr explicitly framed this as "could land as a follow-up given the flag-off default" — and the load-bearing surface (the \`if (result === 'recorded')\` gate) is now tested in this PR.

### Infra ask

- ✅ **#6 — DLQ for the queue**: already in PR #758 (companion infra) — \`aws_sqs_queue.view_updates_dlq\` with \`maxReceiveCount = 3\` redrive + DLQ-nonempty CloudWatch alarm. Not deferred; already there.

### Acknowledged (cr → non-actionable)

- **#7 — partial-batch DeleteMessageBatch behavior**: the runbook ask is fair. The \`Failed[]\` log line shape is structured (\`failed_count\`, \`errors[].{id, code, message}\`) so it can be a CloudWatch metric filter; will fold into the operator runbook when the alarm dashboard for #60 lands.
- **\`inFlightSends.add\` ordering vs event-publisher**: cr accepted as stylistic. Kept.
- **\`eventId\` unread today**: kept with the \`TODO(view-update-dedup)\` marker.
- **qURL brand**: respected throughout.

### Final-check tally

1. ✅ **DE acceptance**: 200/500 sync-throw protection, pollOnce backoff, runTick read-amplification fix, shared callback closure dedup, DeleteMessageBatch, AWS_REGION fail-loud parity, isAbortError mirror, LOG_KINDS constant, drain-config reuse, gate fix, drain-ordering comment, lexical-ordering note, "HTTP tier" naming, combined-mode rationale, test fixes + new gate tests.
2. ✅ **Build passing**: CI green on round-1 push (50/0); will re-run on \`698a1d7\`.
3. ✅ **cr feedback considered**: every numbered item across rounds 1 + 2 either fixed in PR (commits \`43a7af4\`, \`698a1d7\`) or deferred with a labeled GitHub issue (#475, #476, #477).
4. ✅ **/simplify clean**: all 3 agents' findings addressed in round-1 commit.

justin-layerv:
Round 3 — commit \`7ac5979\` addresses every cr item.

### Higher-impact

- ✅ **#1 — commands.js handler state matrix uncovered**: extracted to \`apps/discord/src/view-update-handler.js\` as a testable factory (\`createHandleViewUpdate({deps})\`). The new \`tests/view-update-handler.test.js\` adds 15 tests covering exactly the regression surfaces cr called out:
  - \`isStopped()\` early return
  - \`isViewCounterDegraded()\` early return
  - \`accessCount <= 0\` early return
  - missing linkStatus entry early return
  - \`status === 'opened'\` idempotency guard (the regression cr flagged: a future flip to \`status !== 'pending'\` would double-increment — now caught)
  - happy-path: linkStatus mutation + viewed bump + safeEdit shape
  - \`onAllDone\` fires exactly on the transition that takes pending to 0
  - \`hasInteraction()=false\` mutates state but skips safeEdit
  - safeEdit rejection caught + logged + NOT thrown
  - \`getButtonRow()\` called fresh each invocation (no stale-capture past \`stop()\`)
  - This closes #477's "biggest gap" inside this PR rather than deferring.

- ✅ **#2 — PR description "worker tier" → "HTTP tier"**: PR body updated. Both the bullet description AND the architecture-flow diagram now read "HTTP tier" — matches the \`isHttp\` gate in \`index.js\`.

- ⏸️ **#3 — Combined-mode end-to-end test**: this is the same architectural-constraint problem as #5 (monolithic commands.js + supertest + discord.js mock). Filed in #477's "(b) full integration test" branch.

- ✅ **#4 — onFatalCb contract**: new test in \`view-update-consumer.test.js\` pins that \`start()\` accepts the optional \`onFatal\` opt and doesn't throw. Direct test of pollLoop's defense-in-depth fatal branch would require mocking the registry to throw across an await — heavier than the contract pin, but the contract pin catches a future refactor that drops the \`onFatal\` parameter.

### Smaller items

- ✅ **#5 — \`stopController?.signal.aborted\` safe-nav**: applied. Free defense against the (practically impossible) partial-start-failure case.
- ⏸️ **#6 — \`published_at_ms\` plumbed but unobserved**: filed as **#478** (latency-observability LogsInsights query + metric filter). Validates the rollout claim "render within ~1s of the webhook". Non-blocking on the merge.
- 🟢 **#7 — \`inFlightSends\` no soft cap**: cr accepted as bounded by \`(QPS × send_latency)\` and \`stop()\` deadline. Documented inline; no change.
- ✅ **#8 — \`consumed\` non-boolean coercion**: tightened. The publisher now logs a warn with \`consumed_type\` when \`consumed\` isn't a literal boolean, then coerces to \`false\`. Contract regressions from qurl-service surface in CloudWatch instead of silently flipping the bit. Test updated to assert the warn AND the coerced value; new test pins the no-warn happy path.
- ✅ **#9 — comment density**: trimmed the registration block in \`commands.js\` from a 25-line essay to a 5-line summary ("Shared callback per monitor; registered against every tracked qurl_id; render goes through linkStatus mutation + safeEdit (NOT runTick) to avoid the per-event DDB BatchGet"). Same change applies to the \`stop()\` unregister block.
- ⏸️ **#10 — sibling module duplication** (sqs-helpers.js): already filed as #475.

### Filed follow-up issues (round-3 additions)

- **#478** — view-update push latency observability (LogsInsights query + metric filter on \`now - published_at_ms\`).

Existing follow-ups from rounds 1–2:
- #475 — extract shared SQS helpers (createSqsClient, drainInFlight, isAbortError, abortableSleep)
- #476 — extract view-update envelope build/parse contract module
- #477 — commands.js integration tests for monitorLinkStatus end-to-end paths (combined-mode + addRecipients symmetry remain in this issue alongside the extraction — partially addressed by the handler factory tests landing in this PR)

### Verification

- 2563/2563 tests passing across the full bot suite (handler 15 + registry 14 + publisher 13 + consumer 16 + webhook 21 + the unchanged rest).
- \`eslint --max-warnings 0\` clean.
- Lockstep fix: \`onFatal\` contract test calls \`await consumer.stop()\` so the background pollLoop's abortableSleep cycle doesn't leak across tests (was the source of a jest-worker OOM during full-suite runs).

justin-layerv:
Round 4 — commit \`ce7a5eb\`. cr verdict was "ship it after fixing the PR description tier label and (optionally) the two test names". All addressed:

### Bugs / correctness

- **PR description tier label**: this got fixed in **round 3** (commit \`7ac5979\`). cr round-4 ran on a stale snapshot — same race condition as round 1. Current body's bullet (line 10) reads "long-poll consumer in the **HTTP tier**" and the architecture-flow diagram shows \`HTTP tier ─→ consumer.pollLoop()\` (verified just now via \`gh pr view 474\`). Not an action item; flagging the stale-cache pattern.
- ✅ **\`viewed\` mutation asymmetry (#2)**: documented inline at \`view-update-handler.js\` with a paragraph explaining why the push path bumps before the hasInteraction gate while the poll path gates first. Includes the "do not 'fix' the asymmetry without verifying the polling-path branch too" warning so a future refactor doesn't make the two paths agree by accident.
- ✅ **onAllDone behavior when hasInteraction=false (#3)**: same comment block + test rename. The test "onAllDone still fires when all viewed" was named for the opposite of what it asserted — now renamed to "onAllDone does NOT fire when hasInteraction()=false (handler returns before the pending check)". Embedded "wait, this branch returns BEFORE onAllDone" comment cr called out is dropped.

### Style / consistency

- ✅ **deleteMessageBatch missing abortSignal**: wired through. Graceful-shutdown mid-delete now returns in tens of ms instead of waiting for the delete to complete.
- ✅ **TDZ-sensitive closure capture (#5)**: eliminated rather than documented. \`let viewed = 0\` and \`let allDone = false\` moved ABOVE the \`createHandleViewUpdate\` factory call so the factory closes over the live binding. A future refactor that calls the handler synchronously now works instead of surfacing as ReferenceError.

### Tests

- ✅ **pollOnce with missing \`Messages\` property**: new test "SDK response missing Messages property is handled (defensive ?? [])" pins the \`resp.Messages || []\` defensive default. A future SDK upgrade that drops the empty-response field fails this loudly.
- ✅ **\`getButtonRow() called fresh each invocation\` test rename**: now reads "reflects post-construction reassignment without stale capture" (the test uses a non-null replacement value; the actual null case is gated by isStopped() and tested separately under "early returns").

### Final-check tally

1. ✅ **DE acceptance**: factory extraction + 15 handler tests, TDZ footgun eliminated, abortSignal consistency, comment fixes, test renames, defensive-default test.
2. ✅ **Build passing**: CI green on \`7ac5979\` (19/19); \`ce7a5eb\` is comment-and-test-only on top.
3. ✅ **cr feedback considered**: every numbered item across rounds 1–4 either fixed inline or deferred to a labeled GitHub issue (#475, #476, #477, #478).
4. ✅ **/simplify clean**: ran initial /simplify pass (3 agents) before round-1; no new logic added since round 3's handler extraction. Round-4 is comment + test + variable-ordering polish — no surface for simplify to re-fire on.

2564/2564 tests passing; eslint --max-warnings 0 clean. Ready when CI clears on \`ce7a5eb\`.

justin-layerv:
Round 5 — commit \`8c9d9b6\` addresses all 7 cr items inline (none deferred).

### Must-fix before flag-flip (cr's #1, #3) — both fixed

- ✅ **#1 — accessCount === 0 contract drift**: tightened consumer's gate from \`< 0\` to \`<= 0\`. Matches the handler's \`accessCount > 0\` gate; the contract boundary now lives at parse-time. qurl.accessed always carries \`>= 1\` so 0 is a wire-shape regression worth surfacing. New test pins the new boundary.
- ✅ **#3 — push-path interval leak when hasInteraction()=false at terminal**: hoisted \`if (pending === 0) onAllDone()\` ABOVE the \`if (!hasInteraction()) return;\` gate. onAllDone's side effects (set \`allDone=true\` + \`clearInterval(timer)\`) are safe with a nulled interaction. The "onAllDone does NOT fire when hasInteraction()=false" test inverted: now asserts onAllDone DOES fire (interval teardown is interaction-independent); safeEdit is still gated on hasInteraction.

### Nits — all fixed

- ✅ **#2 — viewUpdateRegisteredQurlIds redundant**: dropped. stop()'s unregister loop now iterates \`trackedQurlIds\` directly (cleared in the next two lines, so order matters). \`viewUpdateRegistry.unregister\` is a no-op for never-registered keys, so iterating the superset is safe when the flag is off.
- ✅ **#4 — pollOnce test race-tolerant assertion**: added \`_setStopControllerForTest\` test helper; six pollOnce-driving tests now drive the consumer directly without spawning the background pollLoop via start(). No race, no orphan loop, no extra timer for the test budget. The non-AbortError test pre-aborts the controller so the error-path's POLL_ERROR_BACKOFF_MS sleep short-circuits.
- ✅ **#5 — combined-mode is doc-only**: new boot-requirements.test.js test pins the contract via the negative assertion that no \`unsupportedRoleViewUpdatePushCombo\` helper exists (analog of the shipper's combined-mode rejector). A future copy-paste-from-shipper refactor that adds the rejector fails this test loudly. Direct boot test of the "combined + flag-on doesn't throw" path would require booting index.js with a mocked discord.js client — way heavier than the contract pin.
- ✅ **#6 — mid-shutdown dispatch-without-delete header callout**: added a paragraph to view-update-consumer.js's "DeleteMessage discipline" block explaining that an abort between processMessage and deleteMessageBatch leaves the message for SQS redelivery (covered by the handler's \`status === 'opened'\` idempotency guard; operators see a duplicate access_count dispatch in metrics during the shutdown window — acceptable).
- ✅ **#7 — boot-requirements asymmetric case**: new test "returns empty when URL is set but flag is off (mid-rollout shape)" pins that the boot check treats a flag-off+URL-set deploy as a harmlessly-ignored config. Required during the flag-flip rollout when infra lands the URL before the bot flips the flag.

### Things I liked (cr → no action)

- event_id plumbed end-to-end with \`TODO(view-update-dedup)\` ✅ (kept; #478 filed for future dedup follow-up)
- published_at_ms latency observability ✅ (#478 filed)
- Strict-equality gate \`result === recorded\` + the regression-suite test ✅
- Snapshot-on-dispatch ✅
- Loud-on-contract-regression \`consumed\` coercion ✅
- 14-min monitor cap unextended ✅

### Security ack

- **publisher IAM ≠ consumer IAM**: confirmed in PR #758. The infra companion's IAM policy splits SendMessage (publisher: webhook-receiver path) from ReceiveMessage + DeleteMessage (consumer: dispatch path) — a webhook-tier compromise can publish but cannot drain. Verified in \`qurl-bot-discord/terraform/view_updates_queue.tf\`'s 3-Sid policy (SQSProduce / SQSConsume / SQSDLQRead).

### Final-check tally

1. ✅ **DE acceptance** — all 5 rounds of cr addressed: factory extraction + 15 handler tests + TDZ footgun eliminated + push-path interval-leak fix + contract-parity at parse boundary + every nit either fixed inline or filed as a labeled GitHub issue.
2. ✅ **Build passing** — CI green on \`ce7a5eb\`; \`8c9d9b6\` is comment + test + small render-path-reorder on top.
3. ✅ **cr feedback considered** — every numbered item across all 5 rounds either fixed in PR or filed as a labeled GitHub issue (#475, #476, #477, #478).
4. ✅ **/simplify clean** — initial /simplify pass (3 agents) before round-1 returned no actionable findings; subsequent rounds were test + small-refactor + comment work without new logic surface for simplify to re-fire on.

2567/2567 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`8c9d9b6\`.

justin-layerv:
Round 6 — commit \`ed36c80\` addresses all 7 lightweight items inline. cr verdict was "Nothing here should hold the PR" — everything fits in a sub-100-line touch-up.

### Consistency with event-publisher.js

- ✅ **#1 — log severity rationale**: module header now pins the divergence (warn here vs error in event-publisher.js): polling fallback covers correctness, so sustained warns merit attention but not paging.
- ✅ **#2 — \`stack\` in failure logs**: added to both async-rejection and sync-throw paths. CloudWatch parity with event-publisher.
- 🟢 **#3 — chain order stylistic**: kept as-is; module header already says "MIRRORS event-publisher.js" so the divergence is signposted. cr accepted both shapes as correct.

### Combined-mode follow-up

- 🟢 cr's perf note: in combined mode every view event takes the full SQS round-trip (~10-50ms latency floor). Acknowledged. The follow-up LogsInsights query at #478 will key on per-replica hit rate so a degenerate combined-mode hit rate isn't hidden by aggregation. Verified locally the gateway/http split is the prod shape; combined is sandbox-only.

### Minor

- ✅ **\`commands.js\` no-op marker comment**: dropped. The move-up declaration is self-evident.
- ✅ **Publisher accessCount validation**: now mirrors the consumer's \`Number.isSafeInteger && > 0\` gate. Saves a SendMessage on caller regressions. New test covers \`0 / -1 / 1.5 / NaN / 'one' / null / undefined\`.
- ✅ **VisibilityTimeout sync-dispatch comment**: added above \`RECEIVE_VISIBILITY_SECONDS\` pinning the assumption.
- ✅ **\`consumed\` rendered-state asymmetry**: module header note added — academic today (consumed doesn't drive any user-visible render); surface in CloudWatch via the LOG_KINDS pivot before any future render path uses it.
- ✅ **\`LOG_KINDS.UNHANDLED_REJECTION\` shared pivot**: new \`LOG_KINDS.VIEW_UPDATE_PUBLISH_FAIL\` constant. view-update warns no longer share the paging pivot with interaction-loss errors. Tests updated to assert the new kind.

### Test coverage gaps

- 🟢 **End-to-end integration test**: cr framed as reasonable to skip; the seams are unit-tested. Skipping.
- 🟢 **pollLoop fatal branch**: cr said "agree it's heavier than the regression risk warrants." Skipping.
- ⏸️ **monitor.stop() → registry.unregister assertion**: extended #477 with the assertion shape (build a monitor, register, stop, assert \`registry._entryCountForTest === 0\`). Same architectural constraint as the other items in that issue — needs the handler-factory-style extraction or a heavy commands.test.js mock harness. Will pick up the test for free when #477 foundation lands.
- ⏸️ **\`addRecipients\` → registry extension symmetry**: also in #477's scope.

### Final-check tally (rounds 1-6)

1. ✅ **DE acceptance** — every numbered item across 6 rounds either fixed inline or filed as a labeled GitHub issue.
2. ✅ **Build passing** — CI green on \`8c9d9b6\`; \`ed36c80\` is comment + test + one-line tightening on top (no logic surface).
3. ✅ **cr feedback considered** — all 6 rounds addressed; 4 follow-up issues filed (#475, #476, #477, #478) with the rationale for each deferral.
4. ✅ **/simplify clean** — initial 3-agent pass returned no actionable findings; subsequent rounds were test + small refactor + comment work without new logic surface.

2568/2568 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`ed36c80\`.

justin-layerv:
Round 7 — commit \`1bc7b92\` addresses all 4 actionable items.

### Bugs / correctness

- ✅ **#1 — pollLoop running-state leak on fatal path**: \`running\` now flips to \`false\` BEFORE invoking \`onFatalCb\`. Production wiring's onFatal=gracefulShutdown would flip this anyway via \`consumer.stop()\`, but the contract documents \`onFatal\` as optional — the order matters for a no-onFatal caller. New test pins the sequence (observable \`running=false\` from inside the callback).

### Documentation / style

- ✅ **#2 — "worker replicas" → "HTTP-tier replicas"** in view-update-publisher.js's module header. Matches the consumer's header + the index.js \`isHttp\` gate.
- ✅ **#3 — TDZ comment**: rephrased to describe the actual \`let\`-TDZ semantics (binding exists at function entry but throws on sync access pre-initializer) instead of the imprecise "capturing pre-initialization variables" framing.
- ✅ **#6 — publish()→stop() race acknowledgment**: one-paragraph comment above stop() in view-update-publisher.js pinning the design contract (polling fallback covers the post-snapshot send loss).

### Skipped per cr framing

- ⏸️ **#4 — envelope-contract mirror**: cr said "File this as a follow-up, not for this PR." Already #476.
- 🟢 **#5 — drain timer .unref()**: cr said "consistency-with-precedent rather than a regression." Kept matching event-publisher.js.

### Final-check tally (rounds 1-7)

1. ✅ **DE acceptance** — every numbered item across 7 rounds either fixed inline or filed as a labeled GitHub issue (#475, #476, #477, #478).
2. ✅ **Build passing** — CI green on \`ed36c80\`; \`1bc7b92\` is +49/-5 of comment + 1-line state fix + 1 test (low surface).
3. ✅ **cr feedback considered** — all 7 rounds + 7 cycles of nits addressed. Convergence: rounds 1-2 had blockers; rounds 3-4 had largest gaps (handler factory + tests); rounds 5-6 were envelope/log polish; round 7 was a single state-leak fix + 3 doc nits.
4. ✅ **/simplify clean** — initial 3-agent pass returned no actionable findings; subsequent rounds were test + small refactor + comment work.

2569/2569 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`1bc7b92\`.

justin-layerv:
Round 8 — commit \`1682412\` addresses the must-fix doc + GC leak.

### Must-fix (per cr)

- ✅ **#1 — config.js "worker tier" doc**: fixed. Now reads "the HTTP tier (same process that owns the webhook receiver and the live monitorLinkStatus instances) drains the queue and dispatches..." Eliminates the IAM-misdirection footgun.

### Fixed in PR (worth the lift)

- ✅ **#2 — addRecipients GC leak surface**: \`if (stopped) return;\` at the top of \`addRecipients\`. Closes both the new (post-#60) leak (registered callbacks never unregistered) AND the pre-#60 leak (expectedCount + linkStatus mutated on a dead monitor). New handler test "stopped handler does not mutate linkStatus, viewed, safeEdit, or onAllDone" is the backstop at the handler layer.
- ✅ **#4 — TODO marker now linked**: \`TODO(view-update-dedup)\` → \`TODO(view-update-dedup, #476)\`. #476 tracks the envelope-contract module extraction; dedup naturally lands there.

### Acknowledged (cr → follow-up or no-action)

- 🟢 **#3 — handler bumps viewed when hasInteraction=false**: pinned by comment + the existing \`onAllDone DOES fire\` test. A dedicated test for the asymmetry would be future-refactor-defensive but the invariant is already in-tree.
- ✅ **#5 — DLQ in infra companion**: already provisioned in PR #758. \`aws_sqs_queue.view_updates_dlq\` (maxReceiveCount=3) + \`view_updates_dlq_nonempty\` CloudWatch alarm. Linked in the PR body's rollout section.
- 🟢 **#6 — combined-mode self-publish/receive**: cr noted as benign (\`status === 'opened'\` idempotency handles it). The boot-requirements test pins the absence of a rejector helper, which is the contract that matters.
- ⏸️ **#7 — CloudWatch alarm on the publish-fail log line**: extended #478 with the metric filter shape (\`{ $.kind = "viewUpdatePublishFail" }\`). Folds into the same observability follow-up as the latency metric.

### Final-check tally (rounds 1-8)

1. ✅ **DE acceptance** — every numbered item across 8 rounds either fixed inline or filed as a labeled GitHub issue (#475, #476, #477, #478).
2. ✅ **Build passing** — CI green on \`1bc7b92\`; \`1682412\` is +36/-4 of comment + 1-line guard + 1 test.
3. ✅ **cr feedback considered** — all 8 rounds + 8 cycles of nits addressed. Convergence pattern: rounds 1-2 surfaced blockers (200/500 bug, pollLoop tight-spin); rounds 3-4 surfaced architectural gaps (handler extraction, integration tests); rounds 5-8 sharpened safety rails (accessCount parity, GC leak, running-state, doc consistency).
4. ✅ **/simplify clean** — initial 3-agent pass before round-1 returned no actionable findings; subsequent rounds didn't add logic surface for simplify to re-fire on.

2570/2570 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`1682412\`.

justin-layerv:
Round 9 — commit \`5e6530b\` addresses all 4 actionable items.

### Code quality / correctness

- ✅ **#1 — addRecipients-after-stop direct test**: added \`"addRecipients() called AFTER stop() is a strict no-op"\` to \`send-pipeline-back-half.test.js\`'s \`addRecipients() + stop() races\` block. Asserts that a post-stop \`addRecipients\` call doesn't change \`monitor.getFullMsg()\` — would catch a regression that drops the early-out at the top of \`addRecipients\`. The handler-layer backstop in \`view-update-handler.test.js\` stays as defense in depth.
- 🟢 **#2 — \`err.stack\` log volume on sustained outage**: kept the stack field for sibling parity with \`event-publisher.js:242, 261\` (also includes \`err?.stack\`). Tightened from \`err.stack\` to \`err?.stack\` for non-Error rejection safety. Parity wins over volume-reduction; if log volume becomes a real cost surface, a follow-up could touch both modules together.
- 🟢 **#3 — onAllDone hoist as contract-by-test**: cr accepted; behavior is defensive against a future refactor. Kept.
- ✅ **#4 — \`consumed\` plumbing**: TODO marker already references #476 (envelope contract module follow-up). Where dedup naturally lands, \`consumed\` semantics get re-examined.
- ✅ **#4 (separate) — \`pollOnce\` safe-nav on \`stopController.signal\`**: tightened to \`stopController?.signal\`. Matches the posture elsewhere in the same file. The test helper \`_setStopControllerForTest\` lets tests drive pollOnce without start(); the safe-nav makes \`pollOnce\` callable without a controller wired (low risk in practice, free defense).

### Comment style (cr round-9 general note)

- ✅ **Stripped \`cr round-N #M\` markers** from all comments I added across rounds 1-8. 5 in src files, 5 in test files. Replaced with semantic descriptions of what the code defends against. Per cr's point: review-round numbers rot, the WHY persists.
- Pre-existing markers in \`qurl-send-map.test.js\` (round-13, round-15) and \`qurl-webhook-registrar.test.js\` (round-8) are from earlier PRs — out of scope for this PR to touch.

### Nits acknowledged

- AWS_REGION boot-time vs first-call: consistent with \`event-publisher.js\` / \`event-consumer.js\`. A boot-test would touch all three modules together.
- \`getButtonRow()\` test uses non-null replacement instead of null: belt-and-suspenders for a refactor that doesn't exist; the null case is gated by \`isStopped()\` (tested separately under "early returns").
- \`onFatal\` contract test simulates the order rather than exercising the real \`pollLoop\` fatal branch: cr accepted as honest framing.

### Final-check tally (rounds 1-9)

1. ✅ **DE acceptance** — 9 rounds of cr; every numbered item fixed inline or filed as a labeled GitHub issue (#475, #476, #477, #478).
2. ✅ **Build passing** — CI green on \`1682412\`; \`5e6530b\` is +62/-30 of test additions + safe-navs + comment polish.
3. ✅ **cr feedback considered** — 9 rounds, all addressed. The convergence trajectory:
   - Rounds 1-2: behavioral blockers (200/500 bug, pollLoop tight-spin)
   - Rounds 3-4: architectural (handler factory, state-matrix tests)
   - Rounds 5-6: safety rails (accessCount parity, log-pivot decouple, render-state asymmetry docs)
   - Rounds 7-8: state-leak fixes (running-state, addRecipients GC, config doc)
   - Round 9: test gap closure + comment-rot scrub
4. ✅ **/simplify clean** — initial 3-agent pass before round-1; no new logic surface in subsequent rounds.

2571/2571 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`5e6530b\`.

justin-layerv:
Round 10 — commit \`310ed40\` addresses all behavioral + doc items.

### Fixed in PR

- ✅ **#1 — \`consumed\` plumbed comment**: tightened to "plumbed end-to-end but discarded at the handler today". Removes the overstated claim that consumed drives linkStatus.
- ✅ **#3 — async onFatal callback**: \`pollLoop\` now \`await\`s \`onFatalCb(err)\`. Production wiring's gracefulShutdown is async — an unawaited reject would have surfaced as UnhandledRejection. await wraps both sync-throw and async-reject shapes.
- ✅ **#4 — \`stop()\` contract on already-stopped path**: documented inline. When pollLoop's fatal branch flipped running=false before onFatalCb → gracefulShutdown → stop re-entry, the early return skips \`await loopPromise\` because the loop has already unwound. Restructuring the function would change semantics; the contract documentation is the right scope.
- ✅ **#5 — \`INITIAL_DRAIN_DEADLINE_MS\` snapshot pattern**: restored. \`_resetStateForTest\` now reads the const (captured at module-load) instead of re-reading config. Matches event-publisher.js's posture; dodges the test-config-mock-drift footgun.
- ✅ **#6 — TDZ comment fix**: dropped buildStatusMsg from the affected-bindings list (function declarations are hoisted; TDZ risk is on the \`let\` bindings only). Per cr's "comment reads correct in spirit, but inclusion of buildStatusMsg muddies the story."

### Acknowledged (cr → no-action or already-deferred)

- 🟢 **#2 — pollLoop fatal-branch direct exercise**: cr accepted the "test pins the order" framing. The simulation catches a swap of the two assignment lines (which is the regression-class the test pins); a deeper test would require mocking pollOnce to throw across an await, which cr called "heavier setup."
- 🟢 **#7 — \`update\` shape at handler boundary**: cr noted as defensive observation. The consumer's parse path already type-checks; an additional validate-at-handler would be belt-and-suspenders for the trivial-cost case.
- 🟢 **#8 — DeleteMessageBatch Failed[] log volume**: bounded by \`MAX_MESSAGES_PER_RECEIVE = 10\`. Worth revisiting only if that cap changes.
- ⏸️ **#9 — \`event_id\` constraint when dedup layer lands**: rolls into #476 (envelope contract module). TODO marker already references the issue.

### Final-check tally (rounds 1-10)

1. ✅ **DE acceptance** — 10 rounds, every numbered item fixed inline or filed as a labeled GitHub issue (#475 / #476 / #477 / #478).
2. ✅ **Build passing** — CI green on \`5e6530b\`; \`310ed40\` is +31/-13 of one await + one snapshot + 3 comment edits.
3. ✅ **cr feedback considered** — 10 rounds, all addressed. The convergence trajectory: rounds 1-2 (behavioral blockers) → 3-4 (architectural gaps + factory extraction) → 5-6 (safety rails + envelope contracts) → 7-8 (state leaks) → 9 (test coverage) → 10 (async-await + sibling-parity refinements).
4. ✅ **/simplify clean** — initial 3-agent pass returned no actionable findings; subsequent rounds were polish work without new logic surface.

2571/2571 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`310ed40\`.

justin-layerv:
Round 11 — commit \`ad1455c\` reconciles the must-fix + all small items.

### Must-fix (cr's "before-merge" item)

- ✅ **#1 — accessCount parity at webhook boundary**: \`qurl-webhook.js\` tightened from \`< 0\` to \`<= 0\`. qurl.accessed events always carry access_count >= 1 by contract; rejecting 0 at the wire boundary makes that the single source of truth. New test "returns 200 invalid-payload when access_count is 0 (parity with publisher gate)" asserts both the response shape AND that neither \`recordQurlView\` nor view-update publish fires.

### Fixed in PR

- ✅ **#2 — stop() drain-race comment**: tightened to acknowledge single-threaded JS doesn't have an actual race here; the shape is kept "for parity with event-publisher.js so a future drain-hardening pass reads both modules identically." Previous wording read like a real bug.
- ✅ **#3 — \`inFlightSends.add\` divergence**: one-paragraph comment at the call site explaining the intentional shape (track the post-cleanup chain vs raw send like event-publisher.js does). Both shapes drain correctly; this trades sibling-symmetry for tidier post-allSettled state.
- ✅ **Id contract pin** (smaller nit): inline comment on \`deleteMessageBatch\` documenting the SQS \`Id: [a-zA-Z0-9_-]+\` + batch-uniqueness requirement satisfied by \`String(i)\` for \`0..MAX_MESSAGES_PER_RECEIVE-1\`.
- ✅ **Hoist \`viewUpdatePushEnabled\` once per monitor**: \`const\` outside the closure instead of re-reading \`config.ENABLE_VIEW_UPDATE_PUSH\` on each \`registerViewUpdateFor\` call (up to 50 times per monitor). Negligible cost; reads cleaner.

### Acknowledged (cr → no action)

- 🟢 **#4 — Hit rate scales as 1/N**: rollout runbook concern. #478 (latency observability) already tracks per-replica hit rate so a scaling event that erodes the optimization becomes visible.
- 🟢 **#5 — Test env \`AWS_REGION\` leak**: cr accepted as harmless + consistent with older suites.
- 🟢 **\`update && update.accessCount > 0\` dead branch in handler**: cr accepted as defensive + consistent with publisher's posture.

### Final-check tally (rounds 1-11)

1. ✅ **DE acceptance** — 11 rounds, every numbered item fixed inline or filed as a labeled GitHub issue (#475 / #476 / #477 / #478).
2. ✅ **Build passing** — CI green on \`310ed40\`; \`ad1455c\` is +59/-11 across the webhook gate + 4 comment edits + 1 hoist + 1 new test.
3. ✅ **cr feedback considered** — 11 rounds, all addressed. Trajectory: rounds 1-2 behavioral blockers → 3-4 architectural + tests → 5-6 envelope safety → 7-8 state leaks → 9 test coverage → 10 async-await + parity → 11 wire-boundary parity.
4. ✅ **/simplify clean** — initial 3-agent pass; subsequent rounds were polish without new logic surface.

2572/2572 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`ad1455c\`.

justin-layerv:
Round 12 — commit \`614feb1\`. cr verdict was "LGTM modulo #1 and #3". Both addressed:

### Fixed in PR

- ✅ **#1 — Handler \`.catch\` on safeEdit**: kept the catch but added an inline comment documenting it as defense-in-depth (catch only fires if \`logIgnoredDiscordErr\` itself throws). cr's "drop the .catch and the test" option would loosen the contract; the comment option preserves the guard.
- ✅ **#2 — \`accessCount > 0\` string-coercion**: tightened to \`Number.isSafeInteger(update.accessCount) && update.accessCount > 0\`. Symmetric with the consumer's parse-time gate; blocks \`"3" > 0\` truthy-coercion regressions.
- ✅ **#3 — Tautological pollLoop test**: deleted per cr's option (a). The contract is one line of production code (running = false BEFORE onFatalCb in pollLoop's catch) — verifiable by inspection. Attempted a real-throw exercise via mocked logger.warn but the microtask-ordering for catch dispatch was brittle in the test environment. Replaced the test with an inline comment pointing at the source contract.
- ✅ **#5 — TDZ comment density**: trimmed from 11 lines to 4. The long-form explanation was working harder than the WHY needed.

### Acknowledged (no action)

- 🟢 **#4 — Webhook \`access_count: 0\` → invalid-payload**: rollout runbook tightening (not a regression). qurl.accessed carries \`>= 1\` by contract.
- 🟢 **#6 — \`_setStopControllerForTest\` exposure**: cr accepted as live-with-it given the long-poll abort plumbing.

### Final-check tally (rounds 1-12)

1. ✅ **DE acceptance** — 12 rounds; every numbered item fixed inline or filed as a labeled GitHub issue (#475, #476, #477, #478).
2. ✅ **Build passing** — CI green on \`ad1455c\`; \`614feb1\` is +23/-33 of test cleanup + 1-line integer check + comment polish.
3. ✅ **cr feedback considered** — all 12 rounds. Trajectory: rounds 1-2 blockers → 3-4 architectural → 5-6 envelope safety → 7-8 state leaks → 9-10 test coverage + async-await + parity → 11 wire-boundary parity → 12 handler tightening + test cleanup.
4. ✅ **/simplify clean** — initial 3-agent pass; subsequent rounds were polish without new logic surface.

2571/2571 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`614feb1\`.

justin-layerv:
Round 13 — commit \`5b5dc52\`. Merge-gate verified at source; all nits addressed.

### Merge-gate ask (#1) — **VERIFIED**

cr asked to confirm \`access_count: 0\` isn't a qurl-service contract case. Pulled the source:

- \`qurl-service/internal/service/resolve_service.go\`: \`AccessCount: rc.token.UseCount + 1\`. The webhook carries the count AFTER the access, so the minimum emitted value is \`0 + 1 = 1\`. qurl-service never sends \`access_count: 0\` for \`qurl.accessed\`.
- \`qurl-service/internal/api/gen/api.gen.go\` contract spec confirms: \`"access count after this access"\`.

The tightening from \`< 0\` to \`<= 0\` is contract-safe.

### Nits fixed

- ✅ **#4 — stopController unaborted in fatal-path stop()**: paragraph at the early-return site explains the stale controller is harmless (no live listeners post-fatal; a subsequent start() replaces it).
- ✅ **#5 — DeleteMessageBatch partial-failure log noise**: TODO marker referencing #475 (sqs-helpers extraction). If soak shows the same message redelivering until retention expires, the dedup-on-f.Code shape change lands there alongside the helper extraction.
- ✅ **#6 — \`.catch\` relies on safeEdit async-return**: note added pointing at the contract and the paranoid-safe \`Promise.resolve(safeEdit(...))\` wrap if a future refactor switches to sync-return.
- ✅ **#7 — boot-requirements absence-test pinning**: paragraph at \`missingViewUpdatePushKeys\` documenting the deliberate absence of a combined-mode rejector. Survives a rename in a way the test alone can't.

### Acknowledged (no action)

- 🟢 **#2 — \`inFlightSends\` unbounded**: cr framed as follow-up. Matches event-publisher.js's posture. A cap + circuit-breaker pair would benefit both modules; rolls into #475 (sqs-helpers extraction) where shared hardening lives.
- 🟢 **#3 — \`LOG_KINDS\` pivot in rollout runbook**: belongs in #478 (latency observability) — the pivot is named + commented in-tree.

### Final-check tally (rounds 1-13)

1. ✅ **DE acceptance** — 13 rounds. Every numbered item fixed inline OR filed as a labeled GitHub issue (#475, #476, #477, #478). Merge-gate verified at the upstream source.
2. ✅ **Build passing** — CI green on \`614feb1\`; \`5b5dc52\` is +26/-0 (comments only, no logic changes).
3. ✅ **cr feedback considered** — all 13 rounds. Trajectory: 1-2 behavioral blockers → 3-4 architectural → 5-6 envelope safety → 7-8 state leaks → 9-10 test coverage + async-await → 11 wire-boundary → 12 handler tightening → 13 source-contract verification + final nits.
4. ✅ **/simplify clean** — initial 3-agent pass; no new logic surface added since round 3.

2571/2571 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`5b5dc52\`.

justin-layerv:
Round 14 — commit \`fe2dd3a\` addresses all 8 items, including both cr's "before flag-flip step" callouts (#3 + #6). Since the flag defaults true per user direction, both must land in PR — done.

### Bug-class / before-flag-flip

- ✅ **#3 — Drop \`consumed\` from envelope**: deleted end-to-end. Publisher signature, envelope construction, strict-boolean-coerce, consumer dispatch payload, webhook publish call, all 4 test files. Pins the absence with \`not.toHaveProperty('consumed')\` on the envelope-shape test. Re-introduce in #476 (envelope contract module) when a render path needs it.
- ✅ **#6 — Parallel drain in gracefulShutdown**: \`Promise.all([viewUpdateConsumer.stop(), viewUpdatePublisher.stop()])\` eliminates 3s of worst-case sequential drain inside the 10s SIGTERM budget. Order-independence is safe (publisher snapshots in-flight; consumer aborts the long-poll; neither depends on the other).
- ✅ **#5 — Backpressure cap**: \`MAX_INFLIGHT_SENDS = 1000\`. Publisher drops-with-log when the cap is reached. Mirrors event-consumer's \`QURL_BOT_MAX_INFLIGHT_HANDLERS\` posture. New test "drops with log when inFlightSends cap reached" pins the shape via a never-resolving mock send.

### Design polish

- ✅ **#1 — Trim unreachable-path hoist framing**: \`view-update-handler.js\`'s hasInteraction-vs-onAllDone block rewritten from 17 lines of "unreachable today / future refactor" framing to 5 lines describing the invariant directly. Per cr's "either delete the hoist + comment, or call it the invariant."
- ✅ **#2 — \`Promise.resolve(safeEdit(...)).catch\` hardening**: wraps the safeEdit call so a future sync-return refactor doesn't surface as \`.catch is not a function\`. Closes the fragility rather than documenting it.
- ✅ **#4 — Cross-module PR-round refs**: deleted the last remaining cr-round-12 simulation test (lived only to pin order via inspection; one-line production contract verifiable at source).

### Acknowledged (no action)

- 🟢 **#7 — sendPromise self-reference comment**: cr accepted the inline comment as documentation; no change needed.
- 🟢 **#8 — Registry footprint at scale (1M entries worst-case)**: rollout runbook concern; rolls into #478 (latency observability + scaling metrics).

### Final-check tally (rounds 1-14)

1. ✅ **DE acceptance** — 14 rounds; every numbered item fixed inline or filed as a labeled GitHub issue (#475, #476, #477, #478).
2. ✅ **Build passing** — CI green on \`5b5dc52\`; \`fe2dd3a\` is +95/-99 (envelope schema cleanup + backpressure cap + parallel drain + tests).
3. ✅ **cr feedback considered** — all 14 rounds addressed. Trajectory: 1-2 behavioral blockers → 3-4 architectural + tests → 5-6 envelope safety → 7-8 state leaks → 9-10 test coverage + async-await → 11 wire-boundary → 12 handler tightening → 13 source-contract verification → 14 envelope hygiene + drain budget.
4. ✅ **/simplify clean** — initial 3-agent pass; no new logic surface added since round 3.

2570/2570 tests passing; eslint --max-warnings 0 clean.

Ready when CI clears on \`fe2dd3a\`.

PR #481, opened 2026-05-20
feat(slack): add once:true flag on /qurl get for one-time-use links

Kevin-layerV:
## Summary

Adds a `once:true` flag to the Slack bot's `/qurl get` command that mints a one-time-use qURL — the link burns on first redemption.

The qurl-service API has long supported `one_time_use: bool`, and the Go SDK already plumbs it (`shared/client/client.go` — `CreateInput.OneTimeUse` with `omitempty`). The Slack bot was the only surface still missing the entry point; this wires the flag through the parser, the mint request, and the reply copy.

### Behavior

- `/qurl get <url|$name> once:true` — mints a single-use qURL; success reply ends with ` (one-time use)` so the requester sees what they shipped.
- `/qurl get <url|$name> once:false` — accepted (same runtime effect as omitting the flag), no copy change.
- `/qurl get <url|$name>` (default) — unchanged; the SDK's `omitempty` keeps `one_time_use` off the wire.

### Strict-value parsing (mirrors `dm:`)

- `once:yes` / `once:1` / `once:please` / `once:` (empty) → reject with `ErrInvalidFlag` and a friendly "use once:true or omit the flag" message.
- Without this gate, a typo'd value would silently no-op (the accessor case-equals against `"true"`), which is exactly the failure mode the parser carefully rejects for typo'd flag keys.

### Composability

Stacks with `dm:` and `reason:` in any order. `/qurl get $name once:true dm:true reason:"incident #123"` is valid.

### Out of scope

Channel-policy default (`once_default`), custom max-uses (`uses:<n>`), and any `qurl-integrations-infra` side changes.

## Test plan

- [x] `go test ./apps/slack/... ./shared/...` — passes
- [x] `go test ./apps/slack/internal/ -race` — passes
- [x] `go vet ./...` — clean
- [x] `golangci-lint run --new-from-rev=origin/main ./apps/slack/...` — 0 issues
- [x] `pre-commit run --files <every changed file>` — passes
- Added 8 new parser-side test rows (happy-path + rejects + composition + `Once()` accessor) and 2 new handler-side test fns (`one_time_use` JSON wire shape + reply-copy suffix presence/absence).

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -752,7 +752,7 @@ func (h *Handler) helpMessage() string {
 		lines = append(lines, "• `/qurl get <url|$name> dm:true` — DM the link to you instead of posting it in-channel")
 	}
 	lines = append(lines,
-		"• `/qurl get <url|$name> once:true` — Mint a single-use link; the first redemption burns it",
+		"• `/qurl get <url|$name> once:true` — Get a single-use qURL; the link burns on first redemption",
 		"• `/qurl get <url|$name> reason:\"…\"` — Annotate the audit log with a reason",
 		"• `/qurl list` — Show your 5 most recent qURLs",
 		"• `/qurl setup` — Connect qURL to your Slack workspace (workspace admin only)",
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -752,7 +752,7 @@ func (h *Handler) helpMessage() string {
 		lines = append(lines, "• `/qurl get <url|$name> dm:true` — DM the link to you instead of posting it in-channel")
 	}
 	lines = append(lines,
-		"• `/qurl get <url|$name> once:true` — The link works exactly once; redemption burns it",
+		"• `/qurl get <url|$name> once:true` — Mint a single-use link; the first redemption burns it",
 		"• `/qurl get <url|$name> reason:\"…\"` — Annotate the audit log with a reason",
 		"• `/qurl list` — Show your 5 most recent qURLs",
 		"• `/qurl setup` — Connect qURL to your Slack workspace (workspace admin only)",
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -752,6 +752,7 @@ func (h *Handler) helpMessage() string {
 		lines = append(lines, "• `/qurl get <url|$name> dm:true` — DM the link to you instead of posting it in-channel")
 	}
 	lines = append(lines,
+		"• `/qurl get <url|$name> once:true` — The link works exactly once; redemption burns it",
 		"• `/qurl get <url|$name> reason:\"…\"` — Annotate the audit log with a reason",
 		"• `/qurl list` — Show your 5 most recent qURLs",
 		"• `/qurl setup` — Connect qURL to your Slack workspace (workspace admin only)",
--- a/apps/slack/internal/handler_get.go
+++ b/apps/slack/internal/handler_get.go
@@ -246,6 +246,7 @@ func (h *Handler) getWork(ctx context.Context, log *slog.Logger, args getWorkArg
 
 	input := client.CreateInput{
 		Reason:         args.cmd.Reason(),
+		OneTimeUse:     args.cmd.Once(),
 		IdempotencyKey: IdempotencyKey(args.teamID, args.channelID, args.userID, args.triggerID),
 	}
 
@@ -302,6 +303,12 @@ func (h *Handler) getWork(ctx context.Context, log *slog.Logger, args getWorkArg
 	}
 
 	message := ":link: *qURL ready:* " + out.QURLLink
+	if args.cmd.Once() {
+		// Make the single-use posture visible to the requester so they
+		// see what they shipped — a link that burns on first redemption
+		// behaves materially differently from the default.
+		message += " (one-time use)"
+	}
 	if args.cmd.DM() {
 		return h.deliverGetDM(ctx, log, args.userID, message), nil
 	}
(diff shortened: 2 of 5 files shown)


Kevin-layerV:
Thanks — addressed all three in `bda83c9`.

**Nit 1 (`flagKeyOnce` asymmetry).** Tried option (a) inline first — golangci's `goconst` rule trips on `"once"` (7 new occurrences across parser.go + parser_test.go, threshold is 3) when run with `--new-from-rev=origin/main`. Option (b) lifting `dm`/`reason` to constants would dedupe but violates the surgical-changes guidance in `CLAUDE.md` (don't refactor adjacent pre-existing code). Kept the const, trimmed its doc to the one load-bearing line.

**Nit 2 (help-line phrasing).** Rewritten as `"Mint a single-use link; the first redemption burns it"` to match the imperative voice of the surrounding lines.

**Nit 3 (redundant `once:` empty-bare-value row).** Dropped; the key-agnostic `empty bare value` row above already covers that branch.

Also took a `/simplify` pass on the diff and trimmed several over-narrated doc comments (preambles on `TestCreateInputJSON_Once*`, the `Once()` accessor, the `applyFlag` arm, the composition test row, `once:false` row, `onceRejectSubstr` const) — net `-38` lines.

Kevin-layerV:
Thanks — round-2 triage:

**Obs #1 (`Once()` called twice).** Acknowledged as intentional symmetry with `DM()`; no change.

**Obs #2 (loose `Contains` assertion).** Tightened in `2d7e1fb` — `TestCreateInputJSON_OnceTrue` now uses `strings.HasSuffix(strings.TrimSpace(async), "(one-time use)")`. The absence test keeps `Contains` because absence-anywhere is the stronger property.

**Obs #3 (idempotency-key + `OneTimeUse`).** Not filing a follow-up; the design comment in `apps/slack/internal/idempotency.go:26-30` already explains the relevant invariant. `trigger_id` rotates per slash invocation (Slack-side, before the user submits), so two distinct user-issued `/qurl get` calls always hash to distinct keys regardless of `(team, channel, user)` repeat. The qurl-service idempotency cache therefore can't see a same-key second request with a different `one_time_use` value unless the bot itself retried — and the bot captures `args.cmd.Once()` once into `client.CreateInput.OneTimeUse` before the upstream call, so a bot-side retry carries the same flag value. The cross-trigger collision path is structurally absent by design; nothing to widen on the slack side without first changing the trigger_id contract.

**Obs #4 (`once:false` not yet useful).** Per the PR body — `once_default` channel-policy is out of scope. The strict gate accepts the literal so users can opt out without an `unknown flag` error today, in anticipation of the policy default landing later.

Kevin-layerV:
Thanks — round-3 triage (commit `87930ab`):

**1. Help-line voice.** Applied — `"Get a single-use qURL; the link burns on first redemption"`. Matches the opening-verb of the first two bullet entries (`Get a qURL for…`) and uses the brand-spelled `qURL` per `CLAUDE.md`.

**2. `once:false` wire-shape gap.** Added `TestCreateInputJSON_OnceFalse` — sits beside the existing `_OnceTrue` / `_OnceAbsent` pair and pins the explicit-false parser branch (Flags["once"]="false" → `Once()` returns false → `omitempty` drops the zero). Distinct from `_OnceAbsent` because that path never populates the flag at all.

**3. DM + once handler composition.** Added `TestHandleGet_OnceTrueDMTrue` — fences that the `(one-time use)` suffix lands in the DM payload (and not in the channel ephemeral) when both flags stack. Modeled on `TestHandleGet_DMVariantPostDMSuccess`.

**4. `flagKeyOnce`-only-constant inconsistency.** Leaving as-is per your explicit "not blocking." The in-tree comment carries the goconst rationale. Lifting `dm`/`reason` to sibling constants would dedupe but violates the CLAUDE.md surgical-changes guidance — happy to file a follow-up if you'd rather it not linger.

Kevin-layerV:
Thanks — round-4 triage (commit `33e70a4`):

**Obs #1 (empty-value coverage).** Applied. Added explicit `once:` and `once:""` empty-rejection rows in `TestParse_GetFlagErrors`. This reverses round-1's "drop the redundant row" note — round-4's refactor-resistance rationale (per-key validation could move in front of the shared empty-value gate) is more durable than round-1's "covered by the key-agnostic branch" framing.

**Obs #2 (double `Once()` call).** Acknowledged as "take or leave"; leaving for symmetry with the established `DM()` two-call pattern in the same function.

**Obs #3 (`dm` bare literal vs. `flagKeyOnce` const).** Acknowledged as "out of scope; flagging for awareness." Not in this PR.

Kevin-layerV:
Thanks — round-5 triage. All three are flagged optional by you; not pushing further changes.

**Nit #1 (flagKeyOnce placement & const-asymmetry).** Leaving as-is. The 7-occurrence count goconst flags spans tests + production (parser_test.go map keys × 5, parser.go accessor + switch case × 2); the format-string substring isn't one of them. Inline-and-nolint would replace the structural fact with a comment carrying the same constraint. Keeping it grouped near the other package-level identifiers in the file.

**Nit #2 (error-message copy doesn't mention once:false accepted).** Leaving — the consistency with `dm:`'s reject copy is load-bearing for grep-ability and you flagged it as fine.

**Nit #3 (help-line tone "burns on first redemption").** Leaving — wording has bounced across three review rounds (round-1 wanted imperative, round-3 wanted "Get a qURL"-prefixed, round-5 finds it slightly informal). Staying with the round-3-converged copy since you marked round-5's call as "either way works."

**Composition wire-shape gap (`once:true reason:"…"`).** Not adding. Parser composition is fenced by `TestParse_HappyPaths` "get with once, dm, and reason flags", and `CreateInput` serialization is plain `encoding/json` per-field — no realistic failure mode where individual-field wire tests + parser composition test both pass and a stacked-flags wire test would catch a regression.

PR is at the convergence point. Final-check status:
- DE acceptance: green (five "ship/approve/merge-ready" notes)
- CI: claude-review green, Lint/Test/Vet/CodeQL/Vulnerability/Analyze all green, Docker still building
- cr feedback: every round considered + replied
- /simplify: converged (re-ran on the final state — no actionable findings)

Issue #265, opened 2026-05-12
SetAPIKey: switch warn-on-overwrite to PutItem + ReturnValues=ALL_OLD (race-free + half the RTTs)


Kevin-layerV:
## Summary

`shared/auth/DDBProvider.SetAPIKey` currently does:

1. `GetItem` (read previous row to preserve `configured_at` + decide overwrite warning)
2. `PutItem` (write new row)

This is two RTTs and is racy under concurrent installs (two parallel calls both observe "no existing row", both write, neither warns).

The DynamoDB-native pattern is `PutItem` with `ReturnValues: ALL_OLD`, which:
- Atomically swaps the item.
- Returns the prior item's attributes in one RTT.
- Avoids the inter-step race window.

## Acceptance criteria

- Replace the `GetItem`-then-`PutItem` flow with a single `PutItem` carrying `ReturnValues: types.ReturnValueAllOld`.
- Use `Attributes` from the response to recover the prior `configured_at` and emit the overwrite warn.
- If `Attributes` is empty, this is a first install: use the current timestamp.
- Update `TestDDBProviderSetAPIKey` and `TestDDBProviderSetAPIKeyPreservesConfiguredAt` to match the new flow (no `getOutput` setup needed).
- No behavior change visible from `APIKey()`.

## Why this is a follow-up

PR #254 added the warn-on-overwrite + configured_at-preservation logic with the two-step pattern as an explicit trade-off (cited in the code comment). The race is real but improbable on the install path. Worth fixing for correctness + RTT, but not blocking the PR.

Surfaced by /pr-ready loop on PR #254 (cr review item #4 + simplify efficiency review).

PR #482, opened 2026-05-20
feat(slack): collapse /qurl setup + admin claim — installer becomes seed admin

Kevin-layerV:
## Summary

Today the Slack bot has two separate bootstrap commands. `/qurl setup` runs OAuth, mints a qURL API key, and stores it in `workspace_keys`. `/qurl admin claim` takes an out-of-band bootstrap code, redeems it, and writes the `workspace_mappings` row that seeds the admin set. Until both run, the workspace is half-installed: `/qurl admin add @user` replies "you are not an admin" because no admin row exists.

The two-actor mental model (qURL account owner ≠ Slack workspace admin, with a bootstrap code as the cross-system handshake) doesn't match how anyone actually installs the bot. In practice it's the same person, and the handshake is theater that produces a confusing failure mode.

This PR collapses both steps into one. After OAuth, the installer is the admin. The bootstrap-code path is removed entirely.

## Behavior

| Scenario | Old | New |
|---|---|---|
| First /qurl setup → /qurl admin add @bob | "you are not an admin" until claim runs | works — installer is the seed admin |
| /qurl setup re-run by same admin | rotates API key, admin set unchanged | rotates API key, admin set unchanged (idempotent) |
| /qurl setup by a *different* Slack user against an already-bound workspace | API key overwritten silently | rebind-refused page; existing admin's stored API key untouched |
| BindWorkspace fails before mint | n/a (path didn't exist) | 500 response (no key minted, no DDB row touched) |
| id_token sub claim verification fails | n/a | 500 response (no key minted; bind requires the OwnerID) |

## Key changes

- `apps/slack/internal/oauth/callback.go` — handler runs `BindWorkspace` BEFORE `MintAPIKey`/`SetAPIKey` so a rebind-refused install can't overwrite the existing admin's encrypted `qurl_api_key` row. The Auth0 `sub` claim is already in hand pre-mint, so there's no ordering blocker. Bind conflicts are classified via `cfg.BindClassifyError` into `BindConflictAlreadyBoundToCaller` (continue to mint = key rotation), `BindConflictAlreadyBound` / `BindConflictUnverified` (rebind-refused page; nothing minted), or default-classify failure (generic 500; nothing minted). The handler is split into `validateCallbackRequest` / `verifyIDTokenClaims` / `checkBindAllowed` / `mintAndPersist` to keep `Callback` straight-line under the gocognit threshold.
- `apps/slack/internal/oauth/jwks.go` — adds `VerifySub` alongside `VerifyEmail`. Both go through a shared `verifiedToken` parse.
- `apps/slack/cmd/main.go` — wires the `slackdata.Store` as the `AdminStore` and supplies `classifyBindError`. JWKS prime now fails-fast at boot when `AdminStore` is wired (mismatched degradation between subsystems was the actual bug class); the sandbox / no-DDB path still warns-and-continues so the API-key surface stays functional. `adminStoreAdapter` construction moved into the caller so `buildOAuthConfig` takes `oauth.AdminStore` directly.

## Removed

- `apps/slack/internal/handler_admin_claim.go` + `handler_admin_claim_test.go` — modal entry, `RedeemBootstrap` call, `surfaceBindError` / `surfaceClaimError` chain.
- `apps/slack/internal/slackdata/bootstrap_codes.go` — `RedeemBootstrap`, `hashBootstrapCode`, `MinBootstrapPlaintextLen`, `ErrCodeBootstrapInvalid`, `ErrCodeBootstrapRowMalformed`.
- `apps/slack/internal/slackdata/external_identity_bindings.go` — the post-redeem qurl-service hop; its only consumer was the redeem chain.
- `AdminClaimModal()`, `callbackIDAdminClaim`, `blockIDClaimCode`, `actionIDClaimCode`, `IsRedactedSubmissionBlock` (no other redacted block remains).
- `AdminClaim` action constant + corresponding parser cases.
- `QURL_BOOTSTRAP_CODES_TABLE` env-var requirement in `slackdata.NewStore` (third `WithTableNames` arg dropped).
- `/qurl admin claim` line from `helpMessage()`.
- The `admin claim` and `admin claim <args>` dispatch branches and `redactSlashCommandText` claim-prefix handling.

## Tests

- `TestCallbackSeedsAdminOnBind` — successful callback calls BindWorkspace with `TeamID = verified.TeamID`, `OwnerID = id_token sub`, `seedAdmin = verified.UserID`.
- `TestCallbackBindFailureSkipsMint` — generic BindWorkspace failure returns 500 WITHOUT minting or persisting; existing key state untouched.
- `TestCallbackBindIdempotentForSameCaller` — `BindConflictAlreadyBoundToCaller` continues to mint (rotates the key) and returns 200.
- `TestCallbackBindRefusedForDifferentAdmin` — `BindConflictAlreadyBound` renders the 409 rebind-refused page (asserts page body), skips mint, skips persist.
- `TestCallbackBindRefusedWhenUnverified` — same posture as the cross-admin arm; fences against a future split that downgrades unverified to idempotent.
- `TestCallbackBindSkippedWhenSubMissing` — id_token sub verification failure → 500 without mint or persist (no key to revoke).
- `TestCallbackSkipsBindWhenAdminStoreNil` — sandbox/no-DDB path: bind skipped, success page still renders.
- `TestBindWorkspaceSeedsAdminThatCheckAdminAccepts` (in `package internal`) — end-to-end fence using real `slackdata.Store` against `fakeDDB`: BindWorkspace's write produces a row CheckAdmin reads as admin.
- `TestBuildOAuthConfigFailsFastOnJWKSWhenAdminStoreWired` — boot-time guard against the mismatched-degradation case.
- `TestAdminStoreAdapterMappingShapesMatch` — reflect-walks both `WorkspaceMapping` types so a future field added to one and not the other breaks the build.

Deleted: all `handler_admin_claim_test.go` cases, all `*Bootstrap*` tests in `slackdata/store_test.go`, the views_test cases that exercised the modal block IDs.

## Infrastructure follow-up

The `bootstrap_codes` DDB table lives in `qurl-integrations-infra` and isn't touched here. After this rolls out, the table can be torn down — no code path reads or writes it anymore.

## Test plan

- [x] `go test ./apps/slack/... ./shared/...` — PASS
- [x] `go test ./apps/slack/... -race` — PASS
- [x] `go vet ./...` — clean
- [x] `golangci-lint run` — clean
- [x] `pre-commit run --files <changed>` — PASS


Kevin-layerV:
Thanks @claude — both reviews addressed in the latest push (a70c4eb), pivoting on the bind-before-mint reorder as the load-bearing fix.

**Blocking (cr2 #1 / cr1 "Important"): rebind-refused overwrites existing admin's key.** Reordered `BindWorkspace` ahead of `mintAndPersist` in `Callback`. On a rebind-refused outcome no key is minted and no DDB row is touched — half-install becomes impossible by construction rather than by revoke-after-the-fact. The entire orphan-revoke wiring from bind-failure paths went away; only the `persist-failure` revoke inside `mintAndPersist` remains.

The bind-failure tests now assert `store.setArgs == nil` and `minter.revoked == false` so any future refactor that re-introduces the overwrite-then-revoke pattern breaks them. Also added the `"qURL setup blocked"` page-body assertion (cr1 nit) so a future swap to bare `http.Error 409` doesn't slip past.

**cr1 stale doc references in `slackdata/store.go` + `workspace.go`.** Cleaned in the previous push (a6fd4fb).

**cr1 JWKS prime-failure warning understates impact.** `cmd/main.go` now fails-fast at boot when `AdminStore` is wired and JWKS prime fails (every callback would otherwise 500). Sandbox / no-DDB deploys (AdminStore=nil) still warn-and-continue so the API-key surface stays functional. New `TestBuildOAuthConfigFailsFastOnJWKSWhenAdminStoreWired` covers both arms.

**cr1 nits.**
- `persistTimeout` → dedicated `bindTimeout` const for the bind path: done.
- `TestCallbackBindRefusedForDifferentAdmin` page-body assertion: done.
- `adminStoreAdapter` field-drift catcher: `TestAdminStoreAdapterMappingShapesMatch` reflect-walks both `WorkspaceMapping` types so adding a field to one and not the other breaks the build.
- Sub-missing routed through `handleBindError`: moot after the reorder (no key to revoke); the sub-missing branch in `checkBindAllowed` now just 500s without revoke wiring.
- `TestCallbackSkipsBindWhenAdminStoreNil`: added.
- `BindConflictUnverified` test: `TestCallbackBindRefusedWhenUnverified` (added in a6fd4fb).

**cr1+cr2 JWKS double-parse.** Confirmed once-per-install path; skipping the interface churn for a sub-millisecond CPU win. Noted in the file-level callback.go comment that the two-method shape is intentional.

**cr2 #2 stale PR-description race narrative.** Updated the PR body — the snapshot pattern in `mintAndPersist` is now load-bearing only for the persist-failure case where `keyID` is a local that's still in scope when the goroutine reads it; the bind-failure paths don't need any keyID at all.

**cr2 #4 no Slack-workspace-admin gate.** Punting on this — explicitly out of scope for this PR (which is about removing the bootstrap-code handshake). Tracking the question in a follow-up issue if the product call goes the other way.

**cr2 other nits.**
- `handler_admin.go:53-58` unused locals — false positive on my read; the `teamID`/`userID` locals ARE consumed by the `handleAdminRevoke/Add/Remove/List` calls below, not just the `default:` slog. Left as-is.
- `helpMessage` stale PostDM comment — code at `handler.go:664` still gates `dm:true` on `h.cfg.PostDM != nil`; the comment matches the code. Left as-is.
- `workspace.go` "message variants" comment — refreshed to say "caller-side outcomes."
- `buildOAuthConfig` taking `*slackdata.Store` — refactored to take `oauth.AdminStore`; adapter construction moved into the caller.

Kevin-layerV:
Round 3 addressed in 63a95af:

**Doc refreshes (cr3 #1-#5):** `workspace.go` BindWorkspace docstring, `callback.go` `verifyIDTokenClaims` (s/seedAdminBind/checkBindAllowed), `fakeddb_fixtures_test.go`, `fakeddb_test.go`, `admin_handler_helpers_test.go` — all the stale bootstrap_codes / pre-reorder references cleaned.

**Coverage gap (cr3 #7) — bind-ok-mint-fail recovery:** `TestCallbackBindSucceedsThenMintFails` walks the two-step recovery: first attempt seeds the admin row but mint 502s (502 response, store untouched); second attempt with same-caller classifier code rotates the key and lands fully installed. Without this the new ordering's recovery path was inspection-only.

**Coverage gap (cr3 #10) — `classifyBindError` table test:** `TestClassifyBindErrorMapping` in `cmd/main_test.go` locks all 7 arms of the slackdata.Error.Code → oauth.BindConflictCode wiring. The reflect-shape test fences the struct; this fences the code-string mapping which is its own drift surface.

**verifyIDTokenClaims nil-verifier note (cr3 #8):** Added a one-line clarification that the nil-verifier branch is reachable only on the sandbox/no-DDB path post-#482, since main.go's JWKS prime fails-fast at boot in production.

**Idempotent-rotation assertion + error-text nit (cr3 #6, #9):** `TestCallbackBindIdempotentForSameCaller` now asserts `SetAPIKey.APIKey == testAPIKey` so the "key rotated" half of the promise is fenced; admin error text changed to "classified as same-caller" so a future reader doesn't think the production classifier reads the message.

**Skipped explicitly:**
- `configuredBy` overwrite confirmation (cr3 #6): the field is set via SetAPIKey's unconditional PutItem; `seed_admin_slack_user_id` on the workspace_mappings row carries the forensic attribution under `attribute_not_exists` first-write-wins. Not adding the docstring line — the workspace.go comment block on `attrSeedAdminSlackUser` already documents this.
- `bindDisambiguationBudget` operator-page note (cr3 #3): agreed signal worth flagging; not adding a code comment for this round, would belong in the runbook.
- JWKS double-parse: agreed to skip, file-level intent comment is in place.

Also addressed from the prior DE re-read (not in cr3):
- `structFieldSet` tightened from `Kind` to `Type.String()` so a future drift like `OwnerID OwnerID` vs `OwnerID string` fails the reflect-shape test instead of the build line.
- Dropped redundant `minter.revoked==false` assertions on bind-failure tests (no key minted post-reorder; sync read against a maybe-spawned goroutine was a flake risk). `setArgs==nil` is the deterministic fence.

Kevin-layerV:
cr round 4 addressed in 2a28a18.

**Real coverage gaps (cr4 sub 1-2 / cr4-alt sub 1):**
- `jwks_test.go`: 4 new `TestJWKSVerifierVerifySub*` cases (happy path, empty-sub rejection, bad-signature rejection, expired-token rejection) mirroring the `VerifyEmail` coverage.
- `slackdata/store_test.go`: `TestBindWorkspace_WritesSeedAdminAttribute` captures the PutItem input via stubDDB and asserts `seed_admin_slack_user_id` lands with the right value.
- `cmd/main_test.go`: `TestAdminStoreAdapterForwardsAllFields` exercises the adapter copy with a non-zero CreatedAt, closing the zero-value-only blind spot.

**Coupling note (cr4 sub 3):** `Config.BindClassifyError` doc now warns that callers wiring `AdminStore` must also wire `BindClassifyError` — a nil classifier silently downgrades rebind-refused (and even idempotent-same-caller) to 500. `cmd/main.go` already wires both; the note exists so future callers don't drift.

**Comment polish (cr4 sub 4 + nits):**
- Softened "half-install becomes impossible by construction" → "refused-bind half-install" so the wording doesn't conflict with the existing #265 TODO for the lost-PutItem-race orphan path.
- `verifyIDTokenClaims` sub-verify slog copy now says "bind will be skipped (fatal in production where AdminStore is wired)" instead of the misleading "non-fatal for mint, blocks bind."
- `verifyIDTokenClaims` doc xrefs `cmd/main.go`'s `buildOAuthConfig`.
- `oauth.WorkspaceMapping` comment xrefs `TestAdminStoreAdapterMappingShapesMatch`.

**Skipped (re-confirmed):**
- JWT double-parse on the install path: sub-millisecond, once per workspace install. The split-method interface (`IDTokenVerifier.VerifyEmail` / `VerifySub`) keeps the callback's distinct best-effort vs. fatal posture clean and the verifier is opaque to consumers.
- "Admin claim is no longer needed" hint on `parseAdmin`: ephemeral rollout-period suggestion; would need a deletion follow-up I can't track from here.

Kevin-layerV:
cr5 addressed.

**Stale example fixed (cr5 nit #1):** `parser.go:41` `AdminAction` doc comment now uses `admin revoke` instead of the retired `admin claim`. (ec91fa5)

**Other nits triaged:**
- **#2 (handler_interaction.go wording):** "no view_submission *handlers* are wired today" vs "no view_submission callbacks routed today" — the current wording is accurate; deferring the rephrase.
- **#3 (dmAdminAsync raw mrkdwn interpolation of keyPrefix):** Low-priority defense-in-depth on a DM message that already came from qurl-service's JSON response (in-scope shape). Not adding a mrkdwn escape this round — the XSS-class risk is on the HTML success page, which is already covered by `html/template` auto-escape.
- **#4 (#265 TODO scope):** Acknowledged in the PR description and in callback.go's softened comment ("refused-bind half-install becomes impossible by construction (lost-PutItem-race tracked at #265)"). The bind-before-mint reorder narrows but does not close #265.

**Post-merge sequencing items (not code):**
- `bootstrap_codes` DDB teardown in `qurl-integrations-infra` happens after rollout is confirmed.
- Slack app manifest admin-only restriction on `/qurl setup` is the load-bearing seed-admin gate; the boot-time `CONFIGURATION REMINDER` log is the in-code signal.

Kevin-layerV:
cr6 addressed in f92497d.

**#1 BindClassifyError nil-coupling, now enforced:** Added `oauth.Config.Validate()` that errors when AdminStore is wired without BindClassifyError. `RegisterRoutes` calls it and panics on failure — boot-time call, so a misconfigured deploy fails before serving callbacks instead of every install attempting and surfacing 500. `cmd/main.go` already wires both; future callers can't accidentally drop the pairing.

Tests: `TestConfigValidateRejectsAdminStoreWithoutClassifier`, `TestConfigValidateAcceptsPairedAdminStore`, `TestConfigValidateAcceptsSandboxConfig`, `TestRegisterRoutesPanicsOnInvalidConfig`.

**#3 Empty id_token slog:** `verifyIDTokenClaims` now logs a distinct warn for the "Auth0 returned 200 + access_token but no id_token" case (likely the openid scope dropped on the consent screen) so on-call doesn't dig through JWKS logs that never fired.

**#4 Adapter exercising:** Refactored `adminStoreAdapter.store` to `slackdataBinder` interface so the production adapter is what `TestAdminStoreAdapterForwardsAllFields` actually drives. Previously the test was building a parallel `capturingAdapter` and `_ = adapter`-ing the production type. Now `*slackdata.Store` satisfies the interface in prod; `capturingSlackdataStore` satisfies it in tests.

**Skipped:**
- #2 (JWT double-parse): triaged earlier, sub-millisecond on once-per-install path.
- The `TestClassifyBindErrorMapping`-lives-in-grab-bag observation: cosmetic file-organization, not actionable.

Kevin-layerV:
cr7 — LGTM noted, addressed the one concrete code change.

**#3 unmapped 409 slog.Warn (4a2d006):** `classifyBindError` now emits a warn when a 409 with an unmapped Code arrives (slackdata producer added a variant without updating the classifier). Default-500 behavior unchanged; this is a diagnostic so on-call sees the drift on CloudWatch before users surface it.

**Explicitly skipped (re-confirmed):**
- #1 JWT double-parse — sub-millisecond on once-per-install; not worth the interface churn.
- #2 timeout headroom — informational only.
- #4 "admin claim was retired" friendly hint — would special-case the parser default arm + need a deletion follow-up. The current "Unknown admin action" reply is honest.
- #5 `QURL_BOOTSTRAP_CODES_TABLE` Fargate env cleanup — infra coordination, owned in `qurl-integrations-infra` follow-up.
- #6 "should-never-happen" 500 surface — informational.

PR #483, opened 2026-05-21
fix(discord): forward selfDestructSeconds as session_duration on mintLinks

justin-layerv:
## Summary

Plumbs the user's chosen self-destruct timer through to the qURL token's `session_duration`, so every minted link's L7 session window matches the fileviewer's client-side blank.

## Why

Surfaced by a user-reported bug on sandbox resource `r_9mii4njollm` (`expire_after=0.5` on the fileviewer URL): **refresh the `*.qurl.site/...` page past the 500ms canvas blank → fileviewer re-renders the full content**.

Root cause is in this repo + [`qurl-integrations-infra`](https://github.com/layervai/qurl-integrations-infra/pull/764):

| Token | `one_time_use` | DDB `session_duration_seconds` | Why null |
|---|---|---|---|
| Upload-time auto-created | false | null (should have been 1) | Connector deploy lag (#540) |
| Bot-minted (one per recipient) | true | null | `mintLinks()` doesn't forward selfDestructSeconds |

So every minted token defaults to `QURL_SESSION_TTL` (sandbox: **3600s**). Recipients can refresh for up to an hour after the fileviewer's 500ms blank — defeating the entire self-destruct UX. The L7 gate from [`traefik-plugins#173`](https://github.com/layervai/traefik-plugins/pull/173) is correctly authorizing because qurl-service correctly says the session is still valid.

## Changes

| File | Change |
|---|---|
| `src/connector.js::mintLinks` | New 5th param `selfDestructSeconds = null`. Formats as `"Ns"`, clamps `0.5 → "1s"` (qurl-service `MinSessionDuration` floor), ceils fractional values >1 (defensive — only the 0.5s preset is fractional today), includes in request body as `session_duration`. Null/undefined → field omitted → server default. |
| `src/commands.js::mintLinksInBatches` | New `selfDestructSeconds` opt, forwarded to every `mintLinks` call inside the batching loop. |
| `src/commands.js` (4 call sites) | Pass `selfDestructSeconds` (send-pipeline file + location) / `inheritedDestruct` (addRecipients file + location). All 4 call sites already had the value in scope. |
| `tests/send-pipeline-helpers.test.js` | 7 existing `mockMintLinks` call-signature assertions updated for the new 5th arg (null in these tests since none set self-destruct). |
| `tests/connector-coverage.test.js` | 5 new tests under "mintLinks — session_duration forwarding" fencing the value mapping. |

## Value mapping (matches `SELF_DESTRUCT_PRESETS`)

| User preset | `selfDestructSeconds` | Sent to connector |
|---|---|---|
| "No timer" | null | (field omitted → server default) |
| "1/2 second" | 0.5 | `"1s"` (clamped to qurl-service floor) |
| "1 second" | 1 | `"1s"` |
| "5 seconds" | 5 | `"5s"` |
| "30 seconds" | 30 | `"30s"` |
| "5 minutes" | 300 | `"300s"` |
| "30 minutes" | 1800 | `"1800s"` |
| "1 hour" | 3600 | `"3600s"` |

The 0.5s preset's fileviewer-side 500ms canvas blank still fires; only the L7 session window floors at 1s. For all other presets, exact match.

## Test plan

- [x] `npm test` — 2548 passed (+5 new), 0 failed (1 unrelated Lambda test-suite fails to load due to missing `@aws-sdk/client-ssm` — pre-existing)
- [x] `npm run lint` clean
- [ ] Deploy after [`qurl-integrations-infra#764`](https://github.com/layervai/qurl-integrations-infra/pull/764) merges + the connector deploys. Reverse order would silently drop `session_duration` at the connector handler (unknown JSON fields default to ignored).
- [ ] Manual sandbox verification: send a file via `/qurl file` with self-destruct=30s, capture the recipient's DDB `qurl-access-tokens` row, confirm `session_duration_seconds=30`. Verify refresh past 30s gets silentDropped.

## Coordinates with

- [`qurl-integrations-infra#764`](https://github.com/layervai/qurl-integrations-infra/pull/764) — connector forwards `session_duration` on mint_link path. **Must land + deploy first**, or the bot's new field is dropped silently.
- [`qurl-integrations-infra#540`](https://github.com/layervai/qurl-integrations-infra/pull/540) — upload-side counterpart (sessionDurationFor → CreateQURL). Already merged.
- [`layervai/traefik-plugins#173`](https://github.com/layervai/traefik-plugins/pull/173) — the L7 gate this fix relies on. Already merged + deployed.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1471,9 +1471,14 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
  * @param {string} opts.expiresAt — ISO string; forwarded to mintLinks.
  * @param {number} opts.recipientCount — number of tokens to mint in total.
  * @param {string} opts.apiKey — QURL API key.
+ * @param {?number} [opts.selfDestructSeconds] — when set, forwarded to mintLinks
+ *   as `session_duration`. Each minted token gets an L7 session window matching
+ *   the fileviewer's client-side self-destruct timer. null/undefined inherits
+ *   qurl-service's QURL_SESSION_TTL default. See mintLinks for value mapping
+ *   (0.5 → "1s", N>=1 → ceil to whole seconds).
  * @returns {Array<{qurl_link: string, qurl_id: string, resourceId: string}>}
  */
-async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, recipientCount, apiKey }) {
+async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, recipientCount, apiKey, selfDestructSeconds = null }) {
   const allLinks = [];
   let currentResourceId = initialResourceId;
   let tokensUsed = 0;
@@ -1485,7 +1490,7 @@ async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, re
       tokensUsed = 0;
     }
     const batchSize = Math.min(TOKENS_PER_RESOURCE, recipientCount - i);
-    const minted = await mintLinks(currentResourceId, expiresAt, batchSize, apiKey);
+    const minted = await mintLinks(currentResourceId, expiresAt, batchSize, apiKey, selfDestructSeconds);
     for (const link of minted) {
       // qurl_id is the join key against qurl.accessed webhooks; empty
       // string degrades the whole monitor to bare base-msg.
@@ -1781,6 +1786,7 @@ async function executeSendPipeline(interaction, {
           expiresAt,
           recipientCount: recipients.length,
           apiKey,
+          selfDestructSeconds,
         });
       } finally {
         bufHolder.buf = null;
@@ -1818,6 +1824,7 @@ async function executeSendPipeline(interaction, {
         expiresAt,
         recipientCount: recipients.length,
         apiKey,
+        selfDestructSeconds,
       });
 
       if (allLinks.length < recipients.length) {
@@ -2485,6 +2492,7 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
           expiresAt,
           recipientCount: newRecipients.length,
           apiKey,
+          selfDestructSeconds: inheritedDestruct,
         });
       } catch (err) {
         // Discord CDN URLs are signed and expire (~24h). If re-download fails,
@@ -2529,6 +2537,7 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
         expiresAt,
         recipientCount: newRecipients.length,
         apiKey,
+        selfDestructSeconds: inheritedDestruct,
       });
 
       if (allLinks.length < newRecipients.length) {
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -323,7 +323,24 @@ async function downloadAndUpload(sourceUrl, filename, contentType, apiKey, viewe
  * one-time tokens, so one recipient opening their link doesn't
  * invalidate anyone else's.
  */
-async function mintLinks(resourceId, expiresAt, n, apiKey) {
+// mintLinks mints `n` access tokens on the connector's `/api/mint_link`
+// endpoint. selfDestructSeconds, when set, is forwarded as
+// `session_duration` so every minted token gets an L7 session window
+// matching the fileviewer's client-side self-destruct timer.
+//
+// Without this, the upload-time auto-created token correctly inherits
+// session_duration via the connector's CreateQURL path
+// (qurl-integrations-infra#540), but every additional token minted on
+// the same resource defaults back to qurl-service's QURL_SESSION_TTL
+// (3600s sandbox). A recipient of the self-destruct send could then
+// refresh `*.qurl.site/...` past the fileviewer blank and see the
+// content again for up to an hour. Tracked: qurl-integrations-infra#764.
+//
+// Value mapping (matches SELF_DESTRUCT_PRESETS in src/utils/time.js):
+//   selfDestructSeconds null     → omit session_duration → server default
+//   selfDestructSeconds 0.5      → "1s" (qurl-service MinSessionDuration)
+//   selfDestructSeconds N >= 1   → "Ns" (whole-seconds rounded up)
+async function mintLinks(resourceId, expiresAt, n, apiKey, selfDestructSeconds = null) {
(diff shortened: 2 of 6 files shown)


justin-layerv:
## Cycle 1 — CR response (LGTM modulo defensive validation)

Commit `d510f2d`. All 3 actionable items + 1 coverage observation addressed.

| # | Item | Action |
|---|---|---|
| 1 | Defensive validation gap — null/undefined check let NaN, ±Infinity, '30', booleans, {}, [], 0, negatives through, producing `"NaNs"` / `"Infinitys"` on the wire | ✅ Gate now uses `Number.isFinite(x) && x > 0`, matching the sibling `appendViewerTtl` contract at `connector.js:164`. Same idiom, same belt-and-suspenders justification per the existing rationale on `appendViewerTtl`. |
| 3 | Two adjacent comment blocks (original JSDoc + new line-comment block) looked split | ✅ Consolidated into a single JSDoc with `@param` entries for all 5 args. `selfDestructSeconds` rationale + value mapping is in the prose; `@param selfDestructSeconds` cross-references. |
| 5 | Coverage: defensive-input loop test | ✅ Added `omits session_duration for non-finite / wrong-type / non-positive inputs` looping over `[NaN, Infinity, -Infinity, '30', '0.5', true, false, {}, [], 0, -1, -0.5]`. Each → `session_duration` absent from body. Mirrors the sibling `viewer_ttl_seconds` defensive-input test (this file's `:243`) for symmetry. |

Acknowledged with no action per CR:

| # | Item | Reasoning |
|---|---|---|
| 2 | Object-property shorthand at call sites | CR: "no change requested." Local from line 1597 is in scope; correct. |
| 4 | Deploy-order risk (this PR after #764) | Operational item, already documented in PR description. |

### Verification

- `npm test` — 2576 passed (+28 from the defensive-input loop's iterations + previously-blocked Lambda suite resolving via fresh worktree install)
- `npm run lint` clean

justin-layerv:
## Cycle 2 — CR response

Commit `1918a9e`. Cycle 2 CR ask + /simplify findings folded in.

| Source | Item | Action |
|---|---|---|
| CR cycle 2 #1 | No end-to-end fence that a non-null `selfDestructSeconds` propagates through `executeSendPipeline` → `mintLinksInBatches` → `mintLinks` | ✅ Extended both existing inheritance tests in `send-pipeline-helpers.test.js` (file-path with 30s, URL/maps with 300s) to ALSO assert `mockMintLinks` was called with the value. Without this, a future rename of the `mintLinksInBatches` opts key would silently drop the value at the boundary. |
| /simplify reuse + quality | Wire-format mapping (`Math.max(1, Math.ceil(x))` + `"Ns"`) lived inline inside `mintLinks` — the JSDoc documented the contract but no helper enforced it; risk of drift with the qurl-s3-connector's `sessionDurationFor()` (Go sibling) | ✅ Extracted `formatSessionDurationSeconds(s)` to `src/utils/time.js`, co-located with `SELF_DESTRUCT_PRESETS` (its only legitimate input source). `mintLinks` collapses to 2 lines. Inline JSDoc value-mapping paragraph trimmed in favor of a cross-reference to the helper. |
| (coverage) | Direct unit-test coverage for the new helper | ✅ Added 4 tests in `tests/time.test.js`: preset round-trip (all 7 presets → expected `"Ns"` strings), 0.5s clamp to `"1s"` (`MinSessionDuration` floor), fractional ceil (2.3 → "3s"), defensive-input loop (null/undefined/NaN/Infinity/strings/booleans/objects/arrays/0/negatives all → null). |

### Acknowledged / deferred per CR

| Item | Reasoning |
|---|---|
| 5-positional-arg `mintLinks` "starting to feel crowded" | CR cycle 2: "Not a blocker, would be churn to refactor." Defer to a follow-up when a 6th param lands. |
| Naming asymmetry `selfDestructSeconds` (number) vs `session_duration` (wire string) | CR: "just flagging." JSDoc cross-reference to `formatSessionDurationSeconds` bridges them. |

### Verification

- `npm test` — 2580 passed (+4 new helper tests + 2 new end-to-end assertions)
- `npm run lint` clean

justin-layerv:
## Cycle 4 — CR response (Approve-with-nits)

Commit `da37e4b`. All 3 actionable nits + 1 file-cross-ref applied.

| # | Item | Action |
|---|---|---|
| 1 | Asymmetry between `viewer_ttl_seconds` (forwarded verbatim, 0.5→"0.5") and `session_duration` (floored, 0.5→"1s") not discoverable from the upload side | ✅ Added a paragraph to `appendViewerTtl`'s docstring noting the verbatim vs. floor asymmetry and pointing at `formatSessionDurationSeconds`. A reader hitting either side now learns the other's handling. |
| 2 | 0.5s preset has a residual 500ms→1s refresh window (fileviewer blanks at 500ms but L7 floors at 1s) | ✅ Inlined a note on the `0.5s` preset entry in `SELF_DESTRUCT_PRESETS` explaining the residual, why the preset is retained, and when the gap would close (qurl-service lowering `MinSessionDuration`). |
| 3 | Stale-history comment in `formatSessionDurationSeconds` (referenced an "earlier `Math.max(1, ...)`" that won't exist for a future reader) | ✅ Replaced authoring-history with just the load-bearing WHY ("Math.ceil of any value in (0, 1] is 1; ... combined with the > 0 guard above, never emits '0s'"). |

### Acknowledged / no action per CR

| # | Item | Reasoning |
|---|---|---|
| 4 | Defensive-case redundancy between formatter unit tests + mintLinks integration tests | CR: "Not blocking." Belt-and-suspenders is justified by `mintLinks` being exported (caller boundary). |
| 5 | Location/maps no-op forwarding | CR: "just confirming intentional." Already documented at the call site with the cross-ref to qurl-integrations-infra#480. |

### Coordinated follow-up

The CR's trust-boundary concern (qurl-service should cap mint-time `session_duration` against the resource's original window) — filed as [**qurl-service#613**](https://github.com/layervai/qurl-service/issues/613). Bot-side `formatSessionDurationSeconds` ensures correctness in practice (presets-only input source); the server-side cap is the architectural close. Captured in the PR #764 cycle 4 response.

### Verification

- `npm test` — 2580 passed
- `npm run lint` clean

### Status

PR is ship-ready. Awaiting cycle 5 review pass + the sandbox verification post-deploy (per cycle 1-4 test plan: `/qurl file` with self-destruct=30s → DDB `qurl-access-tokens.session_duration_seconds=30` on the bot-minted Token A → refresh past 30s → expect silentDrop).

PR #484, opened 2026-05-21
refactor(discord): mintLinks options-bag + shared isPositiveFinite predicate

justin-layerv:
## Summary

Two PR #483-derived follow-ups bundled into one focused refactor:

### 1. `mintLinks(resourceId, opts)` — options-bag for sibling-consistency

`mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, recipientCount, apiKey, selfDestructSeconds })` was already options-bag shaped, but its inner call to `mintLinks(resourceId, expiresAt, n, apiKey, selfDestructSeconds)` was 5 positional args — two adjacent strings (`resourceId` + `expiresAt`) and a 3rd (`apiKey`) all swappable at the call site without a compile error. PR #483 cycle-1 review flagged this as a footgun matching the qurl-s3-connector `MintLinkOptions` Go-side refactor (layervai/qurl-integrations-infra#767).

New shape: `mintLinks(resourceId, { expiresAt, n, apiKey, selfDestructSeconds })`. `resourceId` stays positional (always required, the URL path component); everything else is in the bag.

### 2. `isPositiveFinite(n)` — shared predicate

The `Number.isFinite(x) && x > 0` (and its `!Number.isFinite(x) || x <= 0` negation) idiom was inlined at **11 sites**:

- `connector.js` (×1) — `appendViewerTtl` viewerTtlSeconds gate
- `commands.js` (×3) — self-destruct gates + `selfDestructOptionToSeconds`
- `server.js` (×1) — TRUST_PROXY parse
- `index.js` (×2) — `RATE_LIMIT_WINDOW_MS` / `RATE_LIMIT_MAX_REQUESTS`
- `store/ddb-store.js` (×1) — `PENDING_LINK_EXPIRY_MS`
- `utils/time.js` (×3) — `parseExpiryMs`, `formatSelfDestructSegment`, `formatSessionDurationSeconds`

Extracted into `utils/time.js`, hoisted to file top (so readers don't hit unresolved references before its definition), and fenced by a direct unit test alongside the integration coverage at every call site.

## Why

Both items were `please-fix` deferrals from PR #483's CR cycles 2/3/4. Landing as a focused refactor rather than re-opening #483 keeps the bug-fix commit history clean.

## Test plan

- [x] Full `npx jest`: **2584 / 2584 passed** (added 4 new `isPositiveFinite` tests; no regressions; 22 prior test call sites updated, including 9 `mockMintLinks.toHaveBeenCalledWith(...)` expectations rewritten to the bag shape)
- [x] `npx eslint src/ tests/`: clean
- [x] Behavior identical — pure refactor; no wire-format / runtime semantics changed.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -27,6 +27,7 @@ const {
   expiryToMs,
   formatSelfDestructLabel,
   formatSelfDestructSegment,
+  isPositiveFinite,
   selfDestructSelectValueToSeconds,
   isLegitimateSelfDestructSelectValue,
   SELF_DESTRUCT_PRESETS,
@@ -1490,7 +1491,12 @@ async function mintLinksInBatches({ initialResourceId, reuploadFn, expiresAt, re
       tokensUsed = 0;
     }
     const batchSize = Math.min(TOKENS_PER_RESOURCE, recipientCount - i);
-    const minted = await mintLinks(currentResourceId, expiresAt, batchSize, apiKey, selfDestructSeconds);
+    const minted = await mintLinks(currentResourceId, {
+      expiresAt,
+      n: batchSize,
+      apiKey,
+      selfDestructSeconds,
+    });
     for (const link of minted) {
       // qurl_id is the join key against qurl.accessed webhooks; empty
       // string degrades the whole monitor to bare base-msg.
@@ -4058,7 +4064,7 @@ function renderConfirmCardContent({
     content += `\n**To:** ${validRecipients.length} user${validRecipients.length === 1 ? '' : 's'} (${previewNames}${more})\n`;
   }
   content += `**Expires:** ${EXPIRY_LABELS[expiresIn] || expiresIn}\n`;
-  if (Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0) {
+  if (isPositiveFinite(selfDestructSeconds)) {
     content += `**Self-destruct:** ${formatSelfDestructLabel(selfDestructSeconds)}\n`;
   }
   if (personalMessage) {
@@ -4256,7 +4262,7 @@ function renderConfirmCardRows({
   // corrupted DDB row carrying an off-preset finite value (which would
   // otherwise leave every option un-defaulted and force Discord to
   // render the first option's label — wrong UX).
-  const hasTimer = Number.isFinite(selfDestructSeconds) && selfDestructSeconds > 0;
+  const hasTimer = isPositiveFinite(selfDestructSeconds);
   const hasMatchingPreset = hasTimer && SELF_DESTRUCT_PRESETS.some((p) => p.seconds === selfDestructSeconds);
   if (hasTimer && !hasMatchingPreset) {
     // Symmetric with the expiry off-set warn below: a finite positive
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -3505,7 +3505,7 @@ const SELF_DESTRUCT_PRESET_SECONDS = new Set(SELF_DESTRUCT_PRESETS.map((p) => p.
 function selfDestructOptionToSeconds(value) {
   if (!value || value === SELF_DESTRUCT_NO_TIMER_CHOICE) return null;
   const n = Number(value);
-  if (!Number.isFinite(n) || n <= 0) return null;
+  if (!isPositiveFinite(n)) return null;
   // Match the parsed Number directly — SELF_DESTRUCT_PRESETS contains
   // 0.5 (the "1/2 second" preset), so a Math.floor here would map 0.5
   // → 0 (not in the set) and silently downgrade to "no timer." The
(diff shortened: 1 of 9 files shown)


justin-layerv:
Addressed all feedback (commit 49f6bda):

**Real finding — `formatSelfDestructSegment` was unmigrated.** Fixed; the predicate's "Six call sites" comment was actually 5-of-6. Migrated it.

**Sweep — the 4 additional sites flagged for follow-up.** Took the "happy to migrate now" option rather than letting the count drift further. Migrated:
- `commands.js:3508` (`selfDestructOptionToSeconds`)
- `index.js:253,257` (RATE_LIMIT_WINDOW_MS / RATE_LIMIT_MAX_REQUESTS)
- `store/ddb-store.js:207` (PENDING_LINK_EXPIRY_MS)
- `utils/time.js:14` (`parseExpiryMs`)

Total now: **11 sites**. Predicate godoc updated to match.

**Nit — `= {}` default on `mintLinks` destructure.** Dropped. The next gate (`Number.isInteger(n)`) already catches the forgot-the-bag case, and removing the default surfaces the programming error more directly via `Cannot destructure property 'expiresAt' of 'undefined'`.

**Other nits acknowledged out of scope:**
- JSDoc symmetric `expiresAt` guard — pre-existing behavior; can harden in a future sweep if we ever want server-side rejection vs upstream-determines.

2580/2580 jest tests passing; eslint clean.

justin-layerv:
Addressed nits (commit 6882031):

**Nit 2 (hoist).** `isPositiveFinite` moved to the top of `utils/time.js` so top-to-bottom readers see the definition before its (lexically lower) callers in `parseExpiryMs`, `formatSelfDestructSegment`, and `formatSessionDurationSeconds`.

**Nit 3 (count mismatch).** Updated the PR description to match the in-file comment's "11 inline sites" count.

**Nit 4 (direct unit test).** Added 4 test cases under `describe('isPositiveFinite')` — positive finites, `null/undefined/NaN/±Infinity`, zero/negative finites, non-Number types (verifies no `'0x1'` → 1 coercion). Total: 2584 / 2584 jest tests passing.

**Nit 1 (`={}` destructure default).** Intentionally NOT re-added. Round-1 cr suggested removing ("adds nothing the next gate doesn't catch, programming-error message is clearer without"); round-2 cr suggested re-adding for diagnostic clarity. Either is defensible — sticking with the round-1 decision rather than thrash. Happy to flip if you have a strong preference.

Lint clean. Ready for merge IMO unless there's anything else.

justin-layerv:
Round-3 polish applied (commit 95bbe72):

**Nit 1 — `= {}` destructure default re-added.** Round-3 cr again recommended; consensus across round-2 and round-3 reviewers wins. Now `mintLinks('id')` falls through to the cleaner `Invalid link count (n must be integer 1..100): undefined` validator instead of a cryptic `Cannot destructure of undefined` TypeError.

**Nit 2 — `isPositiveFinite` doc trimmed.** Removed the 11-call-site enumeration from the source comment (would have rotted on the next +1 caller). Kept the rationale: rejection set, `Number.isFinite` strict vs global `isFinite` non-coercion, the hoist explanation. The count breakdown lives in the PR description where it belongs.

**Nit 3 — `appendViewerTtl` qualifier restored.** Comment now leads with "Strict positive-finite: …" so the invariant is explicit without making readers infer it from the predicate name.

2584/2584 jest tests passing. Lint clean. Ready to merge IMO.

justin-layerv:
Round-4 observation #1 addressed (commit 7aaf32d):

**Comment-line added near `formatSelfDestructLabel`'s gate** explaining why it intentionally stays on `!Number.isFinite(seconds)` (the function's contract is "render any finite value as Ns; only NaN/Infinity becomes '(invalid)'"). Switching to `!isPositiveFinite` would silently re-route 0 and negatives, which is the exact "finish the refactor" footgun you flagged.

**Observations 2 and 3 acknowledged as non-actionable:**
- **#2 `mintLinks('res-1', null)` destructure throw** — sole caller passes an object literal; deferring the `opts || {}` belt-and-suspenders until an actual `null` callsite shows up.
- **#3 `parseExpiryMs` gate now only filters `ms === 0`** — true and intentional. The `\d{1,6}` regex already prevents NaN/overflow; the predicate's job there is filtering the `"0m"`/`"0h"`/`"0d"` degenerate case (which would otherwise yield `new Date(Date.now())` as an "expiry"). Comment in `parseExpiryMs` already reflects this; no change.

2584/2584 jest passing. Lint clean.

justin-layerv:
Round-5 verdict: "LGTM, ready to merge" — convergence reached after 5 cr cycles.

Both round-5 observations are explicitly marked "do not address":
- **#1 `selfDestructOptionToSeconds` `Number()` coercion** — defense-in-depth via the closed `SELF_DESTRUCT_PRESETS` set; out of scope and the comment already documents the divergence.
- **#2 `mintLinks('res-1', null)` destructure throw** — no in-tree caller passes `null`; the `={}` default is the correct opts-default; adding `opts || {}` would be belt-and-suspenders.

Both deferred follow-ups are tracked. PR is converged and ready to merge whenever you're set.

PR #505, opened 2026-05-27
feat(slack): add tunnel install flow

justin-layerv:
## Summary
- add /qurl tunnel install <slug> to create/find a tunnel resource, bind the channel alias, and mint a short-lived tunnel bootstrap key
- support /qurl set-alias   by resolving slug targets through tunnel find-or-create
- fold alias writes into slackdata.Store and remove the old aliasstore bridge so reads/writes use one DDB facade
- extend the shared client with tunnel slug/find_or_create fields and API-key creation support

Depends on layervai/qurl-service#758 for purpose=tunnel_bootstrap keys.

## Validation
- go test ./...
- golangci-lint run --timeout=5m --new-from-rev=origin/main ./apps/slack/... ./shared/...


justin-layerv:
Addressed the latest review in 4253f01:\n\n- Slug-targeted set-alias now acks immediately and resolves/binds asynchronously via response_url.\n- Bootstrap API key shell validation now rejects only unsafe shell/control characters instead of assuming a narrow token alphabet.\n- Replaced the shell-toxic <web-container> placeholder with a WEB_CONTAINER variable and guard.\n- Help/usage now includes set-alias and unset-alias, including the hyphenated unset alias form.\n- Added focused tests for shell quoting, nested Slack code fences, alias help visibility, hyphenated unset-alias, and async slug alias resolution.\n\nDeferred the production image fallback policy as follow-up issue #506, because it is deployment hardening outside this Slack flow patch and needs an explicit rollout choice.

justin-layerv:
Addressed the latest Claude review in a989dc9:\n\n- Shortened Slack-generated tunnel bootstrap keys from 24h to 1h to reduce plaintext exposure in the ephemeral install response.\n- Changed tunnel-install API-key idempotency from Slack trigger_id to a stable team/channel/user/slug/hour bucket, so user retries in the same setup window reuse the same qurl-service idempotency key.\n- Removed the mixed #nosec suppression by moving the alias audit emit behind a helper that passes lint without suppression.\n- Tightened tunnel-slug target copy and clarified that bootstrap-key shell validation is defense-in-depth over shellSingleQuote.\n- Updated stale unsetalias copy to the public unset-alias form.\n\nThe production image fallback policy remains tracked in follow-up issue #506. Local validation: go test -race -count=1 -coverprofile=coverage.out -covermode=atomic ./..., golangci-lint.

justin-layerv:
Pushed bd4781d for the CI-only lint failure: replaced the alias audit slog.Info varargs call with typed slog.LogAttrs attributes, which keeps the audit event and avoids G706 without suppression. Local validation after the fix: git diff --check, golangci-lint, go test -race -count=1 ./apps/slack/internal.

justin-layerv:
Addressed the latest review in 71bb4cf:\n\n- Updated the empty-aliases hint to the public /qurl set-alias spelling.\n- Matched tunnel slug copy to the actual regex, including the no-trailing-hyphen rule.\n- Pinned tunnel install response_url delivery as ephemeral in the test that carries the plaintext bootstrap key.\n- Added the next-hour idempotency-key assertion and documented why the bucket is hourly.\n- Added public client docs for Resource.Slug and tunnel bootstrap API-key fields.\n- Replaced the set/unset alias helper duplication with a shared slashVerb helper.\n- Changed slackCodeBlock to panic on nested fences instead of silently rewriting future user-visible content.\n\nThe default image hardening remains tracked in #506. Local validation: git diff --check, golangci-lint, go test -race -count=1 ./apps/slack/internal ./apps/slack/cmd ./shared/client.

PR #507, opened 2026-05-27
feat(gmail-extension): add qURL Gmail Chrome extension for compose uploads

haochangjiu:
## Summary

Introduce the qURL Gmail Chrome Extension (MV3) to the integrations monorepo.

The extension allows users to upload local files to qURL directly from Gmail's compose window and auto-insert secure access links into drafts. This provides a native Chrome extension alternative to the existing Gmail Apps Script add-on.

## Features

- **Local file upload** — select files from your computer, no need to attach to Gmail first
- **Batch upload** — upload multiple files at once
- **Auto-insert links** — qURL access links + expiry times are appended to the active Gmail draft
- **Manual copy fallback** — if Gmail draft insertion fails, copy the generated content from the popup
- **Native Gmail integration** — links are inserted directly into the compose body as HTML
- **Clean popup UI** — simple, focused interface with upload progress and per-file status
- **Configurable server** — supports custom qURL server URLs with per-origin permission prompts

## Technical Highlights

- MV3 service worker with ping→reinject→re-ping pattern for content script reliability
- Three-tier draft insertion fallback (execCommand → Selection API → insertAdjacentHTML)
- Request deduplication in content script with bounded cache and eviction callbacks
- Cross-layer timeout chaining with explicit budget comments
- Trust boundary enforcement via `isTrustedInsertLinksSender`
- UTF-8-aware filename truncation without splitting codepoints
- https-only link validation to prevent javascript: smuggling
- Multipart boundary generation with crypto.getRandomValues fallback
- Build-time `QURL_API_BASE` override support for release packaging

## Changes

- Add Chrome MV3 extension under `apps/gmail-extension/`
- Add `gmail-extension` scope to bug report dropdown and CLAUDE.md
- Include manifest, background service worker, popup UI, Gmail content script
- Include qURL API client, compose formatter, i18n helper
- Include icon assets, localization strings, and documentation
- Add ~3,500 lines of unit test coverage for shared non-DOM modules

## Verification

```bash
cd apps/gmail-extension
npm test
npm run release
npm run package:release
```

## Checklist

- [x] Extension follows monorepo layout convention (`apps/gmail-extension/`)
- [x] Brand spelling compliant (qURL in user-visible text)
- [x] Scope added to bug report dropdown and CLAUDE.md
- [x] Documentation aligned with implementation
- [x] Tests pass (86 tests)
- [x] Release/package flow verified


justin-layerv:
Thanks for the thorough pass. Addressed in `a381facf`:

**Blockers**
- `minimum_chrome_version` 92 → **99** (covers the Promise-form `chrome.runtime.sendMessage`).
- **ESLint added + run in CI**, matching `apps/discord`: `.eslintrc.json` (eslint:recommended + the discord ruleset, browser/webextensions envs), `npm run lint --max-warnings 0`, and a `Lint` step in the workflow.

**Should-fix**
- Dropped the non-crypto `createMultipartBoundary` fallback.
- Progress icons no longer interpolate escaped text into `innerHTML` — static icon markup on a dedicated node + `textContent` for the filename; removed the now-unused popup `escapeHtml`.
- `permission_request_confirm` body now uses the declared `$origin$` placeholder.
- Pinned the sender-trust contract with a comment + a test asserting a content-script sender (matching id, has `tab`) is rejected; commented the `tab.url`-via-host-permission reliance.

**Nits**
- `package-all.sh` derives the ZIP name from `package.json` `name`.
- Log prefixes: all runtime `console.*` calls already use `[qURL]` (build scripts intentionally don't).

**Consciously not changing (rationale):**
- *Static `placeholder` flash*: only the `en` locale ships, so the static value equals the localized one — no visible flash today. Worth revisiting if/when more locales are added.
- *`bump-version.js` rejects pre-release suffixes*: Release Please now owns released versions on `main` (this PR wires the app into `release-please-config.json`); the local script is a convenience for ad-hoc Web Store ZIPs and intentionally supports only simple `patch/minor/major`/exact bumps.

justin-layerv:
Addressed in `4e17a732`:
- **Brand spelling** — standalone `QURL` → `qURL` across the docs, the `gmail-compose.js` doc comment, and the `build-release.js` error message (code identifiers/env vars like `QURL_API_BASE`, `QURLConfig`, `QURL_PING` preserved).
- **`expiry_suffix`** now uses its declared `$EXPIRY$` named placeholder (the last placeholder inconsistency).
- Dropped the stale `activeTab` from `DESIGN.md`'s permissions list.

**On the `1.0.2` vs `0.1.0` Release Please seed (intentional):** the extension already carries `1.0.2` in `manifest.json` and `package.json` (its pre-monorepo version), so seeding `.release-please-manifest.json` at `1.0.2` tracks the app's *actual current* version. Seeding `0.1.0` would make Release Please believe the next release regressed below what's effectively already shipped. The `node` release-type + `extra-files` keeps `manifest.json` `$.version` in lockstep going forward.

**Consciously deferred (non-blocking nits):**
- *RFC 6266 `filename*=UTF-8''…`*: the qURL connector accepts raw UTF-8 `filename="…"` (as the existing Apps Script add-on relies on), so this is a future-compat enhancement, not a correctness fix.
- *MutationObserver discovery window*: already `requestAnimationFrame`-debounced and disconnected on finish; flagged as "keep an eye on," no change needed now.

justin-layerv:
Addressed in `66ff31c0`:
- **`formatExpiry`** now appends the sender's UTC offset (`… -0500`) so the expiry is unambiguous across timezones in an email.
- Dropped the unreachable non-crypto `createRequestId` fallback (Chrome 99 + Node test both guarantee `crypto.randomUUID`).
- **Reinjection edge case** documented: the guard flag and the `onMessage` listener share the isolated-world lifetime, so a torn-down context recovers on reinject. I deliberately did **not** move `addListener` above the guard — that registers a duplicate listener on every legitimate double-injection (manifest auto-inject + reinject) and double-replies to each message. The only unrecoverable case (severed port, live context) is noted; a tab reload recovers it.
- Comments added: `lastSuccessfulResults` is latest-batch-only; `_get` numeric coercion is harmless for URL fields; `_buildMultipartBody` uses an explicit boundary for deterministic sanitized bytes.
- **README** documents the Release Please `1.0.2` seed rationale.

**Consciously not changing (with reason):**
- *Comment in `manifest.json` for the `https://*/*` rationale*: the manifest is parsed as strict JSON by the release build (`JSON.parse`), so a comment would break it. The rationale lives in `docs/chrome-web-store-review.md` and the DESIGN.md permissions section.
- *Happy-path test for the permission-confirm flow*: already covered — `test/popup.test.js` has "saving a custom server shows an inline confirmation…", "continuing the inline custom-server confirmation proceeds with saving", and "denied custom-server permission does not persist the override".
- *`node --test test/` (Windows glob)*: the directory form doesn't discover the suite reliably on the pinned Node; CI is ubuntu and the explicit glob works, so kept as-is.
- *jsdom smoke test / ESM-globals*: out of scope (no jsdom in tree; the script-tag globals are the current architecture).

justin-layerv:
Addressed in `0a838f8f`:
- **Popup-close UX (#2 in your review):** the uploading hint now reads "keep this popup open until it finishes", and the README documents that dismissing the popup cancels an in-flight upload. (It already documents that links are appended at the end of the draft.)
- **Gmail selectors (#4):** added a `TODO(gmail-frontend-classes)` marker on the obfuscated `.Am.Al.editable` selector so "insertion stopped working" reports `git grep` straight to it; the role/contenteditable fallbacks are noted as the stable path.
- **`bump-version.js` (#6):** comment explaining the pre-release-suffix rejection is intentional.

**Deferred to #553** (filed, labeled `pr-polish-followup`) — the two deeper runtime-resilience asks, which are genuine enhancements beyond this PR:
- Relaying uploads through the background service worker so they survive popup teardown (your #2, second option).
- Surfacing a late insertion completion back to the popup after a reported timeout (your #3).

**Noted, no change (with reason):**
- *`https://*/*` UI clarification (#1):* reviewer rationale lives in `docs/chrome-web-store-review.md` + DESIGN.md; a store-listing sentence is a product/listing decision.
- *`_parseExpiry` ms/sec heuristic (#5), `_get` `resource_id: 0` nit:* realistic qurl-service payloads don't hit these; flagged-for-completeness only.
- *Node 18+22 CI matrix (#8):* matches `apps/discord`'s single-`.nvmrc` convention; Node only runs the build/test tooling (the extension runs in Chrome).

justin-layerv:
Addressed in `372824f9`:
- **Eviction race (#1):** `pruneInsertRequestState` now `console.warn`s when it evicts a *pending* entry, mirroring the late-completion warning — "draft updated despite an error" is observable on both paths now.
- **Prune ordering (#2):** comment clarifying "oldest" = first-inserted (Map order), not oldest-by-status.
- **Timeout-budget drift:** the `RUNTIME_MESSAGE_TIMEOUT_MS` test now asserts the documented inequality against `background.js`'s own exported constants (`> 2*ping + relay`, with ≥5s reinject headroom) — the budget can't silently drift now.
- **Release smoke-test note:** `docs/local-unpacked-testing.md` flags a live-Gmail smoke test as a per-release step given the obfuscated-class dependency.

**Noted / accepted (no change):**
- `void chrome.runtime.lastError` — already consumed: the callbacks *read* `chrome.runtime.lastError` in their guard, which marks it checked, so no devtools warning is emitted.
- revoke-then-clear edge case (#4), `+ '/'` URL-join defense (#5), filename `/` (server-side, #6), `_parseExpiry` ms/sec heuristic, ESLint flat-config migration, DESIGN.md flow diagram — accepted as flagged-for-completeness / out-of-scope; the deeper resilience work is tracked in #553.

justin-layerv:
Addressed in `d83ffb45`:
- **[blocking-ish] Permission leak custom-A → custom-B:** `setStoredQurlApiBase` now reads the previous base up front and revokes its host permission (best-effort, after the new value persists) whenever the host changes — covering custom→default *and* custom-A→custom-B, skipping same-host path-only changes. New tests cover A→B (revokes A) and A→A path change (no revoke), alongside the existing custom→"" test.
- **[blocking-ish] No file-size cap:** added `MAX_UPLOAD_FILE_BYTES` (100 MB popup-safety ceiling) with a localized `file_too_large_error` reported per-file *before* `file.arrayBuffer()`, so an oversized file no longer risks OOM-ing the popup.
- **[discussion] `:focus` selector fragility:** wrapped in `:is(${selector}):focus`.
- **[nit] no-requestId branch:** comment downgraded to "legacy fallback, never produced by popup.js today".
- **[nit] `SETTINGS_PANEL_AUTO_CLOSE_MS`:** 1200 → 2000ms for screen-reader read time.
- **[nit] two-window relay:** documented in the README that Gmail must be the active tab in the focused window.

**Noted / accepted (no change):**
- *placeholder flash, `_get` `0`, `compareComposeBodies` z-index tiebreaker* — only-`en` ships / flagged-for-completeness / rare-and-bounded.
- *`console.warn` on race paths* — an earlier review round explicitly asked for these warnings as observability; they fire only on the timeout/eviction races, not normal flow, so I kept `warn` rather than flip to `info`.
- *`https://*/*`* — rationale stays current in `docs/chrome-web-store-review.md`.

justin-layerv:
Addressed in `7689918a`:
- **[#2 strip userinfo]** `normalizeQurlApiBase` (and the build's `normalizeBuildQurlApiBase`) now clear `username`/`password`, so a pasted `https://user:pw@host` never persists credentials to `chrome.storage.local` or gets echoed back. Test added.
- **[#3 expiry boundary]** `_parseExpiry` uses `>= 1e12` (treats exactly `1e12` as ms). Boundary test added.
- **[#6 file cap]** README now notes the 100 MB cap is a popup-safety ceiling and the server may enforce a lower one.

**Noted / accepted (no change):**
- **#1** `optional_host_permissions` rationale lives in `docs/chrome-web-store-review.md`; a comment in `manifest.json` isn't possible since the release build parses it as strict JSON (`JSON.parse`).
- **#4** trimmed-compare for the permission card, **#5** auto-close gating on the grant path, **#7** `isGmailTab` prefix breadth, and the smaller nits (empty-`err.message`, compose-iframe race, ISO-in-README) are minor UX/doc polish; the deeper resilience/observability work remains tracked in #553.

PR #510, opened 2026-05-27
feat(slack): owner-only /setup gate + workspace owner = Slack user, not Auth0 sub

Kevin-layerV:
## Summary

Two coupled changes that move the slack-bot admin model to the **LayerV-admin** intent (workspace owner is the Slack identity that ran first /setup) and fix the long-standing `/qurl admin list` owner-display bug.

### 1. `workspace_mappings.owner_id` = Slack user ID

Was: the JWKS-verified Auth0 id_token sub. The handler-side renderer for `/qurl admin list` already expected a Slack user ID (`looksLikeSlackUserID` guard on the `<@%s>` mention render), so the owner line was rendering "(unknown — workspace owner record is missing; contact support)" for every properly-bound workspace — that's the bug. The Auth0 sub stays a runtime gate at OAuth-callback time (id_token must still verify before bind) but is no longer persisted.

### 2. `/setup` owner-only gate at the slash-command surface

- **Fresh install** (no workspace_mappings row): anyone can run `/setup` — that user becomes the owner.
- **Subsequent runs**: only the owner is allowed; any other workspace user — including admins added via `/qurl admin add` — gets a refusal naming the existing owner.

Without this gate, an added admin could complete OAuth against their own Auth0 account and silently rotate the workspace's qURL credential, locking the original owner out at the qurl-service layer. The OAuth callback's `BindWorkspace` classifier also enforces this as defense in depth (now compares seedAdmin to the row's `owner_id` rather than scanning the admin set), but slash-level rejection means non-owners never get a setup URL minted in their name.

The rest of the admin verbs (`admin add/remove/list/revoke`, `tunnel`, `setalias`, `unsetalias`) continue to use the broader admin_set membership gate, unchanged. Only `/setup` is owner-only.

## Migration (sandbox)

Existing sandbox `workspace_mappings` rows have `owner_id` = Auth0 sub. After this deploys, every `/setup` attempt against an existing row will be classified as a different-owner rebind and refused (the stored Auth0 sub won't equal any caller's Slack ID).

**Prod scope:** the Slack bot is sandbox-only today — the prod Slack-OAuth + infra chain hasn't shipped, so there are no prod `workspace_mappings` rows to migrate. Prod will deploy this (post-pivot) binary from the start and write `owner_id` = Slack ID on the first `/setup`, so a prod row will never carry an Auth0-sub `owner_id`. This migration is sandbox-scoped.

**Operator recovery:**

```bash
aws dynamodb delete-item \
  --table-name <workspace-mappings-table-name> \
  --key '{"slack_team_id":{"S":"<TEAM_ID>"}}' \
  --region us-east-2
# then re-run /qurl-sandbox setup in Slack — fresh row written with owner_id = Slack ID
```

The qurl-service-side `workspace_keys` row stays put through this — the bot re-mints the key on the next /setup against the same workspace, and the existing API key is rotated as usual.

The same `delete-item` recovery applies if the workspace owner later leaves or is deactivated in Slack: `owner_id` is immutable once set, so until an in-band ownership-transfer verb exists (tracked in #539) the operator deletes the row and the new intended owner re-runs `/qurl-sandbox setup`.

## Test plan

- [x] `TestHandleSetup_OwnerGate` (new) — seven subtests: fresh install (anyone allowed) / owner rerun (idempotent) / non-owner refuse (with owner mention) / added admin refuse (load-bearing safeguard) / shape-bad legacy owner_id (refused without a broken mention) / owner not on admin set (gate keys off owner_id alone) / CheckAdmin error (fail-closed)
- [x] `TestBindWorkspace_DistinguishesSameCallerFromDifferentAdmin` (updated) — owner rerun → AlreadyBoundToCaller; added admin → AlreadyBound; different Slack user → AlreadyBound
- [x] `TestCallbackSeedsAdminOnBind` (updated) — asserts OwnerID equals verified.UserID (Slack ID) rather than the id_token sub
- [x] `go test -race ./apps/slack/...` clean
- [x] `golangci-lint run --new-from-rev=origin/main ./apps/slack/...` — 0 new issues
- [ ] Post-merge: deploy slack-bot to sandbox, delete existing workspace_mappings row, re-run `/qurl-sandbox setup`, confirm owner shows in `/qurl-sandbox admin list`

**Fresh-install ownership claim (race / wrong-user):** on an unbound workspace, `/setup` is first-user-wins — any workspace member who runs it first becomes the owner (the slash gate reads eventually-consistent; `BindWorkspace` consistent-read picks the single winner). If the wrong user claims ownership (e.g. a member beats the intended admin to it), recovery is the same `delete-item` above followed by the intended owner re-running `/qurl-sandbox setup`. Restrict who can reach the command at install time if first-claim ownership matters.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -205,14 +205,17 @@ func run() error {
 			StateSecret:  oauthCfg.OAuthStateSecret,
 			SlackBaseURL: oauthCfg.SlackBaseURL,
 		})
-		// Operator reminder: /qurl setup runs whichever Slack user
-		// invokes it. Workspace identity comes from the signature-
-		// verified payload, but ROLE is not enforced in this binary.
-		// The Slack app manifest must restrict /qurl setup to
-		// workspace admins; without that gate, any user could
-		// initiate a flow that overwrites the workspace's qURL key
-		// against their own Auth0 account.
-		slog.Info("CONFIGURATION REMINDER: /qurl setup is workspace-admin-only by manifest, not by code — verify the Slack app manifest restricts the command")
+		// Operator note: /qurl setup re-runs are now gated in code —
+		// the first user to complete setup on an unbound workspace
+		// becomes its owner, and only that owner can re-run setup
+		// afterward (enforced in handleSetup via AdminStore, with
+		// BindWorkspace as the structural backstop). The remaining
+		// exposure is the first-install claim: any workspace member
+		// can be the first to run /qurl setup and thereby claim
+		// ownership. Restrict who can reach the command at install
+		// time (Slack app manifest / onboarding) if first-claim
+		// ownership matters for this deployment.
+		slog.Info("CONFIGURATION NOTE: /qurl setup re-runs are owner-only (enforced in code); first install is first-user-wins — restrict command availability at install time if first-claim ownership matters")
 	}
 	// Else: buildOAuthConfig already logged the specific missing-var
 	// list; nothing more to say here.
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -735,8 +735,22 @@ func (h *Handler) handleSetup(w http.ResponseWriter, values url.Values) {
 		// otherwise → non-owner rebind attempt, refuse here so we
 		// don't even mint the state token / setup URL.
 		if ownerID != "" && ownerID != userID {
-			slog.Warn("/qurl setup: rebind refused at slash-command gate — caller is not the workspace owner", "team_id", teamID, "caller_user_id", userID, "owner_user_id", ownerID)
-			respondSlack(w, fmt.Sprintf("Only the workspace owner can re-run `/qurl setup`. The current owner is <@%s>. Ask them to re-run setup, or use `/qurl admin` for non-owner admin actions.", ownerID))
+			// Shape-guard the stored owner_id before interpolating it
+			// into a `<@%s>` mention. BindWorkspace writes owner_id
+			// from the OAuth callback (a different code path than the
+			// parser), and a pre-pivot row holds an Auth0 sub, not a
+			// Slack ID — exactly the case the migration runbook calls
+			// out. Mirrors the looksLikeSlackUserID guard in
+			// handleAdminList so a malformed value can't break out of
+			// the mention surface or render visibly broken.
+			ownerMention := "the existing workspace owner"
+			if looksLikeSlackUserID(ownerID) {
+				ownerMention = fmt.Sprintf("<@%s>", ownerID)
+				slog.Warn("/qurl setup: rebind refused at slash-command gate — caller is not the workspace owner", "team_id", teamID, "caller_user_id", userID, "owner_user_id", ownerID) //nolint:gosec // G706: slog escapes control bytes in attribute values; owner_user_id is shape-validated by looksLikeSlackUserID above.
+			} else {
+				slog.Error("/qurl setup: rebind refused; stored owner_id is shape-bad — likely a pre-pivot Auth0 sub. Operator must delete the workspace_mappings row to recover.", "team_id", teamID, "caller_user_id", userID, "owner_id_len", len(ownerID)) //nolint:gosec // G706: slog escapes control bytes in attribute values; caller_user_id is the Slack-payload user ID and owner_id_len is an int.
+			}
+			respondSlack(w, fmt.Sprintf("Only the workspace owner can re-run `/qurl setup`. The current owner is %s. Ask them to re-run setup, or use `/qurl admin` for non-owner admin actions.", ownerMention))
 			return
 		}
 	}
@@ -801,7 +815,8 @@ func (h *Handler) helpMessage() string {
 		"• `/qurl get <url|$name> once:true` — Get a single-use qURL; the link burns on first redemption",
 		"• `/qurl get <url|$name> reason:\"…\"` — Annotate the audit log with a reason",
 		"• `/qurl list` — Show your 5 most recent qURLs",
-		"• `/qurl setup` — Connect qURL to your Slack workspace and become its owner. After the first setup only the workspace owner can re-run this — other admins (added via `/qurl admin add`) can use the rest of the admin verbs but cannot rotate the qURL credential.",
+		"• `/qurl setup` — Connect qURL to your Slack workspace (one-time setup; the runner becomes the workspace owner)",
+		"• Re-running `/qurl setup` is owner-only — added admins (via `/qurl admin add`) can use the other admin verbs but cannot rotate the qURL credential",
 	)
 	if h.aliasStore != nil && h.cfg.AdminStore != nil {
 		if h.cfg.OpenView != nil {
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -760,6 +760,12 @@ func (h *Handler) handleSetup(w http.ResponseWriter, values url.Values) {
 		// get a setup URL, but BindWorkspace's consistent owner check is
 		// the structural backstop — that caller just lands on the
 		// generic rebind-refused page instead of the friendly copy here.
+		// The eventual read is deliberate, not an oversight: CheckAdmin
+		// is the shared admin-gate read (same call the admin verbs make),
+		// so the race only ever costs the loser a less-friendly error
+		// page — never security, since the consistent backstop is
+		// authoritative. Upgrading just this caller to a consistent read
+		// would spend 2x RCU on every /setup to improve one racer's copy.
 		if ownerID != "" && ownerID != userID {
 			// Shape-guard the stored owner_id before interpolating it
 			// into a `<@%s>` mention. BindWorkspace writes owner_id
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -794,7 +794,7 @@ func (h *Handler) handleSetup(w http.ResponseWriter, values url.Values) {
 			// owner is the existing workspace owner".
 			if looksLikeSlackUserID(ownerID) {
 				slog.Warn("/qurl setup: rebind refused at slash-command gate — caller is not the workspace owner", "team_id", teamID, "caller_user_id", userID, "owner_user_id", ownerID) //nolint:gosec // G706: slog escapes control bytes in attribute values; owner_user_id is shape-validated by looksLikeSlackUserID above.
-				respondSlack(w, fmt.Sprintf("Only the workspace owner can re-run `/qurl setup`. The current owner is <@%s>. Ask them to re-run setup, or use `/qurl admin` for non-owner admin actions.", ownerID))
+				respondSlack(w, fmt.Sprintf("Only the workspace owner can re-run `/qurl setup`. The current owner is <@%s>. Ask them to re-run setup. For admin actions that don't need workspace ownership, use the other `/qurl admin` commands.", ownerID))
 			} else {
 				slog.Error("/qurl setup: rebind refused; stored owner_id is shape-bad — likely a pre-pivot Auth0 sub. Operator must delete the workspace_mappings row to recover.", "team_id", teamID, "caller_user_id", userID, "owner_id_len", len(ownerID)) //nolint:gosec // G706: slog escapes control bytes in attribute values; caller_user_id is the Slack-payload user ID and owner_id_len is an int.
 				respondSlack(w, "Only the workspace owner can re-run `/qurl setup`, but this workspace's stored owner record is malformed and can't be displayed (likely a legacy record). Please contact support to recover access.")
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -742,15 +742,16 @@ func (h *Handler) handleSetup(w http.ResponseWriter, values url.Values) {
 			// Slack ID — exactly the case the migration runbook calls
 			// out. Mirrors the looksLikeSlackUserID guard in
 			// handleAdminList so a malformed value can't break out of
-			// the mention surface or render visibly broken.
-			ownerMention := "the existing workspace owner"
+			// the mention surface. Branch the whole reply so the
+			// shape-bad copy reads cleanly rather than "the current
+			// owner is the existing workspace owner".
 			if looksLikeSlackUserID(ownerID) {
-				ownerMention = fmt.Sprintf("<@%s>", ownerID)
 				slog.Warn("/qurl setup: rebind refused at slash-command gate — caller is not the workspace owner", "team_id", teamID, "caller_user_id", userID, "owner_user_id", ownerID) //nolint:gosec // G706: slog escapes control bytes in attribute values; owner_user_id is shape-validated by looksLikeSlackUserID above.
+				respondSlack(w, fmt.Sprintf("Only the workspace owner can re-run `/qurl setup`. The current owner is <@%s>. Ask them to re-run setup, or use `/qurl admin` for non-owner admin actions.", ownerID))
 			} else {
 				slog.Error("/qurl setup: rebind refused; stored owner_id is shape-bad — likely a pre-pivot Auth0 sub. Operator must delete the workspace_mappings row to recover.", "team_id", teamID, "caller_user_id", userID, "owner_id_len", len(ownerID)) //nolint:gosec // G706: slog escapes control bytes in attribute values; caller_user_id is the Slack-payload user ID and owner_id_len is an int.
+				respondSlack(w, "Only the workspace owner can re-run `/qurl setup`, but this workspace's stored owner record is malformed and can't be displayed (likely a legacy record). Please contact support to recover access.")
 			}
-			respondSlack(w, fmt.Sprintf("Only the workspace owner can re-run `/qurl setup`. The current owner is %s. Ask them to re-run setup, or use `/qurl admin` for non-owner admin actions.", ownerMention))
 			return
 		}
 	}
@@ -816,8 +817,16 @@ func (h *Handler) helpMessage() string {
 		"• `/qurl get <url|$name> reason:\"…\"` — Annotate the audit log with a reason",
 		"• `/qurl list` — Show your 5 most recent qURLs",
 		"• `/qurl setup` — Connect qURL to your Slack workspace (one-time setup; the runner becomes the workspace owner)",
-		"• Re-running `/qurl setup` is owner-only — added admins (via `/qurl admin add`) can use the other admin verbs but cannot rotate the qURL credential",
 	)
+	if h.cfg.AdminStore != nil {
+		// The owner-only re-run gate only exists when AdminStore is
+		// wired (it's skipped on the sandbox/no-DDB path), so only
+		// advertise it there — matches how the admin verbs below
+		// render conditionally.
+		lines = append(lines,
+			"• Re-running `/qurl setup` is owner-only — added admins (via `/qurl admin add`) can use the other admin verbs but cannot rotate the qURL credential",
+		)
+	}
 	if h.aliasStore != nil && h.cfg.AdminStore != nil {
 		if h.cfg.OpenView != nil {
 			lines = append(lines,
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -730,10 +730,19 @@ func (h *Handler) handleSetup(w http.ResponseWriter, values url.Values) {
 			return
 		}
 		// ownerID=="" → workspace not yet bound → fresh install, allow.
+		// (CheckAdmin reads eventually-consistent, and BindWorkspace
+		// validates OwnerID != "" before PutItem, so an empty ownerID
+		// reliably means "no row yet" rather than a half-written one.)
 		// ownerID==userID → idempotent rerun by owner (rotates the
 		// API key on OAuth-callback success), allow.
 		// otherwise → non-owner rebind attempt, refuse here so we
 		// don't even mint the state token / setup URL.
+		//
+		// This gate is best-effort: in the brief eventual-read window
+		// after a fresh bind a fast second-mover could still see "" and
+		// get a setup URL, but BindWorkspace's consistent owner check is
+		// the structural backstop — that caller just lands on the
+		// generic rebind-refused page instead of the friendly copy here.
 		if ownerID != "" && ownerID != userID {
 			// Shape-guard the stored owner_id before interpolating it
 			// into a `<@%s>` mention. BindWorkspace writes owner_id
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -689,15 +689,20 @@ func (h *Handler) handleSlashCommand(w http.ResponseWriter, body []byte) {
 // (the alternative, taking team_id from an unsigned query param at
 // /start, was the workspace-rebind primitive flagged in PR review).
 //
-// Admin restriction: this handler does NOT verify the invoking user is
-// a workspace admin. That gate lives in the Slack app manifest — the
-// `/qurl setup` command must be declared admin-only (or restricted via
-// channel/role permissions in the install config). Without that gate,
-// any workspace user could initiate setup and overwrite the workspace's
-// qURL key with one minted against their own Auth0 account. Confirm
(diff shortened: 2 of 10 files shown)


Kevin-layerV:
Addressed the review feedback in e2e70b9:

**1. Missing `looksLikeSlackUserID` guard on the refuse mention (the real bug)** — fixed. `handleSetup` now shape-guards `owner_id` before interpolating it into `<@%s>`, mirroring `handleAdminList`. On a pre-pivot (Auth0-sub) row the reply falls back to "the existing workspace owner" and logs the operator hint your runbook describes (delete the `workspace_mappings` row to recover). Added a `TestHandleSetup_OwnerGate` subtest covering the shape-bad path (refused, no raw sub leaked, fallback copy present).

**2. Incorrect test-fixture comment in `store_test.go`** — fixed. Renamed the IDs to real Slack-ID shape (`U_caller` → `UCALLER01`, etc.) so they actually match `looksLikeSlackUserID`, and corrected the comment to state the real reason (these tests assert at the store boundary, not the handler renderer; the shape just keeps the fixtures honest).

**3. `seedAdmin` fixture diverges from the owner-on-admin-set invariant** — pre-existing, filed as #524 (priority: low).

**4. Help-string item too long** — fixed. Split the `/setup` bullet into two.

**5. No in-band ownership-transfer path** — out of scope, filed as #525 (the workspace-bricking-on-staff-turnover case, priority: medium).

Kevin-layerV:
Addressed the second-round feedback in 75a9fab:

**1. Stale operator startup log (`main.go`)** — fixed. The "admin-only by manifest, not by code" reminder is now false; reworded to describe the actual posture (re-runs owner-only in code; first install is first-user-wins, restrict command availability at install time if first-claim matters).

**2. Tautological shape-bad copy (`handleSetup`)** — fixed. Branched the whole reply; the shape-bad path now reads "...this workspace's stored owner record is malformed and can't be displayed (likely a legacy record). Please contact support to recover access." instead of "the current owner is the existing workspace owner". Updated the test assertion to match (asserts no `<@` mention + "malformed" copy).

**3. Fresh-install posture callout** — captured in the merge-commit body and the reworded `main.go` note (first install = first-user-wins; subsequent = owner-only).

**4. Stale cross-reference (`handler_alias.go`)** — fixed. Dropped the claim that `/setup` shares setalias's manifest posture; noted it's now code-gated.

**5. Help-message under `AdminStore=nil`** — fixed. The "re-running is owner-only" line now only renders when `AdminStore` is wired, matching the conditional admin verbs.

**Nits:** `adminGateBudget` doc now notes `handleSetup` reuses it. The `callback.go` qurlSub phrasing the reviewer flagged is left as-is (reviewer judged the wording correct).

**Coverage gap (slash→callback defense-in-depth test)** — filed as #526. The `BindWorkspace` classification is already covered at the store layer (`TestBindWorkspace_DistinguishesSameCallerFromDifferentAdmin`); the callback-level integration test is additive and tracked there.

Kevin-layerV:
Thanks — addressed the round-3 items (55fa39d):

**1. Eventual-consistency window + 3. `ownerID==""` contract** — folded one tight note into the `handleSetup` gate comment: an empty `ownerID` reliably means "no row yet" because `BindWorkspace` validates `OwnerID != ""` before PutItem, and the slash gate is best-effort (eventual read) with `BindWorkspace`'s consistent check as the structural backstop — a post-bind second-mover lands on the generic refused page, not this friendly copy.

**2. Redundant `BindWorkspace` params** — filed as #527 (signature cleanup, best bundled with the next `workspace_mappings` write-path change, as you suggested).

**4. Louder manifest note in `main.go`** — leaving as-is. The note already states first-install is first-user-wins and to restrict command availability at install time; I'd rather keep the log line tight than expand it further.

**5. Migration runbook / sandbox dry-run** — no code change; this is the existing unchecked item in the PR test plan (`Post-merge: deploy to sandbox, delete the workspace_mappings row, re-run /qurl-sandbox setup, confirm admin list renders the owner`). It'll get ticked during the sandbox dry-run before this touches any production table.

**Nits** — `handler.go` three-state readability is already covered by the doc-comment block; `helpMessage` gating you confirmed good.

Kevin-layerV:
Round-4 feedback addressed in f3c1929:

**1. Missing CheckAdmin-error test (the pre-merge ask)** — added `TestHandleSetup_OwnerGate/CheckAdmin error`: a transient `CheckAdmin` failure must fail **closed** (surface the upstream-error reply, mint no setup URL). Mirrors `TestHandleAdminRevoke_CheckAdminError`.

**2. `qurlSub == ""` comment** — reworded for accuracy. `qurlSub` is the verifier's output (failure logged upstream); the empty-check refuses the bind rather than half-installing, and isn't itself the gate.

**3. Fresh-install race recovery** — added to the PR description: first-user-wins, and if the wrong user claims ownership the recovery is the same `delete-item` + re-run setup; restrict command availability at install time if first-claim matters.

**6. Help-text copy** — "cannot rotate the qURL credential" → "cannot reconnect this workspace's qURL account" (plain register, per our user-facing-copy convention).

Declining (with reasoning) rather than adding code:

**4. Refusal reveals the owner's Slack ID** — this is by design: the message exists to tell the requester whom to ask. Telling someone "ask `@owner`" requires naming the owner. The residual concern (sensitive owner identity in a large org) is a deployment guidance point ("don't install qURL under a sensitive identity"), not a code gate. Happy to add a docs line if you want, but I'd rather not gate it.

**5. Empty `owner_id` corruption edge** — structurally impossible today: `BindWorkspace` validates `OwnerID != ""` before PutItem, so a row can't be written with an empty owner. The `handleSetup` gate comment already documents that an empty `ownerID` reliably means "no row yet" for exactly this reason. Rather than add another comment for a state the writer prevents, leaving as-is.

(Holding the line on comment-volume here — these last two are low-severity by-design / can't-happen cases, and I'd rather not bloat the diff with notes for them.)

Kevin-layerV:
Thanks for the approve. Closing out this round's items — all non-blocking, and I'm treating the review as converged rather than churning another CI cycle for cosmetic polish:

- **(1) manual `gateCancel()` vs `defer`** — intentional (release the gate context before `MintState`/`SetupURL`, which don't use it). Leaving as-is.
- **(2) "one-time setup" copy** — fair, mild contradiction with the owner-only re-run line. Minor enough that I'm not resetting an all-green PR for it; happy to fold `(first runner becomes the workspace owner)` into the next touch of this file.
- **(3) write-time `OwnerID` shape validation** — deliberately leaving the read-side `looksLikeSlackUserID` guard as the single Slack-shape coupling point, per your "up to you / couples slackdata to Slack validation" note.
- **(4) callback integration test for added-admin rerun** — already tracked in #526 (the store-level classifier test + existing `AlreadyBound` callback copy test cover it today).
- **(5) runbook link from the handler comment** — no internal ops doc to link; the recovery is in the PR body.

All 16 checks green on f3c1929.

Kevin-layerV:
Addressed the latest `claude-review` round in 992c77f:

- **`defer gateCancel()`** — swapped the inline cancel for `defer` right after `WithTimeout`, matching the `requireAdminSync` style in `handler_admin.go`. Refactor-safe insurance against a future early-return landing between context creation and the cancel.
- **Missing test coverage** — added a `TestHandleSetup_OwnerGate` subtest: `owner_id` set to the caller but the caller NOT on `admin_slack_user_ids` → `/setup` still mints. This locks in that the gate keys off `owner_id` alone and would catch any regression that re-coupled it to admin-set membership.
- **Help-bullet nit** — tightened the conditional owner-only bullet so it reads as a standalone note rather than a continuation of the `/qurl setup` line.

Verified for the two observations that didn't need code changes:

- **Race-window rebind-refused page leak check** — confirmed `renderRebindRefused` passes only `{{.TeamID}}` (the caller's own workspace ID) into an auto-escaping `html/template`; no `owner_id`/Auth0 sub/admin identity is rendered. The second-mover in the eventual-consistency window lands on a generic "connected under a different admin" page with nothing leaked. That path is pre-existing and untouched here.
- **`isAdmin` not consumed** — the new subtest above also pins that the discarded first return from `CheckAdmin` stays discarded.

Deferred (out of scope for this PR, filed as a labeled follow-up):

- **In-band `/qurl admin transfer-ownership`** — #539. `owner_id` is immutable once set, so owner turnover currently needs the operator `delete-item` runbook. The issue captures scope + acceptance criteria for an owner-gated transfer verb.

The migration-destructive and first-claim-ownership notes are already documented in the PR description's Migration / fresh-install sections.

Kevin-layerV:
Addressed the round-3 `claude-review` in 73ed046:

- **Stale callback wording (required)** — `handleBindError`'s `AlreadyBoundToCaller` branch comment + slog said "caller already on admin set"; that branch now fires iff the caller is the workspace **owner**. Updated it to owner semantics, and the adjacent `AlreadyBound` branch too ("owned by a different Slack user" — fires for added admins, not just strangers) so the whole switch matches the new `BindWorkspace` model.
- **Eventual-read tradeoff (question)** — added a sentence to the gate comment in `handleSetup` documenting the positive reason: `CheckAdmin` is the shared admin-gate read, so the race only ever costs the loser a less-friendly error page, never security (the consistent backstop is authoritative). A consistent read here would spend 2x RCU on every `/setup` to upgrade one racer's copy — not worth it.
- **Deactivated-owner runbook nit** — added a sentence to the PR description's Operator recovery section: the same `delete-item` + re-`/setup` path covers an owner who later leaves/is deactivated, until the in-band transfer verb (#539) exists.

Also corrected the test-plan bullet in the description (it said "four subtests"; there are now seven, matching what the review enumerated).

No behavior change in this commit — comment/log wording + docs only.

Kevin-layerV:
Addressed the round-4 `claude-review` in 095535e + a PR-description update:

1. **Help line verbosity** — tightened to "Re-running `/qurl setup` is owner-only — only the workspace owner (whoever first connected qURL) can reconnect the account". Dropped the forward-looking "added admins" clause; the other `/qurl admin` verbs are already listed below in the same help block, so the disambiguation isn't lost. Avoided the gate-on-admin-set-size option since that would add a DDB read to help-text rendering.

2. **Prod migration scope** — confirmed and added to the description: the Slack bot is **sandbox-only today**. The infra (`qurl-integrations-infra/bootstrap`) hardcodes `environment = "sandbox"` for every bot table (prefix `qurl-bot-slack-sandbox-`), and the prod Slack-OAuth + infra chain hasn't shipped. So there are **no prod `workspace_mappings` rows** to migrate — prod will deploy this post-pivot binary from the start and write `owner_id` = Slack ID on the first `/setup`, never an Auth0 sub. The migration block is purely sandbox-scoped.

3. **Concurrent owner `/setup` → two rotations** — agree this isn't a regression from this PR, and it's a qurl-service-side concern (the bot just re-mints). Both rotations belong to the same owner so no key is "lost"; last-write-wins is benign for a same-owner flow. Left as-is (cross-repo, pre-existing).

4. **`handleSetup` doc wording** — fixed. The "qURL OAuth is not configured" reply is gated on `oauthSetup == nil` (checked first), not `AdminStore == nil`; reworded so the two short-circuits read as distinct.

5. **`owner_id_len` log field** — intentional. Logging the length (not the value) is the deliberate privacy choice so a legacy Auth0 sub never lands in logs; the `delete-item` + re-`/setup` runbook doesn't need the stored value to recover, so the conservative field is sufficient.

No behavior change in this commit — comment/help-copy + docs only.

Kevin-layerV:
Addressed the round-5 `claude-review` in f716ce0 (all notes were non-blocking):

2. **slog field drift (fixed)** — harmonized the "owner check failed" log from `user_id` to `caller_user_id`, so all three `handleSetup` gate logs use the same field. `owner_user_id` stays distinct (it's the stored owner, not the caller).

Kept as-is, with rationale:

1. **`CheckAdmin` overloaded** — agree it's two reads in a trench-coat, and agree it's not worth splitting for a single GetItem today. Captured the "introduce a thinner `WorkspaceOwner(teamID)` accessor if a third caller appears" thought; leaving the shared call for now.

3. **Two `/qurl setup` help lines** — kept deliberately. The base `/qurl setup` line always renders, but the owner-only line is gated on `AdminStore != nil` because the owner gate only exists there (skipped on the sandbox/no-DDB path). Folding both into one unconditional line — as suggested — would advertise "only the owner can re-run" on the sandbox path where the gate is actually skipped, which would be inaccurate. The two-line split is what keeps the advertised behavior matching the rendered behavior.

5. **fresh-install test asserts on the `/oauth/qurl/start?state=` substring** — intentional; agreed the lockstep failure on a URL-surface refactor is a feature (loud, not silent).

For the security-section note: filed the "re-audit the consistent-read backstop + eventual gate together once ownership-transfer relaxes `owner_id` immutability" scope onto #539 so it isn't lost.

f716ce0 is a single slog-field rename — no behavior change.

Kevin-layerV:
Addressed the round-6 `claude-review` in 6b78f3d (all non-blocking):

1. **Empty `owner_id` fallthrough is implicit (fixed)** — `BindWorkspace` now logs a `slog.Warn` when a row exists but `owner_id` is empty (a manually edited / truncated row). It still refuses via `AlreadyBound` (the safe default), but the malformed row is now grep-able instead of silently routing to the generic rebind-refused page.

4. **`baseCtx` rationale (fixed)** — added a sentence to the gate comment: the read times off `h.baseCtx` (not the request ctx) so a Slack-side connection-close can't truncate it mid-flight; `adminGateBudget` is the only bound. Same posture as `requireAdminSync`.

5. **First-install race in the doc (fixed)** — the `handleSetup` doc block now states first-install is first-user-wins (consistent `BindWorkspace` read picks the winner, loser gets the rebind-refused page), so the full security-model story is in one place.

**Test-coverage gap (fixed)** — added a `TestHandleSetup_OwnerGate` subtest for the `AdminStore == nil` sandbox/no-DDB path: the owner gate is skipped and `/setup` mints unconditionally (even for a non-owner). The contract is now pinned in this file rather than only implied.

Kept as-is, with rationale:

2. **Shape-bad copy "contact support"** — kept. The malformed-record recovery genuinely requires LayerV-side DDB intervention (`delete-item`), which a Slack workspace admin/operator can't perform — so "contact support" (LayerV) is the accurate routing here, not "workspace operator." (For the *normal* non-owner case the copy already routes to the owner via `<@…>`.)

3. **Two `/qurl setup` help bullets** — kept, same reason as the round-5 note: the owner-only bullet is gated on `AdminStore != nil` because the gate only exists there; folding it into one unconditional bullet would advertise owner-only behavior on the sandbox path where the gate is skipped.

6b78f3d: one defensive log + one test + two doc sentences. No behavior change beyond the new Warn.

Kevin-layerV:
Addressed the round-7 `claude-review` in db7242f (both nits were adjacent to the round-6 additions, neither blocking):

1. **`ownerID == ""` gate-comment claim (fixed)** — the round-6 empty-`owner_id` handling made the old "empty ownerID reliably means no row yet" claim stale. Softened it: a manually-edited blank-owner row also reads as `""` at the gate and slips through, but `BindWorkspace`'s consistent check refuses it (and now Warns). It's DDB-tampering territory, so it isn't worth a second read to distinguish at the gate — the comment now says so.

2. **Help line inaccurate on the AdminStore=nil path (fixed)** — made the "(one-time setup; the first runner becomes the workspace owner — only they can re-run it)" parenthetical conditional on `AdminStore != nil`, so on the sandbox/no-DDB path the bullet is just "Connect qURL to your Slack workspace" (no false ownership claim, since the gate is skipped there). This also folds the previous two setup bullets into one, resolving the earlier "two bullets for one verb" style note.

**Test gap (closed at the safety-critical layer)** — added a `BindWorkspace` subtest pinning that a row with an empty `owner_id` refuses via `AlreadyBound` and never short-circuits to `AlreadyBoundToCaller` (the `"" == ""` hijack the seedAdmin-is-non-empty invariant prevents). I focused the test on the store layer since that's where the safety decision is made; the slash-gate side just mints and relies on this backstop, which is the documented (DDB-tampering-only) behavior.

On the staging/canary question: confirmed earlier — the Slack bot is sandbox-only (infra hardcodes `environment = "sandbox"`; the prod chain hasn't shipped). Pre-pivot rows exist only in the sandbox table, and the documented `delete-item` runbook is the operator's audit/recovery path. There are no prod rows.

db7242f: one conditional help line + one softened comment + one store-layer test. No behavior change.

Kevin-layerV:
Addressed the round-8 `claude-review` in 871b153 (all cosmetic, none blocking):

2. **Contradictory help wording (fixed)** — dropped "one-time setup" since it read contradictorily next to "only they can re-run it." The parenthetical (AdminStore-gated) now reads "(the first runner becomes the workspace owner; only they can re-run it)".

4. **`owner_id` / `seed_admin_slack_user_id` equal-by-construction (commented)** — added a note at the `BindWorkspace` write site: both get `verified.UserID` at first bind by construction, kept distinct because `owner_id` is the mutable-someday gate anchor (the #539 transfer verb) while `seed_admin_slack_user_id` is the write-once forensic record. So a future reader won't collapse them.

Kept as-is, with rationale:

1. **`defer gateCancel()` lifetime** — leaving it, per your own "leave it unless you're already refactoring." Worth noting this is the exact shape an earlier review round asked for (inline cancel → `defer` for refactor-safety / parity with `requireAdminSync`); the lingering-resource cost is microseconds through `MintState`, so the refactor-safety win dominates.

3. **Deactivated-owner `<@DELETED_USER_ID>` greys out** — agreed; the runbook recovery is sound and the in-band transfer is tracked in #539. No change here.

5. **Ordering of the empty-teamID/userID guard before the gate** — confirmed correct; no `CheckAdmin("", "")`. No change.

**Test gap (caller IS the malformed owner)** — agreed it's structurally impossible: `userID` comes from the signed Slack payload (always valid shape) and a pre-pivot Auth0 sub can't equal a Slack ID, so `ownerID == userID` can't hold for a shape-bad owner. Not adding a test for an unreachable state.

871b153: one help-copy fix + one clarifying comment. No behavior change.

Kevin-layerV:
Addressed the round-9 `claude-review` in 576514f (all optional nits):

3. **Empty-owner Warn attribute (added)** — added `admin_count` to the `BindWorkspace` empty-`owner_id` Warn so on-call can distinguish a truncated row from a row where only `admin_slack_user_ids` was set, at a glance.

5. **Non-owner refusal copy (clarified)** — replaced the jargon "use `/qurl admin` for non-owner admin actions" with "For admin actions that don't need workspace ownership, use the other `/qurl admin` commands" — clearer for a workspace admin reading the refusal.

2. **Operator-recovery race (filed as follow-up)** — added the caveat to #539: between `delete-item` and the intended owner's `/setup`, a racing member can claim ownership. The PR body's "restrict command reachability" is the mitigation; the in-band transfer verb removes the `delete-item` path (and the race) entirely.

No change, with rationale:

1. **`CheckAdmin` discards `isAdmin`** — confirmed there's no `isAdmin`-keyed side effect in `CheckAdmin` today, so this caller isn't surprised. Documented inline; leaving the shared read.

4. **`qurlSub == ""` → 500 on a hypothetical JWKS-rotation grace window** — the verifier returns `("", err)` on failure, not `("", nil)` on a valid-but-empty case, so the 500 only fires on genuine verification failure. Refusing the bind is the correct safe default for an identity-proof gate; I'd rather 500 a rare install than persist an unverified owner. No change.

576514f: one log attribute + one copy clarification. No behavior change.

Kevin-layerV:
Addressed the round-10 `claude-review` in eb480d4:

1. **Redundant test coverage (fixed)** — good catch. "owner reruns setup" used `seedAdmin`, which puts a *different* user on the admin set, so the owner was off-set there too — identical to "owner not on admin set." Reseeded "owner reruns setup" with the realistic post-bind state (owner IS the sole admin, which is what `BindWorkspace` writes at first bind). The two subtests now genuinely diverge (owner-on-set vs owner-off-set) and together prove the gate keys off `owner_id` regardless of admin-set membership. Also fixed the comment, which was wrong.

Already covered / no change, with rationale:

4. **Callback `verified.UserID` duplication** — the inline rationale you're asking for already exists in two places: the `checkBindAllowed` doc block (explains `verified.UserID` becomes both `owner_id` and the admin seed, equal at first bind) and the `BindWorkspace` write-site comment added earlier (why they're kept structurally distinct for the #539 transfer verb). A future reader hits one of those before collapsing the params.

6. **`oauthSetup==nil` vs `AdminStore==nil` parallel** — already clarified in the `handleSetup` doc: it explicitly says the `AdminStore==nil` skip is "a separate short-circuit from the `oauthSetup==nil` check below… which fires first, before AdminStore is consulted." That's the distinction you're flagging.

2. **`defer gateCancel()`** — leaving it, as you note ("consistency is the argument for leaving it"); matches `requireAdminSync`.

5. **First-install race / manifest** — one correction worth flagging: Slack does **not** enforce admin-only on slash commands — the "admins only" manifest setting is cosmetic, not an authz gate (which is exactly why this PR moves the gate into code). So the operator mitigation isn't "confirm the manifest restricts `/setup` to admins"; it's restricting who can reach the command at install time (channel/workspace install scoping). The `cmd/main.go` log says "restrict command availability at install time," which is accurate and deliberately doesn't claim the manifest enforces it.

3. **Backfill script vs N `delete-item`s** — genuinely useful operational idea (translate legacy Auth0-sub `owner_id` → Slack ID via `seed_admin_slack_user_id`, which is a Slack ID even on legacy rows). It's a deploy-ops decision, not PR code, and sandbox-only — captured for the operator runbook rather than scripted here.

**End-to-end `/qurl admin list` smoke after fresh bind** — agreed it'd guard against future `BindWorkspace`-writes vs `handleAdminList`-reads drift. That's a broader integration test (needs the full bind→render round-trip) better suited to a follow-up than this PR; the unit layers pin both sides (`BindWorkspace` writes a Slack-ID `owner_id`; `handleAdminList` shape-checks it) for now.

eb480d4 is a single test-fixture fix. No production-code change.

Kevin-layerV:
Considered every round-11 item. Several reverse feedback from earlier rounds, so I'm holding the current shape rather than oscillating — rationale per item:

1. **Sandbox help-text should learn the owner concept** — holding. This would reverse the round-7 change (making the owner parenthetical conditional on `AdminStore != nil`), which was made precisely because the sandbox/no-DDB path *skips* the owner gate — re-running `/setup` there genuinely does just rotate keys with no ownership. Advertising "the first runner becomes the owner" on a deploy where that isn't true would be the inaccuracy round-7 flagged. The help text is intentionally accurate per-deployment.

3. **Runbook downstream-key-impact sentence** — for the Slack bot specifically, the bot *is* the consumer of the `workspace_keys` row and re-mints on the next `/setup`; there's no external consumer holding the prior key to coordinate. The PR description already notes the key is rotated. Adding "downstream consumers break until they re-pull" would be misleading in this bot context, so holding.

4. **`defer gateCancel()` placement** — holding (as you note, "leaving as-is is reasonable"). Matches `requireAdminSync`; the context's work is done before function exit. This has been raised and resolved "leave it" across several rounds.

5. **Comment density in `handleSetup`** — holding. These are the load-bearing security WHYs (eventual-read rationale, the two distinct nil short-circuits) added over prior review rounds; you've also called them "justifiable for security-sensitive code." Moving them to package-doc would separate the rationale from the branch it explains.

6. **Slash-gate empty-`owner_id` subtest** — holding. The authoritative behavior (refuse, never hijack) is already pinned at the `BindWorkspace` layer. At the slash gate an empty owner intentionally reads as fresh-install and falls through to the backstop — documented at `handler.go:761-767`. A test there would pin a deliberately-degenerate, DDB-tampering-only path; low value vs. the store-layer guard that already exists.

Nits:
- **:797 "actions that don't need workspace ownership"** — this line was reworded in round 9 (to de-jargon "non-owner admin actions"); round 11 wants it reworded again for inversion. Both readings are understandable, and re-touching it tends to just shift the nit. Holding the round-9 wording.
- **:800 "likely a legacy record"** — the surrounding sentence already says "malformed and can't be displayed … please contact support," which is self-explanatory for an end user; "likely a legacy record" is a non-load-bearing parenthetical that doesn't impede comprehension. Holding.

2 (CheckAdmin error reply) and the `WorkspaceMapping` doc praise — noted, no change requested.

Net: the security model and tests are settled; the open items are cosmetic and several reverse prior accepted feedback, so this is the convergence point. CI is fully green (16/16). Ready for a human approving review.

Issue #350, opened 2026-05-14
feat(slack): /qurl setalias — gate public-host targets at parse time (SSRF-adjacent)


Kevin-layerV:
Follow-up from [PR #347](https://github.com/layervai/qurl-integrations/pull/347) claude-bot review item #3.

## Background
`/qurl setalias` accepts any `http://` or `https://` URL, including `http://localhost`, `http://10.0.0.1`, `http://169.254.169.254/` (AWS IMDS), `http://[::1]`, etc. The PR is admin-gated at the Slack-manifest level so the immediate risk is low, but if qurl-service ever resolves alias targets server-side at lookup time, internal-network targets become an SSRF surface.

The PR fences the current behavior with a test row (`localhost target ACCEPTED (TODO: SSRF gate)` in `handler_alias_test.go`) so the contract is reviewable.

## Acceptance criteria
- [ ] Parse-time rejection of RFC1918, link-local, loopback, and IPv6 ULA targets in `parseAliasArgs`.
- [ ] Optional allowlist env var for sandbox/dev that opens the gate (`QURL_ALLOW_PRIVATE_ALIAS_TARGETS=true`).
- [ ] Test rows flipped from ACCEPTED to wantErr for the listed cases.

## Out of scope
- Any change to qurl-service's own URL validator.

PR #512, opened 2026-05-27
fix(slack): unwrap <url|display> in setalias target

Kevin-layerV:
## Summary

- Strip Slack's `<url|display>` autolink wrapping from the `setalias <target>` token before `url.Parse`, so the admin command keeps accepting URLs after `/qurl-admin` switches to `should_escape: true` (layervai/qurl-integrations-infra#798).
- Adds 2 unit-test rows in `TestParseAliasArgs_SetAlias` covering the `<url|display>` and `<url>` forms.

## Why now

`/qurl-admin` is being flipped to `should_escape: true` in qurl-integrations-infra#798 so that the new `admin add <@U…>` / `admin remove <@U…>` verbs at `parser.go:243` actually receive wrapped user mentions (the regex requires the `<@U…>` form; raw `@username` text would silently reject). The same flag wraps user-typed URLs as `<url|display>`, which the existing `parseAliasArgs` path at `handler_alias.go:181` rejected via `url.Parse → empty Scheme`. This patch closes that gap.

Resource-id (`r_…`) and tunnel-slug (`$slug`) targets aren't auto-linked by Slack, so they pass through unchanged.

## Test plan

- [x] `go test -race -count=1 ./apps/slack/...` (full slack suite, all green)
- [x] `golangci-lint run --timeout=5m --new-from-rev=origin/main ./apps/slack/...` (0 new issues)
- [ ] After merge, with infra#798 also merged: `/qurl-sandbox-admin setalias $foo https://example.com` round-trips and the stored target is `https://example.com` (not the wrapped form)

### DIFF
--- a/apps/slack/internal/handler_alias.go
+++ b/apps/slack/internal/handler_alias.go
@@ -178,12 +178,32 @@ func parseAliasArgs(text string, wantTarget bool) (parsed *aliasArgs, userMsg st
 		out.Target = "$" + slug
 		return out, ""
 	}
+	// Slack's `should_escape: true` (set on /qurl-admin so `admin add
+	// <@U…>` resolves) wraps user-typed URLs as `<url|display>`. Strip
+	// the wrapping so url.Parse below sees a bare URL and the stored
+	// target stays clean.
+	tgt = unwrapSlackAutolink(tgt)
 	u, err := url.Parse(tgt)
 	if err != nil || (u.Scheme != schemeHTTP && u.Scheme != schemeHTTPS) || u.Host == "" {
 		return nil, msgAliasTargetInvalid
 	}
 	out.Target = tgt
 	return out, ""
+}
+
+// unwrapSlackAutolink strips the `<url|display>` (or `<url>`) wrapping
+// Slack applies to typed URLs when the slash command has
+// `should_escape: true`. Returns the inner URL when wrapping is
+// present, else returns the input unchanged.
+func unwrapSlackAutolink(tgt string) string {
+	if len(tgt) < 2 || tgt[0] != '<' || tgt[len(tgt)-1] != '>' {
+		return tgt
+	}
+	inner := tgt[1 : len(tgt)-1]
+	if i := strings.IndexByte(inner, '|'); i >= 0 {
+		return inner[:i]
+	}
+	return inner
 }
 
 // requireAlias checks that `tok` is `$<alias>` and returns the alias
--- a/apps/slack/internal/handler_alias.go
+++ b/apps/slack/internal/handler_alias.go
@@ -198,7 +198,9 @@ func parseAliasArgs(text string, wantTarget bool) (parsed *aliasArgs, userMsg st
 // substitutions Slack applies under the same flag. The entity decode
 // only runs when the wrap is present — that's the signal `should_escape`
 // fired — so a user who literally typed `&amp;` outside an escaped
-// command keeps it intact.
+// command keeps it intact. Splits on the *first* `|` to match Slack's
+// `<url|display>` grammar; a URL containing a raw `|` would be ambiguous
+// here, and Slack autodetect percent-encodes `|` to `%7C` anyway.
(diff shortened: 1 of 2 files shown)


Kevin-layerV:
Round 2 (entity decode + regression test).

- **HTML-entity gap** — fixed. `unwrapSlackAutolink` now `html.UnescapeString`s the inner only when the wrap was present (the `should_escape` signal). Test row pins `?a=1&amp;b=2` round-trips to `?a=1&b=2`.
- **`<@U…>` rejection** — added as a regression row; relies on the existing scheme gate, no special-case in the unwrap.
- **`|` inside URL** — leaving as-is. Slack percent-encodes `|` to `%7C` in autolinks, so the truncation can't fire from real input; adding a test row would pin behavior we never reach.
- **Display-text differs from URL** — same path as the `|url|display` row already in tests; not adding a redundant row.

Kevin-layerV:
Round 3 (doc tweak + responses to suggestions).

- **(2) First-pipe split intent** — added a one-liner on `unwrapSlackAutolink` noting the deliberate first-`|` split + that Slack percent-encodes raw `|` to `%7C` upstream. That closes the LastIndexByte hazard you flagged.
- **(1) Direct unit test on `unwrapSlackAutolink`** — skipping. The 5 table rows on `parseAliasArgs` already cover every branch; an isolated test would be near-duplicate coverage, and #350's SSRF gate won't affect the unwrap-shaped cases (it tightens the host gate, not the angle-bracket parse). Happy to revisit if #350 turns out to refactor more than expected.
- **(3) Audit-log post-decode** — confirmed, no change.
- **(4) Broader contract outside should_escape window** — acknowledged. The new tolerance is for `<https://x>` only (no `|`, no entities), which is more lenient than before but never reachable from real Slack input. Not regressing anything that was previously rejected for a good reason.
- **(5) `&lt;`/`&gt;` row to back the doc claim** — skipping. The only realistic source of `&lt;`/`&gt;` is a user literally typing `<` / `>` in a URL, which is invalid input regardless. The `&amp;` row pins the realistic case; expanding the table to match the doc's enumeration would be coverage theater.

Kevin-layerV:
Round 4 (entity-decode bypass fix).

Good catch on `html.UnescapeString` decoding the full HTML5 entity set. Swapped to `strings.NewReplacer("&amp;", "&", "&lt;", "<", "&gt;", ">")` — matches Slack's documented set, keeps numeric refs like `&#96;` (backtick) and `&#1;` from sneaking past the upstream byte fence. Added a regression row pinning `&#96;` survives unmodified.

Skipping the naming nit (`tgt` → `s`); it stays consistent with the calling site. `IndexByte` byte-level intent already noted in the docstring.

Kevin-layerV:
Round 5 (push-back on remaining nits + comment trim).

Holding the line on the 4 optional nits — adding 4 more test rows for behaviors the function docstring already pins would be coverage theater:

- **`<url|extra|stuff>` first-pipe row** — docstring explicitly calls out the choice; no future refactor reaches `LastIndexByte` without also revising the comment.
- **`<>` empty wrap row** — `len < 2` check + downstream `url.Parse("")` already reject; no path-not-covered.
- **`<mailto:…>` wrapped form** — same code path as the existing bare `ftp://` row.
- **`slackEscapeUndo` naming** — pure preference, leaving consistent with how the rest of the file names internal helpers.
- **Forward-compat marker re: Slack expanding the escape set** — the `slackEscapeUndo` doc already states "exactly the three substitutions Slack applies"; adding a TODO would just decay.

Also trimmed a duplicate 4-line comment at the call site (the function's own docstring covers the same ground).

Kevin-layerV:
Round 6 (docstring clarification only).

Added the one-line note on `unwrapSlackAutolink` that the entity decode is intentionally gated on wrap presence. Skipping the 3 nice-to-have test rows — they pin behaviors already covered by adjacent rows or by url.Parse's empty-string handling.

Kevin-layerV:
closing as we are only supporting tunneled resource qurls through alias and slug

PR #523, opened 2026-05-28
ci: notify Slack on main CI failures

justin-layerv:
## Summary
- add a centralized `workflow_run` notifier for main-branch CI failures
- notify via the org-level `SLACK_WEBHOOK_URL` so failures land in `#build-notifications` even when deploy dispatch is skipped
- remove the old Discord-only inline notifier to avoid duplicate messages and route Discord/Slack/shared/security CI through one notification path

## Why
The guided Slack installer merged, but `slack: Build and Test` failed on `main` before `trigger-deploy` ran. Since only deploy workflows were posting to Slack, the failure never appeared in `#build-notifications`. This makes upstream main CI failures visible at the source.

## Covered workflows
- `slack: Build and Test`
- `discord: Build and Test`
- `Shared Tests (All Apps)`
- `CodeQL`
- `Secrets Scan`
- `Release Please`
- `Validate issue templates`

## Validation
- `actionlint .github/workflows/*.yml`
- `git diff --cached --check`
- local jq payload smoke test

### DIFF
--- a/.github/workflows/discord.yml
+++ b/.github/workflows/discord.yml
@@ -70,46 +70,6 @@ jobs:
         working-directory: apps/discord
         run: docker build -t discord-bot-test .
 
-  # Pages on post-merge CI failures via Slack #all-layerv so a stale-
-  # merge-ref bug (or any other class of regression that slipped past
-  # PR-time CI) surfaces immediately, instead of staying silent behind
-  # a `trigger-deploy: skipped` like #202 did.
-  #
-  # Routes through SLACK_WEBHOOK_URL repo secret — the team already
-  # uses Slack for prod alerting and has no GitHub-issue alerting
-  # attached, so an issue would have been silent. Webhook posts to
-  # #all-layerv (the channel the secret is bound to).
-  notify-on-main-failure:
-    needs: [build-and-test, docker-check]
-    if: failure() && github.ref == 'refs/heads/main' && github.event_name == 'push'
-    runs-on: ubuntu-latest
-    steps:
-      - name: Notify #all-layerv on Slack
-        uses: slackapi/slack-github-action@45a88b9581bfab2566dc881e2cd66d334e621e2c # v3.0.3
-        with:
-          webhook: ${{ secrets.SLACK_WEBHOOK_URL }}
-          webhook-type: incoming-webhook
-          payload: |
-            {
-              "text": ":rotating_light: discord CI failed on main \u2014 `${{ github.sha }}`",
-              "blocks": [
-                {
-                  "type": "section",
-                  "text": {
-                    "type": "mrkdwn",
-                    "text": "*:rotating_light: discord CI regression on main*\n`build-and-test` and/or `docker-check` failed after merge. `trigger-deploy` is *skipped* \u2192 no new image in ECR; sandbox bot is on the prior image; `promote-bot-discord-to-prod.yml` will refuse at the sandbox-status-check gate. Open a hotfix PR against this SHA. See <https://github.com/${{ github.repository }}/pull/204|#204> for the pattern."
-                  }
-                },
-                {
-                  "type": "section",
-                  "fields": [
-                    { "type": "mrkdwn", "text": "*Commit*\n`${{ github.sha }}`" },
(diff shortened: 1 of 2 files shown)


justin-layerv:
Addressed Claude review feedback in the latest push:

- included scheduled CodeQL failures by allowing `workflow_run.event == schedule`
- removed `action_required` from the alert conclusions
- removed the no-op per-run concurrency block
- restored operational impact text for Slack/Discord deploy-dispatch skips while keeping one central notifier
- relabeled `Actor` to `Triggered by`

Validation after the changes:
- `actionlint .github/workflows/*.yml`
- `git diff --check`
- local jq payload smoke test
- `codex review --uncommitted` simplify-style pass: clean

justin-layerv:
Addressed the second Claude/Codex review pass in the latest push:

- kept `cancelled` in the alert set, with a comment explaining why default-branch cancellations are exceptional and should page
- added top-level rationale for why this notifier lives in the source repo instead of infra
- added a workflow-name coupling comment above the impact `case` block
- switched the branch filter/message from hard-coded `main` to `github.event.repository.default_branch`
- restored the Discord #204 hotfix-pattern link
- added 3-attempt Slack webhook retry with backoff

Validation after this batch:
- `actionlint .github/workflows/*.yml`
- `git diff --check`
- local jq payload smoke test asserting default-branch text and Discord #204 impact link
- `codex review --uncommitted`: clean

justin-layerv:
Addressed the latest Claude review pass in the newest push:

- expanded the workflow-name coupling comment to include the more dangerous trigger-list rename failure mode
- changed the Discord hotfix-pattern link to use `${REPOSITORY_URL}/pull/204` instead of hard-coding the org/repo
- bounded the notifier job with `timeout-minutes: 5` and each Slack POST with `--max-time 30`
- simplified the Slack run link to `#run_number` and removed the unused run-id env/arg

Notes on the remaining review questions:
- `cancelled` is intentionally still included; default-branch app workflows do not auto-cancel, so a cancellation before deploy dispatch should be visible. The workflow now documents that tradeoff.
- I checked the dependency age-check workflows; in this repo they are PR/manual only, not scheduled. CodeQL is the scheduled workflow covered by this notifier.
- The target Slack channel is controlled by the org-level `SLACK_WEBHOOK_URL` incoming-webhook binding; that channel routing is not visible in repo code.

Validation after this batch:
- `actionlint .github/workflows/*.yml`
- `git diff --check`
- local jq payload smoke test asserting default-branch text, the repo-derived #204 URL, and compact run link
- `codex review --uncommitted`: clean

justin-layerv:
Final optional-polish pass pushed:

- documented `workflow_run` first-run semantics: the notifier can only fire after this file exists on the default branch
- added `actor_display=${ACTOR:-unknown}` so scheduled/edge-case events do not render a blank Slack field
- filed follow-up #528 for a future CI assertion that workflow names, trigger-list entries, and impact mappings stay in sync

Validation after this pass:
- `actionlint .github/workflows/*.yml`
- `git diff --check`
- local jq payload smoke test for actor fallback
- `codex review --uncommitted`: clean

justin-layerv:
Addressed the latest LGTM polish notes in the current push:

- documented that missing `SLACK_WEBHOOK_URL` intentionally fails red because the notification surface is broken
- display `Triggered by: schedule` for scheduled workflow failures instead of attributing cron failures to the workflow-file actor
- made the context copy less deploy-specific so scheduled security failures read correctly

Validation:
- `actionlint .github/workflows/*.yml`
- `git diff --check`
- local jq payload smoke test for scheduled actor/context text
- `codex review --uncommitted`: clean

justin-layerv:
Channel-binding follow-up: this workflow intentionally uses the org-level `SLACK_WEBHOOK_URL`; GitHub exposes the secret name but not the Slack incoming-webhook target channel. The operational contract for this PR is that `SLACK_WEBHOOK_URL` is bound to `#build-notifications` before merge. The PR description and Slack message copy both name `#build-notifications` so any future channel drift is visible in review. No workflow code change is needed here because channel routing is owned by the Slack webhook configuration.

PR #530, opened 2026-05-28
feat(slack): list tunnels only with slug tokens; /qurl get one-time by default

Kevin-layerV:
## Summary

Tightens the Slack bot's `/qurl list` and `/qurl get` surfaces so they speak in tunnel slugs instead of opaque internal identifiers, and makes one-time use the default for `/qurl get`.

### `/qurl list` — tunnels only, slug as the token
- Lists **only tunnel resources** (`type=tunnel`). URL/transit resources are filtered out.
- Each row renders the tunnel **slug** as the copy-paste-ready `$<slug>` token, with a `slug → alias → resource_id` fallback chain. The opaque `r_<id>` only appears for a legacy tunnel that has neither a slug nor an alias (vanishingly rare).
- Dropped the now-redundant `(tunnel)` label and `[slug:…]` fragment — the whole list is tunnels and the token **is** the slug.

**Before → After:**
```
• `$r_0pxn89qcm9i` → (tunnel) [slug:kktest] — Slack tunnel install for kktest
• `$kktest` → Slack tunnel install for kktest
```

The slug is the same handle `/qurl tunnel install <slug>` already binds as a channel alias, so a listed `$<slug>` pastes straight into `/qurl get $<slug>` and resolves.

The list fetches one page at the server's max (100 resources) before filtering client-side — the resources API has no `type` filter, and a small page could otherwise surface zero tunnels in a URL-heavy workspace. Channel scoping, fail-closed posture, and the has-more footer are unchanged (footer copy now refers to tunnels).

### `/qurl get` — one-time use is the default
- Every `/qurl get` link is now one-time use, unconditionally. The `(one-time use)` suffix always appears so the expectation is set.
- The `once` flag is **removed**. A stray `once:true`/`once:false` is rejected with a transitional hint ("`once` is no longer needed — every `/qurl get` link is one-time use by default") rather than a generic "unknown flag", for users with saved recipes.
- `/qurl get $<slug>` now resolves **tunnel slugs**, not just channel `set-alias` bindings: when the alias lookup misses, it falls back to a slug→resource_id lookup gated by the same channel allow-set as the `$r_<id>` form (admin bypass / `allowed_resource_ids`). This keeps the `/qurl list` → `/qurl get` round-trip honest for tunnels surfaced via admin-sees-all or cross-channel allow (which have no binding in the current channel) — without re-exposing `r_<id>` in the list.

### Help / usage copy
- Get usage collapses to two forms: `/qurl get <url|$slug>` and `/qurl get <url|$slug> reason:"…"`. The `once:true` line is gone; the conditional `dm:true` line is unchanged.
- No more mentions of resource IDs: the list empty-state now points at `/qurl tunnel install <slug>` (the old copy referenced the deprecated `/qurl create`), and `set-alias` usage reads `<url-or-$slug>`.

## Notes
- **`/qurl list` no longer shows URL shortcuts.** A `set-alias`-bound URL/transit resource used to appear in `/qurl list`; it is now filtered out (tunnels only, per this PR's intent). Such shortcuts stay discoverable via `/qurl aliases` and usable via `/qurl get $<shortcut>`.
- The `$r_<id>` form still works in `/qurl get` (admins / channel allow-set) and `set-alias` still accepts a resource_id target — these are just no longer advertised in user-facing copy.
- `dm:true` is left intact (out of scope; it's a separately-wired, gated feature).
- The deprecated `/qurl create` continues to point users at `/qurl get`.

## Test plan
- Rewrote `handler_list_test.go` for the tunnel-only + slug-token behavior: slug-is-token, slug-wins-over-alias, alias/resource_id fallbacks, no `(tunnel)`/`[slug:]` decorations, URL resources filtered out, description rendering, channel scoping + fail-closed + pagination/footer, stable sort.
- Updated parser/get tests: removed the `once` happy-path/rejection cases and `Command.Once()` test, added regression pins that `once:true`/`once:false` are rejected with the transitional hint, and replaced the once-flag JSON tests with a single "one-time is the default" test. Added `TestTunnelToken`, the slug-fallback round-trip tests (allowed-set, admin-bypass, not-allowed), and a list pagination-gap-with-zero-tunnels pin.
- `go build ./...`, full `go test ./apps/slack/...`, `gofmt`, and `golangci-lint` (0 new issues vs `main`) all pass.

### DIFF
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -760,14 +760,14 @@ func (h *Handler) helpMessage() string {
 		"*/qurl* — Create and manage qURLs from Slack",
 		"",
 		"*Commands:*",
-		"• `/qurl get <url|$slug>` — Mint a one-time qURL for a URL or a `$slug` your Slack admin has configured in this channel",
+		"• `/qurl get <url|$slug>` — Mint a one-time qURL for a URL, or a `$slug` (tunnel or shortcut) configured in this channel",
 	}
 	if h.cfg.PostDM != nil {
 		lines = append(lines, "• `/qurl get <url|$slug> dm:true` — DM the link to you instead of posting it in-channel")
 	}
 	lines = append(lines,
 		"• `/qurl get <url|$slug> reason:\"…\"` — Same, annotating the audit log with a reason",
-		"• `/qurl list` — List the tunnels available in this workspace",
+		"• `/qurl list` — List the tunnels available to you",
 		"• `/qurl setup` — Connect qURL to your Slack workspace and become its qURL admin (workspace admin only)",
 	)
 	if h.aliasStore != nil && h.cfg.AdminStore != nil {
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -760,16 +760,14 @@ func (h *Handler) helpMessage() string {
 		"*/qurl* — Create and manage qURLs from Slack",
 		"",
 		"*Commands:*",
-		"• `/qurl get <url>` — Get a qURL for a URL",
-		"• `/qurl get $name` — Get a qURL for a name your Slack admin has configured in this channel",
+		"• `/qurl get <url|$slug>` — Mint a one-time qURL for a URL or a `$slug` your Slack admin has configured in this channel",
 	}
 	if h.cfg.PostDM != nil {
-		lines = append(lines, "• `/qurl get <url|$name> dm:true` — DM the link to you instead of posting it in-channel")
+		lines = append(lines, "• `/qurl get <url|$slug> dm:true` — DM the link to you instead of posting it in-channel")
 	}
 	lines = append(lines,
-		"• `/qurl get <url|$name> once:true` — Get a single-use qURL; the link burns on first redemption",
-		"• `/qurl get <url|$name> reason:\"…\"` — Annotate the audit log with a reason",
-		"• `/qurl list` — Show your 5 most recent qURLs",
+		"• `/qurl get <url|$slug> reason:\"…\"` — Same, annotating the audit log with a reason",
+		"• `/qurl list` — List the tunnels available in this workspace",
 		"• `/qurl setup` — Connect qURL to your Slack workspace and become its qURL admin (workspace admin only)",
 	)
 	if h.aliasStore != nil && h.cfg.AdminStore != nil {
@@ -795,7 +793,7 @@ func (h *Handler) helpMessage() string {
 		// on "shortcut" even though the admin verbs retain their
 		// historical set-alias/unset-alias names.
 		lines = append(lines,
-			"• `/qurl set-alias $<shortcut> <url-or-resource-id-or-$slug>` — Configure a qURL shortcut in this channel (admin only)",
+			"• `/qurl set-alias $<shortcut> <url-or-$slug>` — Configure a qURL shortcut in this channel (admin only)",
 			"• `/qurl unset-alias $<shortcut>` — Remove a qURL shortcut in this channel (admin only)",
 			"• `/qurl aliases` — List the qURL shortcuts configured in this channel",
 		)
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -766,7 +766,7 @@ func (h *Handler) helpMessage() string {
 		lines = append(lines, "• `/qurl get <url|$slug> dm:true` — DM the link to you instead of posting it in-channel")
 	}
 	lines = append(lines,
-		"• `/qurl get <url|$slug> reason:\"…\"` — Same, annotating the audit log with a reason",
+		"• `/qurl get <url|$slug> reason:\"…\"` — Mint a one-time qURL, recording a reason in the audit log",
 		"• `/qurl list` — List the tunnels available to you",
 		"• `/qurl setup` — Connect qURL to your Slack workspace and become its qURL admin (workspace admin only)",
 	)
--- a/apps/slack/internal/handler_alias.go
+++ b/apps/slack/internal/handler_alias.go
@@ -389,6 +389,11 @@ func (h *Handler) resolveTunnelSlugAliasTarget(ctx context.Context, teamID, slug
 	}
 	for i := range page.Resources {
 		resource := &page.Resources[i]
+		// Defense-in-depth: the server's `?slug=` filter is single-
+		// purpose, but re-assert type/slug/active here so an upstream
+		// regression can't leak a non-tunnel, wrong-slug, or revoked
+		// resource into mintable state (this resolves the resource_id
+		// that both /qurl set-alias and /qurl get then mint against).
 		if resource.Type == client.ResourceTypeTunnel && resource.Slug == slug && resource.Status == client.StatusActive {
 			return resource.ResourceID, nil
 		}
--- a/apps/slack/internal/handler_aliases.go
+++ b/apps/slack/internal/handler_aliases.go
@@ -86,7 +86,7 @@ func (h *Handler) processAliases(ctx context.Context, log *slog.Logger, values u
 		return
 	}
 	if len(entries) == 0 {
-		_ = h.postResponse(log, responseURL, ":mag: No aliases are configured for this channel yet. Run `/qurl set-alias $<alias> <url-or-resource-id-or-$slug>` to add one.")
+		_ = h.postResponse(log, responseURL, ":mag: No aliases are configured for this channel yet. Run `/qurl set-alias $<alias> <url-or-$slug>` to add one.")
 		return
 	}
 
@@ -126,7 +126,7 @@ loop:
 			// Fill un-dispatched tail with id-only fallbacks.
 			for j := i; j < len(entries); j++ {
 				e := &entries[j]
-				lines[j] = formatAliasLine(e.Alias, "", e.ResourceID)
+				lines[j] = formatAliasLine(e.Alias, "", "", e.ResourceID)
 			}
 			break loop
 		}
@@ -137,16 +137,21 @@ loop:
 			e := &entries[idx]
 			alias := e.Alias
 			target := ""
+			slug := ""
 			if e.ResourceID != "" && e.Alias != "" {
 				// GetResourceByAlias is the customer-facing path that
-				// returns Alias + TargetURL. The Store row may carry
-				// alias=="" (legacy shape) — fall back to id-only in
-				// that case rather than swallowing a 404.
+				// returns the full Resource (Alias, TargetURL, Slug, …).
+				// The Store row may carry alias=="" (legacy shape) — fall
+				// back to id-only in that case rather than swallowing a
+				// 404. Tunnel resources have no TargetURL but carry a
+				// Slug, which [formatAliasLine] surfaces in place of the
+				// resource_id.
 				if r, rerr := c.GetResourceByAlias(ctx, e.Alias); rerr == nil {
 					if r.Alias != "" {
 						alias = r.Alias
 					}
 					target = r.TargetURL
+					slug = r.Slug
 				} else if errors.Is(rerr, context.Canceled) || errors.Is(rerr, context.DeadlineExceeded) {
 					// Distinct log from the 404/5xx branch so operators
 					// can tell "request was cut short by SIGTERM /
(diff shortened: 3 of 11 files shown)


Kevin-layerV:
Thanks for the review — addressed in 0d16588:

1. **Curly-quote in godoc** (`handler_list.go:284`) — fixed. Root cause was gofmt's own doc-comment formatter converting `` `` `` (two backticks) into U+201C (TeX-style quoting), then oscillating on it. Rewrote the example lines to plain single-backtick code spans so gofmt stays idempotent.
2. **Empty-state copy in URL-heavy / zero-tunnel workspaces** — acknowledged. Same root cause as #5 (no server-side type filter); tracked in #531. The common case (≤100 total resources) is unaffected.
3. **`/qurl list` help "in this workspace"** — fixed → "List the tunnels available to you" (covers the non-admin channel-scoped case).
4. **`$slug` vs `$<shortcut>` vocab** — clarified the get help to `… or a \`$slug\` (tunnel or shortcut) configured in this channel`, mirroring the usage hint. Kept the `<url|$slug>` command form (explicit product ask).
5. **Page-size 10→100 / `has_more` semantics** — this is the leaky abstraction (`page.HasMore` is a master-list signal, not "more tunnels"). Filed #531 to push a `type=tunnel` filter into the resources API + client, which makes `has_more` accurate and drops the over-fetch entirely. Linked it from the `listResourcesScanLimit` godoc.
6. **Test name lags fixture** — renamed `…WithoutAlias` → `…WithoutAliasBinding` (the slug is a resource attribute; the omitted thing is the channel `alias_binding`).

**Breaking-change note** — `once:true` → "unknown flag" is called out in the PR body and commit message. Not tagging it `BREAKING CHANGE` for Release Please: the slash command still works and this is a beta-bot UX change, not a library API break — a major bump would overstate it.

One deliberate skip: collapsing the local `resourceTypeTunnel` const onto `client.ResourceTypeTunnel`. It's pre-existing (not introduced here) and used by 24 test-fixture lines, so collapsing it would add ~25 lines of unrelated churn to a focused PR. Left for whenever that const is next touched (same call the /simplify altitude pass reached).

Kevin-layerV:
Round-2 review addressed in 0385eea:

- **Medium — `has_more` drift in the non-admin pagination-gap branch**: added a `TODO(#531)` right at the branch documenting that `page.HasMore` is a master-list signal (over-claims with zero tunnels), and a pinning test `TestHandleList_NonAdminPaginationGapZeroTunnels` (URL-only page → gap fires today). I deliberately did *not* add an in-PR heuristic: any guard I could write without the server-side filter is wrong in some case (e.g. gating on "tunnels seen on page 1" mislabels a workspace whose tunnels all sit past row 100). #531 (server-side `type=tunnel`) is the correct fix and flips this test's expectation in lockstep.
- **Minor — `once:true` breaking change**: kept the strict reject (the product direction is to remove the flag), but it now returns a transitional hint — `` `once` is no longer needed — every `/qurl get` link is one-time use by default `` — instead of the generic "unknown flag". Covers `once:false` too. Regression test updated.
- **Minor — non-admin footer "others"**: → "additional allow-listed tunnels may sit past it."
- **Nit — `tunnelToken` called twice/row**: skipped per your "skip if you'd rather not churn" — O(n log n) of three string-nil checks at n≤100 is immeasurable, and a Schwartzian transform is more code than it's worth here.
- **Nit — slug-wins-over-alias**: confirmed intended (it's the stable handle `/qurl tunnel install <slug>` binds, and tunnels are always created with a slug). Added a direct table-driven `TestTunnelToken` pinning the slug→alias→resource_id precedence (test-gap #2).
- **Test-gap #1**: covered by `TestHandleList_NonAdminPaginationGapZeroTunnels` above.

Build/test/gofmt/golangci-lint (0 new issues vs `main`) all green.

Kevin-layerV:
Round-3 addressed in 6e6e24e:

- **Nit #1 (the one ask) — `reason` help line "Same,"**: rewritten to stand alone → `Mint a one-time qURL, recording a reason in the audit log`.
- **Nit #2 — URL shortcuts dropping out of `/qurl list`**: this is intended (the whole point of "tunnels only"), but you're right it deserves to be explicit — added a note to the PR description that `set-alias`-bound URL shortcuts now surface via `/qurl aliases` (and `/qurl get $<shortcut>`), not `/qurl list`.
- **Nit #3 (empty-state only points at tunnel install)**: leaving as-is — the empty-state is deliberately tunnel-scoped, and `/qurl help` lists `/qurl aliases` for the shortcut path.
- **Nit #4 (Slack renders the error backticks as inline code)**: confirmed desirable; noted for future error-copy tweaks.

CI green, `golangci-lint` 0 new issues vs `main`.

Kevin-layerV:
Round-4 review addressed:

- **🟡 List-vs-get divergence on `allowed_resource_ids`-only / admin-sees-all tunnels** (the one to fix before merge) — **fixed** in 20990e8 via your option #2. `resolveTokenForGet` (was `resolveAliasForGet`) now falls back to a tunnel-slug lookup when the channel `alias_bindings` lookup misses, and gates the resolved resource_id through the **same** allow-set as the `$r_<id>` form (extracted into `resourceAllowedForUser`: admin bypass / `allowed_resource_ids`). So the list→get round-trip the footer advertises now holds for tunnels installed in another channel or granted cross-channel, without re-exposing `r_<id>` in the list. The not-allowed rejection is worded with the slug the user typed, not the resolved id. Added round-trip tests: allowed-set (non-admin), admin-bypass, and not-allowed (asserts no `r_<id>` leak + no mint). `TestHandleGet_AliasNotFound` got a `GET /v1/resources` mock since the miss path now does a slug lookup.
- **🟢 Duplicate `resourceTypeTunnel` const** — **done** in 5682c04. Collapsed onto `client.ResourceTypeTunnel` (matches `handler_alias`/`handler_tunnel`); `filterTunnelResources` is the sole consumer now and the test file already imports `client`. (Reconsidered my earlier defer now that the cost is just a mechanical swap.)
- **🟢 Description mrkdwn rendering** — pre-existing (not a regression) and ephemeral-only (no broadcast), so out of scope here; filed #532 to escape server-text in Slack payloads as defense-in-depth.
- **🟢 `listResourcesScanLimit` / `has_more` semantics** — tracked in #531 (server-side `type=tunnel` filter), with the `TODO(#531)` + pinning test already in place.
- **🟢 Help-text consistency across surfaces** — agreed, no change.

CI green, `golangci-lint` 0 new issues vs `main`.

Kevin-layerV:
Round-5 review addressed (plus a new feature commit — `/qurl aliases` now shows alias → `$<slug>` for tunnel-backed aliases, by request):

- **(1) Enumeration oracle on the slug fallback** (the one to fix) — **fixed** in 419b658. The not-allowed-slug branch now returns the **same** "not configured for this channel" copy as slug-not-found, so a non-admin can't distinguish "slug exists elsewhere but not allowed here" from "no such slug". Preserves the mirrored-copy posture the `$r_<id>` form documents; the real reason is logged for operators. Updated `TestHandleGet_DollarSlugNotAllowedNonAdmin` to assert the anti-enumeration copy + no "not allowed" leak.
- **Nit — slog key `alias` on a `token` variable** — fixed → `token`.
- **Test gap — binding-wins-over-slug precedence** — added `TestHandleGet_DollarTokenBindingWinsOverSlug`: a live `alias_binding` is authoritative and the slug fallback is never consulted (a same-named shadow tunnel's mint is asserted un-hit).
- **(2) `has_more` over-claim** — tracked in #531; `TODO(#531)` + the zero-tunnel pin are in place. Agreed it's non-blocking; the copy already hedges with "may".
- **(3) extra round-trip on alias miss / scan-limit amplification / unconditional `(one-time use)` suffix** — noted as forward-looking; no change (the round-trip is inherent to the honesty fix, the scan limit is documented, and the suffix is intentional now that one-time is the default).

`go build`, full `go test ./apps/slack/...`, `gofmt`, `golangci-lint` (0 new vs `main`) green.

Kevin-layerV:
Round-6 addressed in fbd1340:

- **#2 (gap-copy over-claim)** — implemented your tighter heuristic: the non-admin pagination-gap copy now fires only when `len(tunnels) > 0` (the scanned page had tunnels that channel scoping filtered out). A URL-heavy / zero-tunnel page falls through to the plain empty-state. Inverted `TestHandleList_NonAdminPaginationGapZeroTunnels` to pin the suppression; the tunnels-present gap case stays covered by `TestHandleList_NonAdminPaginationGap`. (#531 still closes the residual >scan-limit-with-tunnels-past-page-1 ambiguity.)
- **#1 (extra round-trip)** — added a "don't short-circuit this" cost note to `resolveTokenForGet`'s godoc.
- **#6 (defense-in-depth assertion)** — added the one-liner on `resolveTunnelSlugAliasTarget`'s type/slug/active re-check.
- **#4 (description separator `→` vs `—`)** — keeping `→`. The product direction (and the original request) spelled the row as `$kktest -> Slack tunnel install for kktest`, i.e. an arrow, so I'm matching that intent. Noted your indirection-vs-annotation point; it's captured for a wording pass.
- **#5 (`resolveTunnelSlugAliasTarget` name)** — agreed it's slightly off in the get context; leaving as-is per your "not worth a rename now", will fold into a future touch.
- **Corrupt-binding test gap (`found=true, resourceID=""`)** — impossible by construction today (LookupChannelAlias only returns found for a non-empty mapped value); skipping, but noted.

CI green; `golangci-lint` 0 new issues vs `main`.

Kevin-layerV:
Round-7 addressed in 44b5527 (all were non-blocking; took the cheap wins):

- **#1 (empty-state pointed non-admins at an admin-only command)** — fixed. Branched on `isAdmin`: admins get "Set one up with `/qurl tunnel install <slug>`", non-admins get "Ask a Slack admin to set one up with …".
- **#3 (non-deterministic sort on token collision)** — added a `ResourceID` tiebreaker so two rows sharing a token order deterministically, making the two-call-idempotency claim in the sort comment actually hold.
- **#5 (bare-alias branch)** — commented as the defensive escape hatch (unreachable for a real `PolicyEntry`, since `fanoutAliasRows` only reaches the formatter when `ResourceID != ""`).
- **#2 (slug-fallback hop runs unmetered by the mint rate-limit)** — fair flag. The in-bot rate-limit gate (`CheckRateLimit`) is a known stub today (tracked alongside the `/qurl get` rate-limit TODO in the rollout doc); when it lands it should cover the lookup hop too. Confirming the upstream `GET /v1/resources` quota is decoupled from mint quota is an ops follow-up, not a code change here.
- **#4 (`r_<id>` last-resort token vs the "$slug" footer)** — noted; it's the documented legacy slug-less-tunnel exception in `tunnelToken`'s godoc, and `$r_<id>` still mints (routes to the resource-id form). Leaving the footer as-is since that edge is vanishingly rare.

CI green; `golangci-lint` 0 new vs `main`. (The intermittent `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` is a known 5s-timeout flake, unrelated.)

Kevin-layerV:
Round-8 addressed (all non-blocking per your review):

- **#2 (tiebreaker untested)** — added `TestHandleList_SortTiebreakerOnTokenCollision` in 5c60230: two tunnels yielding the same `$dup` token (one via slug, one via a slug-less resource-level alias) ordered deterministically by resource_id, directly exercising the tiebreaker branch.
- **#1 (slug-fallback fires before the rate-limit gate — fresh probe surface)** — agreed this is a real delta the PR introduces. Filed #534 (medium) with both options you listed (move `CheckRateLimit` ahead of the form-switch, or skip the fallback for empty-allow-set non-admins), tied to the existing `/qurl get` rate-limit work. Not fixing inline since the gate's placement/semantics is its own decision and the in-bot limiter is still early-stage.
- **#3 (bare `once:` → generic "empty value" not the transitional hint)** — leaving as-is; the empty-value gate runs first and `once:` isn't a realistic deliberate input.
- **#4 (empty `channel_id` + token-form → "Could not reach qURL")** — pre-existing for the alias form, inherited by the unified token form (no info leak). Worth distinguishing "needs a channel" from "service down" eventually, but out of scope here.

CI green; `golangci-lint` 0 new vs `main`.

Kevin-layerV:
Thanks — taking "ship-ready" as the convergence point. Round-9 items are all handled without further code churn:

- **#1 (slug fallback before the rate-limit gate)** — tracked in #534 (with both options you and the prior round suggested). Genuinely a follow-up: the gate's placement is its own decision and the in-bot limiter is still early-stage.
- **#2 (rejection copy points to `/qurl aliases`; `/qurl list` is better for a slug)** — good catch; noted for the wording pass. Any change here has to stay uniform across the not-found / not-allowed branches to keep the anti-enumeration collapse intact.
- **#3 (100-row over-fetch on every call)** — `TODO(#531)`; the server-side `type=tunnel` filter removes both the over-fetch and the `has_more` ambiguity.
- **#4 (double `authenticatedClient` on the slug-fallback success path)** — only on the rare resolve-and-allowed path; negligible if `AuthProvider.APIKey` caches. Noting, not refactoring the shared helper's signature for it.
- **Nits (bare `once:`, long TODO block)** — acknowledged as-is.

Holding code here: CI is fully green and the four merge-readiness checks are met. Remaining work is the two filed follow-ups (#531, #534) plus a copy pass. Ready for human review/approval.

Issue #460, opened 2026-05-19
slack: `/qurl list` alias may not resolve via `/qurl get $alias` in caller's channel


Kevin-layerV:
## Summary

Post-#459, `/qurl list` renders every workspace `Resource.Alias` regardless of channel-policy state. But `/qurl get $<alias>` continues to resolve through `channel_policies.alias_bindings` per-channel (`handler_get.go:332` `LookupChannelAlias`). So a user can see `$prod-db` in `/qurl list` from channel C1, run `/qurl get $prod-db` from C1, and hit `noResourceForAliasMessage` if `prod-db` isn't bound in C1 — confusing because they just saw the alias rendered.

The `$r_<id>` form still works (subject to the allow-set check) and surfaces a clearer `notAllowedInChannelMessage`, so the user has a recoverable path. This is not a security issue — capability gating still holds at mint time — but it's a UX rough edge introduced by widening list disclosure beyond what the per-channel alias surface supports.

Surfaced in cr review of #459: <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4490752208>.

## Repro

1. As admin in channel C_admin: `/qurl create https://prod.example.com` → resource gets `r_prod_aa`.
2. As admin in C_admin: `/qurl setalias $r_prod_aa prod-db` → binds `prod-db` in C_admin's `channel_policies` row.
3. As any workspace member in C_other (no `channel_policies` row): `/qurl list` → sees `\`$prod-db\` → https://prod.example.com`.
4. From C_other: `/qurl get $prod-db` → hits `noResourceForAliasMessage` ("No resource bound to alias `prod-db` in this channel").

## Scope

Pick one (not both):

- **(a)** Annotate `/qurl list` rows with the channels where the alias resolves: `\`$prod-db\` → https://prod.example.com (alias resolves in: #channel-a, #channel-b)`. Pros: closes the loop without re-narrowing list disclosure. Cons: requires fanning out `alias_bindings` per row, which can be expensive on workspaces with many channel_policies rows.
- **(b)** Document the asymmetry in `/qurl help` and the empty-state copy on `noResourceForAliasMessage`. Pros: zero infra cost. Cons: doesn't fix the surprise, just labels it.

## Acceptance criteria

[...]

PR #536, opened 2026-05-28
feat(slack): tunnels-only /qurl get & set-alias (drop URL forms), list aliases, help reorg

Kevin-layerV:
## Summary

Makes the Slack bot's `/qurl get` and `/qurl set-alias` verbs tunnels-only (dropping both the raw-URL and the `$r_<id>` resource-id forms), surfaces bound aliases in `/qurl list`, and reorganizes `/qurl help` into user vs admin sections.

The bot now only ever mints one-time qURLs for **tunnel resources**, reached by a tunnel `$slug` or a channel `$alias` an admin configured. Arbitrary-URL and raw-resource-id minting through Slack is gone.

## What changed

**`/qurl list` (already on branch):** header renamed to *Protected Tunnel Resources*; each tunnel now shows its bound channel aliases. A slug-less tunnel promotes a bound `$alias` to its primary token, or — if it has none — renders a non-mintable `(no slug — ask your Slack admin to set one)` row; tokenless rows sort after the legible ones so usable tunnels lead.

**`/qurl get` — tunnels only:** the positional arg must be a tunnel `$slug` or a channel `$alias` — both the raw-URL and the `$r_<id>` resource-id forms are gone. `parseGet` rejects `http(s)://` args with `ErrURLNotSupportedGet` and `$r_<id>` pastes with `ErrResourceIDNotSupportedGet`; both surface via the `:warning:` ack as a friendly redirect to `/qurl list` / the `$slug`. `getWork`'s URL branch is removed; an empty-`Alias` guard refuses to mint defensively. `get` now gates on `AdminStore` in `/qurl help` (token resolution fails closed without it).

**`/qurl set-alias` — tunnels only:** the only accepted target is a tunnel `$slug` (`/qurl set-alias $<alias> $<slug>`). A well-formed URL or `r_<id>` target parses but is rejected synchronously with:
> `` `/qurl set-alias` points an alias at a tunnel slug — `/qurl set-alias $<alias> $<slug>`. (Raw URLs and resource IDs aren't supported targets.) ``

The slug→resource_id resolution is the admin act that authorizes the resource in the channel. The success reply echoes the typed slug ("Alias `$staging` now points to tunnel `$kktest` in this channel"), not the opaque resolved resource_id. The now-unreachable sync-bind arm of `handleSetAlias` is removed (the slug path is inherently async — it lists resources before the write). Usage/invalid-target copy no longer advertises URLs or resource ids. A surviving pre-tunnels-only URL binding is refused at `/qurl get` time with an actionable re-bind hint (`legacyAliasBindingMessage`).

**`/qurl help` — reorg:** split into `*Commands:*` (get, list, aliases, help) and `*Admin commands:*` (setup, tunnel install, set-alias, unset-alias, admin add/remove/list/revoke). Each line only shows when its verb is wired (PostDM, aliasStore, AdminStore, OpenView); `get` and `aliases` both gate on `AdminStore` because each fails closed without it, while `set-alias`/`unset-alias` gate on `aliasStore` (the asymmetry is documented in `helpMessage`). Shapes updated to the tunnels-only forms.

**Copy/docs:** the `/qurl create` deprecation hint now redirects to `/qurl get <$slug|$alias>` (it previously pointed at the removed `get <url>`); the apps/slack README Features list is brought in line.

## Tests

- **`process_test.go`** — the async-infra tests (ack latency, idempotency-key derivation, pool saturation, 5xx correlation handle, orphan-ack-on-shutdown, wait-timeout, panic recovery) switch from the removed `get <url>` vehicle to a `get $<alias>` token form. A new `seedGetAliasBinding` helper wires a fakeDDB-backed `AdminStore` with one channel alias→resource_id binding so `LookupChannelAlias` resolves and the mint lands at `POST /v1/resources/{id}/qurls`. Each test's original assertion is preserved.
- **`handler_get_test.go`** — `TestHandleGet_URLRejected` and `TestHandleGet_ResourceIDRejected` fence the raw-URL and `$r_<id>` redirect copy (mint never reached); `TestHandleGet_LegacyURLBindingRefused` fences the read-side legacy-binding guard; `TestGetWork_EmptyAliasRefusesToMint` drives the defensive refuse-to-mint guard directly.
- **`handler_alias_test.go`** — set-alias URL/resource-id cases convert to rejection assertions (store untouched); happy-path / multi-alias / duplicate-refusal / cross-tenancy / no-store / missing-id cases convert to `$slug` targets via a slug-resolving upstream + the async invoker. Happy-path slug tests assert the slug appears in the reply and the resource_id does not. `TestUnsetAlias_SyncTimeoutExceeded` re-fences the `aliasSyncTimeout` budget on the surviving sync (unset) path; `TestHelpGatesAliasVerbsByBackingStore` fences the help gate asymmetry.
- **`handler_list_test.go`** — `TestFormatTunnelListLine`, `TestChannelAliasesByResourceID`, and `TestHandleList_SlugLessTunnelsSortByPromotedAlias` fence the bound-alias display, the slug-less promotion, and the sort/render agreement (the test arranges resource_ids so a resource_id-keyed sort would invert the assertion).
- **`parser_test.go`** — `FuzzParse` learns the new `ErrURLNotSupportedGet` and `ErrResourceIDNotSupportedGet` sentinels; the `$r_<id>` parser rows assert the resource-id redirect.

`go build ./...`, `go test ./apps/slack/...`, `gofmt -l`, and `golangci-lint run --new-from-rev=origin/main ./apps/slack/internal/` are all clean.


Kevin-layerV:
Thanks for the thorough pass — addressed all 7 in `496298f`:

1. **getWork rate-limit gate** — refreshed the comment to the "defensive only" tone (both reachable arms fail closed on nil `AdminStore`, so the gate is unreachable on real minting paths; kept as a future-form backstop rather than calling `CheckRateLimit` unconditionally).
2. **handleGet usage guard** — labeled as defensive-only with the same rationale.
3. **Wasted `context.WithTimeout`** — split `aliasPreamble` into `aliasValidate` (team/channel + store-wired check) and the timeout wrapper. set-alias now calls `aliasValidate` (no timer it never reads); unset-alias keeps `aliasPreamble`.
4. **Unset-alias sync-timeout fence** — added `TestUnsetAlias_SyncTimeoutExceeded` + a `blockUnbind` hook on the fake store, restoring budget coverage on the surviving sync path.
5. **`parseAliasArgs` doc** — added a note that the tunnels-only POLICY lives in `handleSetAlias`, not the parser (which still parses all three shapes so the handler can reject the unsupported two with a specific message).
6. **Placeholder consistency** — unset-alias is now `$<alias>` in help + README to match set-alias.
7. **`ErrMissingTarget` doc** — dropped the stale `create` mention (it's used by setalias + admin revoke).

The pre-existing unset-alias sync-timeout gap you flagged in (4) is now closed rather than deferred. Unrelated note: `handler_aliases_test.go:65` has a pre-existing curly-quote (U+201C) in a godoc comment from #233 — out of scope for this PR, not touching it here.

Kevin-layerV:
Round-2 addressed in `c1d4dbe`:

1. **Success copy leaked `r_<id>`** — fixed. set-alias now renders the typed slug: "Alias `$staging` now points to tunnel `$kktest` in this channel." Happy-path slug tests now assert the slug appears and the resource_id does *not* (the binding still stores the resource_id).
2. **`bindAliasTarget` misnamed + dead URL redaction** — folded `bindAliasTarget` into `resolveAndBindTunnelSlugAlias` (its only caller). The target is always a server-minted `r_<id>`, so `redactURLForLog` (+ `TestRedactURLForLog`) is removed and the audit log records `(alias, slug, resource_id)` directly.
3. **Double `GetChannelPolicy` read on non-admin `/qurl list`** — added a `TODO(perf)` on `channelAliasesByResourceID` noting `AllowedResourceIDsForChannel` already read the same `(team, channel)` row, and that folding the alias map out of that first read would halve per-list DDB reads.
4. **Resolution-failure log missing context** — added `team_id`/`channel_id`/`slug` to the `resolveAndBindTunnelSlugAlias` failure log.
5. **`getURLCommandTest*` constants misnamed** — renamed to `getTokenCommandTest{ChannelID,UserID}`.
6. **PR description nit** — moved the set-alias test note under a dedicated `handler_alias_test.go` bullet.

All four merge-ready checks remain green: `go build`/`go test ./apps/slack/...`/`gofmt -l`/`golangci-lint --new-from-rev=origin/main` clean.

Kevin-layerV:
Round-3 addressed in `ca3ee2e`:

**Pattern/lint — `ErrURLNotSupportedGet` as a sentinel:** by design, declining to change. Every parser sentinel in `parser.go` follows the same contract — its `.Error()` text *is* the user-facing copy, surfaced via `respondSlack(w, ":warning:"+err.Error())`, and `FuzzParse` asserts each parser error `errors.Is` one of the documented sentinels. A `*userError` (handler-layer type) can't satisfy the parser's sentinel set, and splitting into "generic sentinel + separate copy constant" would make this one sentinel inconsistent with the dozen others. The ST1005-fragility concern is real but applies equally to every sentinel here; the existing convention already accepts it.

**Perf — second `channel_policies` read:** deferred to #537 (labeled `enhancement`/`go`/`priority: low`/`pr-polish-followup`). It needs a new `slackdata.Store` method returning `(allow-set, alias_bindings)` from one `GetItem`, which is out of scope for this tunnels-only PR. The `TODO(perf)` stays inline.

**TOCTOU note:** added — `resolveAndBindTunnelSlugAlias` now documents that a tunnel deleted between resolve and bind lands a stale binding, caught at mint time.

**`logAliasBound` count drift:** fixed (now "Every logged field …").

**`$r_<id>` undocumented in usage:** the `handleGet` usage warning now mentions the `$r_<id>` copy-from-list form alongside `$slug`/`$alias`. (The help message still omits it intentionally — it's a power-user copy-from-`/qurl list` path, not a primary verb.)

**Test gaps:**
- Added `TestFormatTunnelListLine` fencing slug-only, slug+description-no-aliases, and the self-binding-slug exclusion independently of the combined end-to-end tests.
- async `setalias_slug` timeout bound: the ctx is `runAsync`'s `asyncWorkTimeout`, already fenced generically by `process_test.go`'s `TestHandler_WaitTimeout` (a non-ctx-honoring upstream → bounded drain). A setalias-specific duplicate would re-test the same `runAsync` machinery, so leaving it to the generic fence.

All four merge-ready checks green: `go build`/`go test ./apps/slack/...`/`gofmt -l`/`golangci-lint --new-from-rev=origin/main`.

Kevin-layerV:
Round-4 addressed in `c890238`. On reflection items 2 and 3 were worth doing rather than re-defending — thanks for pushing.

**1. Rate-limit nil-check comment overstated protection:** rephrased honestly. It's belt-and-suspenders — if a future minting arm regressed its own AdminStore guard, this skips the rate-limit (qurl-service's per-key quota stays the backstop) rather than nil-panicking. It is explicitly *not* a forward fence that refuses such an arm.

**2. `ErrURLNotSupportedGet` smuggled prose into a sentinel:** fixed. The sentinel is now terse (`"raw URL not supported by get"`) like its siblings; `handleGet` owns the rich fix-naming copy (`urlNotSupportedGetMessage`, mapped via `errors.Is`). Matches the `parseAliasArgs` convention you cited.

**3. Dead URL/`r_<id>` parser accept-branches + divergent copy:** collapsed. `parseAliasArgs` now rejects any non-`$` target with the single `msgAliasTargetNotTunnel` (so `r_<typo>` and a valid `r_<id>` read identically), then validates `$<slug>` against the tunnel-slug grammar. Removed the orphaned `resourceIDPrefix`/`schemeHTTP(S)` consts and `handleSetAlias`'s now-redundant `$`-prefix check. The parser test is rewritten to the tunnels-only contract and asserts which rejection copy fires.

**Perf follow-up (second `channel_policies` read):** filed as #537 (`enhancement`/`go`/`priority: low`/`pr-polish-followup`); `TODO(perf)` stays inline.

Also fixed a stray stale hint you didn't flag: `/qurl aliases` empty-state copy still said `<url-or-$slug>` — now `$<slug>`.

All four merge-ready checks green locally (`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`). Note: the prior CI run's `Test` failure was the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` (5.0s kubectl timeout); this push re-runs it.

Kevin-layerV:
Round-5 addressed in `d6fcf3e`:

**1. `$r_<id>` hint pointed at `/qurl list`, which never shows resource ids** — dropped the `$r_<id>` mention from the get usage hint; it now reads `/qurl get $<slug>` (or `$<alias>`). Added a comment noting the `$r_<id>` form stays parser-accepted but is deliberately undiscoverable (a power-user copy form, and `/qurl list` renders slugs only).

**2. "URLs and resource IDs aren't supported" fired on plain garbage too** — reworded rather than re-splitting the parser (re-splitting would re-add the URL/`r_` accept-branches round-4 removed). `msgAliasTargetNotTunnel` now leads with the tunnel-slug form (`points an alias at a tunnel slug — … $<slug>`) and mentions URLs/resource-ids only parenthetically, so a sigil-less typo isn't told it typed a URL. Also fixed the doc comments that wrongly claimed non-`$` garbage routes to `msgAliasTargetInvalid` — it lands on the uniform not-a-tunnel copy; `msgAliasTargetInvalid` fires only once a `$`-prefixed target fails the slug grammar.

**3. `Command.Target` / `SubcmdSetAlias` / `parseSetAlias` doc comments still described the URL/resource-id forms** — each now notes the parser is intentionally non-strict and the tunnels-only policy is enforced downstream in `handleSetAlias`/`parseAliasArgs`. (Also corrected `Command.Target`'s stale `create` reference to `admin revoke`.)

**4. Parser tests read as end-to-end URL/resource-id happy paths** — added a group comment above the three `setalias` rows fencing "Parse grammar only" from "wired end-to-end," pointing at `TestParseAliasArgs_SetAlias` where the policy is actually tested. Kept the terse row names rather than renaming: renaming made those lines `--new-from-rev`-new and tripped `goconst` on the repeated `$prod-dashboard` / `https://internal.example.com` literals (clean at the PR tip, only the rename surfaced them). The group comment carries the clarification instead — which is the alternative you offered.

**5. `getWork` default-arm reply misleading** — added a distinct `unexpectedGetShapeMessage` ("Couldn't process that command. Please contact your Slack admin for assistance.") so if the unreachable arm ever fires, the user's report correlates to the `log.Error` line instead of looping on "please try again."

**6. Second `channel_policies` read on `/qurl list`** — tracked in #537 (`enhancement` / `go` / `priority: low` / `pr-polish-followup`); the `TODO(perf)` stays inline. Out of scope for this tunnels-only PR — the fix needs a new store method returning the allow-set + `alias_bindings` from one `GetItem`.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit run` are all clean.

Kevin-layerV:
Round-6 addressed in `297b270`.

**1. Legacy URL/`r_<id>` bindings silently mint-failing** — fixed at the read side, the layer you suggested. `resolveTokenForGet` now checks the resolved binding value: anything that isn't a tunnel resource id (doesn't start with `r_`) is refused before the mint with an actionable reply — "`$shortcut` points at a target that's no longer supported. Please ask your Slack admin to re-point it at a tunnel with `/qurl set-alias $shortcut $<slug>`." — instead of the generic "please try again". A legacy `r_<id>` binding is a real resource and still mints fine; only the URL shape (the only structurally-dead case) is caught. New `TestHandleGet_LegacyURLBindingRefused` fences it (mint never reached).

I went with the read-side guard rather than a data migration: there's no migration plan on file, I can't audit prod `channel_policies` from here, and the guard makes the read path fail-safe whether or not any legacy URL rows actually exist. Note on the check — I used `strings.HasPrefix(resourceID, "r_")`, not the parser's `resourceIDPattern`: a *stored* resource id is whatever qurl-service issued and isn't guaranteed to match the 11-char get-token shape (the test fixture `r_prod_db` is one such id), so the strict regex would have rejected valid bindings.

**2. `*Admin commands:*` header always renders on a sandbox** — keeping as-is, by design. `/qurl setup` is gated on `oauthSetup`, not `AdminStore` (it mints an OAuth state token; it never touches AdminStore), and it's the bootstrap verb — a brand-new workspace with no admin yet must be able to discover how to *become* one. The verbs that do need `aliasStore`/`AdminStore` (tunnel install, set-alias, admin add/remove/…) are already gated, so the section degrades to just `setup`, which is exactly the one verb that still works there.

**3. `/qurl help` placement (between user reads and the admin section)** — intentional. The reorg groups by audience: `help` is a user verb, so it closes the user `*Commands:*` block. "help last overall" was the pre-reorg single-list ordering; under the two-section split, "help last within user commands" is the natural home.

**Smaller items:**
- **TOCTOU cross-link** — `resolveAndBindTunnelSlugAlias`'s note now names the re-validation site: the bound resource_id flows to the qurl-service mint call, which rejects a deleted resource via `mapMintError`.
- **`channelAliasesByResourceID` test gap** — added `TestChannelAliasesByResourceID` covering the best-effort arms directly: nil `AdminStore`, empty-channel short-circuit, policy-fetch error → nil, and grouping + sort. The double-read perf `TODO(perf)` is tracked in #537.
- **Parser rows reading as supported forms** — moved the three `setalias` grammar rows out of `TestParse_HappyPaths` into a dedicated `TestParse_SetAliasGrammar` so the "supported commands" table no longer lists URL/`r_<id>` shapes. (Reused existing `testAliasURL`/`testTunnelSlug` consts so the move stayed under goconst.)
- **`TestSetAlias_MissingTeamOrChannelID` noop pin** — fixed the stale `aliasPreamble` reference (set-alias uses `aliasValidate`) and spelled out the ordering fence: it's a slug target, so a guard regressed below the slug-resolve dial would flip the reply to "Tunnel slug not found" and fail the Contains check.
- **`aliasValidate` ctx bound** — doc now names it: the async path runs on `runAsync`'s ctx, bounded by `asyncWorkTimeout` (25s), so a wedged upstream can't hang the resolve+bind.
- **`urlNotSupportedGetMessage`** — now also points at `/qurl aliases` for users with channel shortcuts but not slug fluency.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit run` are all clean.

Kevin-layerV:
Thanks — round-7 addressed in `f713f39`. All six were polish/follow-ups; one turned up a real (latent) gating bug.

**1. Junk `r_<typo>` rows from the old permissive parser** — added the doc note you suggested. A `r_<typo>` row passes the `r_` prefix check, proceeds to mint, and 404s → the generic retry copy. Accepted as a rare residual: an admin re-bind is the same fix as for a URL row, and it's not worth an upstream pre-resolve on every binding hit just to special-case it.

**2. `channelAliasesByResourceID` double-read** — tracked in #537.

**3. `/qurl aliases` help placement** — the reread turned up a real mismatch, now fixed. The help line gated on `aliasStore != nil`, but `processAliases` reads `channel_policies` through the **`AdminStore`** (not the `aliasStore` that set-alias/unset-alias write through) and fails closed when it's nil. So on a deploy with `aliasStore` wired but `AdminStore` nil, help advertised `/qurl aliases` even though it would reply "not configured." Gated it on `AdminStore != nil` to match the runtime, and locked both directions: `TestHelpListsNewVerbs` (aliasStore wired, AdminStore nil → hidden) and `TestHelpListsGuidedAndTypedTunnelInstall` (AdminStore wired → shown). set-alias/unset-alias correctly stay gated on `aliasStore` (their write path).

**4. Double upstream call on duplicate-alias refusal** — filed as #541 (`enhancement`/`go`/`priority: low`/`pr-polish-followup`). Low-priority: admin verb, only on the duplicate (error) path.

**5. `TestHandleGet_LegacyURLBindingRefused` matcher** — narrowed from the blanket `addCustomerPrefix("POST", "/v1/")` to `/v1/resources/` (the family a regressed guard would mint against).

**6. `getWork` default arm test** — added `TestGetWork_DefaultArmRefusesUnknownShape`: constructs the otherwise-unreachable "neither alias nor resource-id" command and asserts the arm returns `unexpectedGetShapeMessage` rather than falling through to a mint.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit run` are all clean.

Kevin-layerV:
Thanks — reads as converged. Dispositions on round-8:

**1. Extra DDB read per `/qurl list`** — ticket filed: #537 (`enhancement`/`go`/`priority: low`/`pr-polish-followup`). The `TODO(perf)` stays inline pointing at it.

**2. `r_` guard permissive by design** — agreed; the doc comment calls out the junk-`r_<typo>` residual. An offline bindings-table scan is the right place to bulk-flag those if we ever do one.

**3. `*Admin commands:*` header always emitted** — intentional, as you note: `/qurl setup` is the bootstrap verb (gated on `oauthSetup`, not a store), so a fresh workspace with nothing wired still needs to discover how to become an admin. Leaving the header so that one verb isn't orphaned.

**4. TOCTOU** — acknowledged; mint-time rejection via `mapMintError` is the fallback, documented on `resolveAndBindTunnelSlugAlias`. No change.

**6. `urlNotSupportedGetMessage` points at AdminStore-gated `/qurl aliases`** — keeping as-is. The message already points at `/qurl list` (same AdminStore dependency), so the `/qurl aliases` hint is consistent rather than newly-misleading; both are the right discovery commands on a configured deploy. Conditionally rendering them on store-presence isn't worth the handler machinery for the rare fully-unconfigured sandbox, where the raw-URL-rejection path is itself a degenerate edge. Not blocking, as you flagged.

The `Test` check hit the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` (5s kubectl dry-run timeout, "signal: killed") — unrelated to this diff; re-running the job. No code change for this round.

Kevin-layerV:
Merged `main` (`c5fd38f`) — it now carries #459 (revert of the `/qurl list` admin gate + channel-policy filter), which overlapped this PR's `/qurl list` work. Resolution:

- **`handler_list.go`** — #459's simplified, unscoped `processListResources` no longer extracts `channelID` (it went with the removed admin gate). Re-added it: the tunnel **listing** is workspace-wide now, but the per-row alias annotations stay **channel-scoped** — `channelAliasesByResourceID` reads *this* channel's `alias_bindings` so each row shows the shortcuts that resolve here. No new disclosure (other channels' aliases aren't shown); empty/DM channels degrade to slug-only rows.
- **Stale `TODO(perf)` rewritten** — it described a *double* `channel_policies` read (allow-set read + alias-display read). #459 removed the allow-set read (`scopeResourcesForUser`), so it's now a single cosmetic read per list with nothing to fold into. **#537 closed as obsolete.**
- **`handler_list_test.go`** — kept this PR's additive tests (`ShowsBoundAliases`, `FormatTunnelListLine`, `ChannelAliasesByResourceID`); dropped `NonAdminFiltersToChannelPolicy` (one of the 9 non-admin tests #459 removed); kept #459's `UnscopedAcrossChannels`. Re-added the `ddbtypes` import the auto-merge dropped.
- **`handler_get.go` / `handler_get_test.go`** — auto-merged cleanly: this PR's legacy-URL guard + default-arm test coexist with #459's `notAllowedInChannelMessage` breadcrumb change and its `#460` TODOs.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit run` are clean on the merged tree.

Kevin-layerV:
Thanks — addressed the one fixable nit in `5b255c6`, rest are dispositioned below.

**3. Formatting drift (`<$slug|$alias>` vs `$<slug>`)** — fixed. The help lines, the `handleGet` doc, and the `create`-deprecation hint all use `<$slug|$alias>`; the defensive usage hint was the lone outlier (`$<slug>` (or `$<alias>`)). Canonicalized it to `<$slug|$alias>` so the get-command form is uniform. (The `$<slug>` in `legacyAliasBindingMessage` stays — it's the *set-alias* form `$<alias> $<slug>`, which matches set-alias copy.)

**1. `r_` prefix check is loose** — agreed it's inert, and slightly more so than noted: a stored legacy URL always carries its scheme (`set-alias` ran `url.Parse` requiring http/https), so the bound value is `http(s)://…` and never starts with `r_` — the guard fires correctly on it. The only values that slip through are junk `r_<typo>` rows, which the doc comment already calls out.

**5. `parseSetAlias` parser-vs-policy split** — the function's doc comment already states it: *"At the parser layer Target may be a URL, raw resource_id, or `$slug` shape; the tunnels-only policy (only `$slug` is bound) is enforced downstream in parseAliasArgs / handleSetAlias."* There's no inline URL/scheme check left inside `parseSetAlias` to annotate — that branch was collapsed in an earlier round; the parser is now a non-strict extractor.

**2. `*Admin commands:*` header always renders** — intentional: `/qurl setup` is the bootstrap verb (gated on `oauthSetup`, not a store), so a fresh workspace with nothing wired still needs to discover how to become an admin.

**4. `/qurl help` placement** — intentional: the reorg groups by audience, so `help` (a user verb) closes the user `*Commands:*` block rather than trailing the whole message.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit run` are clean.

Kevin-layerV:
Thanks — addressed the actionable items in `04e2953`. Dispositions on all 11 observations:

**Fixed**
- **#1 + #7 (slug-less render + coverage):** `formatTunnelListLine` now promotes the first (lexically sorted) bound channel `$alias` to the primary token for a slug-less, resource-alias-less tunnel — so the row shows a usable `$<alias>` instead of `` `r_xxx` (no slug set) `` sitting next to an `(also `$alias`)` that already advertises a working token. Added the two missing `TestFormatTunnelListLine` rows (slug-less with bound aliases → promotes; slug-less with none → bare resource_id).
- **#2 (help gate asymmetry):** Documented it as intentional rather than changing the gate. `set-alias` is admin-gated at the Slack manifest level and writes through `aliasStore` only — it never touches `AdminStore` at runtime. So gating it on `AdminStore` (the "joint gate" option) would *hide a working verb* on an aliasStore-only deploy. Each verb is advertised exactly when its own backing store is wired; the aliasStore-wired/AdminStore-nil shape isn't real (both come from the same `QURL_*_TABLE` env vars). Added a comment in `helpMessage` saying so.
- **#3 (aliasPreamble doc):** Added a cross-reference noting `set-alias` resolves its slug upstream and runs on `runAsync`'s ctx, calling `aliasValidate` directly — it never needs the sync timer `aliasPreamble` layers on.
- **#8 (test name):** Renamed `TestHelpListsNewVerbs` → `TestHelpGatesAliasVerbsByBackingStore` (the verbs aren't new, and the test fences per-verb store gating, including the #2 asymmetry). Updated the cross-reference in `handler_tunnel_test.go`.

**Considered, no change**
- **#4 (legacy-binding `r_` prefix guard):** Confirmed the only writer to `alias_bindings` post-this-PR is `resolveAndBindTunnelSlugAlias`, which writes a server-resolved `r_<id>` resource_id. Nothing writes a non-URL, non-`r_` value today, so the prefix guard is sufficient; the residual junk-`r_<typo>` row 404s at mint, as the docstring notes. A tighter `resourceIDPattern` check stays rejected for the stated reason (stored id shape may differ from the get-token regex).
- **#5 (slug-less rows sort first):** Accepted — the empty-token tie falls through to the existing `ResourceID` secondary sort (`TestHandleList_SortTiebreakerOnTokenCollision` fences it). Slug-less tunnels are vanishingly rare.
- **#11 (URL-bound bindings in prod DDB):** Deployment-coordination item, out of scope for this PR. The read-side guard already degrades raw-URL legacy bindings to `legacyAliasBindingMessage` gracefully; whether any exist in prod is an ops check tracked separately.
- **#6 / #9 / #10:** Noted, thanks — no action (positive observations on coverage, brand spelling, and title).

`go build ./...`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint run --new-from-rev=origin/main`, and `pre-commit run` are all clean.

Kevin-layerV:
Thanks — round-11 addressed in `4b26043` (rebased onto `5d5d7db`).

**Fixed**
- **#1 (doc rot, 4 spots):** Removed the deleted `[resourceIDPattern]` doclink in the `handler_get.go` legacy-binding guard; retargeted the `seedPolicySet` / `seedChannelPolicySet` godoc to the live consumer (`resolveTokenForGet` → `resourceAllowedForUser` → `AllowedResourceIDsForChannel`); renamed `TestGetWork_DefaultArmRefusesUnknownShape` → `TestGetWork_EmptyAliasRefusesToMint` and dropped the stale `Resource.Kind != ResourceTokenResourceID` comment (`getWork` is a single empty-alias guard now, not a switch default).
- **#2 (sort/render asymmetry):** Good catch — this was a gap I introduced with the alias-promotion in the prior round. Extracted `tunnelDisplayToken` (slug → resource-alias → promoted channel-alias) and routed BOTH the `/qurl list` sort key and `formatTunnelListLine` through it, building `aliasMap` before the sort so the sort sees the same string the row renders. New e2e fence `TestHandleList_SlugLessTunnelsSortByPromotedAlias`: two slug-less tunnels whose promoted aliases sort opposite to their resource_ids, so it passes only if the sort follows the displayed alias.
- **#4 (`legacyAliasBindingMessage` charset):** Documented the precondition — `alias` must be a parser-validated token (it's interpolated into the reply and a `set-alias` hint), and noted every caller comes through the validated get path today.

**Deferred (issue filed)**
- **#3 (flag legacy bindings in `/qurl aliases`):** Filed https://github.com/layervai/qurl-integrations/issues/547 (labeled `enhancement` / `pr-polish-followup` / `priority: low`) with repro/scope/acceptance. It's admin-discoverability only — the read-side guard already prevents a dead binding from minting, so not a correctness gap.

Also extracted a `testTunnelAliasDash` const (goconst fired on the `"dash"` literal once the count hit 3 after rebasing onto the new get-by-slug test).

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit` are all clean.

Kevin-layerV:
Thanks — round-12 in `8304d3c`.

**Fixed**
- **#1 (slug-less row looks copyable but isn't):** The merge-worthy one. A tunnel with no slug, no resource alias, and no channel binding now renders `` `r_xxx` (no slug — ask your Slack admin to set one) `` instead of `(no slug set)`. It has no `$`-token `/qurl get` can accept, so the copy now says that plainly and routes to an admin — the discovery surface stays honest. Tests updated to the new marker.
- **#2 (channel-scoped annotations on a workspace-wide list):** The deeper asymmetry (list is workspace-wide, mintability + alias display are channel-scoped) is already tracked by `TODO(#460)` (referenced in `handler_get.go` and `handler_list.go`). For the interim UX surprise you flagged, the footer now notes the `(also …)` aliases shown are channel-specific, so their absence doesn't read as "no aliases anywhere."
- **#3 (sort recomputes token twice per compare):** Precompute each row's `tunnelDisplayToken` once into a resource_id-keyed map; the comparator now does map lookups.
- **README nit:** "raw URLs are not supported" → "aren't supported" to match the in-app copy.

**Considered, no change**
- **#4 (`legacyAliasBindingMessage` precondition doc-only):** You called the doc note acceptable for a one-caller function — agreed; leaving as documented (round 11). The typed-`parsedAlias` option is over-engineering for one caller.
- **#5 (two defensive `cmd.Alias == ""` checks):** Intentional layered defense at different boundaries with deliberately different copy — `handleGet` shows a usage hint (recoverable user error shape), `getWork` returns the internal "contact admin" copy (a shape the parser guarantees is unreachable, so it correlates to the `log.Error`). Both unreachable today; collapsing them would lose that signal distinction.
- **#6 (backtick-before-prefix rejection ordering):** Both paths reject correctly; the differing copy for a backtick-bearing URL vs a clean URL is cosmetic and the backtick path is pathological. Not worth reordering the grammar checks.
- **Security note + test-gap (a) — unset on a legacy URL binding:** Already fenced. `TestUnsetAlias_Happy` seeds `staging → https://example.com` (a raw-URL value) and asserts the "no longer bound" reply + the row is cleared. `UnbindChannelAlias` keys on the alias name, never the value, so the stored URL is irrelevant to removal.
- **Test-gap (b) — cross-channel slug-less-with-alias list rendering:** That's the `TODO(#460)` behavior change; the footer note covers the interim UX. Fencing it now would pin behavior we intend to change there.
- **`urlNotSupportedGetMessage` long literal:** Left as-is — it's a single user-facing sentence; splitting it would just fragment the copy.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit` are all clean.

Kevin-layerV:
Thanks — round-13 in `c49d23d`.

**Fixed**
- **#1 (help advertises `/qurl get` when it can't work):** Confirmed — `resolveTokenForGet` fails closed with `errAdminStoreNotConfigured` when `AdminStore` is nil, and the URL form that once let `get` work without DDB is gone. So an unguarded `get` violated `helpMessage`'s own stated rule. `get` (and its `dm:`/`reason:` variants) now gate on `AdminStore != nil`, matching `/qurl aliases`. This also makes onboarding coherent: a no-DDB deploy shows `list` / `help` / `setup`, and `get` appears once the store is wired. Fenced both directions — `TestHelpGatesAliasVerbsByBackingStore` asserts `get` hidden without `AdminStore`, `TestHelpListsGuidedAndTypedTunnelInstall` asserts it shown with — and retargeted `TestSlashCommand_EmptyBodyShowsHelp` to the unconditional `/qurl help` marker.
- **#3 (`legacyAliasBindingMessage` precondition was doc-only):** Made it structural. `aliasCharsetPattern.MatchString(alias)` is now re-asserted at the message boundary; a non-charset-clean value (e.g. backtick-bearing) falls back to token-free copy, so a future caller can't reopen the Slack-fence-escape surface. No longer convention-dependent.
- **#4 (`tunnelToken` godoc):** Step 3 now says it stops at `""`; the channel-alias promotion and bare-`resource_id` fallback live in `tunnelDisplayToken` / `formatTunnelListLine`, not here — so a reader won't grep for `r.ResourceID` in `tunnelToken`.

**Considered, no change**
- **#2 (backtick guard fires before the `$`-prefix check):** Intentional and documented (`handler_alias.go:154-160`) — the backtick/control-byte rejection is a hygiene-first gate run before any further parsing. A backtick-bearing target is *malformed*, so `msgAliasTargetInvalid` is the more accurate copy than `msgAliasTargetNotTunnel` (which implies "well-formed, wrong type"). The common non-tunnel inputs you'd want uniform — clean URLs, `r_<id>`, sigil-less typos — all already get the single `msgAliasTargetNotTunnel`. Only a pathological control-byte/backtick target diverges, correctly.
- **#5 (stale `r_<id>` binding gets "please try again" vs URL gets a re-bind hint):** Acknowledged migration residual (documented in the guard comment). The broader "proactively flag broken bindings" improvement — including stale-but-valid-shape `r_<id>` rows — is tracked by #547; folding the `r_<id>` 404 case into the same surface is the natural place for it.
- **Positive-side `r_` gate test:** Already fenced. `TestHandleGet_HappyPath` seeds `prod-db → r_prod_db` (via `seedChannelPolicySet`, which binds `alias_bindings[alias] = resourceIDs[0]`) and mints — i.e. a valid `r_<id>` binding value passing the `r_` prefix accept path. The negative is `TestHandleGet_LegacyURLBindingRefused`. Both sides of the asymmetry are locked.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit` are all clean.

Kevin-layerV:
Thanks — round-14 in `f131324`.

**Fixed**
- **#1 (stale `$r_<id>` paste UX — the near-blocking one):** Mirrored the raw-URL redirect. New `ErrResourceIDNotSupportedGet` sentinel + a `$r_` prefix branch in `parseGet` (before the generic charset check) so a pasted `$r_<id>` now gets *"`/qurl get` takes a tunnel `$slug` or a channel `$alias`, not a resource ID. Run `/qurl list` and copy the `$slug` instead."* rather than the opaque `ErrInvalidAlias`. A real alias can never start with `r_` (no `_` in the charset), so it never shadows a valid token. Wired into `FuzzParse`'s known sentinels, converted the four `$r_` `TestParse` rows, and added `TestHandleGet_ResourceIDRejected`.
- **#4 (slug-less rows cluster at the front):** Tokenless rows now sort to the END, so legible `$slug`/`$alias` tunnels lead the list instead of a wall of opaque `r_<id>`s. Flipped `TestHandleList_StableSortByToken` to assert the new order.
- **Nit (`urlNotSupportedGetMessage` breadcrumb points at AdminStore-gated `/qurl aliases`):** Now points only at `/qurl list` — it's ungated and (this PR) renders each tunnel's channel `$alias` shortcuts inline, so it's the single surface that always works and shows both slugs and aliases. Drops the gated breadcrumb and the redundancy at once.

**Considered, no change**
- **#2 (`slog.Warn` when the charset fallback fires):** `legacyAliasBindingMessage` is a pure formatter (called from tests without a logger); threading one through just to warn on a branch that's unreachable today (single validated caller) would trade that purity for a log line. The refusal itself is already logged at the call site (`handler_get.go` Warn "channel alias bound to a non-resource-id … refusing to mint"), so an operator already sees the event; the charset fallback is belt-and-suspenders beneath that.
- **#3 (residual `r_<typo>` rows get generic retry copy):** Acknowledged migration residual, documented in the guard. Both of your suggested directions — a typed `mapMintError` 404 hint and an `alias_bindings` scan script — fit the "proactively surface broken bindings" work already tracked in #547; I've left a note there.
- **#5 (two defensive empty-`Alias` guards):** Intentional layered defense (sync ack vs async mint, distinct copy), `getWork`'s arm independently tested. Keeping both.
- **Nits — section break for 2-3 lines / `tunnelToken` one-liner:** Both sections always carry ≥1 item (no empty-section render), and `tunnelToken`'s doc already points downstream to `tunnelDisplayToken`. Leaving as-is.

Also filed the `r_<typo>` follow-up note on #547. `go build`, `go test ./apps/slack/...` (incl. a 10s `FuzzParse` burst, 1.7M execs clean), `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit` are all clean.

Kevin-layerV:
Thanks — reads as essentially converged. Round-15 in `8ed5a1f`.

**Fixed**
- **#3 (`tunnelDisplayToken` "first lexical alias wins" surprise):** Added a godoc note — a slug-less tunnel's displayed identity can flip if an admin later binds an alphabetically-earlier alias (`$zebra` → `$aardvark`). Rare (installs come with a slug), now documented.

**Considered, no change**
- **#1 (legacy URL bindings linger without a proactive surface):** Tracked by #547 (which I expanded with your `/qurl aliases` marker and boot-time-counter ideas). Fail-soft refusal at `/qurl get` is correct as-is.
- **#2 (`r_<typo>` junk rows 404 at mint):** Documented residual; the `resourceIDPattern` shape-check angle and the migration-scan/typed-404-hint ideas are folded into #547. Not worth a per-binding pre-resolve.
- **#4 (tag `unexpectedGetShapeMessage` with an operator code like `[get-shape]`):** Skipping deliberately — this repo keeps user-facing copy jargon- and operator-code-free (a recurring convention here). The correlation already exists operator-side: `getWork` logs `log.Error("get: empty alias token reached getWork — refusing to mint", "raw", …)`, which is the unique line an escalation maps to. Adding a code to the user string would trade the convention for redundancy.
- **#5 / parseGet godoc removal-history / `SubcmdSetAlias` asymmetry / one DDB read:** All acknowledged-and-accepted in your own notes; leaving as-is (the removal-history prose is useful while the migration is fresh; CHANGELOG move is a fine post-stabilization follow-up).

Separately, I ran `/simplify` (reuse / simplification / efficiency / altitude passes) over the full diff: reuse, efficiency, and altitude came back clean. The simplification pass surfaced three candidates, all skipped with cause — the `aliasValidate`/`aliasPreamble` split is intentional (two distinct callers; collapsing re-introduces the unused-timer coupling an earlier round removed), `bindAliasTarget` is already inlined, and the two empty-`Alias` guards are the deliberate layered defense from the thread above.

`go build`, `go test ./apps/slack/...`, `gofmt -l`, `golangci-lint --new-from-rev=origin/main`, and `pre-commit` are all clean.

Kevin-layerV:
Thanks — agreed this is converged. No code changes this round; dispositions below.

**Verified (your observation #2 — TOCTOU mint copy):** Checked `mapMintError`. A 404 on the resource-scoped mint (deleted-upstream tunnel, or a stale binding) lands in the `default` arm → `commonGetMintFailedMessage` (generic, but correctly *not* "retry forever", and it logs loud). So you're right that it isn't an actionable "re-run `set-alias`" hint today. I left it as a follow-up rather than fixing in-PR because the right fix is path-aware: `mapMintError` is shared by *every* get mint (slug-fallback and alias-binding), so a blanket 404→"re-point with `/qurl set-alias`" message would mislead a direct-`$slug` get whose resource was deleted (no alias involved). Doing it correctly means threading the resolution path into the error mapping — folded into #547 (the typed-404-hint idea is already noted there).

**Confirmed (your #4 — sandbox help lists only `list`+`help`):** Explicit product call, not a side-effect. `helpMessage`'s stated rule is to omit any verb whose only reply would be ":warning: not configured", and post-tunnels-only `get` can't work without `AdminStore`. On a no-DDB sandbox the honest set is `list` / `help` / `setup` — `setup` being the path that gets you configured. Fenced by `TestHelpGatesAliasVerbsByBackingStore` (hidden) and `TestHelpListsGuidedAndTypedTunnelInstall` (shown when wired).

**Noted, no change (#1, #3, #5):** All flagged as review-history notes / already-documented — the per-`list` `GetChannelPolicy` cost (acceptable for the discovery surface; batch-fold if traffic grows), the `tunnelDisplayToken` lexical-alias flip (now in the godoc), and the tokenless-rows-sort-last reordering (intended, friendlier, fenced).

**Nit — second `/qurl aliases` breadcrumb:** Leaving the get-rejection copy pointing only at `/qurl list`. This is the same breadcrumb an earlier review round asked me to *narrow to* `/qurl list` — `/qurl aliases` gates on `AdminStore` (so it replies "not configured" on the very sandbox where a stray paste could happen), and `/qurl list` now renders each tunnel's channel `$alias` shortcuts inline, so it's the single always-working surface that shows both slugs and aliases. Adding `/qurl aliases` back would reintroduce that gated-breadcrumb concern.

**Other nits:** `legacyAliasBindingMessage`'s charset re-check stays (you noted the godoc justifies it); PR scope `feat(slack)` confirmed correct.

CI is green and `/simplify` came back clean (reuse/efficiency/altitude clean; three simplification candidates all skipped with cause). Ready for human review.

PR #538, opened 2026-05-29
feat(slack): split admin verbs into a /qurl-admin slash command

Kevin-layerV:
## What

Introduce a separate **`/qurl-admin`** Slack slash command and move the admin verbs to it. `/qurl` keeps the user commands — **including `setup`**, which is not an admin verb (see "Why setup stays on `/qurl`" below).

- **`/qurl` (user):** `setup`, `get`, `list`, `aliases`, `help`.
- **`/qurl-admin` (admin):** `tunnel install` (guided + typed), `set-alias $<alias> $<slug>`, `unset-alias $<alias>`, `admin add @user` / `remove @user` / `list` / `revoke <qurl_id>`, `help`.

## Security (the headline change, not just UX)

This PR adds an **in-code admin gate** (`requireAdminStoreSync` + `requireAdminSync`) to `handleSetAlias` / `handleUnsetAlias` — closing a real authorization gap, not just splitting the surface. Previously these verbs relied on the Slack app-config "admins only" restriction, but **Slack does not enforce a slash command's admin-only setting at invocation time** (it's a picker hint). So before this PR, any workspace member who knew the verb could mutate a channel's alias bindings. Now a non-admin is denied **before** any tunnel resolve or store write (`TestSetAlias_NonAdminDenied` / `TestUnsetAlias_NonAdminDenied` pin the gate-before-resolve ordering). The same in-code gate already protected `tunnel` and the `admin` membership verbs; this extends it to the alias verbs. Operators on older deploys (where set-alias/unset-alias weren't yet on `/qurl-admin`) should be aware this tightens behavior.

## How

Both commands POST to the same `/slack/commands` endpoint with the same HMAC signature verification — Slack stamps which command was invoked in the `command` form field. `handleSlashCommand` now dispatches on that field **before** the subcommand switch:

- `/qurl-admin <verb>` → admin handlers.
- `/qurl <verb>` → user handlers (`setup`, `get`, `list`, `aliases`, `help`).
- A verb typed on the wrong command gets a friendly redirect ("`tunnel` is an admin command. Use `/qurl-admin tunnel`…" / "`get` belongs on `/qurl`. Use `/qurl get`…") instead of a bare "unknown subcommand". `setup` typed on `/qurl-admin` redirects to `/qurl setup`.
- An unrecognized `command` value falls back to the user surface (defensive; the user surface never mutates admin state).

`helpMessage` is split into `userHelpMessage` (`/qurl help`, now including the `setup` line) and `adminHelpMessage` (`/qurl-admin help`), each preserving the existing conditional gating on `PostDM` / `aliasStore` / `AdminStore` / `OpenView` so help never advertises a verb that would only reply ":warning: not configured". `/qurl help` ends with a pointer line routing admins to `/qurl-admin help`. `adminHelpMessage` carries an always-present `/qurl-admin help` anchor line (mirroring the user surface) so a no-store sandbox deploy still renders a verb.

The verb handlers and the parser are unchanged — they still see the same `text` (e.g. `admin revoke q_…`, `tunnel install <slug>`, `set-alias $a $b`); only the command they arrive under changed.

### Why setup stays on `/qurl`

qURL setup is **first-come-claims**: on an unbound workspace, the first user to complete setup becomes its owner. The admin verbs on `/qurl-admin` are admin-gated in code (`requireAdminSync`), so putting `setup` there would lock out the very first claimant — who is by definition not yet an admin of anything. So `setup` stays on the open `/qurl` command.

This does **not** weaken the overwrite guard. A `setup` run against an *already-bound* workspace is refused at the OAuth callback's `BindWorkspace` pre-flight (`oauth.checkBindAllowed`), which runs **before** any key is minted and rejects a binder who isn't the existing owner. So the open surface can mint a setup link for anyone, but only the first claimant (fresh workspace) or the existing owner (rebind) ever rotates a credential. The protection is structural in the bind layer, not a manifest gate.

### Why the `admin` sub-word is kept

The whole `/qurl-admin` command is admin-scoped, so `admin add/remove/list` reads as stuttering. I kept it anyway because **`list` collides**: `/qurl list` means "list this channel's tunnels" (user), and dropping the sub-word would make `/qurl-admin list` a *second, different* "list" (the admin roster). Keeping `/qurl-admin admin list` distinct from `/qurl list` avoids two same-named surfaces and keeps the membership grammar coherent.

### User-facing copy

Every reply / usage string / error page that names an admin verb is updated to the `/qurl-admin` form: alias usage + not-a-tunnel copy, tunnel-install usage and the guided-setup retry messages, the modal error copy, the Slack-install success page, the `/qurl list` empty-state hint, the cmd operator log reminder, and both READMEs. `setup`-flow copy (the slash reply, the OAuth `/start` + callback browser error pages, the rebind-refused page, the not-connected nudge) stays on `/qurl setup`. User-verb references (`/qurl get`, `/qurl list`) keep their `/qurl` names. qURL brand spelling preserved.

## ⚠️ Deploy prerequisite (NOT code)

`/qurl-admin` must be **registered as a slash command in the Slack app config**, pointing at the **same request URL** as `/qurl`. **The admin verbs in this PR are inert until that registration exists**: until then, Slack never sends a `command=/qurl-admin` payload, so the admin verbs are unreachable.

**Deploy ordering — register the manifest first (or simultaneously) with this code deploy.** In the gap between code deploy and manifest registration, an admin typing the *old* invocation (`/qurl set-alias …`, `/qurl tunnel install`) now gets the wrong-surface redirect to `/qurl-admin` — a command that doesn't exist yet — so the admin verbs are effectively unavailable until the manifest lands. The user surface (`/qurl get`/`list`/`aliases`/`setup`) stays fully functional throughout, and `setup` needs no new registration. The tests assume both commands are registered, so they don't fence this window — it's a deploy-runbook checklist item, not a code change.

Admin **enforcement is in code** (`requireAdminSync` against the qURL admin set), so the AdminStore must be wired. The "admins only" restriction on the `/qurl-admin` registration is a cosmetic Slack-picker hint — **not** the enforcement boundary (Slack does not gate slash-command invocation on workspace-admin role). And do **not** restrict `/qurl` itself to admins: `setup` is intentionally open (first-come-claims), and an admin-only `/qurl` would lock out the first claimant. `setup` works without any new registration.

## Tests / checks

- `go build ./...`, full `go test ./apps/slack/...` + `./shared/...`, `gofmt -l` clean, `golangci-lint run` (v2.10.1, CI's pinned version) — 0 issues on the full tree, pre-commit clean.
- Migrated the existing admin/user tests: the shared slash invoker routes the `command` field by verb (user verbs — now including `setup` — → `/qurl`, everything else → `/qurl-admin`), and the two setup tests post on `/qurl`. Content assertions that pinned `/qurl <admin-verb>` literals were updated to `/qurl-admin <admin-verb>`.
- `TestDispatchSplit_*`: per-command help content (user help advertises `setup`; admin help does not), both wrong-surface redirects (`setup` on `/qurl-admin` → `/qurl`), and the unknown-command-defaults-to-user fallback.

## Stacking

Originally stacked on `feat/slack-tunnels-only-get-list-aliases` (the v2 PR #536); now rebased onto and targeting `main` (#536 merged). `main` was re-merged in to resolve #550's alias-terminology / self-heal-setup-owner conflicts.


Kevin-layerV:
Addressed cr round-2 in 5ca7b1d:

- **🐛 Stale `/qurl set-alias` (handler_get.go `legacyAliasBindingMessage`)** — fixed; now points at `/qurl-admin set-alias`. The other two refs (lines 109/429 here) are historical doc-comment context and left as-is; the doc comment at line 397 already says `/qurl-admin set-alias`.
- **Missing `TestUnsetAlias_NonAdminDenied`** — added, symmetric to `TestSetAlias_NonAdminDenied` (pre-binds an alias, asserts a non-admin is denied before the unbind and the binding survives).
- **firstWord echoes user text** — the redirect-site comment was already corrected in the previous round (round-1, 82de873) to say the full `text` is echoed and is safe only because it's an ephemeral plain-text Slack reply, not because it's a known literal.

Deferred to #546 (non-blocking polish, per your own framing):

- **`ReplaceAll` brittleness → sentinel placeholders** ("worth considering if more copy lands here").
- **Threading `command` to the standalone messages** (`workspaceNotSetupMessage`, the `/qurl create` deprecation reply, the `Try /qurl help` fallback) for non-prod command-name consistency ("Defer-able").
- **Unknown-command default rendering a fake `*-admin` pointer** (cosmetic; only reachable if Slack sends an unregistered command).

All three share the "non-prod env command-name awareness" theme and are tracked together in #546.

Kevin-layerV:
Thanks — round-3 disposition (also re-synced onto base round-13 in 6c3cb5d):

- **#1 ReplaceAll invariant has no test fence** — deferred to #546, which already scopes the sentinel-placeholder rewrite *plus* "a test fails if a raw `/qurl` substring is left in the authored lines." A weak fence now (counting `/qurl-sandbox-sandbox`) only catches a hardcoded sandbox literal; the meaningful fence comes with the sentinel refactor, so I've kept them together there.
- **#2 wrong-surface redirect echoes full `text` (backticks could break mrkdwn)** — added to #546 (same redirect/help copy-robustness theme). Cosmetic, user sees only their own reply, as you noted.
- **#3 `admin` double-word in redirect** — leaving as-is per your read; it's the deliberate `admin` sub-word retention.
- **#4 `create` in `userVerbs`** — considered, keeping it. `create` is the deprecated *user* mint verb, so classifying it as a user verb is correct: `/qurl create` hits the explicit deprecation reply, and `/qurl-admin create` redirects to `/qurl create` (which then shows the deprecation). Dropping it would make `/qurl-admin create` return a bare "unknown admin subcommand" with no hint that create was replaced by `get` — strictly less helpful for someone with create muscle memory. The two-step only affects the rare admin-surface typo.
- **#5 brand spelling** — ✅ no change.
- **Coverage note (empty-text → `adminHelpMessage` on literal `/qurl-admin`)** — noted; the empty-text and `help` branches share the same `case`, so it's covered transitively. Separately, base's round-13 get-gating is now fenced: `TestUserHelpGatesGetAndAliasesOnAdminStore` asserts both `/qurl get` and `/qurl aliases` gate on AdminStore in user help.

Note: this PR is stacked on #536, which is actively moving through its own cr rounds — I've re-synced it three times. It'll need a final rebase/retarget once #536 merges to `main`.

Kevin-layerV:
Round-4 disposition (in f1722d6, also re-synced onto base round-14):

- **#1 stale `adminHelpMessage` comment ("runtime dependency is the aliasStore alone")** — fixed. The comment no longer claims aliasStore is the sole dependency; it now says set-alias also needs AdminStore for `requireAdminSync`, but the two stores are wired in lockstep (same `QURL_*_TABLE` env vars), so gating the help line on aliasStore is equivalent to gating on both. Went with the comment fix rather than tightening the gate to `aliasStore && AdminStore` because the latter inverts the premise of `TestHelpListsAliasVerbsWhenAliasStoreWired` (which unit-tests the aliasStore-only gating path).
- **#2 `/qurl setup` now open to any member** — added an explicit operator warning to the boot-log line ("Do NOT restrict /qurl to admins: /qurl setup is intentionally open … an admin-only /qurl would lock out the first claimant"). Re: rebind — confirmed: `oauth.checkBindAllowed` rejects a binder who isn't the existing owner before any key is minted (see `handleSetup`'s doc comment), so the overwrite guard covers the same surface the manifest gate notionally did.
- **#3 firstWord/backtick redirect aesthetics** — tracked in #546 (redirect/help copy-robustness).
- **#4 empty-`command` defense in help fns** — keeping the centralized normalization at the HTTP entry (`handleSlashCommand`) rather than re-adding the per-function coercion. The /simplify pass deliberately collapsed that coercion from 4 sites to 1; each help fn documents the non-empty contract ("command is non-empty here — handleSlashCommand normalizes…"). Re-adding it would reintroduce the duplication. Optional, as you noted.
- **#5 switch verb-match order** — added the one-line comment: the admin/tunnel/alias sub-word matches must precede the `isUserVerb` fall-through (else `admin list` mis-routes via the `list` collision).

Heads-up: #536 is moving fast (round 11→14 during this review cycle); I've re-synced #538 four times. It'll need a final re-sync/retarget when #536 lands on `main`.

Kevin-layerV:
Round-5 disposition (in 8434e94, re-synced onto base 8ed5a1f):

- **#1 `isAdminCommand` broader than `/qurl*`** — fixed. Now `strings.HasPrefix(command, commandUser) && strings.HasSuffix(command, adminCommandSuffix)`, so a stray `/some-other-admin` falls to the user surface (safe default) instead of admin dispatch with a bogus `userCommandName`.
- **#2 redirect echoes full `text` (backtick breaks code span)** — fixed (rather than keep deferring). Added a `stripBackticks` helper applied to the echoed `text` in both wrong-surface redirects, so a stray backtick can't unbalance the inline-code span. Tightened the comment accordingly.
- **#3 verb-order comment overstates the invariant** — good catch, corrected. You're right that `slashVerb("admin list", "list")` is already false (exact-token or `verb ` prefix), so the ordering isn't load-bearing today; the comment now says it's defensive against a future colliding leading-token verb.
- **#4 ReplaceAll invariant not comprehensively test-fenced** — keeping in #546. The catch: the `/qurl`→command substitution is *reversible*, so output inspection can't distinguish an intended command-literal rewrite from an unintended prose rewrite (a `no bare "/qurl "` assertion always passes, since ReplaceAll rewrites every occurrence). Only the sentinel-placeholder approach makes intent explicit enough to fence — which is exactly what #546 scopes.
- **#5 verb-list / dispatch-case drift** — added a note on `adminVerbs` naming the three sync points (the list, a `dispatchAdminCommand` case, and `adminHelpMessage`).
- **#6 `admin admin` double-word** — leaving per your read (conflicts with the deliberate `admin list` ≠ `list` grammar).

(Still stacked on #536, which advanced to round-14 + doc commits during this cycle; five re-syncs so far. Final retarget pending #536 → `main`.)

Kevin-layerV:
Round-6 disposition (89df437):

- **#1 interleaved `firstWord`/`stripBackticks` doc comments** — fixed; reordered so each comment sits above its own function (stripBackticks had spliced between firstWord's doc and body).
- **#2 `isAdminCommand` prefix too loose** — tightened to `strings.HasPrefix(command, commandUser+"-")`, so `/qurlfoo-admin` and `/some-other-admin` no longer match.
- **#3 ReplaceAll invariant unfenced** — staying in #546. As noted last round, a `no leftover "/qurl "` assertion can't actually fence it: ReplaceAll rewrites every occurrence, so no bare `/qurl ` ever survives in non-prod help regardless of whether a prose `/qurl…` literal was wrongly rewritten — the substitution is reversible, so output inspection can't tell intended from unintended. The sentinel approach in #546 is what makes it fenceable.
- **#4 redirect grammar** — reworded "is a `/qurl` command" → "belongs on `/qurl`".
- **#5 empty-command normalization untested** — added `TestDispatchSplit_EmptyCommandDefaultsToUserHelp`.
- **#6 `firstWord` interpolated unescaped** — wrapped `firstWord(text)` in `stripBackticks` at both redirects, so the inline-code-fence safety is local rather than by chained invariant.

Kevin-layerV:
Round-7 disposition (9202619 + PR-body update):

- **#1 frame the security gate explicitly** — done. Added a **## Security** section at the top of the PR body as the headline, and corrected the body's stale claims that "the admin gate lives in the app config, not in code" (Deploy-prerequisite + "Why setup stays" sections) to reflect the in-code `requireAdminSync` enforcement. Commit messages already carry the `fix`-shaped framing.
- **#2 blind `ReplaceAll` guardrail** — keeping in #546. The two real options are (a) the Sprintf-per-line refactor or (b) a test. (b) can't actually fence it (ReplaceAll rewrites every occurrence, so no bare `/qurl ` ever survives — covered last round). (a) trades the current concise authored-literals form for ~17 `Sprintf` calls across the two help fns — more code for a non-blocking nit — so I've left the design decision to #546 with the invariant documented in-code.
- **#3 unknown-command cosmetic leak** — made the comment honest: an unrecognized command renders help echoing that name (`/qurl-bogus list`), we can't distinguish bogus from a valid non-prod env command without a registry, and it's unreachable since Slack only dispatches registered commands.
- **#4 `stripBackticks` on unknown-subcommand branches** — fixed; both `default:` branches now wrap `text`.
- **#5 setup-link DoS surface** — agreed it's pre-existing (setup was already effectively open); filed #548 to track a per-user/workspace setup-link throttle.
- Smaller observations (dup switches, firstWord micro-opt, slashCommandForVerb help routing) — acknowledged, no action (matches your read).

Kevin-layerV:
Round-8 disposition — no code changes this round; everything is tracked or a held judgment call:

- **#1 negative test for the `ReplaceAll` invariant** — this is the "one suggested follow-up", tracked in #546. As covered in prior rounds, the lightweight variants don't actually fence it (the substitution is reversible / single-pass, so no `qurl-admin-admin` or bare `/qurl ` ever survives regardless of an unintended prose rewrite); an "exactly N literals" count is brittle (every help-text edit updates N). The sentinel-placeholder refactor in #546 is what makes it genuinely fenceable.
- **#2 parser runs before `requireAdminSync` on set-alias/unset-alias** — keeping the current order. The security-relevant property holds: the gate runs before any slug resolve or DDB read/write (fenced by the `resolveHits == 0` assertions). The parser only sees the caller's own raw form text (no alias-state disclosure), so a non-admin getting a grammar error before the admin-only reply isn't a leak. Defensible, as you noted.
- **#3 asymmetric redirect wording** — keeping. "`tunnel` is an admin command" classifies (the verb needs admin); "`get` belongs on `/qurl`" relocates (a user verb on the wrong surface). The two situations differ, so the differing copy is intentional rather than an inconsistency to flatten.
- **#4 verb-list single-source-of-truth** — agreed it's fine for six verbs; the `adminVerbs` comment names the three sync points so a future addition is caught in review.
- **#5 `/qurl-bogus` fallback** and **#6 test rename** — resolved/positive, no action.

Substantively this is converged: conflicts resolved, in-code admin gate fenced, all eight review rounds addressed. Remaining items are tracked (#546 help-rewrite refactor, #548 setup-link throttle). Still stacked on #536 — final retarget to `main` pending its merge.

Kevin-layerV:
Round-9 disposition — all items are polish/tracked; no code change (the substantive review converged at round-8, and `/simplify` is clean after 6c58abc):

- **#1 static error/usage strings not command-aware in non-prod** & **#2 ReplaceAll help-rewrite test** — both folded into #546 (the non-prod command-name-awareness follow-up). Added the static-string surfaces to #546's scope.
- **#3 verb-list drift test** & **#6 `slashVerb` ASCII-whitespace note** — optional polish; tracked in #546.
- **#4 `create` in `userVerbs`** — keeping (the deprecated user verb; `/qurl-admin create` → `/qurl create` preserves the redirect-to-deprecation path). Comment-it-or-drop-it tracked in #546.
- **#5 `slog.Warn` on unrecognized command** — can't be done cleanly: we can't distinguish a bogus command from a valid non-prod env command (`/qurl-sandbox`) without a command registry, so a `Warn` would false-positive on every legit sandbox install. The existing `slog.Info("slash command", "command", …)` already captures the command for debugging.
- **#7 happy-path gate assertion** — the deny-side tests (`TestSetAlias_NonAdminDenied` / `TestUnsetAlias_NonAdminDenied`) already fence "gate exists + runs before resolve"; an admin-path assertion wouldn't trip if the gate were dropped (admin binds fine either way), so it adds little. Leaving as-is, as you noted it's optional.

Thanks for the nine thorough passes. The substantive change — the in-code admin gate on the alias verbs — is fenced; the rest is tracked in #546 / #548.

Kevin-layerV:
Pushed f0b9485. First, the **Lint failure** after the main-merge: two `//nolint:gosec` directives in `handler_alias.go` had gone unused (gosec no longer flags those `slog` lines under the merged-in toolchain), so `nolintlint` failed the build. Removed both. (The `Test` failure was the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` kubectl dry-run timeout, unrelated to this PR — re-running it.)

Also folded in the latest review's suggestions:

1. **ReplaceAll fragility** — added `TestHelpMessagesContainOnlyCommandTokens`: renders both help surfaces fully wired and asserts every `/qurl…` token is a command literal (`/qurl`/`/qurl-admin`), so a future `/qurl-docs`-style token fails the test instead of getting mangled in non-prod envs. Updated the MAINTAINER INVARIANT comment to cite it.
2. **stripBackticks the command in the unknown-subcommand reply** — left as-is, by design: `command` is the Slack-stamped command token (a trusted literal that can't contain backticks), and the existing convention is to hygiene-wrap only user-controlled text, not command literals — the redirect arms right above do the same (`adminCmd`/`userCmd` are raw too). Wrapping a trusted literal in a no-op sanitizer would be inconsistent in the other direction. The user-controlled echo path is now uniformly hardened via `echoText` (see #5).
3. **String-typed AdminAction labels** — added named constants `AdminActionSetAlias` / `AdminActionUnsetAlias` / `AdminActionTunnelInstall` (constified `tunnel_install` too so the group is uniform). These extend the `AdminAction` enum, which makes the two parser-action switches non-exhaustive — handled with a `//nolint:exhaustive` on each, because `requireAdminSync` legitimately takes both parser actions and these gate-only audit labels, so the shared type is correct and the switches genuinely cover only the parser-producible subset.
4. **isUserVerb(\"help\")==false** — added a note in `slashCommandForVerb` explaining help is intentionally not a user verb (dispatched explicitly per surface) and routes to `/qurl-admin` here, which is what the help tests want.
5. **Unbounded echoed text** — centralized echo hygiene in `echoText` (strip backticks + cap at 200 runes), applied to all four wrong-surface / unknown-subcommand echoes.
6. **Verb-slice immutability** — added read-only/do-not-mutate notes on `adminVerbs`/`userVerbs`.
7. **seedAliasAdminGate default wiring** — documented that `newAliasTestHandler` wires `AdminStore` unconditionally, so a test needing the nil path must build the handler directly.

Local `golangci-lint v2.10.1` (CI-pinned) is clean and the slack package tests pass.

Kevin-layerV:
Pushed 9290d55 addressing the review:

1. **Non-prod string drift** — documented near `adminCommandSuffix`: only the help text is rewritten to the invoked command; static error/usage/retry copy intentionally bakes in the prod `/qurl-admin` literal. Accepted (not threaded) because non-prod env-infix installs are operator-only internal sandboxes and customer installs are always prod — so the literal is always correct for the audience that sees it. Went with your leaned-toward option (document, don't thread).
2. **Test regex blind spot** — broadened to `` /qurl[^\s*\x60]* `` so a `/qurl_admin` / `/qurl.docs` token surfaces as its own non-allowed token (and fails) instead of matching a bare allowed `/qurl`. Stops at the formatting delimiters (space, `*`, backtick fence) so it doesn't trip on the `*/qurl*` header.
3. **isAdminCommand boundary** — documented the assumption that Slack only dispatches registered commands; an unregistered `/qurl-admin-extra` falls through to the user surface harmlessly.
4. **Command name not echoText-wrapped** — documented on `userCommandName`/`adminCommandName` that these are Slack-stamped command literals (not user input), so redirects interpolate them raw while only user-typed text goes through `echoText`. Kept raw rather than no-op-wrapping a trusted literal; the trust assumption is now explicit and cross-referenced from `isAdminCommand`.
5. Kept the `seedAliasAdminGate` comment over a `newAliasTestHandlerNoAdminStore` variant (only two tests bypass the helper — less churn than a second constructor). Agreed on the `AdminAction` overload note: the `//nolint:exhaustive` + comment capture it; would split into two types if the gate-label set grows.

Kevin-layerV:
Merged `main` (b04dfdb) — #550 landed there (alias terminology + clearer slug/alias copy + self-heal legacy `/qurl setup` owner rows) and conflicted with this branch. Resolution kept this PR's command split and folded in #550's improvements:

- **Help text**: kept the `userHelpMessage`/`adminHelpMessage` split with per-env command rewrite; adopted #550's `$slug`/`$alias` glossary line, clearer setup-owner copy, "aliases" (not "shortcuts"), and "List who connected qURL (the owner)". Admin verbs keep `/qurl-admin` routing.
- **handleSetup**: took #550's self-heal owner-recovery (shape-bad `owner_id` falls through to reclaim the legacy row; a valid non-owner is refused and returns), with the refusal copy pointing at the `/qurl-admin` commands.
- **handler_get.go / handler_tunnel.go**: #550's "alias" terminology + this PR's `/qurl-admin` command routing.
- Dropped two now-unused `//nolint:gosec` directives (the merged call graph left gosec's G706 taint pass not flagging those two `slog` lines — same nolintlint class as the earlier handler_alias.go fix).

Verified against CI's `golangci-lint v2.10.1` (0 issues) and the full `apps/slack/...` + `shared/...` suites.

Kevin-layerV:
Pushed 0cd7e92 + updated the PR description. Going through the notes:

1. **Deploy ordering / manifest gap** — agreed it's a deploy-runbook item, not code. Expanded the "⚠️ Deploy prerequisite" section to spell out the transitional window: register `/qurl-admin` in the manifest first (or simultaneously); until it lands, old `/qurl set-alias`/`/qurl tunnel install` invocations get the wrong-surface redirect to a not-yet-existing `/qurl-admin`, so admin verbs are unavailable in that window while the user surface stays functional. Not fenced by tests (they assume both commands registered), as you noted.
2. **Static copy bakes in prod literal** — documented in the `SCOPE OF THE REWRITE` comment; operator-only sandbox audience. Would thread the invoked command through those sites only if the env-infix surface ever ships externally; not worth the churn now.
3. **`/qurl-admin help` not a redirect** — added an explicit assertion to `TestDispatchSplit_HelpPerCommand` (`adminHelp` must not contain "belongs on"), so adding `help` to `userVerbs` would now fail a test rather than silently redirect.
4. **`adminVerbs` dual `set-alias`/`setalias`** — correct as-is (`slashVerb` is exact-or-prefix, not dash-tolerant); leaving the dual entries. Low-priority cleanup only if `slashVerb` ever normalizes dashes itself.
5. **`requireAliasAdminGate` user-ID trim** — leaving the local `TrimSpace`; it's a one-liner at the single gate entry and centralizing into `aliasPreamble` would spread the canonicalization across more surface than it saves. Minor.
6. **`AdminAction` constants + `//nolint:exhaustive`** — thanks; no change.
7. **Removed `//nolint:gosec` on the slog lines** — confirmed clean: CI pins `golangci-lint v2.10.1` and runs the *full* tree (no `--new-from-rev`), and it reports 0 issues. The gosec G706 taint pass simply no longer flags those `slog` lines under the post-merge call graph, so `nolintlint` had started failing them as *unused* directives — the removal is that fix (same class as the earlier handler_alias.go removal), not a config change.
8. **Sandbox help discoverability** — noted; current copy is intentional, no change.

Kevin-layerV:
Pushed f26ce03. On the optional polish:

**Implemented:**
- **Constant invariant** — added `TestCommandNameConstantsStayInSync` asserting `commandAdmin == commandUser + adminCommandSuffix` (the load-bearing assumption behind the help-text `ReplaceAll`). A future rename that desyncs it now fails a test.
- **`echoText` rune-budget branch** — commented the final `return s`: byte length over cap but rune count within it (multi-byte text) intentionally returns as-is rather than truncating.

**Confirmed, no change:**
- **`nolint:gosec` removal** — sanity-checked: CI runs full `golangci-lint v2.10.1` and it's green; the directives were unused (`nolintlint`), not a gosec regression.
- **Deploy gap** — runbook ordering (register `/qurl-admin` first/simultaneously) is now explicit in the PR description's deploy-prerequisite section.

**Considered, keeping as-is (with reasons):**
- **Parser-error leak before the admin gate** — intentional: gate ordering is parse → validate → admin (the security assertion is gate-before-*resolve*/-*write*, which the tests pin). The usage hint is public grammar, not alias enumeration (it never reveals which aliases exist), so gating it would only worsen UX for a non-admin typo. Happy to revisit if customer-visible enumeration becomes a concern.
- **`requireAliasAdminGate` `TrimSpace`** — kept: it matches the existing convention (`handleAdmin` also trims `fieldUserID` before calling `requireAdminSync`), so removing it here would make the alias gate the odd one out.
- **`firstWord` helper** — kept: it guards the empty-`Fields` case (returns `""`); inlining `strings.Fields(text)[0]` would reintroduce an index panic risk.
- **Extract `respondUnknownSubcommand` / merge adjacent `AdminStore` blocks** — left explicit: the two unknown-subcommand branches differ in copy ("admin" wording), and the two `AdminStore` blocks in `userHelpMessage` are separated by the `setupLine` append, so neither dedups cleanly. Verb-lookup via map vs slice is noise at this scale, agreed.

Kevin-layerV:
Pushed 5a232e7.

**Implemented (both "worth a doc note"):**
1. **Gate-order in `handleSetAlias`** — added an inline note: parse/validate usage hints are surfaced before the admin gate on purpose (the grammar is public — it's in the ungated `/qurl-admin help`), so the gate's guarantee is "no slug resolve or store mutation before the admin check," not "no parser feedback to non-admins."
2. **`/qurl-admin help` ungated** — added an inline note that `help` is intentionally not admin-gated (discovery, not a privileged action; gating it would obscure rather than protect), so the file no longer reads as "every admin verb gates" with an unexplained exception.

**Considered, no change:**
3. **Bogus-command rendered name** — leaving the coverage as-is. `/qurl-bogus` rendering `/qurl-bogus list` is only reachable if Slack dispatches an unregistered command, which it doesn't; pinning that in a test would assert behavior on an impossible input. The dispatch comment already marks it "unreachable in practice."
4. **`slashCommandForVerb` / `help` routing** — agreed it's a maintainer note; `TestDispatchSplit_HelpPerCommand`'s "belongs on" assertion already fences the dangerous form (adding `help` to `userVerbs`), so no further change.
5. **Deploy ordering → runbook** — the PR's "⚠️ Deploy prerequisite" section documents the manifest-first ordering and that tests don't fence the gap; I'll carry that into the deploy runbook at rollout time (it's an ops/process step, not a code change, and the Slack-app rollout is tracked outside this repo).

PR #542, opened 2026-05-29
fix(slack): actionable error when /qurl setup hits the API-key limit

Kevin-layerV:
## Problem

Running `/qurl setup` returned a bare white page reading:

> could not provision qURL key — run /qurl setup again to retry

That advice is **wrong** for the actual cause. `/qurl setup` mints a **new** qURL API key on every run (it overwrites the stored DDB row but doesn't revoke the prior key on qurl-service). Once an account reaches its plan's key cap (**free tier = 3**), qurl-service refuses the next mint with **HTTP 403** and error code `api_key_limit`. Retrying never clears a quota — the user has to revoke a key. (Confirmed in sandbox: at 3 keys it 403'd; revoking one let setup succeed.)

The callback couldn't tell the user that because (a) the minter collapsed every non-2xx into `returned <code>`, and (b) the `api_key_limit` 403 is indistinguishable on status alone from a genuine missing-scope 403 — the body `code` is the only signal.

## Change

- **`minter.go`**: parse qurl-service's error envelope (`{"error":{"code":...}}`) and return a typed `ErrAPIKeyLimitReached` when the code is `api_key_limit`. A body that isn't that shape (e.g. the existing `{"error":"forbidden"}` test case) falls back to the generic status error.
- **`callback.go`**: replace the bare `http.Error` on the mint-failure path with a styled HTML page (same strict-CSP/no-asset posture as the existing success and rebind-refused pages — no more blank white page). Two cases:
  - **key limit → 409**, names the cause + tells the admin to revoke a key.
  - **other mint failure → 502**, keeps the (correct, for transient errors) retry guidance.

The number of allowed keys is deliberately **not** stated — it's plan-dependent (free 3 / growth 50 / enterprise unlimited) and is qurl-service's to own.

### Before → After (the page text the user sees)

Before (both cases, blank page): `could not provision qURL key — run /qurl setup again to retry`

After, key-limit case:
> **⚠ qURL key limit reached**
> Your qURL account already has the maximum number of API keys allowed on your plan, so a new one couldn't be created. Each run of /qurl setup creates a new key — revoke one you no longer use, then run /qurl setup again.

After, other failure:
> **⚠ Couldn't connect qURL**
> Something went wrong while creating your qURL API key. Run /qurl setup again in a few minutes. If it keeps failing, please contact your qURL administrator.

(Wording is easy to tweak — say the word.)

## Tests

- `TestHTTPAPIKeyMinterMintAPIKeyLimit`: the `api_key_limit` envelope maps to `ErrAPIKeyLimitReached`; the plain `{"error":"forbidden"}` 403 stays generic.
- `TestCallbackMintAPIKeyLimitRendersGuidance`: the limit case renders a 409 page naming the limit + "revoke", and persists nothing.
- `go build`, `go test ./internal/oauth/...`, and `pre-commit` (incl. pinned golangci-lint v2.10.1) all pass.

## Note / possible follow-up

This fixes the *message*. The reason accounts hit the cap is that `/qurl setup` leaks a key per run (mints without revoking the prior — the orphan-key path, qurl-integrations#265-adjacent). Making setup revoke/reuse the previous workspace key would stop the accumulation at the source; out of scope here.

### DIFF
--- a/apps/slack/internal/oauth/callback.go
+++ b/apps/slack/internal/oauth/callback.go
@@ -508,9 +508,10 @@ func mintAndPersist(w http.ResponseWriter, cfg Config, accessToken, teamID, user
 	apiKey, keyID, keyPrefix, err := cfg.Minter.MintAPIKey(mintCtx, accessToken,
 		keyName, apiKeyScopes())
 	if err != nil {
+		limitReached := errors.Is(err, ErrAPIKeyLimitReached)
 		//nolint:gosec // G706: slog escapes control bytes in attribute values.
-		slog.Error("oauth/callback qurl-service mint failed", "error", err, "team_id", teamID, "api_key_limit", errors.Is(err, ErrAPIKeyLimitReached))
-		if errors.Is(err, ErrAPIKeyLimitReached) {
+		slog.Error("oauth/callback qurl-service mint failed", "error", err, "team_id", teamID, "api_key_limit_reached", limitReached)
+		if limitReached {
 			// Quota is a precondition the admin must clear themselves —
 			// retrying does nothing (the old "run setup again" advice was
 			// actively wrong here). 409 so an automated retry surfaces the
@@ -585,17 +586,27 @@ func dmAdminAsync(client SlackClient, userID, teamID, keyPrefix string) {
 	}
 }
 
-// renderRebindRefused writes the rebind-refused page. Same defense-in-
-// depth headers as the success page — the only differences are the body
-// template and a 409 status code (so an automated retry surfaces the
-// conflict rather than silently looping).
-func renderRebindRefused(w http.ResponseWriter, teamID string) {
+// setOAuthPageSecurityHeaders writes the defense-in-depth header set shared
+// by every OAuth-callback HTML response (success, rebind-refused, error).
+// These pages render post-redirect, so they shouldn't be framable
+// (clickjacking), shouldn't leak the callback URL via Referer to anything
+// they link to, and shouldn't load any off-origin resources beyond the
+// inline style.
+func setOAuthPageSecurityHeaders(w http.ResponseWriter) {
 	w.Header().Set("Content-Type", "text/html; charset=utf-8")
 	w.Header().Set("Cache-Control", "no-store")
 	w.Header().Set("X-Frame-Options", "DENY")
 	w.Header().Set("Content-Security-Policy", "default-src 'none'; style-src 'unsafe-inline'")
 	w.Header().Set("Referrer-Policy", "no-referrer")
 	w.Header().Set("X-Content-Type-Options", "nosniff")
+}
+
+// renderRebindRefused writes the rebind-refused page. Same defense-in-
+// depth headers as the success page — the only differences are the body
+// template and a 409 status code (so an automated retry surfaces the
+// conflict rather than silently looping).
+func renderRebindRefused(w http.ResponseWriter, teamID string) {
+	setOAuthPageSecurityHeaders(w)
 	w.WriteHeader(http.StatusConflict)
 	if err := rebindRefusedPageTemplate.Execute(w, rebindRefusedPageData{TeamID: teamID}); err != nil {
 		slog.Warn("oauth/callback rebind-refused page write failed", "error", err)
@@ -607,12 +618,7 @@ func renderRebindRefused(w http.ResponseWriter, teamID string) {
 // renders human-readable, actionable guidance instead of a blank page with
 // raw text. Same defense-in-depth headers as renderRebindRefused.
 func renderOAuthErrorPage(w http.ResponseWriter, status int, heading, message string) {
-	w.Header().Set("Content-Type", "text/html; charset=utf-8")
-	w.Header().Set("Cache-Control", "no-store")
-	w.Header().Set("X-Frame-Options", "DENY")
-	w.Header().Set("Content-Security-Policy", "default-src 'none'; style-src 'unsafe-inline'")
-	w.Header().Set("Referrer-Policy", "no-referrer")
-	w.Header().Set("X-Content-Type-Options", "nosniff")
+	setOAuthPageSecurityHeaders(w)
 	w.WriteHeader(status)
 	if err := oauthErrorPageTemplate.Execute(w, oauthErrorPageData{Heading: heading, Message: message}); err != nil {
 		slog.Warn("oauth/callback error-page write failed", "error", err)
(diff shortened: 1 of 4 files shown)


Kevin-layerV:
Thanks — all five points addressed in e49476c:

1. **Security headers on the new render path** — added an `assertSecurityHeaders(t, rec)` helper and applied it to the 409 (key-limit), 502 (generic), and success render paths. Refactored `TestCallbackHappyPath` onto it too, so the header contract is asserted in one place across all three paths.
2. **502 path is now styled HTML** — `TestCallbackMintFailureDoesNotRevoke` now asserts the styled body (`connect qURL` — avoids the apostrophe in "Couldn't", which `html/template` escapes to `&#39;`) plus the headers, so a revert to bare `http.Error` fails loudly.
3. **Asymmetric body check** — both substrings now match against `strings.ToLower(rec.Body.String())`.
4. **slog field naming** — `api_key_limit` → `api_key_limit_reached`. Also hoisted the duplicated `errors.Is(...)` into a local `limitReached` so the field and the branch share one evaluation.
5. **Template consolidation** — went ahead and extracted the duplicated 6-header block into a shared `setOAuthPageSecurityHeaders(w)` (success/rebind-refused/error now share one header source — this PR added the third copy). The deeper `<style>`/card-shell consolidation across the three page templates is deferred to #543 (labeled `pr-polish-followup`): it touches the XSS-sensitive success/rebind templates that are outside this PR's scope, and your "if a fourth lands" threshold is the right trigger.

The only failing check is the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` (5s kubectl dry-run timeout, `signal: killed`) — unrelated to this diff (the `oauth` package passes); re-running the job.

Kevin-layerV:
Thanks for the second pass. Walking the five observations:

1. **`apiKeyLimitError` ignores the HTTP status** — keeping it code-driven (not gating on 403), by design. The PR's core thesis is that the body `code` is the *authoritative* disambiguator and status alone can't be trusted (scope-gate and quota both surface as 403 today). Gating on 403 would reintroduce a status coupling and is actually more fragile in the other direction: a plan/quota limit is exactly the kind of condition qurl-service could reframe as `402 Payment Required` or `429`, and a hard 403 gate would then silently miss the `api_key_limit` code and fall back to the wrong "try again" page. If qurl-service ever emits the code on a 5xx that's a server bug to fix upstream, not something to paper over here. The check already only runs on the non-2xx branch, so the blast radius is bounded.

2. **Shared CSS skeleton across the templates** — agreed, and tracked: #543 (`pr-polish-followup`) covers extracting a shared base shell for the three page templates, with the same "extract when it stops diverging / a fourth lands" reasoning.

3. **Dead `code{...}` CSS rule in `oauthErrorPageTemplate`** — dropped in 3d4f904. `Heading`/`Message` are plain-text-escaped so no `<code>` ever renders; it was dead style cloned from the success/rebind pages.

4. **`/qurl setup` not wrapped in `<code>`** — leaving as-is intentionally. `Message` is a single plain-text-escaped string with no markup hook, and the whole point of that posture is that html/template auto-escapes it (the load-bearing XSS defense for any future caller passing an upstream string). Wrapping a substring in `<code>` would mean trusting HTML inside `Message`, which undercuts that guarantee for ~one inline-code style. Not worth the safety tradeoff for this page.

5. **Tracking issue for the key-per-run leak** — filed #544 (revoke-then-mint or reuse the workspace key), linked back to this PR so the "why we needed a 409 page" context is preserved.

CI is green (the earlier red was the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl`, which passed on the rerun).

Kevin-layerV:
Thanks — third pass triaged:

1. **Four other bare `http.Error` sites** — good catch; this is the rest of the "no blank white pages on `/qurl setup`" UX. Filed #545 with the full punch list (persist-failure at `callback.go:546` flagged highest-leverage, as you noted — it's a half-installed state the admin must understand). Kept out of this PR to keep the diff scoped to the key-limit case, but the infra (`renderOAuthErrorPage` + `setOAuthPageSecurityHeaders`) is now in place for it.
2. **Test stubs the sentinel directly** — agreed, no action: coverage is complete in aggregate (`TestHTTPAPIKeyMinterMintAPIKeyLimit` locks envelope→`%w` wrap; the callback test locks sentinel→page+no-persist). Noting the implicit `errors.Is` composition for whoever touches either seam.
3. **JSON-shape vs Content-Type / RFC 7807** — added a one-line note to the `apiKeyLimitError` doc in 341a62c: the envelope is qurl-service's nested `{"error":{"code":...}}`, NOT RFC 7807 problem+json, so the match is shape-driven not Content-Type-driven. No behavior change.
4. **Inline-style nits** (`p` font-size divergence; two-`<p>` for mobile) — folded into #545's polish scope; the `<p>` sizing will also fall out of the #543 shared-base extraction. No action here.

CI green; #543 (template-shell base), #544 (key-per-run leak), #545 (remaining error pages) track the deferred work.

Kevin-layerV:
Thanks — fourth pass:

1. **XSS lock for the new template** — added in cc5b571. `TestOAuthErrorPageHTMLEscapesInterpolations` feeds `<script>` (Heading) and `` (Message) payloads through `renderOAuthErrorPage` and asserts they render escaped, symmetric to `TestSuccessPageHTMLEscapesInterpolations`. This was the one item you'd weigh before merge, so it's in.
2. **status-code check on `apiKeyLimitError`** — leaving name/shape-driven by design (covered in the round-2 reply: gating on 403 would miss a future 402/429 reframing of the quota condition). No change.
3. **Loose 502 body assertion** — tightened in cc5b571 to pin the exact escaped heading `Couldn&#39;t connect qURL`, so a reworded heading is caught too.
4. **Body-cap (8 KiB) interaction** — acknowledged; envelope is a few hundred bytes, so the generic exceeded-bytes path is unreachable in practice. No guard.
5. **CSS duplication** — tracked by #543 (shared-base extraction), same "if a fourth lands" trigger you noted.

Follow-ups are all filed: #543 (template base), #544 (key-per-run leak — the root cause), #545 (style the remaining callback error pages). CI is green.

Kevin-layerV:
Thanks — fifth pass:

1. **`mintAndPersist` length** — agreed it's getting long; leaving the `handleMintFailure(w, err)` split for the #265 orphan-revoke follow-up that'll touch this function anyway (noted in #544). No change here.
2. **Wrap-format test** — added in d29f4bd: `TestHTTPAPIKeyMinterMintAPIKeyLimit` now asserts `err.Error()` still contains `403`, so dropping the `(status %d)` suffix is caught.
3. **Wording ("Each run of /qurl setup creates a new key")** — keeping it. From the admin's seat at the cap, the most useful thing is knowing that *re-running setup is what accumulated the keys* — that's what stops them from doing the wrong thing (running setup again) and steers them to revoke. The neutral phrasing loses that causal nudge. It's accurate, and #544 will remove the underlying leak so the sentence becomes moot anyway.
4. **`team_id` on the error page** — skipping, keeping the template minimal. `renderOAuthErrorPage` is a generic heading+message renderer (reused for the 502 path and available for the #545 conversions); threading a workspace ID would couple it to the OAuth-callback context. `renderRebindRefused` shows it because that page is intrinsically workspace-specific.
5. **Negative-case envelope test** — added in d29f4bd: `TestHTTPAPIKeyMinterMintForbiddenEnvelopeStaysGeneric` pins that a well-formed envelope with a non-limit code (`insufficient_scope`) stays generic, so an exact-code → substring loosening breaks loudly.

CI green. Deferred work tracked in #543/#544/#545.

Kevin-layerV:
Thanks. On the one item you wanted explicit acknowledgement of:

1. **Styling scope** — intentionally narrow *and* follow-up tracked: #545 applies `renderOAuthErrorPage` to the remaining bare `http.Error` paths (236 Auth0 token, 304 Auth0 ?error=, 412 identity, 460 bind, 546 persist), with the persist-failure case called out as highest-leverage exactly as you note. This PR keeps to the actionable-error contract for the mint path.

Items 2–5 are all take-it-or-leave-it and I'm leaving them: the `api_key_limit_reached` bool earns its keep as a structured dashboard/alert filter; the escaped-heading assertion is the tighter lock requested earlier; and the `apiKeyLimitError` → code-extractor evolution is the natural shape if qurl-service grows more actionable codes (not needed at one code).

The Test failure is the known-flaky `TestRenderKubernetesPodSpecFragmentDryRunsWithKubectl` (5s kubectl dry-run, `signal: killed`) — unrelated to this diff (oauth package passes at 87.6%); re-ran the job.

PR #549, opened 2026-05-29
fix(slack): drop phantom views:write scope from install flow

Kevin-layerV:
## Problem

The Slack app-install flow requested bot scopes `commands` + `views:write` (`slackinstall.DefaultBotScopes()`), but **`views:write` is not a real Slack OAuth scope**:

- Slack's `views.open` — the call that renders the `/qurl tunnel install` modal — requires **no scope** ([docs](https://docs.slack.dev/reference/methods/views.open): *"Scopes: No scopes required"*).
- The "Add an OAuth Scope" picker on api.slack.com doesn't list `views:write`.

Because the install handler put `commands,views:write` in the Slack `oauth/v2/authorize` URL, Slack would reject the flow with **`invalid_scope`** (an error the callback already allowlists in `safeSlackOAuthErrorCode`), so the workspace app-install flow could never capture a bot token. There was also no env workaround: `Config.Validate()` hard-rejected any scope set missing `views:write`. This never surfaced live only because the flow was never activated (its env vars weren't wired — that wiring is `qurl-integrations-infra#800`).

## Fix

The captured per-workspace token is used **solely** for `views.open` (`apps/slack/cmd/main.go`: *"Slack views.open wired with per-workspace token lookup"*), which needs no scope. `commands` is the minimal scope required to install the slash-command app into a workspace. So `DefaultBotScopes()` now returns just `["commands"]`.

- `routes.go`: remove the `botScopeViewsWrite` const; `DefaultBotScopes()` → `["commands"]` (with a comment explaining why).
- `routes_test.go`: a `commands`-only config now validates; the missing-required-scope callback test grants a scope set **without** `commands` so it still exercises the 422 rejection path.
- `main_test.go`: default-scope and custom-scope assertions use real scopes.
- `README.md`: correct the `SLACK_BOT_SCOPES` default (`commands`) and the bot-scope setup guidance.

## Validation

`go build ./apps/slack/...`, `go test ./apps/slack/internal/slackinstall/... ./apps/slack/cmd/...`, `go vet`, and `pre-commit run` all pass.

## Context

This is the code-side companion to the infra activation of the Slack app install flow (`qurl-integrations-infra#800` wired `SLACK_CLIENT_ID/SECRET`). Without this fix, the install OAuth fails at the authorize step regardless of infra/manifest. Note: the Slack app manifest must **not** carry `views:write` either — `apps.manifest.update` would reject it.

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -156,7 +156,7 @@ For customer Slack installs, configure the Slack app with:
   shapes defensively, but qURL does not request or persist Slack refresh tokens
   yet, so rotation-enabled apps need refresh support before production use.
 
-When moving to per-workspace token storage, existing customer workspaces must
+With per-workspace token storage in place, existing customer workspaces must
 reinstall or reauthorize the Slack app so Slack issues a per-workspace bot
 token. New installs through `/oauth/slack/install` store that token
 automatically, and guided `/qurl tunnel install` uses it for `views.open`
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -148,7 +148,7 @@ For customer Slack installs, configure the Slack app with:
 - Customer install link: `https://<SLACK_BASE_URL host>/oauth/slack/install`
 - Slash command request URL: `https://<SLACK_BASE_URL host>/slack/commands`
 - Interactivity request URL: `https://<SLACK_BASE_URL host>/slack/interactions`
-- Bot scopes: at least `commands` (`views.open` itself requires no scope)
+- Bot scopes: `commands` (the captured token is used only for `views.open`, which requires no scope of its own)
 - Installation mode: workspace-level installs or Enterprise Grid org-level
   installs. Org-level bot tokens are stored under Slack `enterprise_id`; qURL
   workspace setup and admin checks still use workspace `team_id`.
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -14,7 +14,7 @@ Customer onboarding is install-first:
 - `/qurl get <$slug|$alias>` — Mint a one-time qURL for a tunnel `$slug` or a channel `$alias` (raw URLs aren't supported)
 - `/qurl set-alias $<alias> $<slug>` — Point a channel shortcut at a tunnel slug (admin-only)
 - `/qurl unset-alias $<alias>` — Remove a channel shortcut binding (admin-only)
-- `/qurl tunnel install` — Guided tunnel sidecar setup with target-environment choices (admin-only; uses the workspace bot token stored during Slack app install with `views:write`)
+- `/qurl tunnel install` — Guided tunnel sidecar setup with target-environment choices (admin-only; uses the workspace bot token stored during Slack app install)
 - `/qurl tunnel install <slug|$slug> [port:<n>] [alias:$shortcut] [env:<target>] [container:<name>]` — Provision a tunnel from a typed command (admin-only; default local port is 8080)
 - `/qurl list` — List the protected tunnel resources available to you (with their bound channel shortcuts)
 - Link unfurling for `qurl.link` URLs (planned)
@@ -115,7 +115,7 @@ docker buildx build --platform linux/arm64 \
 | `SLACK_CLIENT_ID` | Slack install | Slack app client ID used by `/oauth/slack/install`. Required for customer installs that capture per-workspace bot tokens. |
 | `SLACK_CLIENT_SECRET` | Slack install | Slack app client secret used by `/oauth/slack/callback` to exchange Slack's OAuth code. |
 | `SLACK_INSTALL_STATE_SECRET` | Slack install | HMAC-SHA256 key for Slack install state signing. Must be ≥32 bytes. Use a distinct production secret from `OAUTH_STATE_SECRET`; the fallback is only for local/dev compatibility. |
-| `SLACK_BOT_SCOPES` | No | Comma/space-separated bot scopes requested by `/oauth/slack/install`. Empty defaults to `commands,views:write`; any override must still include both required scopes. |
+| `SLACK_BOT_SCOPES` | No | Comma/space-separated bot scopes requested by `/oauth/slack/install`. Empty defaults to `commands` (the captured token is used only for `views.open`, which requires no scope); any override must still include `commands`. |
 | `SLACK_BOT_TOKEN` | Legacy | Single-workspace fallback token for `views.open` when a workspace has not yet completed Slack install OAuth. Accepts `xoxb-` and `xoxe.xoxb-` token shapes. Production multi-customer installs should not depend on this fallback. |
 | `QURL_ENDPOINT` | Yes | qURL API base URL (e.g. `https://api.layerv.xyz`) |
 | `WORKSPACE_STATE_TABLE` | Yes | DynamoDB table holding per-workspace API keys (provisioned by `qurl-integrations-infra`) |
@@ -148,7 +148,7 @@ For customer Slack installs, configure the Slack app with:
 - Customer install link: `https://<SLACK_BASE_URL host>/oauth/slack/install`
 - Slash command request URL: `https://<SLACK_BASE_URL host>/slack/commands`
 - Interactivity request URL: `https://<SLACK_BASE_URL host>/slack/interactions`
-- Bot scopes: at least `commands` and `views:write`
+- Bot scopes: at least `commands` (`views.open` itself requires no scope)
 - Installation mode: workspace-level installs or Enterprise Grid org-level
   installs. Org-level bot tokens are stored under Slack `enterprise_id`; qURL
   workspace setup and admin checks still use workspace `team_id`.
@@ -156,10 +156,9 @@ For customer Slack installs, configure the Slack app with:
   shapes defensively, but qURL does not request or persist Slack refresh tokens
   yet, so rotation-enabled apps need refresh support before production use.
 
-After adding `views:write` or moving to per-workspace token storage, existing
-customer workspaces must reinstall or reauthorize the Slack app so Slack issues
-a bot token with the new scope. New installs through `/oauth/slack/install`
-store that token automatically, and guided `/qurl tunnel install` will use it
-for `views.open`. If Slack tells a customer guided tunnel setup needs the latest
-qURL Slack app install, send them through this reinstall link and confirm the app
-grants `views:write`.
+When moving to per-workspace token storage, existing customer workspaces must
+reinstall or reauthorize the Slack app so Slack issues a per-workspace bot
+token. New installs through `/oauth/slack/install` store that token
+automatically, and guided `/qurl tunnel install` uses it for `views.open`
+(which requires no scope). If Slack tells a customer guided tunnel setup needs
+the latest qURL Slack app install, send them through this reinstall link.
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -764,16 +764,23 @@ func buildSlackInstallConfig(provider *auth.DDBProvider) (slackinstall.Config, b
 	scopes := slackinstall.DefaultBotScopes()
 	if raw := strings.TrimSpace(os.Getenv(envSlackBotScopes)); raw != "" {
 		scopes = slackinstall.NormalizeScopes([]string{raw})
+		// Strip a stale `views:write` (not a real Slack scope, see
+		// slackinstall.DefaultBotScopes) from operator overrides rather than
+		// forward or reject it: forwarding makes every customer install fail at
+		// Slack's authorize step with invalid_scope until the SLACK_BOT_SCOPES
+		// runbook is patched, while rejecting it in Validate() would abort bot
+		// startup. Stripping keeps installs working off the valid scopes and
+		// warns the operator. EqualFold because NormalizeScopes does not lowercase.
+		kept := make([]string, 0, len(scopes))
 		for _, s := range scopes {
-			if s == "views:write" {
-				// Warn rather than reject: a Validate() failure below aborts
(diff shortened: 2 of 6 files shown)


Kevin-layerV:
Thanks for the reviews — addressing the nits from both passes:

**Fixed in 6951492:**
- **`shared/auth/ddb_provider_test.go` phantom fixture** (flagged in both reviews): swapped the `views:write` dedup/sort fixture and assertion to a real scope (`chat:write`). A repo-wide grep for `views:write` now lands only on the intentional "do not add" warning, not on test data that reads as documentation of a nonexistent scope.
- **`DefaultBotScopes` doc comment** (Review 2's CLAUDE.md "comment that rots" nit): reframed the historical `An earlier revision also requested views:write…` line into a forward-looking `Do not add views:write: it is not a real Slack scope…` warning. This drops the reference-to-this-fix framing CLAUDE.md flags while preserving the load-bearing guidance Review 1 wanted (the next person tempted to re-add the scope still finds the answer in `DefaultBotScopes`).

**Evaluated, keeping as-is:**
- **README "When moving to per-workspace token storage…" paragraph**: still load-bearing. The legacy single-workspace `SLACK_BOT_TOKEN` fallback remains supported, so operators still need the guidance that existing installs must reinstall/reauthorize to capture a per-workspace token. The text no longer references the dropped scope, so it's accurate.
- **`botScopeCommands` single-use const**: keeping per Review 2's own "fine to keep" — `missingRequiredScopes` references it and it documents intent.

**Already covered:**
- The PR body references `qurl-integrations-infra#800`; that carries into the squash-merge commit body at merge time.

Kevin-layerV:
Addressing the round-3 observations:

**Obs 1 (env override can still misconfigure) — implemented as a warn, not a reject (90a9dca).**
Added a `slog.Warn` at config load when `SLACK_BOT_SCOPES` carries a stale `views:write`, so operators get an early, actionable signal instead of waiting for an opaque `invalid_scope` at Slack's authorize step.

I deliberately did **not** hard-reject in `Validate()`. A `Validate()` failure propagates through `buildSlackInstallConfig` to `main.go` (`return fmt.Errorf("slack install config: %w", err)`), which aborts the **entire bot** at startup. Denylisting `views:write` there would turn a narrow, stale-runbook misconfig into a full bot outage for every command — strictly worse than today, where the bot boots and only the install attempt trips `invalid_scope` (already structured-logged via `safeSlackOAuthErrorCode`). That's also the best-effort posture the round-1 review affirmed as correct. New test `TestBuildSlackInstallConfigAcceptsStaleViewsWriteOverride` locks in that non-rejection contract so this can't silently regress into the crash variant later.

(The `scopes` value isn't echoed in the log line — gosec flags it as taint since it's operator-controlled, and the static message already says what to fix, so no `#nosec` needed.)

**Obs 2 (comment precision on "minimal") — already covered.**
The comment reads "`commands` is the minimal scope to install **this slash-command app** into a workspace" — it's already scoped to the slash-command app, not a claim that `commands` is the global install minimum. Leaving as-is.

**Obs 3 (Slack manifest sync) — acknowledged, no in-repo change.**
No manifest is checked into this repo; the PR body already flags that whoever applies the Slack-side manifest must drop `views:write` too (else `apps.manifest.update` rejects it). Surfacing here so it isn't lost at merge.

Kevin-layerV:
Addressing the round-4 notes (02a82d9):

**#1 (warn not asserted) — fixed.** `TestBuildSlackInstallConfigStripsAndWarnsViewsWrite` now routes `slog` through a captured `bytes.Buffer` handler and asserts the warning fires. The diagnostic is load-bearing now, not advisory.

**#2 (case-sensitive match) — fixed.** Switched to `strings.EqualFold`, so `Views:Write`/`VIEWS:WRITE` are caught too. The test uses mixed-case (`commands,Views:Write`) to exercise it. I kept the fold local rather than lowercasing inside `NormalizeScopes`, which would change behavior for every scope globally.

**#3 (forward-with-warn vs strip vs reject) — switched to strip-and-warn.** You convinced me: `views:write` is provably non-functional, so forwarding it held customer installs hostage to runbook hygiene (every Install click → `invalid_scope` until patched). Config load now strips it from the override and warns the operator, so installs keep working off the valid scopes immediately. Reject stays off the table for the startup-abort reason. (If an override is *only* `views:write`, stripping empties it and `Validate()` rightly fails on the missing `commands` — same as today's missing-required-scope handling.)

**#4 (chat:write as load-bearing fixture) — keeping, with reasoning.** `chat:write` is a valid scope and works correctly as dedup/sort filler. Swapping it for a "more neutral" valid scope across three test files (`ddb_provider_test.go`, `main_test.go`, `routes_test.go`) is churn guarding a speculative future where qURL requests `chat:write` — and if that ever happens, it's a one-line fixture update at that point. Leaving it rather than spread the change.

**#5 (README future-tense) — fixed.** Reworded to "With per-workspace token storage in place, …" since that wiring already exists.

Kevin-layerV:
Addressing the round-5 suggestions (115f94d):

**#1 (global slog.SetDefault is fragile) — fixed by extracting a pure helper.** Pulled the strip into `slackinstall.DropUnsupportedScopes(scopes) (kept, dropped []string)` and unit-tested it directly (`TestDropUnsupportedScopes`: nothing-unsupported, strips, case-insensitive, only-views:write). No global state mutation. The config test (`TestBuildSlackInstallConfigStripsViewsWriteOverride`) now just asserts the wiring strips the scope — no slog capture.

**#2 (strip lived in cmd, not the package) — fixed.** `DropUnsupportedScopes` now sits next to `DefaultBotScopes` in `routes.go`, co-located with the scope source of truth; `main.go` calls it. Both sides of the defense are discoverable from the package.

**#3 (warn wording could mislead in the views:write-only case) — fixed.** Reworded to "...dropped it. SLACK_BOT_SCOPES must still include `commands`." so an operator who set only `views:write` understands the bot still won't start without `commands`.

**#5 (structured log field) — added.** `slog.Warn(..., "stripped_scope", "views:write")`. The value is a literal (not the operator-controlled scope slice), so it doesn't reintroduce the gosec log-taint that made me drop the `scopes` field earlier.

**#4 (slices.DeleteFunc) — left as-is.** `DropUnsupportedScopes` returns the dropped scopes too, which both its test and the warn condition rely on; `DeleteFunc` would only give the kept slice. The explicit loop reads clearly and avoids a new import.

Kevin-layerV:
Addressing the round-6 suggestions (4f0820c):

**#1 (Validate accepts views:write from direct callers) + #4 (default path bypasses the strip) — fixed with one mechanism.** `Validate()` now rejects unsupported scopes via `DropUnsupportedScopes` (the single source of truth). The env path still strips *before* `Validate` is reached, so a stale `SLACK_BOT_SCOPES` runbook value stays a non-fatal strip-and-warn. Only programmatic misuse — direct `Config{BotScopes: …}` construction, or a regression that puts `views:write` back into `DefaultBotScopes()` — reaches the new error, and failing loudly there is correct (it's a code bug, caught at startup/CI). New test `TestConfigValidateRejectsUnsupportedScope` covers it.

**#3 (no cmd-layer test for strip-to-empty) — added.** `TestBuildSlackInstallConfigRejectsOverrideThatStripsToEmpty`: `SLACK_BOT_SCOPES=views:write` strips to empty, and `buildSlackInstallConfig` surfaces "at least one bot scope is required" (ok=false) — confirming no silent fallback to defaults.

**Dense main.go comment nit — tightened.** Reduced to a one-line pointer to `DropUnsupportedScopes`, which now carries the full rationale. Avoids duplicating the why in two places.

**#2 (log the dropped slice instead of a hardcoded value) — keeping the literal, with reasoning.** `dropped` derives from the operator-controlled `SLACK_BOT_SCOPES`, so logging it re-triggers the gosec G706 log-taint I fixed earlier (I'm avoiding `#nosec`). `views:write` is the only scope `DropUnsupportedScopes` drops today, so the literal `stripped_scope` field is accurate and complete; if the helper ever drops more, that change can introduce a non-tainted canonical list. Kept the brief remediation hint in the message since operators read logs (not always the README) when debugging a failed start.

**#5 (Slack manifest source-of-truth) — no in-repo change.** No manifest is checked into this repo (confirmed). The PR body already flags that the Slack-side manifest (applied alongside `qurl-integrations-infra#800`) must also drop `views:write`, else `apps.manifest.update` rejects it.

Kevin-layerV:
Addressing the round-7 nits (c09fd01):

**#1 + #3 (slog `stripped_scope` is a hardcoded literal / singular) — removed the field.** I verified empirically: logging the actual `dropped` slice (even `strings.Join(dropped, ",")`) trips `gosec G706` (log-injection via taint) because it derives from the operator-controlled `SLACK_BOT_SCOPES`, and I'm not adding `#nosec`. So the field could only ever be the literal `"views:write"`, which just duplicated the message and — as you spotted — read as inconsistent with the case-insensitive match. The warn message already names the canonical scope for log search, so I dropped the redundant field rather than keep something that misleads. (This field was added per the round-5 suggestion before the gosec constraint surfaced.)

**#2 (comment "warns rather than aborting" is half-true) — fixed.** Tightened to: a stale `SLACK_BOT_SCOPES` *with surviving valid scopes* warns instead of aborting; an override of only unsupported scopes strips to empty and `Validate` still rejects it (exactly what `TestBuildSlackInstallConfigRejectsOverrideThatStripsToEmpty` codifies).

**#4 (README reinstall guidance may be over-stated) — keeping, with reasoning.** That paragraph is pre-existing content; this PR only swapped its scope references (dropped the `views:write` mentions). The reinstall guidance is still correct for any dev/internal workspace that did install, and rewording the rollout-state framing ("going live for the first time") is a broader doc change outside this scope-fix PR. Happy to do it as a follow-up if you'd prefer.

Kevin-layerV:
Addressing the round-8 nits (c118d62):

**#3 (README bot-scopes line reads redundantly) — fixed.** Reworded to mirror the `SLACK_BOT_SCOPES` env-var table row: "Bot scopes: `commands` (the captured token is used only for `views.open`, which requires no scope of its own)".

**#1 (warn hardcodes views:write; log the dropped slice instead) — keeping, with reasoning.** I empirically verified logging `dropped` (even `strings.Join(dropped, ",")`) trips `gosec G706` since it derives from operator input, and I'm not adding `#nosec`. A generic message ("scope(s) Slack rejects") would also be less actionable — naming `views:write` tells the operator exactly what to remove from their runbook. There's one unsupported scope today and `DropUnsupportedScopes`'s doc flags the coupling, so a future addition naturally revisits this message. (This is the same point as round-6 #2 / round-7 #1; the prior structured field was removed in c09fd01 for the gosec reason.)

**#2 (two-step Validate diagnosis for a views:write-only direct config) — keeping.** Return-first-error is the standard Go validation shape and this is the programmatic-only path (no live callers — it's defense-in-depth). Folding multiple problems into one error adds collection/join complexity for a path operators never hit. The env path already strips before Validate.

**#4 (rename DropUnsupportedScopes → StripScopesSlackRejects) — keeping per your own call.** You flagged it "not worth churning the symbol for this PR," and the doc comment already hedges the rule-vs-list distinction. Leaving the name.

Kevin-layerV:
Round-9 nits — all considered; none need a code change:

**#1 (warn hardcodes views:write; log `dropped` for original casing) — gosec-blocked.** I verified `slog.Warn(..., "dropped", dropped)` trips `gosec G706` because `dropped` derives (by data-flow) from the operator-controlled `SLACK_BOT_SCOPES`. Your "Things I checked" note is semantically right — `dropped` only ever holds case-variants of the internal `views:write` constant — but gosec's taint analysis is mechanical and flags it regardless, and I'm not adding `#nosec`. The message naming `views:write` is also the most actionable form for operators today (one unsupported scope; the helper's doc flags the coupling for any future addition).

**#2 (Validate two-step diagnosis) — keeping.** Return-first-error is the standard Go validation shape; this is the programmatic-only/defense-in-depth path with no live callers. Accumulating multiple errors isn't worth the complexity (you flagged the same).

**#3 (godoc: callers shouldn't bypass the helper) — already structural.** `Validate()` calls `DropUnsupportedScopes`, and every config path (`buildSlackInstallConfig` → `Validate` → `RegisterRoutes`) runs through `Validate`, so it's the chokepoint — any scope set that reaches Validate is checked. The two-layer usage is already documented at both call sites (the cmd-layer strip comment and the Validate reject comment), so the helper's godoc doesn't need to enumerate callers.

These three have been stable across the last few rounds (the recurring warn-message point is gosec-bound). I believe the substantive feedback is fully resolved.

Kevin-layerV:
Round-10 nits — all considered; this is the same stable set, none newly actionable:

**#1 (warn message / structured `dropped_scopes` attribute) — gosec-blocked.** Verified empirically that `slog.Warn(..., "dropped_scopes", strings.Join(dropped, ","))` trips `gosec G706` (taint: `dropped` derives from the operator-controlled `SLACK_BOT_SCOPES`), and I'm not adding `#nosec`. The structured field was tried and removed for exactly this reason (c09fd01). The message names the one unsupported scope explicitly, which is the most actionable form today.

**#2 (hardcoded `views:write` / co-locate rationale) — keeping.** Agreed a map is premature for one entry. If a second unsupported scope ever lands, that change can introduce a `unsupportedScopes` set with co-located rationale; not worth pre-building now.

**#3 (Validate ordering) — keeping (you noted it's not a defect).** The `routes.go:171` comment already explains the env-strips-first / direct-callers-reject asymmetry; the per-path error-message detail is a minor elaboration I'd rather not re-expand the comment for.

**#4 (manifest) — out of repo.** No manifest in this repo; the Slack-side manifest is applied with `qurl-integrations-infra#800`, and the PR body already flags it must drop `views:write` too.

I believe the PR is converged: CI green, all review feedback across the rounds either applied or answered, and the remaining suggestions are gosec-bound / premature / out-of-scope.

PR #552, opened 2026-05-29
fix(slack): show tunnel slug in /qurl aliases, never resource_id

Kevin-layerV:
## What

`/qurl aliases` now shows each tunnel by its **slug** and never the opaque resource ID.

**Before** (every row fell back to the resource ID):
```
Aliases configured for this channel:
• r_0pxn89qcm9i (no slug) → $kevin-dashboard, $kevin-test-dashboard
• r_cnaip_fprni (no slug) → $dashboard
```

**After**:
```
Aliases configured for this channel:
Format: $<slug> → the aliases that resolve to it. Run /qurl get with the slug or any alias.
• $kktest → $kevin-dashboard, $kevin-test-dashboard
• $dashboard-slug → $dashboard
```

## Why

The listing resolved each tunnel's slug by calling `GetResourceByAlias`, which hits `GET /v1/resources/by-alias/{alias}`. qurl-service removed that path — alias lookup moved to `GET /v1/resources?alias=`, and `slug` is now surfaced on `ResourceData`. So the call always errored, the slug never resolved, and every row degraded to the `r_<id> (no slug)` fallback. (`/qurl get` and `/qurl list` were unaffected — they resolve via the channel binding store and `ListResources`, not the dead endpoint.)

## How

- **shared/client:** add `GetResource(id)` → `GET /v1/resources/{id}` (returns the full resource, including the tunnel slug). Remove the dead-endpoint `GetResourceByAlias` — its only caller was `/qurl aliases`.
- **handler_aliases:** resolve each group by the `resource_id` already held in `channel_policies` (no reason to round-trip through an alias), then render `$<slug> → $<aliases>` under a format legend.
- **Never surface the resource ID:** a group whose slug can't be resolved (upstream fetch failed, or a legacy resource predating the slug requirement) degrades to listing its channel aliases alone — the user can still `/qurl get $alias`.

## Test

- `go test ./shared/client/ ./apps/slack/internal/` — pass.
- Updated client tests (`GetResource*`) and aliases-handler tests to the by-id path; cancellation/deadline tests now assert the alias-only fallback and that no `r_<id>` ever leaks.
- `golangci-lint` clean; pre-commit green.

### DIFF
--- a/apps/slack/internal/handler_aliases.go
+++ b/apps/slack/internal/handler_aliases.go
@@ -92,14 +92,16 @@ func (h *Handler) processAliases(ctx context.Context, log *slog.Logger, values u
 	// Collapse the per-alias bindings into one group per tunnel: several
 	// aliases can point to the same slug, so the listing shows the slug
 	// once followed by every alias that resolves to it here. Per-group
-	// resource fetch is best-effort — a failed fetch degrades to a
-	// resource-id line rather than dropping the group.
+	// resource fetch is best-effort — a failed fetch degrades to an
+	// alias-only line (never the opaque resource_id) rather than dropping
+	// the group.
 	groups := groupAliasEntriesByResource(entries)
 	lines := fanoutAliasGroups(ctx, log, c, groups, aliasesResourceFanoutLimit)
 	sort.Strings(lines)
 
-	body := "*Aliases configured for this channel:*\n" + strings.Join(lines, "\n") +
-		"\n\n_Each line is a tunnel `$<slug>` followed by the aliases that resolve to it here. Run `/qurl get` with the slug or any alias._"
+	body := "*Aliases configured for this channel:*\n" +
+		"_Format: `$<slug>` → the aliases that resolve to it. Run `/qurl get` with the slug or any alias._\n" +
+		strings.Join(lines, "\n")
 	_ = h.postResponse(log, responseURL, body)
 }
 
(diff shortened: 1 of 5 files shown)


Kevin-layerV:
Thanks for the review. Addressed in `7935af4`:

- **Precise leak fence:** tightened the cancellation + deadline tests from `Contains(l, "r_")` to `Contains(l, "`r_")` (the backtick-prefixed shape a leaked id would actually take), and pinned the deadline test to exactly `numEntries` alias-only fallbacks.
- **Stale client doc:** dropped the "PR-3a.2 has not shipped / returns 404/400" disclaimer — that surface (slug on `ResourceData`, `GET /v1/resources/{id}`) has shipped.

Consciously **not** doing in this PR:

- **Degraded-row cue (`(unresolved)` marker):** the explicit goal here is "never surface the resource id," and the bare-alias fallback honors that. A distinct cue for "tunnel deleted upstream but binding lingers" is a real UX idea but a new behavior beyond this fix — better as its own change so it can be designed (404 vs 5xx vs ctx-cancel) and tested deliberately.
- **Legend covering the legacy-URL row:** legacy URL bindings are a rare pre-tunnels-only shape; the inline `(legacy URL)` label is self-describing, and widening the legend adds noise to the common tunnel case.
- **`/qurl list` still showing `r_…` for slug-less tunnels:** agreed the two surfaces shouldn't drift, but that's a separate change — worth a follow-up rather than folding it in here.

Issue #531, opened 2026-05-28
/qurl list: push type=tunnel filter server-side so has_more reflects tunnels (drop 100-row over-fetch)


Kevin-layerV:
Follow-up from PR #530 (cr review nits #2 and #5, and the /simplify altitude finding).

## Context
PR #530 made `/qurl list` list only tunnel resources. `GET /v1/resources` (qurl-service) exposes only `cursor`/`limit`/`slug` query params — **no `type` filter** — so the Slack handler filters to `type=tunnel` **client-side** after fetching one page at the server max (`listResourcesScanLimit = 100`). The over-fetch exists so tunnels aren't buried behind URL/transit resources on a small page.

## Problem (leaky abstraction)
`page.HasMore` is the **master-list** cursor flag — it means "the workspace has >100 resources of *any* type," not "more *tunnels* exist." The handler currently reuses it as a proxy for "more tunnels may exist," which over-claims:

- **Repro:** a workspace with ~150 URL resources and 3 tunnels, all 3 on the first page.
  - Admin `/qurl list` shows all 3 tunnels but appends `…more resources past the first 100-row scan — some tunnels may not be shown.` (tunnels are provably complete).
  - A non-admin whose channel has zero allowed tunnels on the first page sees the pagination-gap escalation copy (`…may exist past the first page — ask an admin…`) even when no tunnels exist at all.

The copy is hedged with "may," so it's not strictly false, but it's misleading in the common URL-heavy case. (Workspaces with ≤100 total resources are unaffected — `has_more` is false there.)

## Proposed fix (the right altitude)
1. Add a `type` query param to `GET /v1/resources` in qurl-service (OpenAPI + handler).
2. Add `Type string` to `client.ListResourcesInput` (qurl-integrations `shared/client`) and pass `?type=tunnel`.
3. In the Slack handler: request `type=tunnel`, drop the client-side `filterTunnelResources` pass and the 100-row over-fetch (back to a normal page size), and let `has_more` mean "more *tunnels* past this page" — which makes the footer + pagination-gap copy accurate.

## Acceptance criteria
- `GET /v1/resources?type=tunnel` returns only tunnel resources with a `has_more` that reflects tunnels.
- `/qurl list` footer / pagination-gap copy only appears when more **tunnels** exist past the page.
- The 100-row over-fetch (`listResourcesScanLimit`) is removed.
[...]

PR #554, opened 2026-05-29
fix(slack): resolve /qurl aliases slugs via ListResources, not by-id

Kevin-layerV:
## What

`/qurl aliases` now resolves each tunnel's slug the same way `/qurl list` does (via the workspace resource list), so it shows `$<slug> → $<aliases>` instead of bare aliases.

## Why this is a follow-up to #552

#552 stopped `/qurl aliases` from leaking the opaque resource ID, but it resolved each slug with `GET /v1/resources/{id}`. In the deployed qurl-service that **per-id detail path does not return the tunnel slug**, so every row fell through to the "no slug" alias-only fallback — the listing showed bare `$alias` lines with no slug at all:

```
• $ben-app
• $dashboard
• $kevin-dashboard, $kevin-test-dashboard
```

`/qurl list` shows these slugs correctly because it reads the workspace **list** endpoint (`GET /v1/resources`), which *does* return the slug. (Confirmed against the live sandbox: `/qurl list` renders `$kktest (aliases: $kevin-dashboard, $kevin-test-dashboard)` while `/qurl aliases` showed the bare aliases.)

## How

- **handler_aliases:** replace the per-id fanout with a single `ListResources` call (`resourcesByResourceID`), indexed by `resource_id` and joined to the channel's `alias_bindings` — the exact join `/qurl list` already uses. Removes `fanoutAliasGroups` and `aliasesResourceFanoutLimit`. The rendering (`$<slug> → $<aliases>`, never the resource ID; unresolved → alias-only) is unchanged from #552.
- **shared/client:** remove `GetResource` (added in #552, now unused — the per-id path it wrapped doesn't return the slug).
- **Tests:** mock the list endpoint via the existing `writeResourceListFixture`; drop the obsolete fanout/by-id tests and fixtures.

Net −409 lines.

Expected output:
```
Aliases configured for this channel:
Format: $<slug> → the aliases that resolve to it. Run /qurl get with the slug or any alias.
• $kktest → $kevin-dashboard, $kevin-test-dashboard
• $dashboard → ...
```

## Test

- `go test ./shared/client/ ./apps/slack/internal/` — pass.
- `golangci-lint` clean; pre-commit green.


Kevin-layerV:
Thanks for the review — addressed in 6779dd1.

- **Ask #1 (no test for the ListResources-failure → alias-only path):** added `TestHandleAliases_ListResourcesFailureDegradesToAliasOnly` — registers a `500` on `/v1/resources` and asserts every binding renders as a `• `` `$alias` `` line, the opaque `r_…` ID never leaks, and the header still renders (failure is non-fatal).
- **Ask #3 (per-invocation fetch-count invariant):** added an `atomic.Int32` counter to `TestHandleAliases_MultipleAliasesOneTunnelCollapse` asserting `/v1/resources` is hit exactly once per invocation — the structural replacement for the deleted fanout test's `fetches.Load() != 1` assertion, locking the "no accidental per-alias fanout" property.
- **Style nit (`map[string]*client.Resource` → `map[string]client.Resource`):** applied. The call site drops the `r != nil` check in favor of the zero-value `Resource` the formatter already degrades to, and no longer holds pointers into the `ListResources` page slice.

**Ask #2 (incomplete-listing footer when a bound tunnel is past the first page):** deferred to #555. Mirroring `/qurl list`'s footer today would surface the **master-list** `page.HasMore` signal — the exact leaky abstraction #531 exists to remove (it means ">100 resources of any type," not ">tunnels"). #555 depends on #531's server-side `type=tunnel` filter, after which a footer here is trivially accurate. The degradation is cosmetic in the meantime: the slug just doesn't render; `/qurl get $alias` still works. The `resourcesByResourceID` doc comment now points at #555.

Kevin-layerV:
Round 2 addressed in 1ddc7a7.

- **Issue #1 (doc-comment reflow):** reflowed both — `CreateInput.ResourceID` and the `UpdateResource` trim comment no longer wrap mid-clause.
- **Issue #2 (`HasMore` silently discarded):** implemented. `resourcesByResourceID` now returns `(map, page.HasMore)`; `processAliases` counts bindings whose `resource_id` wasn't on the scanned page and emits a single `log.Warn` when `page.HasMore && unresolved > 0` — one triage line for "why doesn't `$foo` show its slug?", arming #555. Purely additive; the rendered rows are unchanged. Fenced by `TestHandleAliases_IncompleteListingLogsTriageWarning` (fires) and `TestHandleAliases_CompleteListingLogsNoWarning` (does not fire when every binding resolves even with `has_more=true`).
- **Issue #5 (zero-value-as-degradation contract):** the call-site comment now spells out that both the nil-map (fetch failed) and missing-key (paginated out) lookups yield a zero-value `Resource` → alias-only.

Issues #3 (`listResourcesScanLimit` coupling) and #4 (ctx-cancel test removal) noted as non-blocking / correct — #3 is folded into #555's scope.

Kevin-layerV:
Round 3 addressed in 3b5d067 (all comment-only — no behavior change).

- **Nit #1 (redundant comment):** tightened the `resourcesByResourceID` doc so it no longer repeats the call-site block's "list vs per-id" rationale — the doc now carries just the mechanics + the `(nil,false)`/`hasMore` contract.
- **Nit #2 (`unresolved` zero-resources edge):** the warn-block comment now acks that the signal also covers a short/empty list returned with `has_more` set, not just genuine pagination past page 1.
- **Nit #3 (HappyPath fixture):** no change — the assertion pins the full rendered line (`` https://prod.example.com (legacy URL) → `$prod-db` ``), so renaming the alias would break it; it's already strong.
- **Product regression flag:** recorded explicitly on #555 — the by-id fanout could resolve a tunnel regardless of page position, so the >100-resource edge case is strictly worse UX; net-correct because the by-id path returned empty slugs anyway. #555 closes it properly once #531 makes `has_more` tunnel-accurate.

Kevin-layerV:
Round 4 addressed in 8240ea2.

- **Issue #1 (stale `listResourcesScanLimit` doc — the one ship-blocker):** updated. The doc now says the constant backs **both** `/qurl list` and `/qurl aliases`, that bumping it affects both, and that the #531 `has_more` "any resource type" caveat applies identically to the new `/qurl aliases` triage warning (cross-referenced from `processAliases`).
- **Coverage gap (partial-unresolved + complete page):** added `TestHandleAliases_PartialUnresolvedCompletePageNoWarning` — a channel binding both a live tunnel and a stale/deleted `resource_id`: the live row renders its slug, the stale row degrades to alias-only, neither `resource_id` leaks, and the triage warning does **not** fire (`has_more=false`). The helper now returns the rendered body too, so rendering + log content are asserted together.

Not taken (with rationale):
- **#3 (broaden `unresolved` to count present-but-slug-less resources):** intentionally declined. The warning means "the listing may be incomplete (paginated out)." A resource that's present on the page but has no slug/target is a data-quality issue, not a pagination one — counting it would make the "may be paginated out" warning fire for the wrong reason, i.e. *less* precise. The new test pins this boundary.
- **#2 (URL-heavy haystack reduces hit-rate):** real, and exactly what #555 fixes — once #531 pushes `type=tunnel` server-side, `/qurl aliases` drops the over-fetch and `has_more` becomes tunnel-accurate. Noted on #555.
- **#4 (extract `resourcesByResourceID` to `*Client`/shared):** you flagged it as not-asking; agreed — it's the right extraction point if a third caller appears, but two render-divergent callers don't justify it yet.

Kevin-layerV:
Round 5 addressed in 5024449.

- **Code quality #1 (skip `byID[""]` for resource-less groups):** adopted your suggested shape — `processAliases` now only does the lookup when `rid != ""`, making "resource-less groups never join" explicit at the call site.
- **Code quality #2 (document the no-`filterTunnelResources` choice):** the `resourcesByResourceID` doc now states it intentionally indexes every resource on the page (no `type=tunnel` filter, unlike `/qurl list`) so a legacy URL binding still resolves its `target_url` — flagged so nobody "fixes" it by adding the filter.
- **Coverage gap (resource-less group through `processAliases`):** added `TestHandleAliases_ResourceLessGroupRendersAliasOnly` — an empty-`resource_id` binding renders alias-only and does **not** trip the pagination warning even with `has_more=true` (it can't be paginated out).
- **Coverage gap (URL-resource indexing):** already pinned by `TestHandleAliases_HappyPath` (a `target_url` binding renders the `… (legacy URL) → $alias` line); the new doc one-liner documents the intent.

Verified, no code change:
- **Potential bug (error body in logs):** `parseError` puts the API's RFC-7807 `Detail` (or raw body on the non-standard fallback) into the error, and the fetch-failure `log.Warn(... "error", err)` does log it — but that's the established operator-log posture across these handlers (slog escapes attribute values; the user-facing wire reply is separately redacted to `serviceUnreachableMessage`). The new triage warning logs only `unresolved_groups/scan_limit/channel_id` — no error, no body.
- **>100-resource silent degradation:** acknowledged and tracked — `#531` (server-side `type=tunnel`) + `#555` (the `/qurl aliases` footer) close it; the triage warning is the interim operator signal.

Kevin-layerV:
Round 6 addressed in 5e31205.

- **Correctness #1 (useless upstream call for all-legacy channels):** `processAliases` now short-circuits the `ListResources` call when no group carries a `resource_id` (new `groupsNeedResolution` helper) — nothing to join, so no fetch.
- **Correctness #2 (triage warning omits `team_id`):** added `team_id` to the warning, matching the `GetChannelPolicy failed` warning earlier in the same function.
- **Test gap:** `TestHandleAliases_ResourceLessGroupRendersAliasOnly` now asserts the `/v1/resources` endpoint is hit **0** times, pinning the short-circuit.

The code-quality / security / perf observations (rationale comments, the `out` pre-size invariant, the preserved no-leak guarantee, workspace-scoped join) need no change — thanks for confirming them.

Kevin-layerV:
Round 7 addressed in cea8d01.

- **Nit #1 (`hasMore` conflates pagination vs stale binding):** the warning message now names both causes — "a bound resource was not on the scanned page (paginated out, or the binding is stale)" — so operators don't chase pagination when the real cause is a dangling DDB binding. The block comment spells out why both share the signal (master-list `has_more`, until #531).

Not taken (with rationale):
- **Nit #4 (rename `resourcesByResourceID`):** keeping the name. It deliberately parallels `channelAliasesByResourceID` in `handler_list.go` (the sibling "index-by-resource_id" helper you noted positively in an earlier round); renaming would break that consistency, and you rate the current name's cost as ~zero.
- **Nits #2, #3, #5:** no change — you flagged each as fine as-is (one caller so inlining the log helper is fine; the warning test is single-concern by design; `processAliases` is readable end-to-end). Agreed.

PR #557, opened 2026-05-29
refactor(slack): rename user-facing "slug" to "name" in command copy

Kevin-layerV:
## What

End users found **slug** confusing, so this renames it to **name** everywhere a Slack user sees it. Copy-only — no logic, grammar, or wire-contract changes.

Changed (user-visible copy):
- `/qurl` and `/qurl-admin` help text and the `tunnel install` usage block
- `/qurl get` / `set-alias` usage + validation errors (`tunnel name must be 3-64 chars…`, `Target must be a tunnel name…`)
- `/qurl list` footer (`$<name>` instead of `$<slug>`), empty-state, and the `(no name — ask your Slack admin to set one)` row
- the guided tunnel-install **modal** field label (`qURL tunnel name`) and hint
- "Tunnel `$x` was not found", "Failed to resolve tunnel name", and the per-tunnel install advisory notes
- the inline `#` comments in the Docker/Compose install snippets (the user pastes these), so they name the generated **per-tunnel** config/fragment consistently with the surrounding prose

Tests that assert this copy are updated in lockstep; `go test ./apps/slack/...`, `gofmt`, and `golangci-lint` all pass.

## What stayed `slug` (internal, intentionally unchanged)

- the `QURL_TUNNEL_SLUG` env var (the tunnel client reads it; it appears verbatim in the generated install blocks)
- the qurl-service API fields/query param (`slug`, `tunnel_slug`)
- Go struct fields, the Slack `block_id`/`action_id` (`tunnel_slug`/`slug_input`), log fields, the validation regex, and Go source comments

The command grammar is unchanged: `$name` is accepted and normalized exactly as `$slug` was, and positional `tunnel install <name>` behaves identically.

> Note: the repo READMEs that document this command grammar are renamed in a separate follow-up PR so this customer-facing copy can ship independently.

### DIFF
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -855,7 +855,7 @@ func (h *Handler) dispatchUserCommand(w http.ResponseWriter, command, text strin
 		// `$slug` or a channel `$alias`. Surface a deprecation hint
 		// instead of an "unknown subcommand" so existing users hitting
 		// muscle memory get a direct redirect to the new shape.
-		respondSlack(w, "`/qurl create` is no longer supported. Use `/qurl get <$name|$alias>` instead — run `/qurl list` to see your tunnels.")
+		respondSlack(w, "`/qurl create` is no longer supported. Use `/qurl get <$id|$alias>` instead — run `/qurl list` to see your tunnels.")
 	case text == "list":
 		// Exact match only: the looser `HasPrefix(text, "list")` form
 		// matched `listing`, `lists`, `list-foo` (silently routing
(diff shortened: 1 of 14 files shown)


Kevin-layerV:
Thanks for the thorough review passes. Disposition of every point, including nits:

**Fixed in c5a39c0 (pushed):**
- **Heredoc comments still said `per-slug`** (`handler_tunnel_docker.go:32-33`, `handler_tunnel_compose.go:64`). These shell-script comments are pasted/read by the user, and the matching prose was already `per-tunnel`, so they're swept now: `per-tunnel config` / `per-tunnel Compose fragment` and "deterministic name and port values".
- Also tidied a tautology the mechanical rename introduced: help text `\`$name\` is a tunnel's name` → `\`$name\` identifies a tunnel`, matching the list-footer phrasing.

**Deferred to #559 (genuinely a separate concern):**
- **`<name>` placeholder collision** in `tunnel install` usage/help — the leading tunnel-name slot (`<name|$name>`) now shares the `<name>` token with `container:<name>`/`service:<name>`. A consistent fix redesigns placeholder copy across `handler_tunnel.go`, `handler.go`, `handler_interaction.go`, and `handler_tunnel_compose.go` (+ tests) — copy that was never "slug", and doing it half-way introduces fresh inconsistency (e.g. `<tunnel-name>` in install but `$<name>` in set-alias for the same concept). Filed as #559 with both disambiguation options, sites, and acceptance criteria.
- **Internal stragglers** (`handler_alias_test.go` case names + adjacent comment; `handler_get.go` "shows both slugs and aliases" doc comment) — folded into #559 as a low-priority item. #557 intentionally scoped to user-facing copy; these are internal.

**Intentional, no change (with rationale):**
- **Footer "identifies a tunnel" vs old "is a tunnel's name":** the old wording is tautological after the rename (`$name` *is a tunnel's name*), which is exactly why both the footer and the help text use "identifies a tunnel" instead. It's clear and non-circular; the concept is also introduced in adjacent help copy.
- **Prose-says-name / env-says-`SLUG` juxtaposition:** inherent to keeping `QURL_TUNNEL_SLUG` literal (per `CLAUDE.md`), as you noted — not a blocker.
- **"Tunnel `$x` was not found"** dropping "slug" entirely (vs swapping to "name"): intentional — it reads more naturally and the `tunnel install %s` hint that follows uses the bare token, matching what `parseTunnelInstall` accepts.

CI + claude-review are green on c5a39c0.

PR #558, opened 2026-05-29
docs(slack): rename user-facing "slug" to "name" in READMEs

Kevin-layerV:
## What

Follow-up to the customer-facing copy rename (slug → name). Updates the command-grammar documentation so the READMEs match what Slack users now see:

- `README.md` — "customer-facing tunnel **name**", "per-tunnel durable agent-state directory"
- `apps/slack/README.md` — `/qurl get <$name|$alias>`, `set-alias $<alias> $<name>`, `tunnel install <name|$name>`, and the prose ("tunnel name", "the name doubles as the channel shortcut", etc.)

## Kept internal

The `QURL_TUNNEL_SLUG` env var name is unchanged — only its placeholder value now reads `<name>` (`QURL_TUNNEL_SLUG=<name>`).

## Merge order — depends on #557

The runtime copy rename lives in #557 (`refactor(slack): rename user-facing "slug" to "name" in command copy`), which flips the `/qurl help`, `/qurl-admin help`, alias-error, and `views.go` modal-label strings the bot actually emits.

**This docs PR must merge after #557**, not before. If it lands first, the README will say `$name` while a user running `/qurl help` still sees `$slug` — two vocabularies for one concept. #557 is split out so it can merge and ship on its own without waiting on doc review; this PR then follows to keep the docs in lockstep. No temporary "the bot may still say slug" disclaimer is added to the README, because it would become stale debt the moment #557 ships.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,9 +42,9 @@ All integrations connect to the qURL API. The endpoint is configurable via the `
 
 Customer onboarding is install-first: install the qURL Slack app, run `/qurl setup`, then use `/qurl-admin tunnel install` or `/qurl get`. The Slack app install stores the workspace bot token for modal support; customers do not manually provide a bot token. Admin verbs live under a separate `/qurl-admin` slash command (registered against the same request URL as `/qurl`); user verbs — including `setup` — stay on `/qurl`.
 
-Slack admins can run `/qurl-admin tunnel install` to generate a guided install for the qURL reverse-tunnel sidecar. The workflow asks for the customer-facing tunnel slug, an optional channel shortcut, the local HTTP port, and the target runtime. Slack then returns copy/paste output tailored for Docker, Docker Compose, AWS ECS/Fargate, or Kubernetes.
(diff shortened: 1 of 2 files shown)


Kevin-layerV:
Thanks — confirmed the merge-order concern and resolving it via **option 1 (coordinate merge order)** rather than the parenthetical.

The runtime rename you flagged (`views.go:170/172` modal labels, the `handler.go` help/error bodies, `handler_alias.go` usage strings) is all done in **#557**, which is `MERGEABLE` with every check green. So both halves of the rename are ready; the only thing left is ordering.

I did **not** add a "the bot may still call this a slug" disclaimer to the README (your option 2): it would be correct for only the window between these two merges and would itself need a cleanup PR the moment #557 ships. Instead I made the dependency explicit in the PR body — **this docs PR must merge after #557** — so the human merging respects the order. Both are tiny and ready, so the intended path is #557 first, then #558 right after.

No README content change needed: the prose is already correct for the post-#557 world.

Issue #199, opened 2026-05-08
discord: Revoke and Add-Recipients are not mutually exclusive


vikramlayerv:
**Tracking item from #198 pr-polish (round-18 #5)**.

Pre-existing concurrency: a user can have an Add Recipients select-menu pending and click Revoke on the original `/qurl send` confirmation. New recipients written by a post-Revoke `mintLinksInBatches` would dodge revocation.

`revokeInFlight` (single-button-click dedup) and `addRecipientsLocks` (per-send Add semaphore) don't cross-block. Not regression-introduced by #198 — flagging here so the user-centric revoke shape from #198 doesn't outpace the underlying concurrency contract.

**Fix shapes** (pick whichever fits the broader `/qurl send` collector lifecycle):
- Cross-block: `revokeInFlight` checks `addRecipientsLocks` (and vice versa); whichever hits first short-circuits the other with a "try again" message.
- Sequence: queue Add behind in-flight Revoke (Add becomes a no-op if Revoke completes first).
- Disable the buttons: when one is in-flight, dim the other in the components row.

**Acceptance**: unit test that simulates both clicks racing → asserts the post-Revoke count + names list matches what was sent (no orphan recipients silently surviving the revoke window).

PR #561, opened 2026-05-31
feat(slack): set link/session/max-session limits on tunnel qURL mints

justin-layerv:
## What

Every `/qurl get` mint (tunnel-only — raw URLs are unsupported) now sets three access limits, bounding a shared tunnel link to a single short-lived viewer:

| field | value | meaning |
|---|---|---|
| `expires_in` | `1m` | link only admits a **new** visitor session for 60s after minting |
| `session_duration` | `1h` | an admitted visitor session lasts up to 1 hour |
| `max_sessions` | `1` | one concurrent visitor session |

Also wires `SessionDuration` through `shared/client` (`CreateInput` + `createForResourceBody`) — it was missing — mapping to qurl-service's `session_duration` on both create wire shapes.

## Why

Tunnel qURLs are shared links to a customer's exposed service. Without limits, a link is reusable and concurrently accessible indefinitely until revoked. These bound exposure to **one viewer, for one hour, started within one minute of minting**.

## ⚠️ Deploy-order dependency (not a flag) — qurl-service#778 first

Enforcement is entirely **server-side** (qurl-service + qurl-router). This PR only sets the policy at mint time. The *pre-#778* qurl-service API **rejects** `max_sessions`/`session_duration` on tunnel resources with **400**, so this MUST deploy **after** qurl-service#778 is live in prod — else every `/qurl get` 400s. This PR is **ready** (cr-approved, all green, the #778 string-duration create contract is verified); merge/deploy it only once #778 is deployed. There is **no feature flag** — #778's session enforcement is unconditional on deploy.

- Per-visitor identity is **IP-based** in the interim; cookie follow-up tracked in **layervai/qurl-service#777**.

## Tests

- `apps/slack` — `TestCreateInputJSON_TunnelSessionLimits`: fences that the mint wire body carries `expires_in:1m`, `session_duration:1h`, `max_sessions:1` end-to-end through the resource-scoped create.
- `shared/client` — `TestCreateSessionDurationOnWire` (both wire shapes) + `TestCreateSessionDurationOmittedWhenEmpty` (omitempty contract).
- Full `shared/client` + `apps/slack/internal` suites pass; `go vet` clean.

### DIFF
--- a/apps/slack/internal/handler_get.go
+++ b/apps/slack/internal/handler_get.go
@@ -52,6 +52,10 @@ const (
 	tunnelLinkExpiry      = "1m"
 	tunnelSessionDuration = "1h"
 	tunnelMaxSessions     = 1
+	// tunnelLinkExpiryHuman is the user-facing rendering of tunnelLinkExpiry
+	// for the Slack reply — "1 minute" reads clearer to a recipient than the
+	// terse "1m" duration syntax. Keep in sync with tunnelLinkExpiry above.
+	tunnelLinkExpiryHuman = "1 minute"
 )
 
 // urlNotSupportedGetMessage is the user-facing copy for a raw-URL
@@ -404,7 +408,7 @@ func (h *Handler) getWork(ctx context.Context, log *slog.Logger, args getWorkArg
 	// minting. That admit window is tight, so surface it at the point of
 	// sharing: a recipient who clicks after it lapses gets a dead link, and
 	// the suffix tells them why.
-	message := ":link: *qURL ready:* " + out.QURLLink + " (one-time use · link expires in " + tunnelLinkExpiry + ")"
+	message := ":link: *qURL ready:* " + out.QURLLink + " (one-time use · link expires in " + tunnelLinkExpiryHuman + ")"
 	if args.cmd.DM() {
 		return h.deliverGetDM(ctx, log, args.userID, message), nil
 	}
--- a/apps/slack/internal/handler_get.go
+++ b/apps/slack/internal/handler_get.go
@@ -42,6 +42,12 @@ const commonGetMintFailedMessage = "Failed to mint qURL. Please try again."
 // session-limit support to be deployed (otherwise create returns 400 —
 // hence the gating in the PR). Per-visitor identity is IP-based for now;
 // layervai/qurl-service#777 tracks the cookie follow-up.
+//
+// As of qurl-service#778 these values are ALSO the server-side tunnel
+// defaults (session_duration→1h, and one_time_use→single-visitor when
+// max_sessions<=1), so setting them explicitly here is belt-and-suspenders:
+// it pins the bot's intent on the wire and decouples it from any future
+// change to the server defaults, rather than being load-bearing.
 const (
 	tunnelLinkExpiry      = "1m"
 	tunnelSessionDuration = "1h"
--- a/apps/slack/internal/handler_get.go
+++ b/apps/slack/internal/handler_get.go
@@ -19,6 +19,27 @@ import (
 // because three different mapMintError branches need it.
 const commonGetMintFailedMessage = "Failed to mint qURL. Please try again."
 
+// Tunnel access limits applied to every `/qurl get` mint. `/qurl get` is
+// tunnel-only (raw URLs are unsupported — see urlNotSupportedGetMessage), so
+// these bound a shared tunnel link to a single short-lived viewer:
+//   - tunnelLinkExpiry: the link only admits a NEW visitor session for this
+//     long after minting (qurl-service `expires_in`).
+//   - tunnelSessionDuration: how long an admitted visitor session lasts
+//     (qurl-service `session_duration`).
+//   - tunnelMaxSessions: max concurrent visitor sessions (qurl-service
+//     `max_sessions`).
+//
+// Enforcement is entirely server-side (qurl-service + qurl-router); this just
+// sets the policy at mint time and requires the qurl-service tunnel
+// session-limit support to be deployed (otherwise create returns 400 —
+// hence the gating in the PR). Per-visitor identity is IP-based for now;
+// layervai/qurl-service#777 tracks the cookie follow-up.
+const (
+	tunnelLinkExpiry      = "1m"
+	tunnelSessionDuration = "1h"
+	tunnelMaxSessions     = 1
+)
+
 // urlNotSupportedGetMessage is the user-facing copy for a raw-URL
 // `/qurl get`. The parser flags the case with the terse
 // [ErrURLNotSupportedGet] sentinel; this is the rich reply the handler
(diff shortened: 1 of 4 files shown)


justin-layerv:
Server-side dependency is now open: qurl-service #778 makes tunnel qURLs accept + enforce `session_duration`/`max_sessions` (and `access_policy`). Un-draft this once #778 is merged + deployed.

justin-layerv:
Addressed cr round 1 in 2ffb5c19:
- **Q2 (surface the expiry)** — the reply is now `(one-time use · link expires in 1m)` so the tight admit window is visible at share time; added an assertion that the window is surfaced (mirrors the one-time-default test).
- **Q3 (one_time_use vs max_sessions stacking)** — added a const-block comment: OneTimeUse burns the link on first redemption, so `max_sessions=1` is belt-and-suspenders (explicit "one viewer" on the wire, and stays correct if the one-time default is ever relaxed). Keeping it since it's an explicit product requirement and harmless.
- **Q1 (1-minute window)** — deliberate, product-confirmed. The admit window is intentionally tight; Q2's visible expiry suffix mitigates the late-click confusion.
- **Minor (hardcoded constants)** — no per-team/plan variance anticipated yet; the const block is the natural seam if that changes. No TODO added to avoid a speculative marker.

justin-layerv:
### `/simplify` pass — clean. No reuse/simplification/efficiency/altitude findings; the SessionDuration plumbing, the const block, and the string-concat suffix are all idiomatic and minimal.

justin-layerv:
On the load-bearing semantics question (admit-window vs total-lifetime): **confirmed correct** against the qurl-service implementation (#778). `expires_in` maps to the qURL token's `ExpiresAt`, which is checked at **resolve** time (`loadAndValidate` → token expiry) — i.e. it bounds when a *new* session can be admitted. `session_duration` is enforced separately at **authorize** time via `FirstAuthorizedAt + session_duration`, independent of the token's `ExpiresAt`. So a visitor admitted at 0:59 keeps their session until ~1:59 — the 1m window does NOT cap the 1h session. The two are orthogonal, exactly as the `CreateInput.SessionDuration` doc states.

On the 1-minute window: deliberate, product-confirmed (kept). The suffix surfaces it. An e2e that admits late-in-window and checks the session survives past link expiry would be a nice belt-and-suspenders once the stack is deployed together — I can add it to the smoke suite after #778 lands.

justin-layerv:
Thanks — addressed #2 and #4a in `9179225`:

**2. QURL response struct gains `SessionDuration`** — added it (seconds), symmetric with `MaxSessions`/`ExpiresAt`, so a created/fetched qURL can surface its session duration (the admin "show limits" case you flagged). `CreateInput` already sent it; read-back was silently invisible.

**4a. "1m" → "1 minute"** — the Slack reply now renders the admit window as "1 minute" (`tunnelLinkExpiryHuman`) for recipient clarity; the wire value (`expires_in`) stays `"1m"`. Tests split accordingly.

**1. Deploy-ordering (unconditional limits → 400 if integrations ships first)** — real sharp edge. The **draft** gates the merge, and the rollout sequence (#778 deploy → #199 → flip flag → **then** un-draft #561) gates the deploy. I considered an env-gated fallback so the repos could deploy independently, but the bot is intentionally simple and the coordinated sequence is the operational plan — un-drafting #561 only after #778 is live in the target env. Flagged in the PR description.

**3. `max_sessions`/`omitempty` (0=unlimited future)** — noted; `tunnelMaxSessions=1` is non-zero so it reaches the wire today. If `0`-means-unlimited is ever needed it'd need a `*int` (mirrors qurl-service#779).

**4b. 60s admit window friction** — product-confirmed and deliberate; re-flagged for the record. If real-world late-click friction shows up we can revisit the 1m default.

Staying draft until qurl-service#778 deploys.

justin-layerv:
Thanks — addressed in `2a0e95f`:

**1. SessionDuration read-back decode risk** — dropped the `QURL.SessionDuration int` field I'd added. You're right that it's the higher-risk move: the field is speculative (no consumer yet), and an `int` field couples the **whole** `QURL` decode to session_duration's wire type — a string-typed echo would fail `json.Unmarshal` for the entire struct, not just that field. (I did verify qurl-service serializes it as a JSON **number** today — `QurlSummary.session_duration` is `*int` in the generated types, distinct from the `*string` *input* on `CreateQurlRequest` — but the field had no consumer to justify the coupling.) Left a comment to add it WITH a decode test against the confirmed number shape when an actual consumer (e.g. an admin "show limits" view) needs it. Net: back to the bot safely ignoring the field on read.

**3. Stale test doc-comments** — updated both (`DMRidesOneTimeSuffix`, `OneTimeDefault`) from the old `(one-time use)` to the full `(one-time use · link expires in 1 minute)` suffix; the assertions were already correct, only the prose had drifted.

**2. 60s admit window** — product-confirmed and deliberately kept; the "link expires in 1 minute" copy is the mitigation. Re-flagged for the record; if real-world late-click friction shows up we can revisit the 1m default.

Staying draft until qurl-service#778 deploys.

justin-layerv:
Thanks — addressed in `43152f5`:

**2. Const sync untested** — added `TestTunnelLinkExpiryConstsInSync`, a tripwire pinning the `tunnelLinkExpiry` ("1m", wire) ↔ `tunnelLinkExpiryHuman` ("1 minute", copy) pair. A future bump to one that forgets the other now fails the test, so the user-facing window can't silently diverge from `expires_in`.

**3. Heavy deferred-field comment** — trimmed from 6 lines to 4 and pointed at qurl-service#778 for the wire shape rather than restating the full string-vs-number rationale inline.

**1. 60s admit window vs Slack read-latency** — agreed it's the substantive one; product-confirmed and deliberate, with the "link expires in 1 minute" suffix as the mitigation. Re-flagged for the record — if real-world late-click friction shows up post-launch we can revisit (e.g. lengthen the admit window, or the "start the window at first reach" idea you raised). I'll sanity-check it against real Slack usage before un-drafting, alongside confirming qurl-service#778 is deployed (the 400 risk).

Staying draft until #778 deploys.

justin-layerv:
Thanks — and I verified the gating concern (#1) directly against qurl-service#778:

**1. session_duration string-duration acceptance — CONFIRMED.** #778's `CreateQurlRequest.session_duration` (and the for-resource variant) is a `*string`, and the create handler validates it via `ValidateSessionDuration` → `domain.ParseDuration`, i.e. it parses Go-style duration strings ("15m"/"1h"/"8h" — the error message literally suggests those). So the bot sending `session_duration: "1h"` (and `expires_in: "1m"`) is exactly the shape #778 accepts on create — no string-vs-number mismatch, no 400. (The response serializes it back as a *number*, which is why I dropped the speculative int read-back field last round rather than risk the decode coupling.)

So the precondition you flagged is satisfied; this is good to un-draft once #778 is **deployed** (the only remaining gate is the deploy, not the contract).

**3. Comment density (belt-and-suspenders 3×)** — leaving as-is per your lean; the rationale is genuinely non-obvious (belt-and-suspenders vs load-bearing + the post-#778 server-default interaction), so I'd rather keep it explicit than consolidate-and-indirect. I did trim the deferred-read-back comment last round.

**2 / 4. 1m window + max_sessions omitempty** — window is product-confirmed (will sanity-check against real Slack latency before un-draft); `max_sessions=1` is non-zero so omitempty is fine today, noted as a latent footgun if it ever becomes 0.

Staying draft until #778 deploys.

justin-layerv:
**Un-drafted — ready to merge.** cr is satisfied (the gating concern, qurl-service#778 accepting `session_duration` as a Go-style duration string on create, is verified), all checks green, `/simplify` clean.

⚠️ **One hard deploy-order rule (not a flag):** qurl-service#778 MUST be deployed to prod **before** this ships. The bot sends `session_duration`/`max_sessions`/`expires_in` on tunnel creates, which the *pre-#778* API rejects with 400 — so deploying this first would 400 every `/qurl get`. Merge+deploy #778 first, then this. (If qurl-integrations auto-deploys on merge, hold the merge until #778 is live in prod.)

PR #564, opened 2026-06-01
fix(slack): accept $-prefixed id on display-name verbs; richer bare-verb hints

Kevin-layerV:
## What
- `set-display-name` / `unset-display-name` now accept an **optional** leading `$` on the tunnel id, so the `$<id>` form `/qurl list` prints works alongside the bare form (`$prod-dashboard` and `prod-dashboard` both resolve). Previously only the bare form was accepted, so a user who copied the `$<id>` straight out of `/qurl list` hit a confusing rejection.
  - Note: `/qurl get` and `/qurl-admin set-alias` *require* the `$` (a missing sigil is `ErrMissingSigil`). These display-name verbs are deliberately the **lenient superset** — sigil optional — not strict parity. The point is to accept the form `/qurl list` shows, not to mirror those verbs' strictness.
- Bare `/qurl get` (no token) now replies with the `$<id>|$<alias>` usage hint instead of the terse `missing or empty $alias argument`.
- Bare `/qurl-admin admin` (no action) now lists the available actions (add / remove / list / revoke) instead of `missing admin action`.

## Why
Command-UX papercut pass. Accepting only the bare id, and the terse bare-verb errors, were footguns: a user who typed the id the way `/qurl list` shows it, or ran a verb to discover its arguments, hit confusing errors instead of guidance.

## Notes
- `$` is stripped before the slug-pattern check rather than widening `tunnelSlugPattern` (which the slug grammar shares with install). The strip + empty-check + slug-validate path is shared by both display-name verbs via the `parseDisplayNameID` helper.
- The invalid-id error echoes the stripped id (e.g. `$Prod` → `` `Prod` `` ), matching `parseAliasToken`, which echoes its post-`$` value.
- The bare-`admin` hint says "qURL bot admin", not "workspace admin" — membership is the bot's own admin set, not Slack's workspace-admin role.

## Tests
- Parser unit tests cover `$`-prefixed and lone-`$` ids for both display-name verbs.
- Handler tests updated to assert the richer bare-`get` and bare-`admin` hints.
- `go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` reports 0 new issues.

### DIFF
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1276,8 +1276,9 @@ func (h *Handler) adminHelpMessage(command string) string {
 		lines = append(lines,
 			"• `/qurl-admin set-display-name <id> <display name>` — Set a tunnel's friendly Display Name shown in `/qurl list` (admin only)",
 			"• `/qurl-admin unset-display-name <id>` — Reset a tunnel's Display Name to the default (admin only)",
-			// admin add/remove/list/revoke: keep this roster in sync with
-			// adminUsageMessage (handler_admin.go), the bare-`admin` arg hint.
+			// admin add/remove/list/revoke: keep this action set in sync with
+			// adminUsageMessage (handler_admin.go), the bare-`admin` arg hint —
+			// the verb roster must match, not the prose.
 			"• `/qurl-admin admin add @user` — Promote a Slack user to bot admin (admin only)",
 			"• `/qurl-admin admin remove @user` — Demote a Slack user from bot admin (admin only)",
 			"• `/qurl-admin admin list` — List who connected qURL (the owner) and the current bot admins (admin only)",
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1276,6 +1276,8 @@ func (h *Handler) adminHelpMessage(command string) string {
 		lines = append(lines,
 			"• `/qurl-admin set-display-name <id> <display name>` — Set a tunnel's friendly Display Name shown in `/qurl list` (admin only)",
 			"• `/qurl-admin unset-display-name <id>` — Reset a tunnel's Display Name to the default (admin only)",
+			// admin add/remove/list/revoke: keep this roster in sync with
+			// adminUsageMessage (handler_admin.go), the bare-`admin` arg hint.
 			"• `/qurl-admin admin add @user` — Promote a Slack user to bot admin (admin only)",
 			"• `/qurl-admin admin remove @user` — Demote a Slack user from bot admin (admin only)",
 			"• `/qurl-admin admin list` — List who connected qURL (the owner) and the current bot admins (admin only)",
--- a/apps/slack/internal/handler_admin.go
+++ b/apps/slack/internal/handler_admin.go
@@ -20,6 +20,10 @@ import (
 // user learns the grammar rather than seeing the terse sentinel. "qURL bot
 // admin" (not "workspace admin") is deliberate: membership is the bot's own
 // admin set, not Slack's workspace-admin role.
+//
+// Keep this action roster in sync with the `/qurl-admin help` listing in
+// handler.go — the two are maintained independently and would otherwise drift
+// when an admin action is added or renamed.
 const adminUsageMessage = "Usage:\n• `/qurl-admin admin add @user` — promote a Slack user to qURL bot admin\n• `/qurl-admin admin remove @user` — demote a qURL bot admin\n• `/qurl-admin admin list` — show who connected qURL (the owner) and current bot admins\n• `/qurl-admin admin revoke <qurl_id>` — revoke a single qURL"
 
 // handleAdmin parses the `admin <verb> ...` form via the shared parser
--- a/apps/slack/internal/handler_admin.go
+++ b/apps/slack/internal/handler_admin.go
@@ -14,6 +14,14 @@ import (
 	"github.com/layervai/qurl-integrations/shared/client"
 )
 
+// adminUsageMessage is the arg hint shown when `/qurl-admin admin` is
+// invoked with no action (bare `admin`, which parses to
+// [ErrMissingAdminAction]). Lists the membership + revoke actions so the
+// user learns the grammar rather than seeing the terse sentinel. "qURL bot
+// admin" (not "workspace admin") is deliberate: membership is the bot's own
+// admin set, not Slack's workspace-admin role.
+const adminUsageMessage = "Usage:\n• `/qurl-admin admin add @user` — promote a Slack user to qURL bot admin\n• `/qurl-admin admin remove @user` — demote a qURL bot admin\n• `/qurl-admin admin list` — show who connected qURL (the owner) and current bot admins\n• `/qurl-admin admin revoke <qurl_id>` — revoke a single qURL"
+
 // handleAdmin parses the `admin <verb> ...` form via the shared parser
 // and dispatches to the action-specific handler. Every recognized verb
 // is sync — one DDB GetItem/UpdateItem or one qURL API DELETE — so the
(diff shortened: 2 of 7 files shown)


Kevin-layerV:
Thanks — both points addressed:

**Consistency framing.** You're right that `/qurl get` and `set-alias` *require* the `$` (`ErrMissingSigil`) while these verbs make it optional, so it's a lenient superset, not strict parity. Corrected the PR description and tightened the `parseDisplayNameID` doc comment (`ba92fa4`) to call out that the sigil is optional here and that the goal is accepting the form `/qurl list` prints, not mirroring those verbs' strictness.

**Invalid-id echo (nit).** Keeping the stripped echo (`echoText(id)`) on purpose — it matches the canonical convention: `parseAliasToken` (parser.go:646,648) also echoes its *post-`$`* value (`truncateForError(alias)`), not the raw `$`-token. Echoing the raw token here would make display-name the lone outlier. Documented the rationale in the same comment so it's a conscious choice rather than an accident.

Also folded out the strip+validate duplication the branch introduced into the shared `parseDisplayNameID` helper (`e9f6654`) — `go test ./apps/slack/...` green, `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Both non-blocking notes addressed in `6561df1`:

1. **`echoText` vs `truncateForError`.** Tightened the `parseDisplayNameID` comment to say the invalid-id echo matches `parseAliasToken` on the *value* (the post-`$` id), and to call out explicitly that the Slack-escaping helper differs (`echoText` here, `truncateForError` there) — so a future reader doesn't expect identical formatting.

2. **End-to-end `$`-coverage.** Added handler-level smoke tests for both verbs (`TestSetDisplayName_DollarPrefixedID`, `TestUnsetDisplayName_DollarPrefixedID`) that drive the full path — parser → admin gate → resolve → PATCH. They assert the `$` is stripped, the bare slug resolves the tunnel, the `description` is PATCHed (`calls == 1`), and the id echo carries no `$`.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Thanks — addressed in `28893fc`:

1. **`/qurl get $` (lone sigil).** Good catch that this is a distinct parse path (`parseAliasToken` → `ErrEmptyResource`) from bare `get`. Added `TestHandleGet_LoneSigil` asserting `get $` hits the `getUsageMessage` hint (`$id|$alias` + `/qurl list`) and never kicks off async work.
2. **Long single-line consts.** Leaving as-is per your note — consistent with the existing style in these files.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Addressed in `6b27d06`:

1. **`unset $` message pinned.** Added `wantMsgSub` to the `TestParseUnsetDisplayNameArgs` table so each rejection path's copy is locked: lone `$` → "Missing tunnel id" (the shared `parseDisplayNameID` path), `""`/trailing-args → "Provide exactly one tunnel id" (the arity path), invalid slug → "valid tunnel id". The divergence is intentional and now drift-fenced.
2. **Comment density / parity story.** Keeping per your recommendation — the three surfaces serve different readers (helper comment = code rationale, `displayNameUsage` = end-user copy, PR body = reviewer context), so it's not collapsible duplication.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Thanks for the thorough pass.

- **lone-`$` vs no-token copy.** Intentional and test-pinned, as you noted — leaving the friendlier "Missing tunnel id" for the lone-`$` path.
- **`parseDisplayNameID` comment length.** You've flagged this a few rounds running, so I trimmed it from ~13 lines to 8 (`1a53836`), keeping the load-bearing whys (lenient-superset framing, strip-beats-widening-`tunnelSlugPattern`, stripped-id echo matching `parseAliasToken`) and dropping the restated detail.

Re-ran `/simplify` over the final diff: reuse / efficiency / altitude all clean, and it approved the trimmed comment. No further changes outstanding.

Kevin-layerV:
Both addressed in `2c80052`:

1. **Echo escaper precision.** Took your suggested framing verbatim — the comment now reads "matching parseAliasToken on the value (the Slack escaper differs)", so it's honest about value-match-but-not-escaper while staying a short parenthetical.
2. **Roster sync.** Confirmed: `adminUsageMessage` lists `add` / `remove` / `list` / `revoke <qurl_id>`, which matches the `/qurl-admin help` text at `handler.go:1279-1282` (same four actions, same `<qurl_id>` arg). In sync.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Thanks — addressed in `fd5d346`:

- **Bare-`$` (no name) coverage.** Added the `"$"` case to the set parser table, pinning that the empty-id check fires before the missing-name check (so it returns "Missing tunnel id", not "Missing Display Name"). The `"$ Prod API"` case already covered the `found=true` variant; this completes the pair.
- **`$$prod` single-strip.** Agreed — intentional and matches `parseAliasToken`; leaving as-is.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Addressed in `60cff49`:

- **Double-`$` test.** Added `$$prod-dashboard` → strips to `$prod-dashboard` → fails the slug check, pinning the single-strip contract the comment claims (matching `parseAliasToken`).
- **Lone-`$` set behavior shift** — conscious change, already covered by `lone dollar id rejected` / `lone dollar no name rejected`. ✅
- **Comment density** — leaving as house-consistent per your note.

That closes out the sigil input-space coverage (bare `get`, `get $`, `$`-prefixed set/unset, lone `$`, lone `$` no-name, `$`+invalid, `$$`). `go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Addressed in `b10ac9b`:

1. **Roster drift.** Added a `keep in sync` cross-reference comment on both `adminUsageMessage` (handler_admin.go) and the `/qurl-admin help` listing (handler.go), so a future editor adding/renaming an admin action updates both rosters.
2. **Invalid-id echo drops `$`** — deliberate parity with `parseAliasToken`, acceptable as you note. Leaving as-is.
3. **`unset $ <id>` → "Provide exactly one tunnel id"** — edge-of-edge `strings.Fields` tokenization; not worth special-casing, leaving as-is.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Addressed in `97ce510`:

1. **Ungated bare-`admin` on no-DDB.** Kept the behavior (returning not-configured first would contradict the file's deliberate "parser feedback is useful regardless of wiring" design), but fixed the comment: it claimed the roster "is in `/qurl-admin help`", which isn't true on a no-DDB deploy where help gates its admin lines behind AdminStore. The comment now documents that the discovery hint stays shown there while help hides them, and that the divergence is deliberate.
2. **Sync-comment scope.** Tightened both cross-references to "keep the action set (add/remove/list/revoke) in sync, not the prose" — so a future editor doesn't over-correct toward identical wording (help capitalizes + appends "(admin only)").
3. **`getUsageMessage` sharing** — thanks, leaving as-is.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

Kevin-layerV:
Done — `3b23323` adds `TestAdminRosters_InSync`, the roster-sync guardrail you suggested. Rather than just asserting both contain `{add, remove, list, revoke}`, it extracts the `admin <verb>` set from each roster (`adminUsageMessage` and the rendered `adminHelpMessage`) and asserts the two sets are *equal* — so a future 5th action added to one but not the other fails CI, and so does dropping one. The comment is now backed by an enforced invariant.

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` 0 issues.

PR #566, opened 2026-06-01
feat(slack): admin Edit button + edit modal on /qurl list

Kevin-layerV:
## What
Adds a per-row **Edit** button to `/qurl list`, shown only to **qURL bot admins** (the ones managed by `/qurl-admin admin add` — not Slack workspace admins), beside *Create qURL*. Tapping it opens a modal to rename the tunnel's **Display Name** and **configure its channel aliases**, both pre-filled with the current values.

## How it works
- **Visibility:** the list reply is ephemeral (only the caller sees it), so rendering the Edit button is gated on the caller being a bot admin (`CheckAdmin`) with the modal/alias wiring present. Non-admins see only *Create qURL*; admins see *Create qURL* + *Edit* in an `actions` block.
- **Pre-fill without a read:** the Edit button carries a small JSON snapshot (resource id, token, current Display Name, current extra aliases) so the modal opens with `views.open` immediately, inside Slack's ~3s trigger window — no extra API round-trip. (Snapshot is bounded to Slack's 2000-char button-value cap; an over-cap row falls back to a Create-only button.)
- **Aliases box:** a multiline field pre-loaded with the tunnel's *extra* channel aliases — i.e. every bound alias except the row's primary `$<token>`, which is the tunnel's canonical name and is managed automatically (never offered for removal here). One alias per line, `$` optional.
- **Submit reconciles:** the submitted alias set is authoritative — new lines bind, removed lines unbind, and an alias already used by a *different* tunnel in the channel is reported and skipped. The Display Name PATCHes **only when changed**, so an alias-only edit can't clobber a concurrent rename.

## Security posture
- **Opening** the modal isn't admin-re-gated: it shows only what's already visible via `/qurl list` + `/qurl aliases`.
- **Mutating** is gated: the `view_submission` re-checks `CheckAdmin` (with team/user cross-checks against modal replay) before any PATCH or alias write — same posture as the tunnel-install modal.
- Display Names run through the **same character fence** as `/qurl-admin set-display-name` (no backticks / angle brackets / control bytes — they're stored in `description` and rendered to every member). That fence is now extracted into `validateDisplayNameChars` and shared by both the slash verb and the modal.
- Alias reconcile is **last-write-wins** on the submitted set; two admins editing the same tunnel's aliases concurrently is an accepted, documented rare race for an admin tool.

## Tests
Snapshot building (token excluded, over-cap → fallback), alias-line parsing (`$` optional, dedup, token dropped, too-many), modal-arg validation (required name, mrkdwn-injection rejected, bad alias), list render (Edit shown for admin, hidden without wiring + Create preserved), modal open (pre-filled with name + aliases; unparseable value → error, no `views.open`), and submission (happy path: PATCH + bind/unbind + summary, verified against the alias store; non-admin → denied, nothing mutated).

`go test ./apps/slack/...` green; `golangci-lint --new-from-rev=origin/main` reports 0 new issues.

### DIFF
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1334,6 +1334,10 @@ const (
 	respFieldResponseType = "response_type"
 	respFieldText         = "text"
 	respTypeEphemeral     = "ephemeral"
+	// respFieldResponseAction / respFieldView are the view_submission reply
+	// keys (response_action: "errors"|"update" + the replacement view).
+	respFieldResponseAction = "response_action"
+	respFieldView           = "view"
 )
 
 func respondSlack(w http.ResponseWriter, text string) {
--- a/apps/slack/internal/handler_display_name.go
+++ b/apps/slack/internal/handler_display_name.go
@@ -71,25 +71,31 @@ func parseSetDisplayNameArgs(text string) (id, name, userMsg string) {
 	if name == "" {
 		return "", "", "Missing Display Name.\n\n" + displayNameUsage
 	}
-	if len([]rune(name)) > displayNameMaxLen {
-		return "", "", fmt.Sprintf("Display Name is too long (max %d characters).\n\n%s", displayNameMaxLen, displayNameUsage)
-	}
-	// Reject backticks, angle brackets, and control bytes. The Display Name
-	// is stored in `description` and rendered to EVERY user who runs
-	// `/qurl list` or `/qurl aliases` (not just the admin who set it), inside
-	// an inline-code fence (`` `<id>` — <name> ``). A backtick breaks the
-	// fence (running an unterminated code span into the next row); a control
-	// byte garbles the line; and Slack mrkdwn `<…>` is an injection vector —
-	// `<https://evil|Prod>` renders a disguised link and `<!channel>` /
-	// `<!here>` ping the workspace. Admins are trusted, but a stored value
-	// shown to all users is worth fencing at the parser (the alias-target
-	// parser rejects backticks for the same hygiene reason).
-	for _, r := range name {
-		if r == '`' || r == '<' || r == '>' || !unicode.IsPrint(r) {
-			return "", "", "Display Name can't contain backticks, angle brackets, or control characters. Use plain text only.\n\n" + displayNameUsage
-		}
+	if msg := validateDisplayNameChars(name); msg != "" {
+		return "", "", msg + "\n\n" + displayNameUsage
 	}
 	return idTok, name, ""
+}
+
+// validateDisplayNameChars rejects a Display Name that is too long or carries
+// characters unsafe in the shared `/qurl list` / `/qurl aliases` surfaces: a
+// backtick breaks the inline-code fence the id is rendered in, mrkdwn `<…>` is
+// an injection vector (`<https://evil|Prod>` disguised links, `<!channel>` /
+// `<!here>` pings), and a control byte garbles the line. The Display Name is
+// stored in `description` and shown to EVERY workspace member, so this fence is
+// applied wherever a name is set — the set-display-name verb here AND the
+// `/qurl list` Edit modal (handler_edit.go). Returns "" when ok, else the
+// user-facing reason WITHOUT a usage dump (callers append their own context).
+func validateDisplayNameChars(name string) string {
+	if len([]rune(name)) > displayNameMaxLen {
+		return fmt.Sprintf("Display Name is too long (max %d characters).", displayNameMaxLen)
+	}
+	for _, r := range name {
+		if r == '`' || r == '<' || r == '>' || !unicode.IsPrint(r) {
+			return "Display Name can't contain backticks, angle brackets, or control characters. Use plain text only."
+		}
+	}
+	return ""
 }
 
 // parseUnsetDisplayNameArgs validates a `unset-display-name <id>` body:
--- a/apps/slack/internal/handler_display_name.go
+++ b/apps/slack/internal/handler_display_name.go
@@ -67,7 +67,7 @@ func parseSetDisplayNameArgs(text string) (id, name, userMsg string) {
 	if !found {
 		return "", "", "Missing Display Name.\n\n" + displayNameUsage
 	}
-	name = strings.TrimSpace(stripSurroundingQuotes(strings.TrimSpace(rest)))
+	name = normalizeDisplayNameInput(rest)
 	if name == "" {
 		return "", "", "Missing Display Name.\n\n" + displayNameUsage
 	}
@@ -109,6 +109,15 @@ func parseUnsetDisplayNameArgs(text string) (id, userMsg string) {
 		return "", fmt.Sprintf("`%s` isn't a valid tunnel id. Run `/qurl list` to see your tunnel ids, then retry.\n\n%s", echoText(tokens[0]), displayNameUsage)
 	}
 	return tokens[0], ""
+}
+
+// normalizeDisplayNameInput canonicalizes a raw Display Name: trim, drop a
+// matched surrounding quote pair, trim again. Shared by the set-display-name
+// verb and the `/qurl list` Edit modal so both interpret a typed/pre-filled
+// name identically — the modal also diffs against this form to decide whether
+// the name actually changed (see parseTunnelEditModalArgs).
+func normalizeDisplayNameInput(s string) string {
+	return strings.TrimSpace(stripSurroundingQuotes(strings.TrimSpace(s)))
 }
 
 // stripSurroundingQuotes removes one matched pair of surrounding single or
--- a/apps/slack/internal/handler_edit.go
+++ b/apps/slack/internal/handler_edit.go
@@ -392,11 +392,12 @@ func (h *Handler) reconcileChannelAliases(ctx context.Context, log *slog.Logger,
 }
 
 // formatTunnelEditSummary renders the admin-facing ephemeral summarizing an
-// edit. Aliases are shown as `$<alias>` code spans; the token is echoed as the
-// tunnel's id. Everything interpolated here is charset-validated (token/alias)
-// or char-fenced (display name handled upstream), so it's safe in mrkdwn.
+// edit. Aliases are shown as `$<alias>` code spans (charset-validated, so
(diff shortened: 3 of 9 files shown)


Kevin-layerV:
Thanks — addressed in `0ea7f50` (cleanups in `6bb63be`):

- **#1 legacy/external display names blocking an alias-only edit** — Fixed. `parseTunnelEditModalArgs` now validates the Display Name **only when it actually changed**, diffing the submission against the identically-normalized pre-filled value. An untouched name that predates the char fence (backtick / `<…>` / control byte, e.g. set via the qURL-service API) no longer blocks an alias-only edit. New unit test: `untouched legacy name skips validation (alias-only edit)`.
- **#2 whitespace → spurious PATCH** — Fixed by the same normalize-both-sides diff: surrounding whitespace/quotes on the stored value no longer register as a change. Extracted a shared `normalizeDisplayNameInput` (trim → strip surrounding quotes → trim) used by both the set-display-name verb and the modal so they interpret a name identically. New unit test: `whitespace-only difference is not a change`.
- **#3 extra `CheckAdmin` per `/qurl list`** — Acknowledged, no change. It's one bounded (`adminGateBudget`), fail-closed, eventually-consistent single-key GetItem on a human-triggered, already-async path (different table/key from the existing `GetChannelPolicy` read, so it can't be folded in). The cost is immaterial at this surface's QPS; parallelizing the two reads would add concurrency for a sub-round-trip win that isn't worth it.
- **#4 `respondTunnelEditModalError` method → free function** — Done; now matches its sibling `respondTunnelInstallModalError`.
- **#5 over-cap render test** — Added `TestHandleList_OverCapRowFallsBackToCreateOnly`: a tunnel with a large bound-alias set degrades to the Create-only accessory button while a normal sibling keeps Edit.

Also, a `/simplify` pass earlier (`6bb63be`) dropped an unreachable `meta.ChannelID==""` guard in `processTunnelEdit`, derived `listEditButtonMaxRows` from `listCreateButtonMaxRows/2` (so the two block-budget caps can't drift), and removed a stale `CreatedAtUnix` doc sentence.

Kevin-layerV:
Thanks — addressed the substantive finding in `3267234`:

- **#1 pre-fill exceeding `listEditMaxAliases` blocking otherwise-valid edits (medium)** — Fixed per your suggestion. The pre-filled extra-alias set is now carried through the modal's `private_metadata` (`TunnelEditModalMetadata.Aliases`), and `parseEditAliasLines` caps only **newly-added** aliases against it. A tunnel already carrying more than the cap stays editable for a name-only or removal-only change (an untouched aliases field has 0 newly-added → never rejected). New binds remain bounded; removals are unbinds of the pre-filled set, itself bounded by the button-value cap that gated the Edit affordance. New tests: `untouched over-cap pre-filled set is allowed` and `too many aliases ADDED on top of pre-filled rejected`.
- **"No changes." coverage (nit)** — Added `TestHandleTunnelEdit_NoOpSubmission`: submitting the pre-filled values verbatim skips the PATCH (name unchanged) and the alias reconcile is a no-op, so the admin sees "No changes." and the bindings are untouched.
- **#2 validation feedback precedes the admin gate (low)** — Leaving as-is for parity with `handleTunnelInstallSubmission`; the parse is side-effect-free and the cross-checks/admin gate still block all mutation, as you noted.
- **#3 concurrent-edit race** — Confirmed intentional; documented as an accepted rare race for an admin tool.
- **PATCH-fail ordering nit** — Intentional (name PATCH failure short-circuits before the alias reconcile, the safer ordering), as you called out.

Kevin-layerV:
Thanks — addressed in `e024513`:

- **#1 (medium) alias-only edits impossible on a tunnel with no Display Name** — Fixed. The Display Name input is now `optional=true`. With the changed-only diff, an empty submission on an empty-named tunnel normalizes to `"" == ""` → `nameChanged=false` → PATCH skipped, so an admin can add/remove aliases without inventing a name. New test: `TestHandleTunnelEdit_EmptyNameAliasOnly`.
- **#2 (low–med) admins with 23–45 tunnels losing all row buttons** — Fixed. Decoupled "show buttons" (`len ≤ listCreateButtonMaxRows`) from "show Edit" (`editable && len ≤ listEditButtonMaxRows`). Past the edit cap an admin now keeps the same Create-only buttons any caller gets — only past `listCreateButtonMaxRows` does the list degrade to text. New test: `TestHandleList_AdminPastEditCapKeepsCreateButtons`.
- **#3 concurrent unbind of a concurrently-added alias** — Documented explicitly on `reconcileChannelAliases`: the submitted set is authoritative (last-write-wins), so an alias bound by another admin after the modal opened can be unbound — broader than the rename race, an accepted documented race for an admin tool. (The newly-added cap is scoped to the pre-filled set; the bind/unbind reconcile is intentionally not.)
- **Test gaps** — Added all three: conflict path (`TestHandleTunnelEdit_AliasConflictReported`), `hadError` warning path (`TestHandleTunnelEdit_BindErrorSurfacesWarning`, via injected UpdateItem failure), and the empty-name + alias-only scenario above.
- **#2-validation-before-gate / PATCH-fail ordering nits** — Left as-is (intentional, parity with the install modal / safer ordering), as previously noted.

Kevin-layerV:
Thanks — addressed in `a2d1a75` + `68daa05`:

- **(medium) rename-only edit unbinding aliases on an empty snapshot** — Fixed at the deeper of the two suggestions. `reconcileChannelAliases` now scopes removals to the pre-filled snapshot (`meta.Aliases`): only an alias the admin **saw and deleted** — and that's still bound — is unbound. Additions still diff against the fresh set. This closes the transient-empty-snapshot data loss (a rename-only edit with an empty snapshot removes nothing) **and** the broader concurrent-add case from the previous round (an alias another admin bound after open isn't in the snapshot, so it survives). New regression test `TestHandleTunnelEdit_EmptySnapshotRenamePreservesAliases`; `HappyPath` now carries a realistic snapshot for the dropped alias. Updated the doc comment to describe the snapshot-baseline semantics.
- **(minor) CheckAdmin on every `/qurl list`** — Now gated behind the size check: `showEdit := len(resources) <= listEditButtonMaxRows && h.listCallerCanEdit(...)`, so a list too large to carry Edit buttons skips the read entirely (the answer can't change the output).
- **(minor) no way to clear a Display Name** — Intentional: the modal renames; clearing entirely stays with `/qurl-admin unset-display-name`. The field error now points there ("Enter a display name, or run /qurl-admin unset-display-name to remove it"). The field is `optional` only so an unnamed tunnel stays editable for alias-only changes.
- **(minor) field-validation precedes the admin gate** — Left as-is (parity with the install modal; parse is side-effect-free, no mutation either way), as you noted it's not exploitable.

Kevin-layerV:
Thanks — all flagged as minor; handled in `312970c`:

- **#3 `hadError` + "No changes." co-occurring** — Fixed. The "No changes." line is now gated on `!res.hadError`, so a failed bind/unbind shows only the `:warning:` line.
- **#2 failed Display Name PATCH aborts the alias reconcile** — Keeping the fail-fast ordering (name first; a transient failure stops before touching aliases, and the error tells the admin to retry — cleaner than a partial apply). Added the missing coverage: `TestHandleTunnelEdit_NamePatchFailureLeavesAliasesUntouched` asserts the aliases are untouched and the failure surfaces.
- **#1 rename-into-conflict dropping the old alias** — Leaving as-is by design. The modal edits an alias **set**, not paired renames, so there's no "A↔B" pairing to key suppression off — removing A and adding B are independent edits. Silently keeping A because B conflicted would ignore the admin's explicit removal. The summary already surfaces both outcomes on separate lines ("Removed alias(es): `$A`" and "Skipped (already used by another tunnel…): `$B`"), so the net effect is visible; the admin can re-add A or pick a non-conflicting alias. Honoring the removal + reporting the conflict is the most predictable behavior for the set model.

Confirmed your other checks (byte-vs-char cap is conservative, sigil-free/token-excluded/dedup parsing, snapshot matches the list line, disclosure-vs-mutation split) — appreciated the thorough pass.

Kevin-layerV:
Thanks — all three were minor; addressed the two doc asks in `ae8f631`:

- **#1 block-ceiling coupling** — Added the explicit invariant. `listEditButtonMaxRows`'s comment now states `2*listEditButtonMaxRows + 3 <= 50`, and `listCreateButtonMaxRows` carries a note that the edit cap is derived as `/2` so bumping it (or adding an always-on block) erodes the 3-block margin.
- **#3 byte-vs-char cap** — Added a comment on `slackButtonValueMaxBytes` noting the byte-length check is intentionally conservative (bytes >= runes), so it can't pass an over-limit value even with a multibyte Display Name — and not to "fix" it to rune-counting.
- **#2 conflict mislabel on a same-tunnel concurrent bind** — Agreed, leaving as-is: cosmetic, vanishingly rare, and the alternative (distinguishing same-tunnel vs other-tunnel on the race) isn't worth the code. The outcome is still correct; only the skip message would read slightly off in that race.

Kevin-layerV:
Thanks — all non-blocking; handled in `c44f65a`:

- **#1 unescaped token in `formatTunnelEditSummary`** — Fixed. The token now runs through `escapeMrkdwnCode` before the code span, matching the modal's `tunnelEditTokenLabel`. Since the token can be an upstream slug that never passed the alias charset fence, this closes the backtick-breaks-the-fence case as defense-in-depth. Locked by `TestFormatTunnelEditSummary_EscapesToken`.
- **#3 untested replay cross-checks** — Added `TestHandleTunnelEdit_ReplayCrossChecks`: team mismatch, user mismatch, and unparseable/incomplete `private_metadata` all reject with an error view before any PATCH/bind. These are the mutation access boundary, so good to have them locked.
- **#2 unbind TOCTOU is alias-name-scoped, not resource-scoped** — Confirmed intentional and acceptable for this PR. It's pre-existing `slackdata` store behavior (not introduced here), the reconcile's snapshot-baseline scoping already closes the practical edit-modal data-loss cases, and the residual is documented as accepted last-write-wins for an admin tool. Adding a resource-id compare-and-delete to `UnbindChannelAlias` (so a concurrent rebind to a different tunnel can't be clobbered) is a reasonable store-level hardening, but it touches `unsetalias`/`setalias` callers too, so it belongs in a separate follow-up rather than this feature PR. Recommend tracking it as a `pr-polish-followup` for the store.

PR #568, opened 2026-06-01
feat(slack): require email for setup

justin-layerv:
## Summary
- require `/qurl setup <email>` before minting a Slack setup OAuth state
- update setup/help/error copy to point users at the required email form
- keep email-bound setup behavior and coverage, including invalid/bare setup paths
- remove the duplicate `blockKitFieldElements` const currently present on `main`; the referenced declaration remains in place, and this restores Slack package compilation for the current `main` merge result

## Validation
- `go test -count=1 ./...`
- `go test -count=1 ./apps/slack/internal`
- `go test -count=1 ./apps/slack/internal/oauth`
- `golangci-lint run --new-from-rev=origin/main`
- GitHub checks pass, including Slack lint/test/Docker and Claude review

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ All integrations connect to the qURL API. The endpoint is configurable via the `
 
 ## Slack Tunnel Onboarding
 
-Customer onboarding is install-first: install the qURL Slack app, run `/qurl setup`, then use `/qurl-admin tunnel install` or `/qurl get`. The Slack app install stores the workspace bot token for modal support; customers do not manually provide a bot token. Admin verbs live under a separate `/qurl-admin` slash command (registered against the same request URL as `/qurl`); user verbs — including `setup` — stay on `/qurl`.
+Customer onboarding is install-first: install the qURL Slack app, run `/qurl setup <email>`, then use `/qurl-admin tunnel install` or `/qurl get`. The Slack app install stores the workspace bot token for modal support; customers do not manually provide a bot token. Admin verbs live under a separate `/qurl-admin` slash command (registered against the same request URL as `/qurl`); user verbs — including `setup` — stay on `/qurl`.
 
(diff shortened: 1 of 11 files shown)


justin-layerv:
Claude follow-up:

- Added a doc note on `oauth.MintState` clarifying that the no-email minter is legacy/test support now; new `/qurl setup` dispatches use `MintStateWithEmail`.
- Keeping the `views.go` hunk intentionally: current `origin/main` has `blockKitFieldElements` declared twice in the same const block. This branch removes only the duplicate declaration and leaves the referenced const in place. Dropping that hunk reintroduces the local compile error (`blockKitFieldElements redeclared`).

justin-layerv:
Final Claude follow-up:

- Fixed the stale setup log string to say `MintStateWithEmail failed`.
- Re-confirmed the `views.go` hunk is intentional: current `main` has two `blockKitFieldElements` declarations in the same const block. The PR removes only the duplicate and leaves the used declaration intact; dropping it would reintroduce the compile error.

justin-layerv:
@claude /simplify

justin-layerv:
/simplify follow-up:

- Applied the behavior-preserving `parseSetupSubcommand` simplification and pushed it.
- Tested the suggested test-table literal simplification locally, but `golangci-lint run --new-from-rev=origin/main` rejects it (`goconst` wants the existing subcommand constants), so I kept the lint-safe table.

Local validation after the accepted simplification:
- `go test -count=1 ./apps/slack/internal`
- `go test -count=1 ./...`
- `golangci-lint run --new-from-rev=origin/main`

justin-layerv:
Addressed latest Claude nit by syncing the slackdata workspace comment to `/qurl setup <email>`.

Also confirming the `/qurl-admin setup <email>` redirect echo is intentional: the response is ephemeral to the same Slack caller, and command logging continues to redact setup email input.

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the latest review follow-ups:

- Updated the PR body to call out that the `blockKitFieldElements` removal restores Slack package compilation for the current `main` merge result.
- Filed #569 for the branch-protection / merge-result CI drift question that allowed PR #567 to be green pre-merge but fail `slack: Build and Test` on the merge commit.

The process gap is tracked there; this PR contains the direct code fix and all current PR checks are passing.

justin-layerv:
@claude /simplify

PR #570, opened 2026-06-01
fix(slack): owner sees /qurl list Edit button (+ unbreak duplicate-const build)

Kevin-layerV:
## Summary

Two fixes that together explain why the bot-assigned **owner** sees no **Edit** button on `/qurl list`.

### 1. Unbreak the `main` build (P0)
`#565` (rich_text install snippets) and `#566` (admin Edit button) each added a `blockKitFieldElements = "elements"` const to the same block in `views.go`. The two merges didn't conflict textually, so `main` landed with the const declared **twice** and `go build ./...` fails. Every image build — and therefore every deploy — from `main` HEAD has failed since both landed (2026-06-01), which means `#566`'s Edit button isn't actually deployed yet. First commit drops the duplicate (value unchanged). _This commit is cleanly cherry-pickable if the build needs unblocking ahead of the rest of this PR._

### 2. Treat the workspace owner as an admin
`CheckAdmin` only honored `admin_slack_user_ids`. The owner recorded in `owner_id` by `/qurl setup` was never treated as admin unless they also happened to be on that set. A fresh bind seeds the owner onto the set, but that link is written **once** — legacy rows and idempotent setup reruns can leave the owner off it, locking them out of every admin affordance, including the new per-row Edit button.

The fix grants admin off `owner_id` directly, so the gate self-heals on read and the change flows through every admin surface (the list Edit gate, `requireAdminSync`, and the edit-modal submit check) — i.e. the owner can both **see and use** Edit, alongside admins who already could.

**Safety:** `owner_id` is the write-once Slack user-id anchor. A shape-mismatched legacy `owner_id` (an Auth0 `sub` from before the Slack-id migration) can never equal a real Slack user id, so this can't over-grant.

## Tests
- `slackdata`: `TestCheckAdmin_OwnerIsAdminOffOwnerID` — owner absent from the admin set is admin; a stranger who is neither owner nor on the set stays denied (over-grant fence).
- `internal`: `TestHandleList_RendersEditButtonForOwnerNotOnAdminSet` — owner not on the admin set gets the Edit button on `/qurl list` (the reported regression).
- Updated the now-stale "owner is not auto-admin" comments.
- `go build`, `go vet`, full `apps/slack/...` suite, and `pre-commit` all pass locally.

## Not included (by design)
The raw `r_…` rows in `/qurl list` are slug-less **legacy** tunnels created before qurl-service persisted slugs (~2026-05-24); the renderer is already slug-first and correct, and new tunnels always get a slug. Those rows are being cleared by deleting the stale test tunnels, not by a code change.

### DIFF
--- a/apps/slack/internal/fakeddb_fixtures_test.go
+++ b/apps/slack/internal/fakeddb_fixtures_test.go
@@ -42,9 +42,10 @@ func seedWorkspaceAdmin(teamID, ownerID, slackUserID string, configuredAt time.T
 }
 
 // seedWorkspaceAdmins is seedWorkspaceAdmin with more than one user on the
-// admin set. CheckAdmin only honors the admin SS (the owner is not
-// auto-admin), so the alias-gate helper needs both of its test callers
-// ("U_admin" and "U_alias_admin") listed here.
+// admin set. CheckAdmin grants admin to the owner (off owner_id) AND to any
+// user on the admin SS, so a caller that isn't the owner must be listed here
+// — hence the alias-gate helper lists both of its test callers ("U_admin"
+// and "U_alias_admin").
 func seedWorkspaceAdmins(teamID, ownerID string, adminUserIDs []string, configuredAt time.Time) map[string]ddbtypes.AttributeValue {
 	at := configuredAt.UTC().Format(time.RFC3339)
 	return map[string]ddbtypes.AttributeValue{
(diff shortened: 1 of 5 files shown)


Kevin-layerV:
Thanks for the thorough pass. On the two non-blocking nits:

- **Verbose `CheckAdmin` comment** — leaving as-is; it matches the heavy doc-comment style throughout `slackdata`, and you flagged you'd keep it too. No churn.
- **Bind-time owner seed now partly redundant for the gate** — agreed it's no longer load-bearing for the gate, but the seed still feeds `admin list` display, so removing it is a behavior-touching change beyond this gate fix. Tracked for follow-up in #571 (labeled `pr-polish-followup`).

Issue #555, opened 2026-05-29
/qurl aliases: surface an incomplete-listing footer when a bound tunnel is past the first resource page


Kevin-layerV:
Follow-up from PR #554 (cr review ask #2).

## Context
PR #554 made `/qurl aliases` resolve each bound tunnel's slug from a single `ListResources` page (the same source `/qurl list` reads), joined by `resource_id`. A binding whose `resource_id` isn't on that first page (workspace has >`listResourcesScanLimit` = 100 resources, and the tunnel sorts onto a later page) degrades that row to **alias-only** — the slug just doesn't render. The alias stays fully usable (`/qurl get $alias` works regardless); only the cosmetic slug display is lost.

## Problem
Unlike `/qurl list` — which appends `…more resources past the first 100-row scan — some tunnels may not be shown.` when `page.HasMore` is true (`apps/slack/internal/handler_list.go:202`) — `/qurl aliases` swallows the `page.HasMore` signal entirely. A user whose bound tunnel fell past the first page sees an alias-only row with no indication the listing may be incomplete.

## Why not fixed in #554
The available signal, `page.HasMore`, is the **master-list** cursor flag — "the workspace has >100 resources of *any* type," not "more *tunnels* exist." Mirroring `/qurl list`'s footer today would propagate the exact leaky abstraction that #531 exists to remove (push the `type=tunnel` filter server-side so `has_more` reflects tunnels). The right ordering is: land #531 first, then a tunnel-accurate `has_more` makes a footer here trivially correct.

## Proposed fix (after #531)
- When at least one alias group fails to resolve its slug AND `page.HasMore` is true (tunnel-accurate, post-#531), append an incomplete-listing footer to `/qurl aliases`, mirroring `/qurl list`'s copy.
- Until #531 lands, do **not** add a `page.HasMore`-driven footer here — it would over-claim in the common URL-heavy workspace.

## Acceptance criteria
- After #531: `/qurl aliases` surfaces an incomplete-listing footer only when a bound tunnel is genuinely past the scanned page.
- A workspace with ≤100 total resources (or where every bound tunnel resolves) shows no footer.
- Test asserting the footer appears for an unresolved binding with `has_more=true`, and is absent when all bindings resolve.

## Dependency
Blocked on #531 (server-side `type=tunnel` filter so `has_more` reflects tunnels).

Kevin-layerV:
## Regression context (from #554 cr review)

[...]

PR #572, opened 2026-06-01
fix(slack): hide revoked tunnels from /qurl list; "No aliases found" for /qurl aliases

Kevin-layerV:
## Summary

Two small `/qurl list` + `/qurl aliases` display-correctness fixes.

> **Stacked on #570** (base = `fix/slack-owner-is-admin`) so CI builds over the duplicate-const fix in that PR — `main` doesn't currently compile. Merge #570 first, then retarget this to `main` (the diff is already clean).

### 1. Hide revoked tunnels from `/qurl list`
The resource list endpoint is status-visible for **active AND revoked** rows — a revoke is a soft delete; the row is hard-deleted only after a ~90-day retention window. `filterTunnelResources` keyed only on `type`, so a revoked tunnel kept appearing, complete with a "Create qURL" button that can't mint against it. Now we also drop rows whose status is `revoked` (a missing/empty status is treated as live, so the guard can only ever hide an explicitly revoked row).

This is also what makes deleting a stale tunnel actually clear it from the Slack list immediately, instead of lingering as a revoked row for up to 90 days.

### 2. `/qurl aliases` skips slug-only rows and reports the empty state
The install flow auto-binds `$<slug>` as a channel alias, so every installed tunnel carries a binding equal to its slug even with no user-defined alias. `/qurl aliases` listed those as bare `$<slug>` rows, making it ambiguous whether a token was an ID or an alias. Now we skip groups whose only alias equals the slug and render the existing "No aliases are configured for this channel yet…" empty state when none remain. Resource-less / slug-unresolved groups (`slug == ""`) have no slug to exclude, so their aliases still list.

## Tests
- `TestHandleList_ExcludesRevokedTunnels` — a revoked tunnel is dropped; the active sibling still renders.
- `TestHandleAliases_SlugOnlyBindingReportsNoAliases` — a channel whose only binding equals the slug reports no aliases.
- `TestHandleAliases_DropsSlugOnlyKeepsRealAlias` — slug-only row dropped, real-alias sibling kept.
- `go build`, `go vet`, full `apps/slack/...` suite, and `pre-commit` pass locally.

### DIFF
--- a/apps/slack/internal/fakeddb_fixtures_test.go
+++ b/apps/slack/internal/fakeddb_fixtures_test.go
@@ -42,9 +42,10 @@ func seedWorkspaceAdmin(teamID, ownerID, slackUserID string, configuredAt time.T
 }
 
 // seedWorkspaceAdmins is seedWorkspaceAdmin with more than one user on the
-// admin set. CheckAdmin only honors the admin SS (the owner is not
-// auto-admin), so the alias-gate helper needs both of its test callers
-// ("U_admin" and "U_alias_admin") listed here.
+// admin set. CheckAdmin grants admin to the owner (off owner_id) AND to any
+// user on the admin SS, so a caller that isn't the owner must be listed here
+// — hence the alias-gate helper lists both of its test callers ("U_admin"
+// and "U_alias_admin").
 func seedWorkspaceAdmins(teamID, ownerID string, adminUserIDs []string, configuredAt time.Time) map[string]ddbtypes.AttributeValue {
 	at := configuredAt.UTC().Format(time.RFC3339)
 	return map[string]ddbtypes.AttributeValue{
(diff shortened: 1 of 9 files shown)


Kevin-layerV:
Thanks for tracing it through. On the non-blocking callout — the unresolved-slug path (a slug-only binding whose resource is paginated out past the scan limit, or a stale binding): agreed it's the correct trade-off since the code can't know the slug there, and it's already tracked by the two follow-ups this file references. #531 (server-side `type=tunnel` filter) removes the over-fetch/scan-limit entirely, which is what creates the paginated-out gap in the first place; #555 surfaces the incomplete-listing signal for a bound tunnel that's past the first page. Both are open. No code change here — leaving it to ride those two.

PR #574, opened 2026-06-01
fix(slack): align tunnel install output with client id

justin-layerv:
## Summary
- Render the generated tunnel route config with `id` instead of `name`.
- Emit `QURL_TUNNEL_ID` (was `QURL_TUNNEL_SLUG`) in the Docker, Compose, ECS, and Kubernetes install snippets.
- Add a regression test (`TestRenderTunnelConfigYAMLUsesRouteID`) asserting the route emits `id:` and never the legacy `name:`.

## Why — client contract (minimum sidecar client: v0.3.0)
Aligns the Slack install output with the released sidecar client. `qurl-reverse-tunnel-client` v0.3.0 (released 2026-06-01, see layervai/qurl-reverse-tunnel-client#233 "feat(cli)!: rename tunnel identity to id") renamed the customer-facing route identity to `id` and the headless env-var fallback to `QURL_TUNNEL_ID`, and **removed `QURL_TUNNEL_SLUG` outright** — the old env var is no longer read (hard cut; `QURL_TUNNEL_NAME` is deprecated/ignored). The route `name:`/`slug:` YAML keys survive only as read-only legacy shims, but `id` is canonical.

The snippets' default sidecar image is `ghcr.io/layervai/qurl-reverse-tunnel-client:latest`, which now resolves to v0.3.0, so a snippet still emitting `QURL_TUNNEL_SLUG`/`name:` would silently no-op and fail bootstrap. Verified against the released client: `EnvTunnelID = "QURL_TUNNEL_ID"`, and the route struct's primary field is `id` (`name`/`slug` demoted to read-only `LegacyName`/`LegacySlug`).

## Tests
- `go test ./apps/slack/...`
- `go test ./...`
- `git diff --check`


justin-layerv:
Addressed the review items:

- Added a renderer comment clarifying that the Slack/Admin API slug is emitted verbatim as the sidecar route `id`.
- Updated the PR body to call out that this targets the released sidecar client contract (`id` / `QURL_TUNNEL_ID`).

Re-ran `go test ./...` and `git diff --check` before pushing.

Kevin-layerV:
Thanks — addressed the three items.

**1. Negative assertion for the legacy env var (done).** `0299fe7e` adds `QURL_TUNNEL_SLUG` to the forbidden-substring slices of the Docker (`TestTunnelInstallCreatesResourceBindsAliasAndMintsBootstrapKey`) and Kubernetes (`TestTunnelInstallModalSubmissionMintsKubernetesInstructions`) install-path tests, so a stray reappearance in the rendered async reply now fails the suite.

I deliberately did **not** add a bare `name:` forbid at the integration level: `name:` legitimately appears there for the K8s env/volume name fields (`- name: QURL_TUNNEL_ID`, `- name: qurl-agent-state`) and the ECS container env. The route-config `- name:` negative is already pinned at the unambiguous level by `TestRenderTunnelConfigYAMLUsesRouteID`.

**2. `id` (install) vs `slug` (`/qurl list`) label divergence.** Agreed it's worth a UX-consistency pass, and it's out of scope for this rename. Filed #575 (`enhancement` / `pr-polish-followup` / `priority: low`) with scope + acceptance criteria.

**3. `:latest` drift.** Pre-existing and client-side, as you note. It's already partially mitigated on this surface — `tunnelImageNote` warns operators to set `QURL_TUNNEL_IMAGE` to an immutable release tag or digest before production rollout, and the `:latest` default is explicitly the dev/sandbox fallback. A pinned major tag on the client side would harden future contract cuts further; that's a client-repo concern, not tracked here.

Kevin-layerV:
Addressed the two non-blocking notes from the latest review:

1. **Minimum client version (done).** `888cbd66` documents that the rendered snippets require sidecar client **≥ v0.3.0**, in the `QURL_TUNNEL_IMAGE` env-var row next to the pin-to-immutable-tag guidance — so an operator overriding the image is steered away from pinning a pre-v0.3.0 client.
2. **"slug" vocabulary in internal comments.** Leaving as-is per your note — it's the deliberate, documented split (Admin-API resource `slug` = the client's route `id`, same verbatim value). The customer-facing label divergence (`/qurl list` says "slug", install says "id") is tracked separately in #575.

Kevin-layerV:
Latest two notes:

1. **Symmetric negative coverage (done).** `681fb337` adds the `QURL_TUNNEL_SLUG` forbidden assertion to the Compose and ECS renderer tests' existing leak-check slices, so all four rendered surfaces (Docker, Compose, ECS, Kubernetes) now guard against the legacy var reappearing.
2. **Comment terminology (`handler_list_test.go:368-369`).** Leaving as-is per your note — the slug↔id equivalence is the deliberate, documented split. Customer-facing relabeling is tracked in #575.

Kevin-layerV (approved):

PR #584, opened 2026-06-02
feat(slack): polish /qurl list layout (header, bold rows, primary Create)

Kevin-layerV:
## What

Redesigns the interactive `/qurl list` output so it reads as an organized list rather than a cramped wall of bullets + uniform buttons.

**Before:** each tunnel was a bulleted single line (`• \`\$id\` (aliases: …) — name`) followed by a row of two same-weight buttons.

**After:**
- A `header` block titles the list (`🔒 Protected Tunnel Resources`).
- Each tunnel is a bold, multi-line section: the `$id` on its own line, the Display Name beneath it, and a faint `aliases:` line only when channel aliases are bound.
- **Create qURL** is the primary (filled) button so the headline action reads above the secondary, admin-only **Edit**.

## Why this shape (and not a table / dividers)

- Slack table/`data_table` blocks can't hold interactive buttons (cells are text/number/rich_text only), so the per-row Create/Edit actions rule out a real table.
- Per-row `divider` blocks looked nice but cost one block per row. Slack caps a message at **50 blocks**, and the row caps (`listCreateButtonMaxRows=45`, `listEditButtonMaxRows=22`) are derived against a stated `2*x+3 ≤ 50` invariant. Dividers would ~halve capacity and force reworking that math, so they're omitted; the header + bold layout carry the visual structure instead.

## Behavior preserved

- **Block-count per row is unchanged** (no-ID row = 1, create-only = 1, admin = 2), so the ceiling caps and their invariant hold without edits.
- The plain-text fallback is **byte-identical** — `formatTunnelListLine` still builds `body`; the new multi-line layout (`formatTunnelListSection`) is block-only. Notification/accessibility text and all existing text-based assertions are untouched.
- Admin gating, the Edit-button snapshot, the slug-less "(no ID …)" row, sorting, and the >45-row text fallback are all unchanged.

## Changes

- `views.go`: add `headerBlock` + `primaryButtonElement`; generalize `sectionWithButton` → `sectionWithAccessory` (the list was its only caller).
- `handler_list.go`: render a `header` block and a block-only, multi-line `formatTunnelListSection`.
- Tests: `TestFormatTunnelListSection` (unit) + `TestHandleList_PolishedDesignMarkers` (header block, bold layout, primary Create).

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` — all pass.
- `pre-commit run --files …` (gofmt, go mod tidy, golangci-lint, etc.) — all pass.

### DIFF
--- a/apps/slack/internal/handler_edit_test.go
+++ b/apps/slack/internal/handler_edit_test.go
@@ -373,8 +373,8 @@ func TestHandleList_OverCapRowFallsBackToCreateOnly(t *testing.T) {
 	if snap.ResourceID != normalRID {
 		t.Errorf("Edit button is for %q, want the normal row %q", snap.ResourceID, normalRID)
 	}
-	// The over-cap row degrades to a Create-only accessory button (sectionWithButton,
-	// not an actions block), so it's still mintable — just without Edit.
+	// The over-cap row degrades to a Create-only accessory button (a section
+	// accessory, not an actions block), so it's still mintable — just without Edit.
 	createVals := createQurlButtonValues(t, blocks)
 	if len(createVals) != 1 || createVals[0] != bigTok {
 		t.Errorf("over-cap row Create-only accessory = %v, want [%s]", createVals, bigTok)
--- a/apps/slack/internal/handler_list.go
+++ b/apps/slack/internal/handler_list.go
@@ -340,22 +340,25 @@ func (h *Handler) processListResources(ctx context.Context, log *slog.Logger, va
 			blocks = append(blocks, sectionBlock(sectionText))
 			continue
 		}
-		// Create qURL is the headline action, so it renders as the primary
-		// (filled) button. Admin rows pair it with Edit in an actions block; the
+		// Create qURL is the row's headline action. It renders as the primary
+		// (filled) button ONLY when an Edit button sits beside it — admin rows,
+		// where primary expresses the Create-over-Edit hierarchy. A create-only
+		// row has nothing to outrank, and a whole column of lone primaries reads
+		// as noise (Slack advises using `primary` sparingly), so it gets a
+		// default-style button. Admin rows pair the two in an actions block; the
 		// Edit button carries the row's edit snapshot so opening the modal needs
 		// no extra read. A snapshot too large for a button value falls through to
 		// the Create-only accessory path below.
-		create := primaryButtonElement(listCreateButtonLabel, listCreateQurlActionID, tok)
 		if showEdit {
 			if editVal, ok := buildTunnelEditButtonValue(resources[i].ResourceID, tok, resources[i].Description, aliasMap[resources[i].ResourceID]); ok {
 				blocks = append(blocks, sectionBlock(sectionText), actionsBlock(
-					create,
+					primaryButtonElement(listCreateButtonLabel, listCreateQurlActionID, tok),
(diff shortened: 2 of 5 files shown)


Kevin-layerV:
Thanks for the thorough reviews — addressed in `95ea69de` (and the earlier `d2e89124`):

- **No-ID rows lose their Display Name (Review 2, the one flagged to resolve before merge)** — fixed. `formatTunnelListSection`'s `token == ""` branch now keeps `r.Description` above the "No ID set" note, restoring parity with the old `sectionBlock(formatTunnelListLine(...))` path (and with the text fallback). Added a no-ID-**with**-Display-Name case to `TestFormatTunnelListSection` that pins it (the prior no-ID case used an empty description, so it didn't catch the drop).
- **Redundant token recompute (Review 1, note 1)** — already addressed in `d2e89124`: `formatTunnelListSection` now takes the precomputed `tok` instead of recomputing `tunnelDisplayToken`, which also guarantees the section names the same token the row's button mints against.
- **Display Name rendered raw in mrkdwn (both reviews, "pre-existing / no action for this PR")** — agreed it's out of scope (it predates this PR and affects `formatTunnelListLine` too), so filed as a tracking follow-up: #585 (escape mrkdwn metacharacters in the Display Name across both formatters).
- **Slice mutation in `extraAliasTokens`** — confirmed safe as you noted (`aliasesExcluding` allocates a fresh slice). No change.

Kevin-layerV:
Addressed in `ab151cff`:

- **Multi-primary-button density** — good call. Scoped the primary (filled) Create button to admin rows where it pairs with Edit (so the filled style actually expresses Create-over-Edit hierarchy); create-only / non-admin rows now use a default-style Create. That kills the "lone primary with nothing to outrank" and the column-of-green at scale, while keeping the admin row look intact. Pinned both branches: `TestHandleList_PolishedDesignMarkers` asserts the lone create-only button is *not* primary, and the new `TestHandleList_PrimaryCreateScopedToAdminRows` covers admin-paired → primary vs create-only → default.
- **Wording drift between the two no-ID strings** — leaving as-is intentionally: the inline parenthetical `(no ID — ask your Slack admin to set one)` suits the single-line fallback, while the standalone `_No ID set — ask your Slack admin to set one._` reads as a status line under the multi-line section. A shared constant would only cover the trailing clause and would force both call sites to concatenate around different surrounding punctuation, which is more awkward than the small duplication. Low churn risk since neither string is volatile.

Kevin-layerV:
Addressed in `91524d37`:

- **`headerBlock` 150-char cap** — added the caveat to the doc comment so a future caller passing a dynamic string knows to keep it short (today's only caller uses the short `listHeaderBlockText` constant).
- **`strings.Builder` consistency** — `formatTunnelListSection` now uses a single `strings.Builder` across both the no-ID and token branches. Output is byte-identical (covered by `TestFormatTunnelListSection`).
- **Unescaped `r.Description` in mrkdwn** — pre-existing (affects `formatTunnelListLine` too), already tracked as a follow-up in #585, so left out of this PR's scope.

Kevin-layerV:
Addressed in `4461ef64`: added a comment on `extraAliasTokens` documenting that its in-place mutation is safe only because `aliasesExcluding` returns a freshly-allocated slice — hardening that contract against a future change. The 150-char `headerBlock` note is covered by the doc caveat added in `91524d37`.

PR #589, opened 2026-06-02
feat(slack): channel-scope /qurl list & get; add expose-to-channels Edit field

Kevin-layerV:
## Problem

The Slack bot's tunnel permissions had a disclosure/capability gap:

- **`/qurl list`** was **workspace-wide for everyone** — every member (and admin) saw every tunnel in every channel, including DMs.
- **`/qurl get`** had an **admin bypass** — an admin could mint any tunnel's `$slug` from *any* channel, even one the tunnel was never exposed to.
- `/qurl aliases` was already channel-scoped (point read on the current channel's policy).

So a tunnel installed in `#ops` showed up — and, for admins, was mintable — in unrelated channels and DMs.

## Fix: one channel-scoped definition, enforced everywhere

A tunnel is **"available in channel C"** iff its id is in `AllowedResourceIDsForChannel(C)` — the union of the channel's `allowed_resource_ids` and its `alias_bindings` values. This was already the *non-admin* mint gate; now `list`, `aliases`, and `get` all agree on it, for admins and non-admins alike.

| Requirement | Change |
|---|---|
| install in C → only C lists it | `/qurl list` filters to the channel allow-set. Install already binds `$slug` in C (covered by the union), so **no install change**. |
| `/qurl aliases` scoped to channel | already true — pinned with a regression test. |
| can't `/qurl get` outside the channel — even admins, even knowing the `$id`/`$alias` | removed the admin bypass in the mint gate (`resourceAllowedForUser` → `resourceAllowedInChannel`). `$r_id` was already rejected at parse; `$alias` is already per-channel. |
| Edit modal → expose to more channels, pre-filled | new **Channels** `multi_conversations_select` on the `/qurl list` Edit row, pre-filled from a new `ChannelsForResource` enumeration. |

Behavior details:
- `/qurl list` fails **closed**: empty `channel_id` and a scope-read error refuse rather than fall back to an unscoped list; a channel with nothing exposed short-circuits to a channel-framed empty state (no upstream call).
- The empty state names the recovery path (install here, or expose from `/qurl list` → **Edit** in a channel where it already appears), addressing the dead-end that motivated the #459 revert of #234 — but now with the gate applying to admins too.
- The Edit modal's channel reconcile exposes newly-selected channels (`allowed_resource_ids`) and revokes de-selected ones. The channel being edited from **always keeps access** and is never revoked; removals are diffed against the channels the modal actually *showed*, so a partial enumeration can't drop a channel the admin never saw.
- This also closes the former `TODO(#460)` list/mint asymmetry: you see exactly what you can mint here, and mint only what you can see here.

## New store ops (`channel_policies`)

`ExposeResourceToChannel` / `RevokeResourceFromChannel` (idempotent SS add/delete) and `ChannelsForResource` — the package's first DDB **`Query`** caller (a base-table query on the `slack_team_id` partition key).

> [!NOTE]
> **No infra change is needed to ship this.** `ChannelsForResource` uses `dynamodb:Query`, which the bot's task role **already grants** on `channel_policies`: the `qurl-bot-slack` data-plane policy lists `Query` and its `Resource` covers all three Slack base tables (landed on infra `main` 2026-05-12, before this PR). The modal pre-fill is also best-effort by design — if that grant were ever absent it degrades to showing only the current channel, and **no access is ever lost** (the reconcile acts only on the channels it showed). List-scoping, get-blocking, and exposing-to-channels don't depend on it.

## Tests

- Rewrote the two tests that encoded the old behavior into the scoped behavior: `TestHandleList_ScopedToChannel` (admin **and** non-admin scoped; exposed shows, unexposed hidden, no-policy/DM empty) and `TestHandleGet_DollarSlugAdminAlsoChannelScoped` (admin blocked when not exposed, mints when exposed).
- Added: store-op tests (expose/revoke expression shapes; `ChannelsForResource` union + sort + LastEvaluatedKey paging + Query-error), `/qurl list` fail-closed / empty-channel / exposed-via-allow-set, `/qurl aliases` channel-scoping pin, modal-render (channels select + omit-empty-initial), and edit-reconcile (expose / revoke / current-channel-never-revoked / enumeration-degrades) coverage.
- `golangci-lint --new-from-rev=origin/main`: 0 new issues. Full Go suite green (the two `apps/cli` failures are a local config artifact, not from this change — `apps/cli` is untouched).

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -30,10 +30,12 @@ Customer onboarding is install-first:
 ### User commands (`/qurl`)
 
 - `/qurl setup <email>` — Connect qURL to the workspace (one-shot OAuth flow against Auth0; Auth0 starts login with that address prefilled; first-come-claims — the first user to run it becomes the workspace's qURL admin, and only they can re-run it)
-- `/qurl get <$id|$alias>` — Mint a one-time qURL for a tunnel `$id` or a channel `$alias` (raw URLs are not supported)
-- `/qurl list` — List the protected tunnel resources available to you (with their bound channel shortcuts)
+- `/qurl get <$id|$alias>` — Mint a one-time qURL for a tunnel `$id` or a channel `$alias` (raw URLs are not supported). Channel-scoped: you can only mint a tunnel from a channel it's been exposed to — including admins, and regardless of knowing the `$id`/`$alias`.
+- `/qurl list` — List the protected tunnel resources available **in this channel** (with their bound channel shortcuts). A tunnel appears only in the channel it was installed in, plus any channels an admin later exposes it to via the **Edit** button.
 - `/qurl aliases` — List the qURL shortcuts configured in the current channel
 - `/qurl help` — Show the user command help
+
+A tunnel's visibility and mintability are the same channel-scoped set: `/qurl list`, `/qurl aliases`, and `/qurl get` all agree on which tunnels are available in a given channel. Admins manage where a tunnel reaches with `/qurl-admin tunnel install` (installs into the current channel) and the **Edit** button on a `/qurl list` row (a "Channels" multi-select that exposes the tunnel to additional channels; the channel it was installed in always keeps access).
 
 ### Admin commands (`/qurl-admin`)
 
--- a/apps/slack/internal/admin_test_helpers_test.go
+++ b/apps/slack/internal/admin_test_helpers_test.go
@@ -181,6 +181,18 @@ func (ts *adminTestServers) seedPolicyAliasBindings(t *testing.T, teamID, channe
 	ts.ddb.seedItem(t, ts.tableNames.channelPolicy, seedChannelPolicyAliasBindings(teamID, channelID, bindings))
 }
 
+// seedChannelExposure exposes resourceIDs in (teamID, channelID) by seeding the
+// channel's allowed_resource_ids SS — the channel-scoping that `/qurl list`
+// (disclosure) and `/qurl get` (mint) now require for a tunnel to appear/mint
+// there. Thin alias for seedPolicySet(alias="") that reads as intent at the
+// call sites updated for channel-scoped `/qurl list`. Pass every resource_id
+// the test's /v1/resources fixture returns that should be visible in the
+// channel.
+func (ts *adminTestServers) seedChannelExposure(t *testing.T, teamID, channelID string, resourceIDs ...string) {
+	t.Helper()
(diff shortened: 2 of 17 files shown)


Kevin-layerV:
Thanks for the thorough reviews — all points addressed in `5aa296fe` (which also sits on top of the `origin/main` merge `25a00839` that cleared the semantic conflict with #584's list-polish tests).

**Fixed**
- **Alias-bound revoke reporting (the flagged blocker).** De-selecting a non-current channel that's exposed via an alias binding used to clear only `allowed_resource_ids`, leaving the tunnel listable/mintable there while the summary claimed *"Revoked from"*. The reconcile now re-checks the alias surface after the delete (`GetChannelPolicy`, which the SS `DELETE` doesn't touch — so no eventual-consistency race) and reports such channels as *"Still available via a channel alias (remove the alias there to fully revoke)"*. A channel is only reported revoked when it actually leaves `AllowedResourceIDsForChannel`'s union. Pinned by `TestHandleTunnelEdit_AliasBoundChannelReportedRetainedNotRevoked`.
- **DM/non-channel hardening.** `parseEditChannelSelection` now drops conversation ids that aren't `C…`/`G…` channels, so a hand-crafted `view_submission` can't write a tunnel into a DM's policy row — the exact leak the scoping closes. New sub-test.
- **`ChannelsForResource` projection.** Added a `ProjectionExpression` (mirrors `LookupChannelAlias`) so the team-wide Query pulls only the channel id + the two membership attributes instead of full rows.
- **Stale comments in `handler_get.go`.** Corrected the "`/qurl list` is workspace-wide" wording and the open-`TODO(#460)` breadcrumb; both now reflect that list / aliases / get share one channel-scoped set.

**Deferred (tracked in #590)**
- **Scan-limit × channel-scope.** Only the first 100 upstream resources are scanned before allow-set filtering, so a tunnel sorting past row 100 won't render in a channel. It's fail-safe (under-disclosure, never a leak) and doesn't affect mintability — `/qurl get` gates on the allow-set directly, not the scan. Driving the listing by allow-set ids is a design change out of scope here; added an inline comment + filed #590 with repro/scope/acceptance.

**Acknowledged**
- The `dynamodb:Query` IAM grant remains the documented infra follow-up (PR-body callout). Until it lands, the modal pre-fill degrades to current-channel-only, which is safe — the reconcile only acts on the channels it showed, so nothing is wrongly revoked.
- Post-fix, the only residual over-report is the *expose* side reporting an idempotent re-add as "Exposed to" when the open-time enumeration was degraded/truncated — cosmetic, with no incorrect resulting state.

PR #591, opened 2026-06-02
fix(slack): rename user resources copy

justin-layerv:
## Summary
- Rename non-admin Slack copy from tunnel language to protected/resource language
- Keep tunnel terminology on admin-only setup/help surfaces
- Update Slack README and pinned copy assertions

## Business relevance
This keeps the Slack app clearer for end users by hiding implementation details from non-admin workflows while preserving precise tunnel terminology for admins who configure resources.

## Validation
- go test ./apps/slack/...
- git diff --check


justin-layerv:
@claude /simplify

PR #596, opened 2026-06-03
fix(slack): page /qurl list until the channel allow-set is satisfied (#590)

### LINKED ISSUE #590: /qurl list channel-scope: a tunnel exposed beyond the 100-row upstream scan limit silently doesn't render
## Context
Surfaced by Claude review on #589 and expanded by #599 (`/qurl list` + URL-resource `/qurl get`).

`processListResources` (`apps/slack/internal/handler_list.go`) fetches a single upstream page (`ListResources{Limit: listResourcesScanLimit}` = 100) and then filters that page down to the channel allow-set (`AllowedResourceIDsForChannel`). The 100-row cap is pre-existing, but channel scoping and URL-resource alias support make the cap load-bearing in more places.

## Problem / repro
- A workspace has >100 tunnel+URL resources.
- A resource legitimately exposed/installed in channel C sorts beyond row 100 in the upstream `ListResources` response.
- `/qurl list` in C filters only the first 100 rows to the allow-set, so the exposed resource is omitted from the list. If that is the only exposed resource, the channel renders the empty-state despite being exposed.
- For tunnel slugs and channel-bound aliases, `/qurl get` remains targeted and is not affected by the first-page list scan.
- For URL resources minted by intrinsic resource alias, #599's fallback also scans the first 100 rows. A URL resource exposed to the channel but sorted beyond row 100 can be unmintable by intrinsic alias.
- Duplicate-alias ambiguity detection is also first-page scoped. If a second channel-allowed URL resource sharing an alias sorts beyond row 100, the ambiguity guard can see only one match and mint instead of failing closed as ambiguous.

This remains fail-safe for disclosure (under-disclosure, never a cross-channel leak), but the URL-alias get path makes the scan limit a correctness limitation, not just a display limitation.

## Scope / options
- Drive Slack listing and URL-alias resolution by the known allow-set resource IDs rather than discovering them in a scanned page, for example with per-ID resource fetches or pagination until every allow-set member is found.
- Make duplicate-alias ambiguity detection complete across the channel-allowed resource set, not just the first upstream page.
- Preserve the common-case cost profile for workspaces with <=100 resources where practical.

## Acceptance criteria
- A resource exposed to channel C renders in `/qurl list` in C regardless of its position in upstream resource ordering / total resource count.
- `/qurl get $alias` can resolve a channel-exposed URL resource by intrinsic alias even when it sorts beyond the first 100 upstream rows.
- Duplicate URL-resource aliases fail closed as ambiguous across the full channel-allowed set, even when one duplicate sorts beyond the first page.
- No regression to fail-closed channel scope behavior or to targeted tunnel slug/channel-alias minting.

Reported against #589; expanded by #599 review follow-up.

Kevin-layerV:
## What & why

`/qurl list` is channel-scoped: it shows the tunnels exposed to the current channel (the same set `/qurl get` mints against). It built that list by fetching a **single 100-row page** of all the owner's resources and filtering it down to the channel allow-set. Two problems fell out of that (#590):

1. **Incompleteness.** A tunnel legitimately exposed to the channel but sorting past row 100 in the owner's full resource set silently didn't render. Fail-safe (under-disclosure, never a leak) and it didn't affect mintability — `/qurl get` gates on the allow-set directly — but the discovery listing was wrong.
2. **A misleading footer.** The "…more resources past the first 100-row scan — some resources may not be shown" note keyed on the master-list `has_more`, i.e. "more resources of *any* type." In a URL-heavy workspace it fired even when every exposed tunnel was already shown.

## The fix

Drive the listing off the channel allow-set instead of a fixed page. `fetchAllowedTunnels` pages `/v1/resources`, matches allow-set members as they stream past, and **stops as soon as the last one is found** (or the listing is exhausted).

- A workspace whose exposed tunnels all land on the first page still costs **one request** — no common-case cost regression.
- A tunnel past the old window now renders — #590 fixed.
- The footer is removed: a satisfied allow-set means the listing is complete, so there's nothing to warn about.
- `listMaxResourcePages` (50 × 100 = 5,000 rows) is a safety backstop against a pathological walk (e.g. a stale channel binding to a deleted resource that is never found). It sits far above any real workspace's resource count, so it can only ever drop an id the owner's list no longer contains — it never reintroduces the window for a live tunnel.

## No behavior change beyond the two intended fixes

The entire rendering pipeline — tunnel/revoked filtering (`filterTunnelResources`, reused as-is), sort order, row/section formatting, Create qURL / Edit buttons, empty state, error handling — is untouched. Only the *source* of the resource set changed. The whole existing `/qurl list` test suite passes **unchanged**, which is the regression proof. `/qurl aliases` is not touched (it still does its single-page join).

## Testing

- Existing `/qurl list` + `/qurl aliases` suites pass unchanged.
- New tests:
  - `TestHandleList_FindsTunnelPastFirstPage` — the #590 regression: a tunnel exposed to the channel but on page 2 now renders; a page-1 tunnel not exposed to the channel still doesn't.
  - `TestHandleList_NoTruncationFooterWhenExposedTunnelsFound` — `has_more=true` with the allow-set already satisfied → no footer, and exactly one upstream fetch (early-exit).
  - `TestHandleList_ExhaustsWhenExposedIDMissing` — a stale binding to a missing id pages to exhaustion and terminates; the live tunnel still renders.
- `go vet`, `gofmt`, `golangci-lint` (pinned, via pre-commit), and the full package `go test` are all clean.

## Alternative considered

Per-ID hydration (`GET /v1/resources/{id}` for each allow-set member) was the other sanctioned option for #590. Pagination was chosen because it needs no new client code, keeps the existing HTTP-level test suite as a regression harness, and is cheaper in the common case (the detail endpoint also hydrates each resource's qURLs server-side).

Related: #531 (push a server-side `type=tunnel` filter) would let `/qurl aliases` drop to a small page too; out of scope here.

Fixes #590.


Kevin-layerV:
Thanks — both actionable points addressed:

**#2 (page-cap test gap)** — fixed in e607e58e. `TestHandleList_StopsAtPageCap` serves a fresh cursor + `has_more=true` on every page with an always-missing allow-set id, so natural termination never fires and the walk can only stop at `listMaxResourcePages`; it asserts exactly 50 fetches and that the live tunnel still renders. That closes the one previously-uncovered `fetchAllowedTunnels` exit.

**#1 (stale-binding fan-out)** — confirmed real and deferred to #597. There's no qurl-service→Slack resource-lifecycle signal: `RevokeResourceFromChannel`/`UnbindChannelAlias` fire only from the admin Edit-modal paths, so a hard-deleted upstream resource leaves a dangling `channel_policies` id indefinitely. For a >100-resource workspace that forces the 50-page walk every `/qurl list` in that channel. It's bounded + fail-safe (capped, logged, under-disclosure only), so not a blocker for this PR — but worth fixing properly (reactive prune on revoke/delete, or lazy self-heal when the listing exhausts with ids still pending). #597 has the repro + both candidate fixes.

Leaving the dual `!HasMore || NextCursor == ""` guard as-is per your #3 — it's the load-bearing defense against a malformed `has_more=true, next_cursor=""` page (re-requesting page 1 forever otherwise).

Kevin-layerV:
-

PR #599, opened 2026-06-04
fix(slack): restore URL resources in qurl list/get

justin-layerv:
## Summary
- Show channel-available URL resources in `/qurl list` alongside tunnel resources.
- Allow `/qurl get $alias` and list-row Create qURL buttons to mint qURLs for URL resources.
- Add a guided admin expose surface, all single-word verbs:
  - `/qurl-admin expose` — chooser: pick *qURL Connector* or *URL*, then a short Block Kit form.
  - `/qurl-admin expose-connector` — bare opens the guided connector modal; `expose-connector <id> [env:…] [port:…] [alias:$alias] [container:|service:…]` is the typed power-user form (skips the modal).
  - `/qurl-admin expose-url` — bare opens the guided URL-resource picker; `expose-url $<resource-alias> [as:$channel-alias]` / `expose-url url:<target-url> as:$channel-alias` is the typed power-user form (skips the modal).
  - These replace the former two-word `tunnel install` / `resource expose` grammar — the slash surface is single-word / hyphenated-word only. Typed paths need no `views.open`, so no-guided deploys still work via the typed forms.
- Keep raw ad hoc URL input rejected while preserving the standard `/qurl get` one-time/expiry/session policy for resource-backed links.
- `/qurl list` escapes a resource's target URL without the cosmetic-lookalike substitutions (`_`→＿ etc.), so underscore/tilde URLs stay copy-pasteable; structural escaping (`&`/`<`/`>`/backtick/newline) is preserved.

## Business relevance
Restores Slack discovery and qURL creation for existing URL resources, including resources created in the website dashboard, so users can generate protected links from Slack without switching tools — behind a consistent, single-word, guided admin command surface (`expose`, `expose-connector`, `expose-url`).

## Validation
- `go build ./apps/slack/...`
- `go vet ./apps/slack/...`
- `go test ./apps/slack/...`
- pinned `golangci-lint run` (v2.10.1, matching CI) — 0 issues


justin-layerv:
Follow-up on the latest Claude review notes for `5af893a`:

- Rate-limit-before-resolution is intentional. This is now command-level rather than mint-only so typoed/unknown aliases cannot amplify `ListResources` before the limiter; support should know repeated fat-fingered aliases can consume the per-user limit.
- Verified the Docker builder tag/digest with `docker buildx imagetools inspect golang:1.26.4-alpine`; the registry index digest matches the pinned `sha256:f23e8b227fb4493eabe03bede4d5a32d04092da71962f1fb79b5f7d1e6c2a17f`.
- The first-page scan limitation for URL resource aliases remains tracked by #590; comments in this PR call out that alias fallback and ambiguity detection inherit that boundary.
- The pre-existing tunnel-only token collision is tracked separately in #600 with repro scope and acceptance criteria, since it is not introduced by the URL-resource restoration.

justin-layerv:
Review-note confirmations for the latest pass:

- The first-page URL-alias ambiguity edge is tracked in #590, with an explicit get-side duplicate-alias note added there.
- `resource expose url:<target>` intentionally uses exact `target_url` matching for the bare-minimum Slack-first exposure path; the `$alias` form is the preferred admin path.
- URL resources intentionally do not use the tunnel-shaped Edit modal. Channel exposure is managed through the alias binding (`/qurl-admin resource expose ...`) and removed with `/qurl-admin unset-alias`.
- Moving the rate-limit gate before token resolution is intentional so typoed/unknown tokens cannot fan out into ungated resource scans.
- The mrkdwn homoglyph substitutions are accepted as display-only safety escaping; stored/matched resource values remain unchanged.

justin-layerv:
@claude Please run a final simplify-style review pass on the current PR head. Focus on whether a DE would accept this as merge-ready, whether any simplifications/nits/optional polish remain, and whether the existing #590 deferral is adequately tracked. Please call out anything actionable; if there is nothing actionable left, say it is clean.

PR #254, opened 2026-05-11
feat(slack): per-workspace OAuth — DDBProvider + /oauth/qurl/{start,callback}

Kevin-layerV:
## Summary

- Per-workspace Slack OAuth runtime that mirrors Discord's LIVE flow (`apps/discord/src/routes/qurl-oauth.js`)
- New `auth.DDBProvider` replaces `auth.EnvProvider` — DDB-backed lookup with field-level envelope encryption on `qurl_api_key` via AWS KMS (GenerateDataKey + AES-256-GCM)
- `apps/slack/internal/oauth/{start,callback}` HTTP handlers — CSRF cookie + HMAC state token → Auth0 authorize → code exchange → JWKS-verified id_token → POST /v1/api-keys → DDB upsert → success HTML page

## Pairs with

`qurl-integrations-infra` `feat/slack-oauth-wireup` (TF: `workspace_state` DDB + CMK + Auth0 SSM params). **The two PRs MUST land together** or the ECS task fails on missing env vars.

## Library choices

- **Field encryption**: raw `github.com/aws/aws-sdk-go-v2/service/kms` + `crypto/cipher` (AES-256-GCM with workspaceID as AAD). AWS does not ship a maintained Go binding of the DB Encryption SDK; same posture, different implementation. `TODO(claude-followup)` marker left in `shared/auth/ddb_provider.go`.
- **JWKS verify**: `github.com/lestrrat-go/jwx/v2` (no JWT lib was already in go.mod; this is the most actively maintained option).
- **AWS SDK**: `aws-sdk-go-v2` (config + dynamodb + kms + attributevalue).

## Test plan

- [x] `go vet ./...` — clean
- [x] `go test ./shared/auth/... ./apps/slack/...` — all green (CLI test `TestNewClient_MissingAPIKey` was already failing on `main`, unrelated)
- [x] `golangci-lint run ./apps/slack/internal/oauth/... ./shared/auth/...` — clean
- [ ] Manual OAuth flow against sandbox Auth0 once `qurl-integrations-infra` PR lands
- [ ] Confirm DDB row written with field-encrypted `qurl_api_key` (binary attribute, not plaintext)
- [ ] Confirm second workspace install isolated from first (separate row, separate ciphertext)

## TODOs left for follow-up PRs

1. Slack-API `chat.postMessage` DM client wiring — `SlackClient` interface is in place; production code passes `nil` for now so the success HTML page is still the user-visible confirmation. Discord's pattern is fire-and-forget DM after success.
[...]

PR #608, opened 2026-06-06
fix(slack): stop setup reruns from leaking API keys

### LINKED ISSUE #544: fix(slack): /qurl setup leaks an API key per run — revoke-then-mint (or reuse) the workspace key
## Background

`/qurl setup` mints a **new** qURL API key on qurl-service every run. It overwrites the stored DDB workspace row with the new key but **does not revoke the previously-minted key** on qurl-service. So each re-run leaks one key toward the account's plan cap (free tier = 3). Once the cap is hit, qurl-service refuses the next mint with HTTP 403 + body code `api_key_limit`.

PR #542 fixed the **symptom**: the OAuth callback now renders an actionable "revoke a key, then run setup again" 409 page instead of the old (misleading) "run setup again to retry" blank page. But the reason accounts hit the cap at all is this per-run leak.

## Proposed fix (the deeper layer)

Make `/qurl setup` stop accumulating keys at the source. Options:

1. **Revoke-then-mint**: before persisting the new key, revoke the key currently stored for the workspace (the prior `key_id`), so steady-state key count stays at 1 per workspace.
2. **Reuse**: if a valid workspace key already exists, reuse it instead of minting a new one (only mint on first setup or after an explicit rotate).

Revoke-then-mint is likely simpler and matches the existing revoke-on-failure paths already in `callback.go`/`minter.go`.

## Acceptance criteria

- Re-running `/qurl setup` on an already-configured workspace does **not** increase the account's live API-key count (verify in sandbox: count keys before/after a re-run).
- The previously-stored key is revoked (or reused) — no orphaned keys accumulate.
- Existing revoke-on-persist-failure / mint-failure behavior is preserved.
- Failure modes are safe: a revoke failure must not leave the workspace without a working key (mint new → persist → revoke old ordering, or equivalent).

## Context

- Root cause confirmed in sandbox during #542 (at 3 keys it 403'd; revoking one let setup succeed).
- Related to the orphan-key path (#265-adjacent).
- This is what made the #542 409 page necessary in the first place.

justin-layerv:
## Summary

Fixes #544 by making `/qurl setup` reuse a healthy stored workspace qURL API key instead of minting a new account-level key on every rerun.

The callback now:
- Reads the existing workspace key after the Slack/user bind gate passes.
- Validates that key with a cheap qurl-service authenticated read.
- Completes setup without minting when the stored key is healthy.
- Mints and persists a replacement only when no key is stored or the stored key is definitely invalid/revoked.
- Preserves the existing best-effort revoke path when a newly minted key cannot be persisted.

## Business relevance

This prevents repeated Slack setup runs from consuming a customer's finite qURL API-key quota, which avoids unnecessary setup failures at the plan limit and reduces support friction for already-configured workspaces.

## Validation

- `go test -count=1 ./apps/slack/internal/oauth`
- `go test -count=1 ./apps/slack/...`
- `go test -count=1 ./shared/...`
- `go test -count=1 ./...`
- `go test -race -count=1 -coverprofile=coverage.out -covermode=atomic ./apps/slack/... ./shared/...` (total coverage 80.0%)
- `go vet ./...`
- `golangci-lint run --timeout=5m --new-from-rev=origin/main ./apps/slack/... ./shared/...`
- `govulncheck ./apps/slack/... ./shared/...` (no called-code vulnerabilities)
- `CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -trimpath -ldflags="-w -s -X main.version=$(git rev-parse --short HEAD)" -o /tmp/qurl-bot-slack-test ./apps/slack/cmd/`

Note: a raw whole-repo local `golangci-lint run --timeout=5m ./apps/slack/... ./shared/...` with local golangci-lint 2.12.2 reports pre-existing baseline issues outside this diff. The new-code lint run above is clean; CI uses pinned v2.10.1.

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -57,7 +57,7 @@ modifiers enabled by the current bot deployment.
 - **Runtime:** AWS Fargate (arm64, distroless container) behind an
   ALB that terminates TLS and routes `/slack/*`, `/oauth/slack/*`,
   `/oauth/qurl/*`, and `/health`.
-- **Auth:** Per-workspace qURL API key, minted via `/qurl setup <email>` →
+- **Auth:** Per-workspace qURL API key, established via `/qurl setup <email>` →
   `/oauth/qurl/start` → Auth0 → `/oauth/qurl/callback`. Supplying
   an email address on setup stores it in signed state, sends Auth0
   `login_hint`, and requires the verified Auth0 email claim to match before any
@@ -69,9 +69,14 @@ modifiers enabled by the current bot deployment.
   connection for the same audience. `AUTH0_EMAIL_CONNECTION` is an optional
   recovery override when a deployment must force a specific connection. The
   callback's security gate is the verified email claim, not the connection hint
-  by itself. Keys are
-  field-level encrypted in the `workspace_state` DynamoDB table using
-  KMS envelope encryption with `workspace_id` bound as AAD.
+  by itself. If a workspace already has a qURL API key and qurl-service still
+  accepts it, setup reuses that key instead of minting another one; missing or
+  revoked stored keys mint a replacement. Rerunning setup is intentionally not
+  a healthy-key rotation or qURL-account switch command; revoke the workspace
+  key from qURL's API-key management/dashboard or operator tooling first, then
+  rerun setup to mint a replacement under the newly authenticated account. Keys
+  are field-level encrypted in the `workspace_state` DynamoDB table using KMS
+  envelope encryption with `workspace_id` bound as AAD.
 - **Slack app install:** Customer workspaces install qURL through
   `/oauth/slack/install`, which redirects to Slack OAuth with the bot scopes
   needed by the slash command and modal surfaces. The callback stores Slack's
--- a/apps/slack/cmd/main_test.go
+++ b/apps/slack/cmd/main_test.go
@@ -294,7 +294,7 @@ func TestClassifyBindErrorMapping(t *testing.T) {
 		want oauth.BindConflictCode
 	}{
 		{
-			"already bound to caller (idempotent rotation)",
+			"already bound to caller (idempotent re-entry)",
 			&slackdata.Error{StatusCode: http.StatusConflict, Code: slackdata.ErrCodeWorkspaceAlreadyBoundToCaller},
 			oauth.BindConflictAlreadyBoundToCaller,
 		},
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1100,17 +1100,18 @@ func setupVerbRest(text string) (rest string, matched bool) {
 // runs only the owner is permitted; other workspace members (including
 // admins added via `/qurl-admin admin add`) get an "owner-only" reply.
 // Without this gate, any added admin could complete OAuth against their
-// own Auth0 account and silently rotate the workspace's qURL credential —
+// own Auth0 account and silently re-point the workspace's qURL credential —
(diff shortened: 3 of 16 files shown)


justin-layerv:
Addressing the latest Claude review items on the current head:

- Confirmed the qurl-service integration contract on `layervai/qurl-service` `main` at `330cb65`: `GET /v1/quota` is protected by `RequireScopes("qurl:read")` ([router.go#L488-L489](https://github.com/layervai/qurl-service/blob/330cb65f487d8d0cb2b1c5448d2e0ab35d4341c0/internal/api/router.go#L488-L489)); non-active/revoked API keys abort with 401 ([apikey.go#L266-L275](https://github.com/layervai/qurl-service/blob/330cb65f487d8d0cb2b1c5448d2e0ab35d4341c0/internal/api/middleware/apikey.go#L266-L275)); and the smoke lifecycle test explicitly polls a revoked API key until it returns 401 ([lifecycle_test.go#L656-L670](https://github.com/layervai/qurl-service/blob/330cb65f487d8d0cb2b1c5448d2e0ab35d4341c0/tests/smoke/lifecycle_test.go#L656-L670)).
- Pinned that contract in `ValidateAPIKey` and added `TestAPIKeyScopesIncludeReadForStoredKeyValidation` so Slack-minted keys cannot silently lose the `qurl:read` scope needed by the validation probe.
- Made the behavior change explicit in the Slack README: `/qurl setup` reruns are intentionally idempotent reuse, not healthy-key rotation. For the separate self-service rotation affordance, opened #609 with `enhancement`, `security`, and `priority: medium`.
- Transient validation failures intentionally fail setup re-entry closed and do not mint; normal bot operations keep using the stored key. This avoids the quota leak/orphan risk when qurl-service is flaky or a scope check returns 403.

justin-layerv:
Follow-up on the account-switch/rotation product question: updated the README and `authorizeURL` docs to state that setup reruns are intentionally not a healthy-key rotation or qURL-account switch. The operator path for #544 is idempotent reuse by default; if a workspace needs to move to another qURL/Auth0 account, the stored key must be revoked first so setup validation returns 401 and the callback mints under the newly authenticated account.

Also updated #609 to explicitly track an owner-only Slack rotation/account-repoint flow with safe old-key revocation/key-id handling.

justin-layerv:
Addressed the latest Claude nits on `62fa066`:

- Replaced the tab-indented README continuation lines with normal two-space list continuation; verified with `! rg -n "^\t" apps/slack/README.md`.
- Made the revoke instruction actionable by naming qURL API-key management/dashboard or operator tooling.
- Added a `keyPrefixLength` comment tying the reuse-path prefix derivation to qurl-service's displayed `key_prefix` contract.

justin-layerv:
Answering the latest Claude question about historical workspace-key scopes:

- The Slack per-workspace OAuth mint path was introduced in commit `14693eba` (`feat(slack): per-workspace OAuth — DDBProvider + /oauth/qurl/{start,callback}` / #254). In that introducing commit, `apiKeyScopes()` already returned `[]string{"qurl:read", "qurl:write"}`.
- `git blame` still points the scope set at `14693eba`, and `git log -G "qurl:read|qurl:write|apiKeyScopes" -- apps/slack/internal/oauth/...` shows only #254, #542, and this PR touching the relevant files; #542 did not alter `apiKeyScopes()`.
- So repo history says every Slack workspace key minted by this OAuth path carried `qurl:read` from day one. I do not see a pre-#254 Slack workspace-key mint path that could have stored a key without that scope.

Docker is the only remaining CI check still running on the current head.

Issue #318, opened 2026-05-14
discord: drop formatRevokeLabel non-user branch (T+7d after PR #313)


justin-layerv:
## Context

PR #313 hard-removed \`/qurl send\`, so all new \`qurl_sends\` rows are written with \`target_type='user'\`. Historical rows from before the merge can still have \`target_type='channel'\`.

\`formatRevokeLabel\` (\`apps/discord/src/commands.js:1948-1950\`) has a branch for the non-\`'user'\` case:

\`\`\`js
const channelRef = s.target_type === 'user'
  ? 'DM'
  : (channelName ? \`#\${channelName}\` : \`\${s.target_type} channel\`);
\`\`\`

The maximum \`EXPIRY_LABELS\` window is \`7d\`. Seven days after PR #313 merges to prod, no \`qurl_sends\` row can still have \`target_type !== 'user'\` (any older rows are past their TTL and either already revoked or hidden by the revoke flow's \`revoked_at\`-aware query). At that point the non-\`'user'\` branch becomes unreachable dead code.

## Acceptance criteria

- [ ] Confirm PR #313 has been in prod for at least 7 days + a small buffer (verify via deploy logs).
- [ ] Drop the ternary's non-\`'user'\` branch in \`formatRevokeLabel\`; collapse to \`const channelRef = 'DM';\` (or simplify further if the variable becomes single-use after the rewrite).
- [ ] Update the comment at \`commands.js:1218-1223\` that flags this followup — either remove it or note it's been completed.
- [ ] Verify the surviving tests still pass without rewriting (the historical-row tests in \`qurl-send.test.js\` at lines 1072, 1121 may need their fixture \`target_type\` value changed).

## Related

- PR #313 (the trigger): refactor(discord): hard-remove /qurl send front-half

justin-layerv:
Refining the acceptance criteria per cr round 12 on PR #313:

The original framing was date-based (\"T+7d post-deploy\"). cr observed that's an over-approximation — what actually makes the branch dead is **\"no revoke-visible row has \`target_type !== 'user'\`\"**, which is condition-driven rather than calendar-driven. The 7-day \`EXPIRY_LABELS\` max window just bounds when that condition is safe to assume post-deploy (because the renderer filters on \`expires_at\`).

Updated acceptance criteria framing:

[...]

PR #614, opened 2026-06-07
feat(discord): edit DM tense on qurl.expired webhook

justin-layerv:
## Summary

Subscribe the bot to `qurl.expired` alongside `qurl.accessed` and PATCH each recipient's existing DM so Discord's `<t:UNIX:R>` relative-time marker flips from "Closes in 24 hours" to "Closed N ago" without further edits. Recipients see the qURL is dead the moment the upstream fires the event, instead of finding out by tapping a Step Through button that now 404s.

## Why this matters

Pre-this-PR, recipients of expired qURLs had no in-DM signal that the link was dead. Tapping Step Through fell into a 404 — the bot had no way to surface expiry without polling, which we deliberately moved away from. Eager DM-edit on the upstream event closes that gap and matches what the revoke path already does for sender-initiated closures.

## What it does

- Subscribes the bot to **`qurl.expired`** alongside `qurl.accessed` (both via the existing webhook-registrar).
- On `qurl.expired`, looks up the recipient row via a new sparse `qurl_sends.qurl_id-index` GSI and PATCHes the DM body so Discord's `<t:UNIX:R>` relative-time marker flips tense client-side. No follow-on edits needed.
- Threads `qurl_id` through the write path (`recordQURLSend` + `recordQURLSendBatch`) at both call sites (`executeSendPipeline`, `handleAddRecipients`) so the GSI is populated on new sends.
- Reconstructs the absolute expiry instant from `row.created_at + row.expires_in` — the wire payload is exactly `{qurl_id, resource_id}` (transit-safe by design; an absolute `expires_at` isn't carried on the wire).

### Collision handling (revoke vs. expired)

If the sender already ran `/qurl revoke`, the DM already reads "Alice closed the door". The handler reads `qurl_send_configs.revoked_at` and skips rather than overwriting that more-specific copy with a less-informative timeline marker. A new `expired_edited_at` attribute on the `qurl_sends` row guards against duplicate edits across distinct delivery IDs.

### Status policy

- **200** on permanent can't-action conditions: no matching row, send was revoked first, DM was already edited, `dm_status` not SENT, hard shape-mismatch, recipient blocked the bot / deleted the DM, marker-rollback failed.
- **503** on transient infra failures a retry can recover from: pre-marker GSI Query throw, pre-marker non-CCFE UpdateItem throw on the idempotency marker, and post-marker `editDM` failures (`ok:false && expected:false`, or thrown error) with a successful marker rollback. The upstream's 5-attempt `1+2+4+8+16=31s` backoff handles the recoverable cases.

`editDM`'s `expected` discriminator (set on the not-ok path) drives the post-edit branch: a permanent recipient-side failure (blocked the bot, deleted the DM) keeps the marker and returns 200; a transient Discord-side failure rolls the marker back and returns 503. If the rollback itself fails, the marker stays in place and we return 200 — the 8-day S3 lifecycle bounds the missed-edit blast radius.

### Latent bug also fixed

`reconcileEvents` short-circuited on `events.includes(QURL_ACCESSED)`, which silently left an accessed-only subscription in place after the target event set grew. Replaced with strict set-equality vs. `TARGET_EVENTS`, so any future event addition can never silently drift again.

### Forward-only by design

qURLs minted before this deploy lack the `qurl_id` attribute on their `qurl_sends` row and don't surface in the sparse GSI. The handler returns `200/no-recipient-row` and skips. The 8-day S3 lifecycle policy renders the missed edits moot.

## What changed

| File | Change |
| --- | --- |
| `apps/discord/src/constants.js` | `QURL_WEBHOOK_EVENTS.EXPIRED = 'qurl.expired'` |
| `apps/discord/src/qurl-webhook-registrar.js` | `TARGET_EVENTS = [ACCESSED, EXPIRED]`, set-based reconcile |
| `apps/discord/src/store/ddb-store.js` | sparse `qurl_id` writes; `findSendsByQurlId` / `markExpiredDMEdited` / `clearExpiredDMEdited` / `isSendRevoked` |
| `apps/discord/src/store/contract.js` | contract entries for the 4 new methods |
| `apps/discord/src/dm-payloads.js` | NEW leaf module — `buildExpiredDMPayload` (extracted to dodge the `commands.js` require graph at route load) |
| `apps/discord/src/routes/qurl-webhook.js` | `handleQurlExpired` branch + reconstruction helper |
| `apps/discord/src/commands.js` | pass `qurlId` at both `recordQURLSendBatch` call sites |
| `apps/discord/src/utils/time.js` | export `parseExpiryMs` (null-on-bad-input) |
| tests | full suite for the expired handler, set-based reconcile tests, store helper tests |

## Prod Rollout Tasks

- [ ] Added a prod rollout ledger entry
- [x] Confirmed this PR has no prod rollout tasks

Bot-code-only deploy; the companion GSI add (sparse `qurl_sends.qurl_id-index`) and the upstream flag-flip are tracked separately.

## Test plan

- [x] Handler suite — happy path, GSI miss, ambiguous `Count>1`, revoke gate, idempotency collision, dm-not-editable, four edit-failure branches, reconstruction failures, per-EXPIRY_LABELS marker pinning, off-set `expires_in` regression, 503-vs-200 split.
- [x] Set-based `reconcileEvents` — missing accessed / missing expired / extra event / order-independent / string-not-array.
- [x] Store helpers — `qurl_id` sparse writes (Send + Batch), `findSendsByQurlId` (happy + Count>1 + empty-input no-op + `Limit:2`), `markExpiredDMEdited` (first-call vs. already-edited vs. non-CCFE rethrow), `clearExpiredDMEdited`, `isSendRevoked` (revoked / not-revoked / missing row / empty input).
- [x] Existing fixtures updated to the two-event target set in registrar + lambda/webhook-registrar suites.
- [x] Full bot test suite: **2696 passed, 80 suites**, lint clean.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1909,6 +1909,11 @@ async function executeSendPipeline(interaction, {
     await db.recordQURLSendBatch(qurlLinks.map(link => ({
       sendId, senderDiscordId: interaction.user.id, recipientDiscordId: link.recipientId,
       resourceId: link.resourceId, resourceType, qurlLink: link.qurlLink,
+      // qurl_id is the GSI lookup key the qurl.expired webhook handler
+      // uses to find the recipient row for DM editing. Pulled from
+      // mintLinksInBatches return value above. Sparse on the write
+      // side — see recordQURLSendBatch.
+      qurlId: link.qurlId,
       // CONTRACT: targetType is always 'user' for new rows post-PR
       // #313 (the only caller is executeSendPipeline via the confirm
       // card on /qurl send + /qurl map, both of which DM individual
@@ -2618,7 +2623,12 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
     for (const link of links) {
       batchSends.push({
         sendId, senderDiscordId, recipientDiscordId: rid, resourceId: link.resourceId,
-        resourceType: link.resType, qurlLink: link.qurlLink, expiresIn: sendConfig.expires_in,
+        resourceType: link.resType, qurlLink: link.qurlLink,
+        // qurl_id threads through addRecipients the same way it does
+        // through executeSendPipeline — needed for the qurl.expired
+        // webhook handler's GSI lookup. See recordQURLSendBatch.
+        qurlId: link.qurlId,
+        expiresIn: sendConfig.expires_in,
         channelId: originalInteraction.channelId, targetType: 'user',
       });
     }
--- a/apps/discord/src/constants.js
+++ b/apps/discord/src/constants.js
@@ -494,6 +494,7 @@ Object.freeze(AUDIT_EVENTS);
 // silently matching nothing. Mirrors qurl-service WebhookEventType.
 const QURL_WEBHOOK_EVENTS = Object.freeze({
   ACCESSED: 'qurl.accessed',
+  EXPIRED: 'qurl.expired',
 });
 
 // Discord gateway dispatch event names (the `t` field on op=0 frames).
--- a/apps/discord/src/dm-payloads.js
+++ b/apps/discord/src/dm-payloads.js
@@ -1,0 +1,51 @@
+// DM payload builders for messages that EDIT an already-delivered DM
+// (PATCH /channels/{id}/messages/{id}). Pure functions — no I/O, no
+// process state. Lives in its own leaf module so the qurl-webhook
+// receiver route can import it without pulling in commands.js's
+// require graph (Discord REST, view-update-registry, flow-state,
+// etc.) at module-init time during tests.
+//
+// Today only buildExpiredDMPayload lives here. buildRevokedDMPayload
+// stays in commands.js next to its sole caller (the /qurl revoke
+// editTargets loop) until a second consumer needs it.
+//
+// CONTRACT: every builder MUST pass `components: []` explicitly.
+// Discord's PATCH /messages does NOT clear fields that aren't
+// supplied — omitting components would leave the original Step
+// Through button live in the recipient's DM, pointing at a now-dead
+// qURL resource.
+
+const { EmbedBuilder } = require('discord.js');
+const { COLORS } = require('./constants');
+
+// Companion to buildRevokedDMPayload for the qurl.expired webhook
+// path: the qURL hit its expiry without a sender-initiated revoke.
+// The expiry instant is rendered with Discord's `<t:N:R>` relative-
+// time marker so the client-side tense flips automatically as time
+// passes (no follow-on edits needed).
+//
+// UX choices (intentional):
+//   - The replacement embed wholly replaces the original delivery
+//     embed — sender alias / resource label / Trust footer are
+//     dropped. The PATCH semantic only supports whole-array embed
+//     replacement, and once the qURL is dead the original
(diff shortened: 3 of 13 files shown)


justin-layerv:
Thanks for round 5. Responses:

**GSI projection — confirmed `ALL`.** Restating from round 2: `modules/qurl-bot-ddb/main.tf` on `qurl-integrations-infra@1c0ab15d` (merged #913) declares the `qurl_id-index` GSI with `projection_type = "ALL"`. The five attributes the handler reads (`dm_channel_id`, `dm_message_id`, `dm_status`, `created_at`, `expires_in`) are all on the projected page. Bootstrap-deploy sandbox apply is green and the GSI is ACTIVE.

**Transient infra → 503 (2f763993).** Good catch. `lookup-error` and `mark-error` both fire *before* the idempotency marker is claimed, so a qurl-service retry can re-enter cleanly and recover the edit. Routed both to 503 (qurl-service retries 5x with 1+2+4+8+16=31s backoff — exactly the window a DDB throttle needs). Post-marker failures (`editDM` throw / `ok:false`) stay 200 since a retry would short-circuit at `already-edited` without recovering the failed PATCH. CCFE still returns false from `markExpiredDMEdited` and stays on the permanent already-edited path — not routed through the throw branch.

**Owner-scoping — same trust model as `qurl.accessed`.** The GSI lookup doesn't re-check `sender_discord_id` against the resolved owner, but the trust chain holds: HMAC verification proves the event came from the registered owner; qurl-service's subscription model delivers only that owner's qURLs to that owner's webhook; the matched recipient row is therefore intrinsically scoped to the owner. Same model `qurl.accessed` operates under (`recordQurlView` also doesn't ownership-check). If we ever multi-tenant on a single `owner_id`, both handlers would need to add the check together.

**30d-cap edge — noted inline (2f763993).** `parseExpiryMs` caps at `MAX_EXPIRY_MS`. Unreachable today (`EXPIRY_LABELS` max is 7d); flagged in `rowExpiresAtSeconds` for the future label-set-expansion case.

**Comment density nit** — the `target_type` block at commands.js:1917 is pre-existing (PR #313 / #318 context), not from this PR. Leaving the marker-before-edit tradeoff block in place — inline is where future readers look first, and PR descriptions rot faster than code comments.

CI re-running on 2f763993.

justin-layerv:
Thanks for round 9.

**#1 — `Limit: 2` (79cee42d).** Cheap hardening — added. The handler only needs the 0 / 1 / >1 trichotomy, so bounding the read at 2 keeps the RCU floor on the hot path while still surfacing ambiguity at the existing `> 1` branch. Pinned by a Limit assertion in the store test.

**#2 — Static "expired" vs marker tense.** Real edge at the firing instant. Bounded and rare enough at the 30m–7d horizon that I'd rather see it in soak before tweaking the copy (alternatives — `"This qURL has now expired"`, or omitting the verb and letting `<t:N:R>` carry the tense — both have their own awkwardness). Will revisit if it surfaces.

**#3 — `recordQURLSend` symmetry.** Keeping for now; the contract-shape parity with `recordQURLSendBatch` is the only reason. Cheap dead surface area.

**#4 — Claim-before-act gap.** Conscious tradeoff, confirmed.

CI re-running on 79cee42d.

PR #615, opened 2026-06-07
feat(slack): customer-facing docs + rename "bot" to "Secure Access Agent"

justin-layerv:
## What

Two related changes to the `apps/slack` integration's customer-facing surface.

### 1. Customer-facing docs (`docs(slack)`)
- **`apps/slack/README.md`** rewritten for the customer (Slack users + workspace admins): value prop + an **install → setup → protect → get** quickstart, a Concepts primer, scannable `/qurl` and `/qurl-admin` command tables **verified against the bot's own help text** (`userHelpMessage`/`adminHelpMessage`), a "Protecting a resource" walkthrough, and an FAQ mapped to the actual error strings.
- **New `apps/slack/docs/operating.md`** holds the operator/self-hosting material (architecture, endpoints, env vars, local dev, Slack app config) — *moved, not deleted*. The customer README links to it.
- **Top-level `README.md`** "Slack Connector Onboarding" trimmed to a short summary that links to the two app docs (it had duplicated the connector detail now canonical in `operating.md`).
- **`.gitignore`**: scoped rules for `.claude/scheduled_tasks.lock` + `.claude/worktrees/` (a runtime lockfile got swept in by a `git add -A` on this branch; these guard against recurrence without shadowing the tracked `.claude/skills/`).

### 2. Rename "bot" → "Secure Access Agent" (`feat(slack)`)
Per product terminology, "bot" is retired. Renamed across the **user-visible surface**: Slack-facing handler strings (help text, error/ack copy, startup log), the admin role ("bot admin" → **"admin"**; `*Bot admins*` header → `*Admins*`), the docs, the tests asserting those strings, and the code comments that directly narrate a renamed string/role. The back-pressure "busy" copy was unified behind a single `modalBusyMsg` const.

**Left literal on purpose:** Slack API terms (`bot token`, `bot scopes`, `SLACK_BOT_*`), example hosts (`slack-bot.example`), and the `qurl-bot-slack` image tag. The rename was scoped to user-visible strings + docs + tests — internal identifiers and generic "the bot" implementation comments are out of scope.

## Why
The old README was infra-first (rating ~5/10 for a customer) and had drifted from the live command surface (stale `admin add` / `admin revoke <qurl_id>` verbs; missing `protect-url`, `protect`, `set-/unset-display-name`, `get` flags, `feedback`). The rewrite targets a 10/10 customer doc; the rename aligns the product with current terminology.

## Notes for review
- **Two logical changes, one PR:** the rename was added at the maintainer's request mid-review. Kept as separate commits for reviewability.
- **Preserve vs. delete:** operator content (incl. the AWS runtime description) was *moved* to `docs/operating.md`, not deleted. Say the word if you'd rather it trimmed or dropped.
- Command names reconciled against the live dispatch/help in `apps/slack/internal/` (the old README had drifted post-#607).
- Go changes are user-visible string/comment renames + a const extraction; `go test ./apps/slack/...`, `go vet`, and CI `slack / *` are green.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,11 @@ coverage.html
 .DS_Store
 Thumbs.db
 
+# Claude Code runtime artifacts (the tracked .claude/skills/ live alongside
+# these — keep this list specific so it doesn't shadow them)
+.claude/scheduled_tasks.lock
+.claude/worktrees/
+
 # Build artifacts
 dist/
 release/
--- a/README.md
+++ b/README.md
@@ -40,11 +40,13 @@ All integrations connect to the qURL API. The endpoint is configurable via the `
 
 ## Slack Connector Onboarding
 
-Customer onboarding is install-first: install the qURL Slack app, run `/qurl setup <email>`, then use `/qurl-admin protect-connector` or `/qurl get`. The Slack app install stores the workspace bot token for modal support; customers do not manually provide a bot token. Admin verbs live under a separate `/qurl-admin` slash command (registered against the same request URL as `/qurl`); user verbs — including `setup` — stay on `/qurl`.
+Onboarding is install-first: install the qURL Slack app, run `/qurl setup <email>`,
+then an admin runs `/qurl-admin protect` to expose a resource in a channel and
+anyone runs `/qurl get` to mint a one-time link.
 
-Slack admins can run `/qurl-admin protect-connector` to generate a guided install for the qURL Connector sidecar. The workflow asks for the customer-facing qURL Connector ID, an optional channel shortcut, the local HTTP port, and the target runtime. Slack then returns copy/paste output tailored for Docker, Docker Compose, AWS ECS/Fargate, or Kubernetes.
-
-The Docker and Docker Compose paths create the proxy config, bootstrap-key secret, and per-connector durable agent-state directory in one guarded shell block. ECS/Fargate and Kubernetes output runtime-native task or pod artifacts, including a per-sidecar persistent state mount. Bootstrap API keys are short-lived and should be removed from the runtime after the sidecar logs show a successful connection.
+See [apps/slack/README.md](apps/slack/README.md) for the full command reference
+and onboarding walkthrough, and [apps/slack/docs/operating.md](apps/slack/docs/operating.md)
+for deploying and operating the bot.
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This monorepo contains Go-based platform integrations (Slack, Teams, Discord), a
 
 ```
 apps/           Per-integration applications (independent release tracks)
-  slack/        Slack bot — slash commands, link unfurling, notifications
+  slack/        Slack Secure Access Agent — slash commands, link unfurling, notifications
   cli/          CLI tool for qURL
   teams/        Microsoft Teams (planned)
   discord/      Discord bot (planned)
@@ -46,7 +46,7 @@ anyone runs `/qurl get` to mint a one-time link.
 
 See [apps/slack/README.md](apps/slack/README.md) for the full command reference
 and onboarding walkthrough, and [apps/slack/docs/operating.md](apps/slack/docs/operating.md)
-for deploying and operating the bot.
+for deploying and operating the Secure Access Agent.
 
 ## Development
 
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -16,7 +16,7 @@ longer than it needs to.
    Slack. The first person to run this becomes the workspace **owner**.
 3. **Protect a resource** (admins): run `/qurl-admin protect` and follow the
    guided form to expose a service or an existing URL in the current channel.
-4. **Share a link**: anyone in that channel runs `/qurl get $your-resource`
+4. **Share a link**: anyone in that channel runs `/qurl get $id`
    to mint a one-time qURL.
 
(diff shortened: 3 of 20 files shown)


justin-layerv:
Thanks for the thorough verification-against-help-text pass. Addressed the optional nits in `91b8bbb3`:

- **Quickstart/sharing placeholder** → now uses `$id` consistently with the term defined in Concepts (was `$your-resource`).
- **`protect-url` rows** → spelled out when each form applies (alias already exists vs. protect-by-URL with no alias yet).
- **`operating.md` connector-onboarding density** → restructured the dense paragraph into a scannable nested list, content preserved verbatim.

Intentionally kept as-is (both reviewers flagged these as "no change required / your call"):
- Customer command tables list conditionally-gated verbs flat — this is correct for a capability doc and is already qualified by the "run `/qurl help` for the exact commands your workspace supports" line.
- Generalized `set-alias` wording ("a resource" vs. "a qURL Connector ID") — kept, since aliases can point at URL resources too.
- Typed-connector `env:`/`container:`/`service:` enumeration stays in `/qurl-admin help` + `operating.md` rather than the customer table.

justin-layerv:
Good catch on the 🔴 — that `.claude/scheduled_tasks.lock` was a runtime session-lock artifact a `git add -A` swept in, not part of the docs. Fixed in `e6c22931` (amended out, so the PR diff no longer contains it):

- Removed `.claude/scheduled_tasks.lock` from the PR.
- Added scoped `.gitignore` rules for `.claude/scheduled_tasks.lock` and `.claude/worktrees/` — kept specific rather than ignoring all of `.claude/`, since the repo intentionally tracks `.claude/skills/`.

The PR diff is now just `.gitignore`, `apps/slack/README.md`, and `apps/slack/docs/operating.md`.

justin-layerv:
Thanks — re the two non-blocking notes, both are deliberate and I'm keeping them as-is:

1. **Owner semantics stated unconditionally.** The README targets the production deployment, where `AdminStore` is wired and the first-runner-owns-it gate is real. The no-DDB sandbox path having no owner is a deployment/operator detail — surfacing it in the customer README would re-introduce exactly the kind of internal nuance this PR set out to remove. The operator-facing `docs/operating.md` already covers setup's first-come-claims behavior and the OAuth-callback bind guard at the right altitude.
2. **Typed `protect-connector` options trimmed.** Intentional for the customer audience — the full `env:`/`container:`/`service:` surface lives in `/qurl-admin help` and `docs/operating.md`, both of which the README points to.

No code/doc change for these.

justin-layerv:
### Added: rename "bot" → "Secure Access Agent" (`3fefd470`)

Per product terminology, "bot" is retired — this is the qURL **Secure Access Agent**. Renamed end-to-end across the user-visible surface:

- **Slack-facing Go strings:** help text, error/ack copy (`Secure Access Agent is busy`, `… not configured on this Secure Access Agent deployment`), and the startup log line.
- **Admin role drops the qualifier** ("bot admin" → "admin"): help header `*Bot admins*` → `*Admins*`; `add`/`remove` confirmations now read "as an admin" / "from admins". (The `/qurl-admin admins` roster already rendered `Owner:`/`Admins:`.)
- **Docs** (README, operating.md, root README) updated to match, including the operating-guide title.
- **Tests** updated for the renamed help header; `go test ./apps/slack/...` green, `gofmt` clean.

**Left literal on purpose:** Slack's own API terms (`bot token`, `bot scopes`, `SLACK_BOT_*`), example hosts (`slack-bot.example`), the `qurl-bot-slack` image tag. Per the scope chosen, internal code comments/identifiers are out of scope for this rename.

PR title bumped to `feat(slack)` since this changes user-visible copy.

justin-layerv:
Thanks — all four notes considered, no change needed:

1. **Deployment-conditional commands listed flat** — intentional for a canonical customer doc; the *"Some commands aren't in `/qurl help`"* FAQ entry covers deploys without `PostDM`/`PostFeedback`. (Consistent with the call made in earlier review rounds.)
2. **Two logical changes in one PR** — the `feat(slack)` rename was added at the maintainer's request mid-review ("bot" is retired terminology), so it's bundled deliberately rather than split. Kept as separate commits for reviewability.
3. **`.gitignore`/lockfile cleanup** — agreed it's tangential; it's here because a `git add -A` in this branch swept the lockfile in, so the fix + guard belong with it.
4. **`go vet`** — run locally (`go vet ./apps/slack/...` exits 0), and CI `slack / test` + `slack / lint` + `slack / govulncheck` are green.

justin-layerv:
Addressed in `0a2e0521` + a PR-description fix:

1. **Self-contradictory PR description** — fixed. The body now accurately describes both the docs rewrite and the Go-touching `feat(slack)` rename (the stale "no Go code touched" note is gone).
2. **Inconsistent "busy" copy** — unified behind a single `modalBusyMsg` const used at all five modal-error sites (the feedback path's "Try again" is now "Retry"). `ackBusy` stays separate since it's the slash-ack surface (`:warning:` + "please retry"), not a modal error.
3. **`.gitignore` as a separate chore** — fair, but keeping it here: the lockfile was swept in by a `git add -A` *on this branch*, so the fix + guard belong with the branch that introduced it. Scoped narrowly so it doesn't shadow the tracked `.claude/skills/`.
- **`dm:true` listed unconditionally** — intentional for a canonical customer doc; the *"Some commands aren't in `/qurl help`"* FAQ covers deploys without `PostDM`.

PR #616, opened 2026-06-07
docs(discord): focus README on customer-facing qURL commands

justin-layerv:
## What

Rewrites `apps/discord/README.md` to be a **customer-facing** doc for the qURL
sharing product, and removes content that does not apply to customer
deployments.

## Why the removed commands are not a regression

The old README documented ten community-bot commands — `/link`, `/unlink`,
`/whois`, `/contributions`, `/stats`, `/leaderboard`, `/forcelink`, `/bulklink`,
`/unlinked`, `/backfill-milestones`. These are **OpenNHP-community features**,
gated behind `ENABLE_OPENNHP_FEATURES` and registered **only** for the internal
community guild. For every customer deployment (multi-tenant or single-guild),
the bot registers only the `/qurl` command set:

```js
// apps/discord/src/commands.js
const CUSTOMER_SAFE_COMMANDS = new Set(['qurl']);

function getActiveCommands() {
  return config.isOpenNHPActive
    ? commands
    : commands.filter(cmd => CUSTOMER_SAFE_COMMANDS.has(cmd.data.name));
}
```

A customer never sees those commands in their slash-command picker, so a
customer-facing README documenting them was the bug. The OpenNHP community bot
keeps every feature it has today — no behavior changes.

## Critical rating of the old README

**3/10 as a customer-facing doc.** It was accurate-ish for an internal operator
but actively misleading and buried for a customer:

- **Led with the wrong product.** Opened on "GitHub OAuth linking and auto
  Contributor-role assignment" — not the qURL sharing the bot exists to do.
- **Documented 10 commands customers can't run** (the OpenNHP set above).
- **~50 lines of internal operational detail** a customer/operator of the qURL
  product doesn't need: local DynamoDB provisioning gymnastics, gateway
  RESUME/hot-standby/SQS internals, internal DDB table names, `host.docker.internal`
  Linux notes, Jest coverage thresholds (`78/68/78/78`), a module-by-module
  source listing.
- **No clear customer path** — three audiences mixed together with no "what is
  this / how do I use it" thread.
- **Buried the genuinely good features** (one-time links, expiry, self-destruct,
  personal message, private DM delivery) under setup minutiae.

## What the new README does (target: 10/10)

- Leads with what the bot does and a 3-step customer getting-started flow
  (invite → `/qurl setup` → share).
- Documents only the customer-safe `/qurl` commands, with an options table.
- Notes `/qurl map` as *"where enabled"* (`MAP_COMMAND_ENABLED` is default-off).
- Describes `/qurl setup` path-agnostically — admins authorize qURL or enter an API key depending on the deployment. The bot ships both an OAuth flow and a legacy modal-paste fallback (the fallback is active until Auth0 is registered in prod, per `src/config.js:366`).
- Calls out the **Server Members Intent** privileged gateway intent the bot
  requires to boot (asserted in `src/discord.js`) — an operator-critical fact the
  old README mentioned but the first draft of this rewrite dropped.
- Marks `QURL_API_KEY` as optional (not production-required): `prodRequired()` in
  `src/boot-requirements.js` only demands it in OpenNHP mode; customer servers
  connect their own key via `/qurl setup`.
- Trims `## Configuration` to the variables most deployments need and points to
  `.env.example` for the full reference (nothing is lost — `.env.example` keeps
  every advanced/OpenNHP knob, fully documented inline).
- Keeps a trimmed Development / Architecture / Troubleshooting at the same
  altitude as `apps/slack/README.md` (house style).
- Uses `qURL` brand casing throughout (per `CLAUDE.md`).

## Also: removes a dead npm script

The README documented `npm run register`, but its target
`src/register-commands.js` no longer exists — the file is untracked in git and
`npm run register` could only error. Slash commands register automatically at
boot (`registerCommands` in `src/index.js`). This PR removes the dead `register`
script from `package.json` and the README note that referenced it.

## Scope

`apps/discord/README.md` plus the one dead-script line in
`apps/discord/package.json`. `.env.example` and `apps/discord/docs/*.md`
(internal design docs) are intentionally left untouched. No source or test
changes; all 79 suites / 2652 tests pass locally.

### DIFF
--- a/apps/discord/README.md
+++ b/apps/discord/README.md
@@ -1,216 +1,160 @@
 # qURL Discord Bot
 
-Discord bot for qURL-powered secure resource sharing, plus GitHub OAuth
-linking and auto Contributor-role assignment for community members.
+Share files and locations in Discord as **one-time, expiring qURL links** —
+delivered privately to each recipient's DMs, never posted in the channel, and
+revocable at any time.
 
 ## Features
 
-- **`/qurl send`** — share a file as a one-time qURL link, delivered to each
-  recipient via DM. Recipients picked via @mentions or a user-select menu.
-- **`/qurl map`** — share a Google Maps location as a one-time qURL link,
-  delivered to each recipient via DM.
-- **`/qurl revoke`** — revoke all links from a previous send.
-- **`/qurl help`** — command reference.
-- **`/qurl setup`** / **`/qurl status`** — admin-only, configure the
-  guild's qURL API key (stored AES-256-GCM encrypted at rest).
-- **GitHub OAuth Linking**: `/link` verifies GitHub identity; the callback
-  is session-cookie-bound to prevent leaked-URL takeover.
-- **Auto Role Assignment**: merged PRs in allowed orgs award the
-  `@Contributor` role automatically.
-- **Contribution Tracking + Badges**: first PR, docs hero, bug hunter,
-  on-fire, streak master, multi-repo — awarded on merged PRs.
-- **Good-first-issue feed + release announcements + star milestones**
-  in configurable channels.
+- **One-time links** — each recipient gets their own link that works exactly
+  once.
+- **Private delivery** — links arrive as a DM, never in the channel.
+- **Expiry & self-destruct** — links expire (default 24 hours) and can start a
+  countdown after the first open.
+- **Personal message** — attach a note shown to each recipient.
+- **Revoke anytime** — kill every link from a previous share with one command.
+- **Per-server setup** — each server connects its own qURL account; keys are
+  encrypted at rest.
 
 ## Commands
 
 | Command | Description |
 |---------|-------------|
-| `/qurl send` | Send a file as one-time qURL links to picked recipients |
-| `/qurl map` | Send a Google Maps location as one-time qURL links to picked recipients |
-| `/qurl revoke` | Revoke links from a previous send |
-| `/qurl help` | Usage reference |
-| `/qurl setup` | *(admin)* Configure the guild's qURL API key |
+| `/qurl send` | Share a file as one-time qURL links, DM'd to the recipients you pick |
+| `/qurl map` | Share a Google Maps location as one-time qURL links *(where enabled)* |
+| `/qurl revoke` | Revoke every link from a previous send |
+| `/qurl help` | Show the command reference |
+| `/qurl setup` | *(admin)* Connect this server to qURL |
 | `/qurl status` | *(admin)* Check whether qURL is configured |
-| `/link` | Link your GitHub account to Discord |
-| `/unlink` | Unlink your GitHub account |
-| `/whois [@user]` | Look up a member's GitHub handle |
-| `/contributions [@user]` | Show a member's merged-PR count + badges |
-| `/stats` | Bot-wide contribution statistics |
-| `/leaderboard` | Top contributors |
-| `/forcelink` | *(admin)* Manually link a Discord user to a GitHub username |
-| `/bulklink` | *(admin)* Bulk-link from a `discordId:github,...` list |
-| `/unlinked` | *(admin)* List contributors who haven't linked |
-| `/backfill-milestones` | *(admin)* Re-announce star milestones |
 
-## Setup
+### `/qurl send` options
 
-### Prerequisites
+| Option | Required | Description |
+|--------|----------|-------------|
+| `attachment` | Yes | The file to share |
+| `recipients` | No | Paste `@mentions`. Leave blank to pick from a menu. |
+| `expires-in` | No | How long the links stay valid (default: 24 hours) |
+| `self-destruct` | No | Countdown after the first open (default: no timer) |
+| `personal-message` | No | A note included in each recipient's DM |
 
-- **Node.js ≥ 22** (see `package.json` engines)
-- A Discord bot application
-- A GitHub OAuth App
-- A hosting target with a public HTTPS URL (ECS, Railway, Fly, etc.)
-- A qURL API key from https://layerv.ai
+`/qurl map` shares a location instead of a file: it takes a required `location`
+(a Google Maps URL, or a place/address to search) in place of `attachment`, the
+same `recipients` / `expires-in` / `self-destruct` / `personal-message` options,
+and an optional `location-name` to override the label recipients see.
 
-### 1. Configure Discord
+## Getting started
 
-1. https://discord.com/developers/applications → your bot
-2. Enable **Server Members Intent** under Bot → Privileged Gateway Intents.
-3. Copy the bot token.
+### 1. Add the bot to your server
 
-### 2. Configure GitHub OAuth
+Invite the qURL bot using the install link from your qURL operator. The bot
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks for the thorough source-cross-check. Addressing each nit:

**1. Brand casing** — confirmed no lowercase `qurl`-in-prose (`grep -n qurl README.md | grep -v qURL` is clean; the only lowercase hits are code/paths/env-vars/`/qurl` command names, which `CLAUDE.md` keeps literal). "backed by DynamoDB" is the AWS product name, not a brand slip. No change needed.

**2. `DISCORD_CLIENT_ID` marked "Yes"** — keeping "Yes". You're right it isn't a `boot-requirements.js` blocker, but it's genuinely required for a functioning deployment: it's the OAuth `client_id` used to build the bot install/invite link (`src/routes/discord-install.js:153`), which is exactly the "add the bot to your server" step the Getting Started section depends on. A footnote about the boot-requirements nuance would re-introduce the kind of internal detail this rewrite is trimming, so I'd rather keep the table simple and accurate-in-practice.

**3. `npm test  # jest`** — kept as-is per your note; the comment names the tool, not its flags.

**Beyond the nits**, the cross-check surfaced a 4th issue I fixed: the README documented `npm run register`, but its target `src/register-commands.js` is gone (untracked in git), so the script could only error — slash commands register automatically at boot via `registerCommands` in `src/index.js`. This PR removes the dead `register` script from `package.json` and the README line that referenced it. Details in the updated PR description.

All 79 test suites / 2652 tests pass locally; lint clean.

justin-layerv:
Addressed the optional nit: scoped the "up to an hour" caveat to the multi-tenant public bot and noted that single-server deployments register per-guild and appear right away. Pushed in `ce4a1e0d`.

justin-layerv:
Both optional nits addressed in `ef8a5291`:

1. **Map enablement** — added a `MAP_COMMAND_ENABLED` row to the Configuration table (`Set to \`true\` to enable \`/qurl map\`, default off`) and clarified `GOOGLE_MAPS_API_KEY` as "needed when map is enabled", so the "where enabled" hedge is now actionable from the README.
2. **Setup callback wording** — reworded the Architecture HTTP-surface bullet from "the \`/qurl setup\` callback" to "the OAuth callback that completes the \`/qurl setup\` flow", so it reads as a purpose-level description rather than a literal path.

justin-layerv:
Addressed in `8173e9ba`:

1. **"Slash commands don't appear" troubleshooting** — reworded to lead with the actual cause (global registration propagation, ~1h on first invite; single-server installs appear immediately) and to point at the **Use Application Commands** permission, rather than the View/Send/Embed perms which govern whether the bot can *respond*, not whether commands *appear*.
2. **"encrypted at rest" in Getting Started** — leaving as-is intentionally: Getting Started is the customer/production path, where keys are always encrypted. The dev-only plaintext exception is documented in the Configuration section, which is the right place for that operator detail. No real inconsistency for the customer audience.

justin-layerv:
Good catch on `BASE_URL` — verified and addressed.

The qURL OAuth routes are mounted unconditionally (`src/server.js:229`) and `/qurl setup` takes the OAuth path whenever `isQurlOAuthConfigured`, so a customer deployment using guided setup does consume `config.BASE_URL` (defaulting to `localhost:3000`). Added a `BASE_URL` row to the Configuration table in `89bc59f9`, marked required for `/qurl setup` with the `https://` requirement and the "can't complete if left at the localhost default" caveat.

The reviewer also surfaced a real **code-side** gap that's out of scope for this docs PR: there's no boot guardrail requiring `BASE_URL` when OAuth setup is configured (the https check only fires in OpenNHP mode), and the `src/index.js:184` comment claiming "BASE_URL is unused in non-OpenNHP modes" is stale. Filed with repro + acceptance criteria: layervai/qurl-integrations#619 (`bug`, `priority: medium`). This PR deliberately doesn't touch boot validation.

The "Required" column mixing yes/no with conditional values (`Production`, `/qurl map`, `/qurl setup`) is intentional per your note — kept as-is since it's more informative than a plain Yes/No.

justin-layerv:
Addressed in `93ed5921`:

1. **`/qurl map` options** — reworded so it's clear map takes a required `location` *in place of* `attachment` (not in addition to it): "it takes a required `location` … in place of `attachment`, the same `recipients` / `expires-in` / `self-destruct` / `personal-message` options, and an optional `location-name`…".
2. **"Required" column legend** — added a one-line legend under the table: Yes/No = always/never; **Production** = required when `NODE_ENV=production`; a command name = required for that feature. Removes the scan-time ambiguity.
3. **KEK outside production** (`missingKekRequiredKeys` + `GITHUB_CLIENT_SECRET`) — leaving "Production" as-is per your note; that path only applies to OpenNHP/community staging, which this customer-facing doc intentionally excludes.

justin-layerv:
Good nuance — addressed in `2aab498b`.

You're right that `/qurl setup` isn't purely guided-OAuth today: `isQurlOAuthConfigured = Boolean(AUTH0_*)` (`src/config.js:264`), and when those are unset it falls back to the legacy modal-paste path (`src/config.js:366` comment: "falls back to the legacy modal-paste path … until Justin registers the Auth0 application + drops these into prod SSM"). So a current customer may well get the paste UX.

Reworded the README to be path-agnostic rather than over-committing to one UX: "runs `/qurl setup` once and follows the prompts to connect this server to its own qURL account — by authorizing qURL or entering an API key, depending on the deployment." Updated the PR-description bullet to match (it previously said "not a paste UX", which was the imprecise bit you flagged).

On the `BASE_URL` "must be https in production" point — agreed the new "guided setup can't complete at the localhost default" framing is more precise given there's no hard boot enforcement; the missing guardrail is tracked in layervai/qurl-integrations#619.

justin-layerv:
Addressed in `e9ecec8a`. Scoped the `BASE_URL` row to the OAuth setup path: "Builds the OAuth `/qurl setup` callback, so the OAuth setup path can't complete at the `localhost` default. (The modal-paste setup fallback doesn't use it.)" — and set its Required marker to "OAuth setup" so it doesn't imply the modal-paste path needs it. Generalized the Required-column legend to cover feature labels like `/qurl map` and OAuth setup.

justin-layerv:
Both optional nits addressed in `b015336d`: moved the Required-column legend *above* the table so the key precedes the cells, and trimmed the dense `BASE_URL` row to a single scannable line ("Public `https://` origin of the bot; required to complete the OAuth `/qurl setup` flow (defaults to `localhost`)"), with the modal-paste detail left to `.env.example`.

justin-layerv:
Addressed in `1c80d5a8`:

1. **`BASE_URL` default precision** — changed "defaults to `localhost`" to the exact `http://localhost:3000` (`src/config.js:381`).
2. **`DISCORD_CLIENT_ID` "Yes"** — leaving as-is per your note. It's required for the documented default (the multi-tenant public bot uses it as the install-flow OAuth `client_id`, `src/routes/discord-install.js:153`); the manual-invite single-guild edge case that can boot without it isn't the path this customer-facing README centers on, and keeping the column a simple Yes/No there reads cleaner than a conditional caveat.

PR #617, opened 2026-06-07
docs(chrome-extension): rewrite README as customer-facing; add dev guide

justin-layerv:
## What

The `apps/chrome-extension` README mixed three audiences (Gmail end-users, packagers, and maintainers) and leaked implementation internals. This makes the README customer-facing and moves the developer material into a sibling guide — the same split applied to `apps/slack` in #615.

### Customer-facing `README.md`
- Opens with a value prop and an **install → use** quickstart.
- A **Using the extension** walkthrough plus the caveats that actually trip users up (keep the popup open, Gmail must be the active tab, the 100 MB per-file cap, the copy fallback).
- A **"Pointing at a different qURL server"** guide for the settings panel.
- A **Troubleshooting** table mapped to the popup's real messages ("Could not find compose window", "Active tab is not Gmail").
- A short **Privacy** section.
- **No internal detail** — no Gmail CSS selectors, service-worker/content-script mechanics, multipart wire format, file-structure tree, Release Please seed mechanics, or the `https://*/*` permission rationale.

### New `docs/development.md` (developers)
- Build/test/lint, packaging, release, versioning, icons, plus pointers to the architecture (`DESIGN.md`), install + build-time override (`installation.md`), manual QA (`local-unpacked-testing.md`), and Web Store review note (`chrome-web-store-review.md`) — **moved, not deleted**. The README links here once.

### Fixes carried along
- `installation.md` referenced a `.env.example` that didn't exist. Added a committed `.env.example` for the build-time `QURL_API_BASE` override (mirrors `apps/discord`, with the matching `!.env.example` gitignore negation) and corrected the instruction.

## Why

Critical rating of the old README: **~4/10** — accurate but wrong altitude for a customer (it led the reader through a wire-level "How It Works" diagram, a `POST /api/upload` contract, a file-structure tree, and Chrome Web Store packaging before ever explaining how to *use* the thing), no troubleshooting, internal selectors and memory/implementation notes in user-facing prose, and a broken `.env.example` reference. The rewrite targets a 10/10 customer doc: correct audience, quickstart, usage walkthrough, real troubleshooting, zero implementation leakage — with the developer content preserved one click away.

## Notes for review
- **Preserve vs. delete:** developer content was *moved* into `docs/development.md`, not dropped. `DESIGN.md`, `local-unpacked-testing.md`, and `chrome-web-store-review.md` are unchanged (internal docs, where implementation detail belongs).
- Troubleshooting strings and the 100 MB cap were reconciled against `_locales/en/messages.json` and `lib/qurl-api.js`.
- No install URL was invented: there's no public Chrome Web Store listing in the repo, so install is framed as operator-provided / Load unpacked, mirroring slack's "install link your operator provided".
- Docs-only change; no extension code touched.

### DIFF
--- a/apps/chrome-extension/.env.example
+++ b/apps/chrome-extension/.env.example
@@ -1,0 +1,11 @@
+# qURL Chrome Extension — build-time configuration
+#
+# Copy this file to `.env`, set a value, then rebuild with `npm run release`
+# or `npm run package:release`. The build regenerates `lib/qurl-config.js` and
+# the manifest host permission from this value. Leave it unset to keep the
+# built-in production default.
+#
+# Only the build scripts read this file; the extension runtime does not.
+# The value must start with https://. Example sandbox value:
+#   QURL_API_BASE=https://getqurllink.layerv.xyz
+QURL_API_BASE=
--- a/apps/chrome-extension/.gitignore
+++ b/apps/chrome-extension/.gitignore
@@ -1,4 +1,6 @@
 node_modules/
 .env
+.env.*
+!.env.example
 release/
 dist/
--- a/apps/chrome-extension/README.md
+++ b/apps/chrome-extension/README.md
@@ -47,8 +47,7 @@ covered in the [developer guide](docs/development.md).
 A few things worth knowing:
 
 - **Keep the popup open while uploading.** Clicking elsewhere or switching
-  windows closes the popup and cancels any upload still in progress. The popup
-  shows a "keep this popup open" reminder while files are uploading.
+  windows closes the popup and cancels any upload still in progress.
 - **Gmail must be the active tab** in the focused window when you open the
   popup. A Gmail tab in a different window won't be picked up.
 - **Files are capped at 100 MB each.** Larger files are reported individually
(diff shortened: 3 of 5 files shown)


justin-layerv:
Thanks for the thorough review — addressed the nits from both passes:

1. **Privacy wording** (the "doesn't read … web pages" being too broad, since the content script does read the Gmail compose DOM): reworded in b19ba4e6 to focus on exfiltration — *"The only thing the extension uploads is the files you choose — it never sends the contents of the pages you browse."* This is verified accurate against `content/gmail-compose.js` (the content script only runs on `mail.google.com`, reads the compose DOM to insert links, and makes no network requests of its own; uploads originate solely from user file selection in the popup).

2. **Abbreviated troubleshooting strings**: kept intentionally as recognizable substrings of the full toasts (e.g. `qURL: Could not find compose window. Please open a compose window and try again.`). Quoting the full strings in a customer doc reads worse and would couple the README to exact punctuation; the substring is what a user scans for. If these strings start drifting upstream we can revisit, but they're stable today.

3. **`.xyz` sandbox vs `.ai` prod domain divergence**: by design — the production default is intentionally omitted from customer docs, and `.env.example` only carries the sandbox example, matching the comment in `lib/qurl-config.js`.

`/simplify` returns clean (reuse / simplification / references / altitude), and all checks are green.

justin-layerv:
Latest review nits addressed:

1. **Custom-server permission flow** — fixed in 930e0260 to describe both dialogs (the extension's own origin-naming confirmation, then Chrome's native host-permission prompt). Now verified accurate against `permission_request_confirm`.

2. **Install path uses Developer mode / Load unpacked** — this is the one spot where customer/developer altitude blurs, and it can't be cleanly resolved until there's a public Chrome Web Store listing to link (none exists today; no URL was invented by design). Tracked as a deferred follow-up: #618 — swap in the Web Store install link once the extension is listed.

3. **`.env.example` sandbox value / default URL omission** — intentional and non-secret, no action (noted in the thread above).

All four merge-ready criteria are green: DE-acceptable approach (mirrors #615), CI passing, cr complete with every nit/optional either fixed or tracked, and `/simplify` returns clean.

PR #626, opened 2026-06-07
test(ci): pin upload-time viewer_ttl_seconds acceptance in /qurl send e2e smoke (#283)

### LINKED ISSUE #283: Prod E2E smoke does not exercise viewer_ttl_seconds < 300s — session_duration regressions land green

justin-layerv:
## What

Adds e2e smoke coverage for the `viewer_ttl_seconds < 300s` (auto-destruct / session_duration) path on `/qurl send`.

The prod E2E smoke uploaded with **no** `viewer_ttl_seconds`, taking the connector's "use server default" branch and bypassing the `session_duration → qurl-service` contract entirely. A regression in that contract (a value below qurl-service's floor returning 400) landed green. The 2026-05-13 incident exercised exactly this gap.

## Change

- Optional `viewerTtlSeconds` on the `uploadFile` e2e helper (forwards `viewer_ttl_seconds` to the connector).
- New `qurl-send-viewer-ttl.smoke.test.ts` pinning both paths:
  - **TTL'd** (`viewer_ttl_seconds=30`) — the regression-class catcher.
  - **No-TTL control** — so a refactor that breaks the "use server default" branch also fails loudly.

## Provenance

Split out of the stale contractor PR #300 (which bundled this with unrelated Discord observability work, issue #276). This half has **no infra dependency** and ships on its own. The #276 half is being handled separately (it's blocked on a CloudWatch alarm that was never filed).

## Honest status

`tsc --noEmit` clean; the helper signature and assertions are correct. The smoke itself needs live infra to **run** — it is compile-validated here, not executed in this PR. It runs in the prod e2e-smoke pipeline.

Closes #283.

---
**⚠️ Before merge (cr follow-up):** this smoke is compile-validated but has never gone green (e2e tests run only in the prod e2e-smoke pipeline, not PR CI). Recommend one sandbox run first to confirm the connector's `/upload` actually accepts the `viewer_ttl_seconds` multipart field — so a green here means real coverage, not an unverified placeholder.

### DIFF
--- a/e2e/helpers/qurl-api.ts
+++ b/e2e/helpers/qurl-api.ts
@@ -19,26 +19,50 @@ export interface LinkAccessResult {
   body?: string;
 }
 
-/** Upload a file to the connector and get a resource_id */
+/** Upload a file to the connector and get a resource_id.
+ *
+ * `viewerTtlSeconds` forwards as `session_duration` (the field #283 pins).
+ * Handles the connector's documented 200-with-error-body convention like
+ * google-maps/file-revoke's `uploadImage`: a transient 429 (HTTP 200 +
+ * `error` string + no `resource_id`) is retried with backoff; any other
+ * missing-`resource_id` body throws WITH the connector's `error` string so a
+ * real regression is legible, not a bare "expected undefined to be truthy". */
 export async function uploadFile(
   uploadUrl: string,
   filePath: string,
   apiKey: string,
-): Promise<{ resource_id: string; hash: string }> {
+  opts?: { viewerTtlSeconds?: number },
+): Promise<{ resource_id: string }> {
   const fileBuffer = fs.readFileSync(filePath);
   const fileName = path.basename(filePath);
+  const maxAttempts = 4; // align with google-maps/file-revoke uploadImage
+  const baseDelayMs = 1500;
 
-  const formData = new FormData();
-  formData.append('file', new Blob([fileBuffer]), fileName);
+  for (let i = 0; i < maxAttempts; i++) {
+    const formData = new FormData();
+    formData.append('file', new Blob([fileBuffer]), fileName);
+    if (opts?.viewerTtlSeconds !== undefined) {
+      formData.append('viewer_ttl_seconds', String(opts.viewerTtlSeconds));
+    }
 
-  const res = await fetch(`${uploadUrl}/upload`, {
-    method: 'POST',
-    headers: { Authorization: `Bearer ${apiKey}` },
-    body: formData,
-  });
+    const res = await fetch(`${uploadUrl}/upload`, {
+      method: 'POST',
+      headers: { Authorization: `Bearer ${apiKey}` },
+      body: formData,
+    });
 
-  if (!res.ok) throw new Error(`Upload failed: ${res.status} ${await res.text()}`);
-  return res.json() as Promise<{ resource_id: string; hash: string }>;
+    if (!res.ok) throw new Error(`Upload failed: ${res.status} ${await res.text()}`);
+    const data = (await res.json()) as { resource_id?: string; error?: string };
+    if (data.resource_id) return { resource_id: data.resource_id };
+
+    const errStr = typeof data.error === 'string' ? data.error : '';
+    if (/429|rate.limit|too many requests/i.test(errStr)) {
+      await new Promise((r) => setTimeout(r, baseDelayMs * (i + 1)));
+      continue;
+    }
+    throw new Error(`Upload returned no resource_id: ${JSON.stringify(data)}`);
+  }
+  throw new Error(`uploadFile: still rate-limited after ${maxAttempts} attempts`);
 }
 
 /** Mint a one-time qURL link for a resource */
--- a/e2e/tests/qurl-send-viewer-ttl.smoke.test.ts
+++ b/e2e/tests/qurl-send-viewer-ttl.smoke.test.ts
@@ -1,0 +1,109 @@
+/**
+ * Smoke for the auto-destruct UPLOAD path (qurl-integrations#283).
+ *
+ * Pins the upload-time `session_duration` contract the 2026-05-13 incident
+ * broke: the connector forwards an upload's `viewer_ttl_seconds` to
+ * qurl-service as `session_duration` (connector.js `appendViewerTtl`), and a
+ * sub-floor value was rejected for hours, un-caught, because the only smoke
+ * took the no-TTL "use server default" branch. This sends a TTL'd upload
+ * (must be accepted) plus a no-TTL control. Full incident write-up + design
+ * rationale are in the PR description and #283.
+ *
+ * Honest boundary: catches a floor regression only IF qurl-service enforces
+ * the floor at upload (CreateQURL). The bot also forwards session_duration
+ * at MINT; pinning that needs a connector-mint helper + live run — #631.
+ * `30` is a future-regression guard kept clear of today's 1s
+ * `MinSessionDuration` floor (tighten to 1 once #631's live run confirms 1 is
+ * accepted). Does NOT catch: mint-time floor (#631), per-recipient
+ * view-window (Traefik), or the UI countdown.
+ */
+
+import * as dotenv from 'dotenv';
+import * as fs from 'fs';
+import * as os from 'os';
+import * as path from 'path';
+dotenv.config({ path: path.resolve(__dirname, '..', '.env') });
+
+import { loadEnv } from '../helpers/env';
+import * as qurl from '../helpers/qurl-api';
+
+const env = loadEnv();
+
+describe('Smoke: /qurl send with viewer_ttl_seconds (Snapchat path)', () => {
+  // Per-test unique fixture content to avoid the connector's md5
+  // content-addressed dedup shortcut. If both tests uploaded the
+  // same bytes, the second call would short-circuit to the cached
+  // resource from the first call (uploaded WITH viewer_ttl_seconds=30)
+  // and the "control" no-TTL case would silently fail to exercise
+  // the "use server default" branch it's supposed to pin.
+  const tempFiles: string[] = [];
+  const resourceIds: string[] = [];
+  function freshFixture(label: string): string {
+    // os.tmpdir() rather than __dirname so a killed-mid-test run doesn't
+    // leak fixture files into the e2e source tree (and from there into
+    // `git status`). afterAll cleanup is still best-effort.
+    const f = path.join(
+      os.tmpdir(),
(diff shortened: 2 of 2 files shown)


justin-layerv:
Addressed the review + ran /simplify:

- **Main concern (upload-only doesn't reproduce the mint-time incident):** took option (a) with verified citations + filed a tracked follow-up for option (b). The upload path **does** forward `session_duration` (connector.js appends `viewer_ttl_seconds` to POST /api/upload → CreateQURL; time.js:154), which is the field infra#540 added and the incident broke — so this smoke pins that upload-time contract. I added an honest boundary in the docstring: it catches a floor regression *iff* qurl-service enforces at CreateQURL; mint-time enforcement needs a connector-mint e2e helper + live run, tracked in #631.
- **Nit — misleading name ("mints a link" but never mints):** renamed to 'accepts + forwards viewer_ttl_seconds=30 at upload'.
- **Nit — 30 isn't sub-floor today:** documented as a future-regression guard (today's floor is 1s).
- **Nit — 'minted WITH viewer_ttl_seconds=30' dedup comment:** the dedup is upload-cache only; changed to 'uploaded WITH'.

/simplify came back clean (only suggestion was extracting a 2-line shared assertion helper — kept inline for smoke self-containedness, matching house style). tsc clean.

justin-layerv:
Round 2: renamed the TTL'd test to `connector accepts upload with viewer_ttl_seconds=30` — you're right that the assertions only observe a 2xx accept, not forwarding (which is verified only transitively via a qurl-service floor rejection). Added a one-line note at the test making that explicit, and trimmed the assertion comment to keep just the implicit-throw rationale.

justin-layerv:
Round 3 (nits):
- **Never-executed concern:** added a '⚠️ Before merge' note to the PR body recommending one sandbox run to confirm the connector accepts the `viewer_ttl_seconds` field before this counts as real coverage (can't run e2e from PR CI here).
- **30 vs 1 (loose guard):** justified the choice in the docstring — 1 (the `MinSessionDuration` floor) would be the tightest guard but risks flaking on boundary semantics I can't verify from here; 30 catches the incident-scale (5-min) recurrence either way, and #631's live run can tighten to 1 if 1 proves safely accepted.
- **Rot-prone line numbers:** swapped `time.js:55`/`connector.js:367` etc. for symbol refs (`appendViewerTtl`, `formatSessionDurationSeconds`, `MinSessionDuration`).
- **No-TTL control marginal value:** kept as the colocated dedup-defeat + use-server-default control (you noted it's fine to keep).

justin-layerv:
Round 4: added a best-effort `afterAll` revoke (tracks each upload's `resource_id`, revokes via `revokeLink(MINT_API_URL, ...)` before unlinking fixtures, all swallowed) — stops the new path leaking two S3 resources per run. Left the present-day-signal / value-tightening items as deferred to #631 (you confirmed that's fine), and the dotenv redundancy as-is for consistency with smoke.test.ts.

justin-layerv:
Round 5:
- **hash assertion risk:** good call — made `resource_id` the load-bearing assertion (`toBeTruthy()`, it's what proves an accepted upload + what afterAll revokes) and softened `hash` to `if (result.hash) expect(...)`. Since this is the first consumer of `uploadFile`, a hash key/shape surprise on the first live run no longer fails the contract this smoke actually pins.
- **dead-defensive guard:** dropped the `if (result.resource_id)` guard on the push — `uploadFile` throws on `!ok` and its return type guarantees the field.
- **tighten to 1:** added it to #631's checklist (once a live run confirms 1 is accepted, 30→1 catches any floor above 1s).

justin-layerv:
Round 6 (the substantive trio): made `uploadFile` handle the connector's documented 200-with-error-body convention, mirroring `google-maps`/`file-revoke`'s `uploadImage`:
- **429 false-RED flake:** a transient rate-limit comes back as HTTP 200 + `error` string + no `resource_id`. The helper now retries those with backoff (`maxAttempts` default 3) instead of failing RED in a pipeline that pages.
- **Detection mechanism was mis-documented:** corrected — a sub-floor session_duration may come back as a wrapped 200+error, not a 4xx. The helper now throws `Upload returned no resource_id: {error: ...}` on any non-429 missing-resource_id body, so the regression surfaces via a THROW (with the connector's message) either way. Comment rewritten to match.
- **Dropped `data.error` → poor diagnostics:** fixed — the throw now includes the connector's `error` string instead of a bare 'expected undefined to be truthy'.

tsc clean. (On #928's dashboard-widget suggestion: noted as optional; the alarm is the deliverable.)

justin-layerv:
Round 7:
- **hash likely-dead assertion:** good evidence (siblings read `resource_url`/`resource_id`, never `hash`). Dropped `hash` from the helper return type and both assertions — `resource_id` is the load-bearing contract (and what afterAll revokes). If a sandbox run later shows /upload returns something useful, we add it back deliberately.
- **comment-to-code ratio:** trimmed the docstring from ~48 lines to ~16 — kept what it pins, the honest boundary, the `30`/#631 note, and the not-catches; moved the incident essay to the PR description + #283 (where it already lives).
- **429 regex drift:** aligned to the siblings' `/429|rate.limit|too many requests/i`.
- Three-copies-of-retry-logic + jitter-less backoff: noted as out-of-scope follow-ups (the shared helper now exists to migrate toward).

justin-layerv:
Round 8 (final nits):
- **YAGNI `maxAttempts`:** dropped from `opts` (no caller passes it) — hardcoded per CLAUDE.md 'no features beyond the task.'
- **backoff alignment:** `maxAttempts=4` / `baseDelayMs=1500` to match `file-revoke`/`google-maps` `uploadImage`, so flake behavior is uniform across the suite.
- **helper docblock:** trimmed from 18 lines to the contract (the 'why' lives in the test + PR body).

Verified earlier: `uploadFile` had zero callers on main, so the return-narrowing + signature change break nothing. Remaining open item is the documented merge gate — one sandbox run to confirm the public `/upload` honors the `viewer_ttl_seconds` field (in the PR body).

justin-layerv:
Round 8 follow-up: renamed the PR title from 'pin viewer_ttl_seconds < 300s' to 'pin upload-time viewer_ttl_seconds forwarding' — you're right that '< 300s' overstates what the test catches today (with the 1s floor, 30 is accepted; the <300s-class catch is the future-regression guard tracked in #631). The tautological-assertion and retry-duplication points are acknowledged/deferred as noted (the real detection is the throw-with-context; retry-dedup is a follow-up now that the shared helper exists).

justin-layerv:
Two corrections from a final review pass:
- **Converted to draft.** It was the odd one out (READY) while #627/#928 are correctly drafts — and the conditional approval is 'once a sandbox run confirms /upload honors the field.' Since with today's 1s floor the TTL'd and control cases assert identically and the smoke has never executed live, marking it draft is the honest state. The one sandbox run (which I can't do from here) is the merge gate; it's in the PR body.
- **Title 'forwarding' → 'acceptance.'** You correctly had me rename the *test* from '+forwards' to 'accepts' because it can't observe forwarding; the title was reintroducing the same overclaim. It now says 'acceptance.'

justin-layerv:
**Contract verified against the connector source** (live run blocked — see below). Read `qurl-s3-connector/internal/handler/handler.go`:
- `/api/upload` parses `viewer_ttl_seconds` (handler.go:1580) → `sessionDurationFor` → forwards as `session_duration` to qurl-service's CreateQURL. **The field is real and honored.**
- Success response includes `resource_id` (+ `hash`, `success`, …) — the smoke's `resource_id` assertion is correct.
- **A CreateQURL rejection (the 2026-05-13 sub-floor-session_duration shape) surfaces as HTTP 200 + `success:true` + `error` + NO `resource_id`** — *not* a 4xx. So `uploadFile` catches it via throw-on-missing-resource_id (the round-6 fix), confirmed correct. The upload-only smoke DOES catch a re-tightened-floor regression.
- Floors: connector 0.5s–3600s, qurl-service 1s–24h → tightening to `viewer_ttl_seconds=1` (#631) is safe at both layers.

**Live run access limitation:** I can't execute the smoke end-to-end from here — the sandbox connector + `QURL_API_KEY` live in AWS account `730883236711`, and my access is a standalone account (`084375564552`) with no path to it. The static verification above is conclusive for the *contract*; a live green still needs either a sandbox `QURL_API_KEY` (for a minimal probe against `getqurllink.layerv.xyz/api/upload`) or a run of the e2e-smoke pipeline against this branch.

justin-layerv:
**Live sandbox run — 2/2 green. Merge gate cleared.** (I have profile access to the sandbox account 730883236711.)

Ran `qurl-send-viewer-ttl.smoke.test.ts` against the real sandbox connector:
```
✓ connector accepts upload with viewer_ttl_seconds=30 (567 ms)
✓ connector accepts no viewer_ttl_seconds (control case) (230 ms)
```
Direct probe of `POST getqurllink.layerv.xyz/api/upload` (viewer_ttl_seconds=30) → `200 success:true`, returns `resource_id` (r_…) + 32-char md5 `hash` + `qurl_link`; `DELETE /v1/resources/{id}` → `204` (cleanup works); sub-floor `0.3` → `400 success:false` ('must be …at least 0.5'). The connector contract the smoke depends on is confirmed end-to-end. Flipped out of draft.

PR #627, opened 2026-06-07
feat(discord): audit-event observability on /qurl send link-creation failures (#276)

### LINKED ISSUE #276: Bot "Failed to create links" path doesn't log or increment any error metric

justin-layerv:
## What

Adds a `QURL_SEND_CREATE_LINK_FAILURE` audit event to the discord bot's "Failed to create links" catch blocks so a sustained mint/upload outage becomes metric-filterable. Closes the observability gap from issue #276 (the 2026-05-13 incident ran ~4h before anyone noticed).

**This is the rebased, re-tested replacement for the stale PR #300** (contractor, ~1 month old, `CONFLICTING`). #300 bundled this with an unrelated e2e smoke (issue #283) — that half shipped separately as #626. Once this merges, #300 can be closed.

## Change

- New `QURL_SEND_CREATE_LINK_FAILURE` audit event + `classifyMintFailure` (reason taxonomy: `timeout` / `upstream_4xx` / `upstream_5xx` / `unknown`).
- A single `emitMintFailureAudit` helper owns the emission contract across all three catch sites (initial `executeSendPipeline` send + `addRecipients` file + `addRecipients` location), so the `quota_exceeded` skip (a viral upload is a normal condition, not a page) and field shape can't drift. `activeKind` labels mixed-send failures by the branch that actually failed.

## Rebase notes

Rebased onto current main. Two drifts since #300 branched, both handled:
- `mintLinks` options-bag (#484) + the `newResourceIds`→`newLinks` rename (#465) moved the catch sites — the emitter re-targets cleanly; the return shape kept as `newLinks` (the stale diff carried `newResourceIds`, which would feed `undefined` to the view counter).
- #300's emission tests targeted `qurl-send*.test.js`, deleted in #313/#315/#319 — re-homed into `send-pipeline-back-half.test.js`.

## Tests

- All three emission sites **asserted** (not just executed): primary `executeSendPipeline` catch (`kind` via `kindMap[resourceType]`), both `addRecipients` catches (`kind` via `activeKind`), plus the `quota_exceeded` skip at each. Classifier taxonomy pinned in `classify-mint-failure.test.js` (14 cases).
- Full discord suite green: **81 suites / 2716 tests**, lint clean.

## Merge coordination (paired with infra#928)

This is the **emitter half** of #276 and merges **first** — infra#928's alarm-verification (synthetic-match) requires this to be *deployed* before it can run. Sequence: merge + deploy this → merge + `terraform apply` infra#928 → run the live synthetic-match. Keep the un-alarmed window short (merge #928 promptly after this), and **do not close #276** until the alarm is confirmed live. The event is harmless-but-inert in the interim (logged, not yet alarmed).

Part of #276. Paired with layervai/qurl-integrations-infra#928. Supersedes #300.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -78,6 +78,80 @@ function largeSendThreshold() {
   return Math.min(LARGE_SEND_RECIPIENT_FLOOR, half || 1);
 }
 
+// classifyMintFailure maps a thrown error from the upload + mint phase
+// of /qurl send into a small enum of reason strings. Used as the
+// `reason` field on the QURL_SEND_CREATE_LINK_FAILURE audit event
+// (qurl-integrations#276) so the CloudWatch metric filter can split
+// failures by class — 4xx (likely client/contract problem) vs 5xx
+// (likely upstream outage) vs timeout (likely network/cold-start).
+//
+// Don't add per-API-code branches here — that explodes the metric
+// cardinality. New reasons should map to one of the existing
+// categories or earn a new top-level category with a clear
+// dimensional purpose (e.g. a future "rate_limit" if we add upstream
+// 429 handling).
+function classifyMintFailure(error) {
+  if (!error) return 'unknown';
+  // AbortError is ambiguous: undici/fetch raises it on deadline-fired
+  // aborts AND on user-cancellations. error.cause is the only reliable
+  // disambiguator — bucket as `timeout` only when cause corroborates a
+  // timeout signal; otherwise fall through to `unknown` so a future
+  // cancel-button adoption doesn't silently mis-bucket cancellations
+  // as timeouts.
+  if (error.name === 'AbortError') {
+    const causeStr = String((error.cause && (error.cause.message || error.cause)) || '');
+    // Word-anchored to drop strange concatenations like "rtimeout" or
+    // "timeouted" — but it does NOT defend against "not due to timeout"
+    // because "timeout" is still a complete word in that string. The
+    // false-positive class for negation-containing causes is acceptable
+    // because real timeout-driven aborts in undici raise TimeoutError
+    // (handled by the name/code branch below) rather than AbortError; this
+    // branch defends against libraries that use controller.abort('timeout')
+    // as their deadline signal, where the cause is the deliberate marker.
+    // This returns BEFORE the status check below — an abort's error.status
+    // (if it somehow carried one) is not meaningful, so abort disambiguation
+    // deliberately takes priority over status classification.
+    if (/\btimed?\s*out\b/i.test(causeStr)) return 'timeout';
+    return 'unknown';
+  }
+  // libuv socket codes + undici/fetch TimeoutError DOMException — all
+  // unambiguous deadline-shaped failures. The message-regex branch was
+  // removed: `/timeout/i.test(error.message)` over-matched
+  // any string containing the substring "timeout" (e.g. "not a timeout
+  // related error") and a future upstream message like "completed
+  // after timeout retry" would mis-bucket. The name/code paths above
+  // cover every real shape we've seen.
+  if (error.code === 'ETIMEDOUT' || error.code === 'ECONNABORTED' ||
+      error.name === 'TimeoutError') {
+    return 'timeout';
+  }
+  const status = error.status ?? 0;
+  if (status >= 500 && status < 600) return 'upstream_5xx';
+  if (status >= 400 && status < 500) return 'upstream_4xx';
+  return 'unknown';
+}
+
+// emitMintFailureAudit centralizes the QURL_SEND_CREATE_LINK_FAILURE
+// emission contract so the three catch sites (initial /qurl send +
+// addRecipients file branch + addRecipients location branch) cannot
+// drift on the skip rule or field shape. Adding a fourth call site
+// goes through here too — owns both the contract and the cardinality
+// discipline in one place. See #276.
+function emitMintFailureAudit(error, { sendId, kind }) {
+  if (error && error.apiCode === 'quota_exceeded') return;
+  logger.audit(AUDIT_EVENTS.QURL_SEND_CREATE_LINK_FAILURE, {
+    send_id: sendId,
+    reason: classifyMintFailure(error),
+    // `?? null` (not `|| null`) preserves an empty-string `apiCode` —
+    // `|| null` would collapse it to null and lose a forensic dimension.
+    // (Status 0 doesn't apply: native fetch throws on transport errors
+    // rather than returning a 0-status response.)
+    api_code: error?.apiCode ?? null,
+    status_code: error?.status ?? null,
+    kind,
+  });
+}
+
 // Shared helper: many Discord API calls (edits, updates, follow-ups) are
 // best-effort — if the interaction token expired or Discord is briefly
 // degraded, we log a warning and continue rather than fail the whole flow.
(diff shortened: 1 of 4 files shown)


justin-layerv:
Addressed the review + ran /simplify:

- **Pool-exhaustion still emits (consistency w/ quota_exceeded skip):** resolved on the alarm side. The paired alarm (infra#928) now counts **all** reasons and pages on a *sustained spike* (volume), not on reason class — so a rare pool-exhaustion event can't false-page, while the genuinely-viral `quota_exceeded` is skipped at source here so it never emits. Updated the constants docstring to say this explicitly (the alarm does not split on reason; reason/kind are forensic/dashboard).
- **`status_code` 'closed enum' wording:** fixed — `reason`/`kind` are closed enums; `status_code` is bounded HTTP-status but not a closed enum, documented as forensic/dashboard-only.
- **Comment volume:** /simplify trimmed the two clearly-redundant blocks (the incident narrative that's duplicated in the constants docstring, and the activeKind explanation duplicated across decl + catch site). The 4 cleanup agents found the code itself well-factored — no reuse/efficiency/altitude changes.

2716 tests green, lint clean.

justin-layerv:
Round 3 (nits): stripped the iteration-history labels (round-11/12, 'PR #300 review (Justin)') from comments and test names — kept the durable rationale (why the message-regex was dropped, why AbortError needs cause-corroboration). Trimmed the pool-exhaustion paragraph in constants.js to the load-bearing reasoning. Left the pre-existing `'Failed to prepare QURL links'` log string as-is (out of scope — I only added fields, and the infra connector_no_resource_id comment references that exact header string).

justin-layerv:
Round 4: good catch on the expiry asymmetry. **Skipped the expired-CDN-URL case at source** — the `addRecipients` inner file catch now gates `emitMintFailureAudit` on `!isExpired`, symmetric with the `quota_exceeded` skip (an expired Discord CDN URL on a >24h-old send is an expected, user-recoverable condition, not an outage signal for the alarm). Documented the carve-out in the constants docstring alongside quota/pool-exhaustion, and pinned both sides with tests: the 403/expired re-download asserts **no** emission; the generic 'something else broke' re-download asserts it **does** emit (reason=unknown). 2717 green, lint clean.

justin-layerv:
Round 5: real catch — fixed the regex coupling. The audit-skip no longer piggybacks on the broad user-message `isExpired` (which includes `network`). It now uses a dedicated narrow signal `isExpectedCdnExpiry = /403|expired|CDN/i` — a transient **network** blip during re-download/re-upload is an outage shape #276 must catch, so it still emits (and the user still gets the soft 'expired' message). Decoupled so a future edit to the user-message regex can't silently widen what's suppressed from the alarm. Updated the docstring + added a test pinning 'network error → expired message BUT emits'. 2718 green.

On the two minors: 429 pool-exhaustion contributing is the documented, conscious tradeoff (rare/auto-handled, threshold absorbs); comment density is consistent with the file's house style and the remaining blocks carry the non-obvious why.

justin-layerv:
Round 6: another real one — the skip was scoped to the wrong axis. The inner file `try` wraps **both** the CDN re-download and the mint, so a mint-phase 403 (revoked key / auth outage) matched `/403/` and got swallowed — exactly the outage #276 must catch, hidden on the addRecipients file path only.

Fixed by scoping to the **phase**: added a `downloadDone` flag, and the expiry skip is now `downloadFailure && /403|expired|CDN/` — i.e. only a CDN re-download failure with an expiry-shaped message. A mint-phase failure (any status) always emits; a `network` download blip still emits. This also fixes the pre-existing user-message twin you flagged (a mint 403 no longer tells the user 'attachment URL has expired'). Added a regression test: 'MINT 403 after successful re-download → still emits (reason=upstream_4xx, status_code=403), not swallowed'. 2719 green.

justin-layerv:
Round 7 (nits): consolidated the skip-rationale duplication — the constants.js `QURL_SEND_CREATE_LINK_FAILURE` docstring is now the canonical explanation of the skip carve-outs, and the catch-site comments point to it instead of restating 'normal product condition, not a page'. Filed #634 for the pre-existing network-UX copy decoupling (the audit side is already correctly decoupled; only the user message bucket is affected). The AbortError-string-cause branch is intentional defensive coverage (the comment frames it as such), left as-is.

justin-layerv:
Round 9 — real bug, good catch. My round-6 `downloadDone` phase flag was too coarse: `downloadAndUpload` bundles the CDN download AND the connector re-upload, so a connector **403 during the re-upload** (revoked key / auth outage) threw before `downloadDone` with `/403/` in its message → mis-bucketed as CDN-expiry and **swallowed** — exactly the #276 outage shape.

Fixed by discriminating on **origin, not phase position**: a connector error carries `err.status` (via `throwConnectorError`); a Discord CDN-download failure throws a bare Error with no status (`connector.js:305`). The skip now requires `!downloadDone && err.status == null` — so a connector 403 from the re-upload (or mint) emits, and the user no longer sees a misleading 'expired' message for it. Added your suggested regression test (`connector re-upload 403 → still emits, reason=upstream_4xx, status_code=403`); the genuine no-status CDN-403 expiry case still skips. 2720 green.

justin-layerv:
Round 10 — stepping back instead of patching again. The CDN-expiry **skip** I added (at the round-3 prompt) spawned three separate bug-rounds in the same spot (network-coupling, mint-403, re-upload-403). That's the signal: the skip itself was the problem. It's also inconsistent with the pool-exhaustion decision in this same PR — I argued *don't* skip rare conditions at source ("the threshold absorbs it, skipping risks suppressing real failures"), then did exactly that for CDN-expiry, and it suppressed real connector 403s three times.

**Deleted the skip entirely.** The addRecipients file catch now always emits (removed `downloadDone`, both phase regexes, `isCdnDownloadFailure`). A rare CDN-expiry (Add Recipients on a >24h-old send) is absorbed by the alarm's sustained threshold, exactly like pool-exhaustion — and `quota_exceeded` (the one genuinely high-volume condition) stays skipped in the helper. This kills the whole bug class and simplifies the code + tests (5 skip-centric tests → 3 behavioral ones: no-status→unknown, generic-message, status-carrying→classified). The pre-existing user-message wart (network → 'expired') is unchanged and tracked in #634. 2718 green, lint clean.

justin-layerv:
Round 11 (brand nit): fixed `per-QURL` → `per-qURL` in the constants doc-comment prose (CLAUDE.md). Left `'Failed to prepare QURL links'` as-is — pre-existing log string that the infra `connector_no_resource_id` comment references (out of scope, not introduced here).

On your other notes: (2) the connector-403-shows-'expired' user-copy divergence — expanded #634 to cover it (not just network). (4) confirmed the cardinality discipline holds — infra#928's filter pattern is `{ $.audit.event && $.audit.agent="discord" }`; it does NOT dimension on `api_code`/`status_code` (it counts all reasons, no reason/forensic split in the filter), so the high-cardinality forensic fields never become metric dimensions.

justin-layerv:
Un-drafted — it's ready for review/approval. Corrected the merge note: this is the **emitter half** and merges **first**, because infra#928's alarm-verification (the live synthetic-match) needs this deployed before it can run (the gate confirms 'against #627's merged commit / the deployed emitter'). The old 'do not merge until #928 lands' was backwards.

Sequence: approve + merge + deploy this → then un-draft + approve + apply infra#928 → run the live synthetic-match (the pattern is already AWS-verified; this just confirms the deployed bot produces it). The event is harmless-but-inert in between (logged, not yet alarmed) — keep that window short, and don't close #276 until the alarm is confirmed live.

Issue #462, opened 2026-05-19
ops: surface /qurl list disclosure-widening in Slack-app release notes (PR #459)


Kevin-layerV:
## Summary

PR #459 (`revert(slack): drop /qurl list admin gate and channel-policy filter`) widens what every workspace member can see in `/qurl list`:

- Target URLs of the 10 most recent Resources (e.g. `https://prod.example.com`) become visible workspace-wide.
- Resource IDs (`$r_<id>`) of un-aliased Resources are surfaced to all members, including IDs the caller cannot mint for in their current channel.
- Aliases (`$<alias>`) are surfaced workspace-wide even when bound only in other channels.

Capability boundary is unchanged — minting still routes through `AllowedResourceIDsForChannel` (`handler_get.go:368`) and `LookupChannelAlias` (`handler_get.go:332`). The widening is a *disclosure* change, not a *capability* change.

## Problem

release-please's default behavior for a `revert(slack)` commit is a single-line CHANGELOG entry from the commit subject. The detailed "Rollout note (disclosure widening)" in PR #459's body will NOT auto-propagate to `apps/slack/CHANGELOG.md` or to any customer-facing release notice. Workspaces that have been treating list-scoping as a confidentiality boundary on target URLs will not see this surfaced without an explicit comms step.

Surfaced in cr round-3 review of #459: <https://github.com/layervai/qurl-integrations/pull/459#issuecomment-4491087612>.

## Action items

- [ ] When the release-please PR for `apps/slack` fires after #459 merges, hand-edit the CHANGELOG entry to include the disclosure-widening rollout note from PR #459 (lines under `## Rollout note (disclosure widening)`).
- [ ] Surface the disclosure change in whatever customer/operator-facing release channel is in use for the Slack app (release notes, customer email, ops Slack channel — whichever matches the team's release cadence).
- [ ] (Optional, if friction reports come in post-deploy) consider routing follow-up #460 (alias-discoverability vs channel-scope mismatch) to address the most common confusion case.

## Acceptance

A workspace operator reading the `apps/slack` release notes for the version that ships #459 sees the disclosure-widening explicitly called out, not just `revert(slack): drop /qurl list admin gate and channel-policy filter`.

## Out of scope

- Reintroducing the channel-policy filter on `/qurl list` (the whole point of #459 is to drop it).
- Code changes to `/qurl list` output shape.

Kevin-layerV:
[...]

PR #642, opened 2026-06-08
docs(slack): document /qurl list disclosure model

Kevin-layerV:
## What

Adds a durable, in-repo reference doc for the `/qurl list` disclosure model, plus a pointer to it from the app README.

- New `apps/slack/docs/list-disclosure.md` — what `/qurl list` discloses to a workspace member and how that's bounded (channel-scoped, fail-closed, DM behavior, and the disclosure-vs-capability boundary).
- `apps/slack/README.md` "Channel scope" bullet gains a one-line pointer to the doc. The existing "`/qurl list` / `/qurl aliases` / `/qurl get` all agree on channel scope" wording is accurate and is left unchanged.

## Why

`/qurl list` is **channel-scoped and fail-closed** on `main` today (`listChannelScope` in `handler_list.go`): a member sees only the resources protected in the current channel — the same allow-set `/qurl get` mints against — DMs return the empty state, and there is no admin bypass. That model is enforced in code but wasn't documented anywhere durable, and its history is easy to misread:

- per-channel scoping was added in #234,
- **reverted in #459** (which briefly widened `/qurl list` to a workspace-wide list — the state #462 was opened against),
- **re-introduced in #589** — the current behavior (with a later paging-completeness fix, #590).

An operator who saw a #459-window release note ("`/qurl list` shows the full workspace list to every member") needs a durable, force-push-proof place that states the current model and notes the widening was already undone. release-please regenerates the CHANGELOG from commit subjects, so this doc is that home.

## Note on #462

#462 was filed to surface the #459 disclosure-*widening*. That widening was undone by #589 before this PR — so rather than warn about a widening that no longer exists, this doc documents the **current** channel-scoped model and notes the earlier widening was already reverted — without reciting the PR-by-PR history in the doc itself (per Justin's review) — which resolves the same operator confusion #462 was worried about. The widening-specific follow-ups named in #462 (a CHANGELOG annotation / customer comms about a *widening*) are therefore moot.

## Scope

Docs-only. No code changes to `/qurl list`.

Refs #462

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -32,12 +32,10 @@ workspace's Secure Access Agent supports.
 - **Alias** — an alternate name for a resource within a channel. Several
   aliases can point at one resource. Use an alias anywhere you'd use a `$id`.
 - **Channel scope** — resources are available per channel. A resource shows up
-  only in the channels it's been protected in. `/qurl aliases` and `/qurl get`
-  agree on what's available in a given channel — including for admins.
-  `/qurl list` is the exception: it shows the full workspace tunnel list to
-  every member, regardless of channel. Seeing a tunnel there does not grant the
-  ability to mint a link for it — minting is still gated per channel. See
-  [docs/list-disclosure.md](docs/list-disclosure.md).
+  only in the channels it's been protected in. `/qurl list`, `/qurl aliases`,
+  and `/qurl get` all agree on what's available in a given channel — including
+  for admins. For why this scoping is a confidentiality boundary and how it
+  fails closed (e.g. in DMs), see [docs/list-disclosure.md](docs/list-disclosure.md).
 - **Owner & admins** — the owner is whoever first connected qURL to the
   workspace. Owners and admins can run the `/qurl-admin` commands.
 
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -32,9 +32,12 @@ workspace's Secure Access Agent supports.
 - **Alias** — an alternate name for a resource within a channel. Several
   aliases can point at one resource. Use an alias anywhere you'd use a `$id`.
 - **Channel scope** — resources are available per channel. A resource shows up
-  only in the channels it's been protected in. `/qurl list`, `/qurl aliases`,
-  and `/qurl get` all agree on what's available in a given channel — including
-  for admins.
+  only in the channels it's been protected in. `/qurl aliases` and `/qurl get`
+  agree on what's available in a given channel — including for admins.
+  `/qurl list` is the exception: it shows the full workspace tunnel list to
+  every member, regardless of channel. Seeing a tunnel there does not grant the
+  ability to mint a link for it — minting is still gated per channel. See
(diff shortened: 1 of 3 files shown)


Kevin-layerV:
Good catch — you were right, and this was a real blocking correctness problem, not noise. I traced the history against `main`:

- #234 added per-channel scoping to `/qurl list`.
- #459 reverted it to workspace-wide. Issue #462 (and this doc's first draft) was written about that state.
- **#589 re-introduced channel scoping**, and #590/#596/#599/#602/#613/#615 built on it. So on current `main`, `/qurl list` is channel-scoped, admins included, and fails closed — `TestHandleList_ScopedToChannel` pins the DM case (`D_direct_msg_1to1` → `No protected resources are available in this channel`, no leak).

Fix pushed (141567e1):

- Rewrote `list-disclosure.md` to describe the current **channel-scoped + fail-closed** model (the inverse of the first draft): channel-only allow-set for everyone including admins; empty `channel_id` / nil admin store / read error / nothing-protected all fail closed; DM returns the channel empty state.
- Kept the capability-boundary section you flagged as accurate, reframed as defense-in-depth: disclosure and mint are governed by the *same* per-channel allow-set, and `/qurl get` re-checks at mint time.
- Added a **History** section (#234 → #459 → #589) so an operator reconciling the #459-era release note isn't misled.
- Restored the README `Channel scope` line to the accurate "`/qurl list`, `/qurl aliases`, and `/qurl get` all agree … including for admins" wording (the original was correct against #589) and pointed it at the note.

Kept `Refs #462` rather than `Closes #462`: the issue also has an at-merge CHANGELOG hand-edit on the release-please PR and human customer/ops comms that this docs PR does not cover.

justin-layerv (changes requested):
`/qurl get $<slug>`

users should not know about slug, right?

```
If you are reading a release note that says `/qurl list` shows the full
workspace list to every member, it refers to the #459 window and no longer
matches current behavior.

```
Strange to refer to ald PRs in a doc like that...
```
`/qurl get` also requires channel context and rejects raw internal
  `r_...` identifiers.
```
Probably don't want to expose users to r_ ID verbiage

Kevin-layerV:
Addressed your review in ea98c600 (doc-only, no code change):

1. **slug** — the capability-boundary section no longer mentions `$<slug>`. It now describes minting in terms of the tokens/aliases that `/qurl list` and `/qurl aliases` already surface, so a reader never has to learn the term "slug."
2. **`r_...`** — dropped the raw `r_...` identifier verbiage; the bullet now just says `/qurl get` accepts only what those commands surface, "never an internal resource identifier."
3. **Old PR refs** — replaced the PR-by-PR `## History` (#234/#459/#589/#590/#596) with a `## Past behavior` note. It still reconciles a stale release note against current behavior (the original #462 worry), but cites zero PR numbers.

Left the `policies.go` code comments as-is — they're developer-facing source, not the operator doc your review was about. Re-requesting review.

PR #643, opened 2026-06-08
fix(slack): close unmetered cold-channel slug-probe surface in /qurl get

### LINKED ISSUE #534: /qurl get: slug-fallback lookup runs before the rate-limit gate (unmetered probe surface)
Follow-up from PR #530 (cr review). Not a merge blocker; a fresh attack-surface delta that PR introduces.

## Context
PR #530 added a tunnel-slug fallback to `/qurl get`: when the channel `alias_bindings` lookup misses, `resolveTokenForGet` calls `resolveTunnelSlugAliasTarget` → `GET /v1/resources?slug=…` (one upstream hop) before authorizing.

## Problem
In `getWork` the per-user `CheckRateLimit` gate runs *after* token resolution (it gates the mint). So the slug-fallback upstream hop is **unmetered**: a user spamming `/qurl get $typo1`, `/qurl get $typo2`, … fires N `GET /v1/resources?slug=…` calls against the workspace's API key without burning their mint-rate budget. The mint itself stays gated (no auth/credential concern), but it's a cheap probe / noisy-neighbor surface that can burn the workspace's per-key qurl-service quota.

This hop did not exist before #530 — a binding miss used to short-circuit to "not configured" with no upstream call.

## Options (from cr)
1. Move the `CheckRateLimit` call ahead of the form-switch in `getWork` so every form (URL, alias, resource-id, slug-fallback) gates on the same per-user budget. (Broadest fix; changes the gate from mint-scoped to request-scoped.)
2. Skip the slug fallback for non-admins whose `AllowedResourceIDsForChannel` is empty — the fallback can never succeed for them, so the hop is pure waste; skipping closes the cold-channel probe case cheaply (a DDB read, no upstream hop).

Note the in-bot rate-limit (`slackdata.CheckRateLimit`) is a stub/early-stage gate today — this should be revisited together with the `/qurl get` rate-limit work tracked in `SLACK_QURL_ROLLOUT.md`.

## Acceptance criteria
- Repeated `/qurl get $<unknown>` from one user is bounded (rate-limited or short-circuited) so it can't fan out unbounded `GET /v1/resources?slug=…` calls.
- Confirm whether qurl-service's per-key quota for `GET /v1/resources` is decoupled from mint quota (informs how urgent option 1 is).

Kevin-layerV:
## What

In `resolveTokenForGet` (the `/qurl get $token` resolver in `apps/slack/internal/handler_get.go`), fetch the channel allow-set **before** the tunnel-slug fallback and short-circuit when it is empty. A "cold" channel — one with no `channel_policies` row, i.e. no protected resources — now returns the existing "not configured for this channel" copy from the DDB allow-set read alone, with **zero** upstream hops.

Net cost:
- **Cold channel:** the same single allow-set DDB read it already did downstream, now done up front, and **0** upstream `GET /v1/resources` calls.
- **Warm channel:** identical upstream behavior and the same single allow-set DDB read per request as before — the fetch is just hoisted to one site up front and threaded through both the slug-miss alias fallback and the slug-not-allowed branch (previously each of those branches fetched it once on its own path; the read count per request is unchanged, the call site is now singular).

No signature changes to `getWork`/`resolveTokenForGet`; reuses the existing `noResourceForAliasMessage` copy (no new user-facing string).

## Why

On an alias-binding miss, `resolveTokenForGet` fired `GET /v1/resources?slug=…` (one upstream hop against the workspace API key) **before** the per-user rate-limit gate in `getWork`. So from a cold channel, `/qurl get $typo1`, `$typo2`, … fanned out one **unmetered** upstream probe each, throttled only by Slack's own slash-command limit — a probe surface against the upstream API (security issue #534).

When the allow-set is empty, **both** fallbacks (tunnel slug and resource alias) are gated against that set and can never match, so the upstream lookup is pure waste in addition to being unmetered. Short-circuiting on the empty set removes the waste and closes the probe surface.

This is deliberately **narrow**. The resolver's docstring warns against "optimizing away the fallbacks on a binding miss" — that warning still holds: for a **configured** channel (non-empty allow-set) the fallbacks run exactly as before, preserving the "spend a read rather than burn the user's quota on a fat-fingered alias" tradeoff. The hop is suppressed **only** for the empty-set case the downstream channel gate would reject anyway. The expanded code comment and docstring call out this distinction explicitly to preempt that concern.

## Test

- **New** `TestHandleGet_UnknownSlugColdChannelNoUpstreamHop`: a cold channel (workspace seeded, no channel policy) firing `get $typo1/$typo2/$typo3` from one user asserts the upstream `GET /v1/resources` counter stays at **0** across all three, the not-configured copy is returned each time, and the mint route is never hit.
- **Reconciled** `TestHandleGet_AliasNotFound` (cold channel): it now also asserts **zero** upstream `GET /v1/resources` hits and documents the #534 short-circuit — it previously relied on an empty-list fixture for that hop, which no longer runs.
- `TestHandleGet_DollarSlugNotAllowedNonAdmin` (warm channel, non-empty allow-set) passes **unchanged**, proving the hop still runs and is suppressed only for the empty set — not for warm-channel rejections.

Verification (in worktree):
- `go build ./apps/slack/... ./shared/...` — clean
- `go test -race -count=1 ./apps/slack/... ./shared/...` — green, 83.9% slack-internal coverage (>40% floor)
- CI-pinned `golangci-lint run ./apps/slack/...` (v2.10.1) — 0 issues
- `pre-commit run --files …` — all hooks pass

Closes #534

### DIFF
--- a/apps/slack/internal/handler_get.go
+++ b/apps/slack/internal/handler_get.go
@@ -437,15 +437,28 @@ func (h *Handler) getWork(ctx context.Context, log *slog.Logger, args getWorkArg
 //
 // Cost note: this runs BEFORE the per-user rate-limit gate in getWork — the gate
 // only fires once a token resolves, so a typo never burns the user's quota (see
-// getWork). A binding miss therefore incurs its upstream resource lookups (slug,
-// then the first-page alias scan when needed) un-throttled by that gate,
-// including for plain typos. That's the deliberate "spend a read rather than burn
-// the user's quota on a fat-fingered alias" tradeoff — don't "optimize" it away
-// by short-circuiting the fallbacks on a binding miss. Each scan is bounded by
+// getWork). For a CONFIGURED channel (one with a non-empty allow-set), a binding
+// miss therefore incurs its upstream resource lookups (slug, then the first-page
+// alias scan when needed) un-throttled by that gate, including for plain typos.
+// That's the deliberate "spend a read rather than burn the user's quota on a
+// fat-fingered alias" tradeoff — don't "optimize" it away by short-circuiting the
+// fallbacks on a binding miss in a configured channel. Each scan is bounded by
 // listResourcesScanLimit and Slack's own per-user slash-command throttle bounds
 // the request rate; if the in-bot limiter (CheckRateLimit, a stub today) ever
 // needs to shed this resolution cost too, add a cheap token-shape pre-filter
 // before the alias scan rather than moving the gate back ahead of resolution.
+//
+// One NARROW exception (closes #534): when the channel allow-set is EMPTY (a
+// "cold" channel with no protected resources), BOTH fallbacks below would be
+// gated out anyway — neither a tunnel slug nor a resource alias can match an
+// empty set — so the upstream GET /v1/resources?slug= would be pure waste. Worse,
+// it's an UNMETERED probe surface: those upstream lookups run before the per-user
+// rate-limit gate, so `/qurl get $typo1`, `$typo2`, … from a cold channel fan out
+// one upstream hop each against the workspace API key, throttled only by Slack's
+// own slash-command limit. So the cold-channel case is short-circuited on the
+// allow-set DDB read alone. This is NOT the blanket "skip the fallbacks on a
+// binding miss" the tradeoff above warns against: it suppresses the hop ONLY when
+// the downstream channel gate is guaranteed to reject every fallback result.
 //
 // Caller must have already checked AdminStore (getWork does this before
 // resolving). Returns a [*userError] on lookup failure, not-a-known-token, or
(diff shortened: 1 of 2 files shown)


Kevin-layerV:
Thanks for the review. Addressed both non-blocking items in 1d9419b7:

- **PR-description "duplicate read" wording** — tightened. You're right that no single request path fetched the allow-set twice in the old code; each branch fetched it once. Reworded the warm-channel bullet to say the read count per request is unchanged and the only change is consolidating the fetch to one hoisted call site (the real win being the avoided upstream hop on cold channels).
- **`seedNonAdmin` vs `seedAdmin` note** — added a comment in `TestHandleGet_UnknownSlugColdChannelNoUpstreamHop` explaining the deliberate pick: the allow-set gate is channel-scoped with no admin bypass, so a non-admin is the more security-relevant caller to pin against the probe surface; the two tests together cover both an admin and a non-admin caller hitting the same short-circuit.

No logic changes this pass — the static invariant + read-count analysis you confirmed still holds.

PR #644, opened 2026-06-08
feat(slack): per-IP rate limit on public OAuth endpoints

### LINKED ISSUE #264: Rate-limit /oauth/qurl/{start,callback} (per-IP token bucket)
## Summary

PR #254 wires `/oauth/qurl/start` and `/oauth/qurl/callback` on the Slack bot's public HTTP surface. Both endpoints do:

- HMAC state verify (CPU)
- Cookie set/check
- Auth0 `/oauth/token` POST (outbound RPS to Auth0)
- qurl-service `/v1/api-keys` POST (outbound RPS to qurl-service)

Without an in-app rate limiter, a tight loop against either endpoint can:
1. Burn CPU on HMAC-SHA256 verifies.
2. Burn the bot's Auth0-tenant rate quota (token endpoint is rate-limited per Auth0 tenant; saturating it locks out legitimate installs).
3. Cause qurl-service quota churn for downstream rate-limiting decisions.

The signed-state contract means `start` returns 400 before any outbound call — that's the moat — but the CPU + per-IP allocation cost is still unbounded.

## Acceptance criteria

- Per-IP token-bucket rate limiter on `/oauth/qurl/start` and `/oauth/qurl/callback`.
- Bucket: e.g. 10 requests per minute per IP, with reasonable burst.
- Reject with 429 + `Retry-After` header.
- Mirrors the Discord pattern at `apps/discord/src/utils/oauth-rate-limit.js` (use that as the structural reference, not a direct port).
- Unit tests: under-limit accepts, over-limit rejects, separate IPs maintain separate buckets.

## Why this is a follow-up

The Auth0-tenant and qurl-service quotas are external concerns and the bot's ALB / WAF is the production moat. In-app rate limiting is defense-in-depth, not the only line of defense. Discord's rate limiter is also additive defense-in-depth there. The CRITICAL signed-state-token defense already landed in PR #254 itself.

## Out of scope

- Distributed (cross-task) rate limiting — single-task is sufficient for the per-workspace install rate.
- Sliding-window vs token-bucket — pick whichever is simpler.

Surfaced by /pr-ready loop on PR #254 (cr review item #4).

Kevin-layerV:
## What

Adds a per-IP **sliding-window** rate limiter (**10 requests / 60s / IP**) in front of both public OAuth routes — `/oauth/qurl/start` and `/oauth/qurl/callback`.

- New file `apps/slack/internal/oauth/rate_limit.go`: unexported `rateLimiter` (a mutex-guarded `map[string][]time.Time` with an injectable clock) + its `middleware`.
- `RegisterRoutes` now constructs **ONE shared** limiter and wraps each route with it **OUTERMOST** (limiter outside the existing `http.TimeoutHandler`) so a rejected request never enters the timeout-tracked handler.
- Over-budget → `429` with a `Retry-After` header (whole seconds until the oldest in-window timestamp ages out, floored at 1), set **before** the status is written, with a bare `http.Error` plain-text body matching the rejection style already used in `start.go` / `callback.go` (no HTML page).
- `clientIP` resolves the client for a **single-hop ALB**: when `X-Forwarded-For` is present, take the **LAST** (ALB-appended, trusted) entry — earlier entries are client-supplied and spoofable; else the `RemoteAddr` host; else an `"unknown"` sentinel so keyless traffic still shares one bucket instead of escaping the limit.
- Memory is bounded **without a background sweep**: stale entries evicted opportunistically under the same lock on each access; a per-IP timestamp cap (`maxRequestsPerIP`) trims abusive slices; a global distinct-IP cap (`maxStoreSize = 20000`) sheds previously-unseen IPs with a `429` once reached so a distributed flood can't grow the map toward OOM.

## Why

`Closes #264`. Both OAuth endpoints were registered with only `http.TimeoutHandler` — no per-IP / HTTP-edge throttle. A tight loop against either endpoint burns CPU on HMAC-SHA256 state verifies and, past the signed-state moat, can saturate the bot's **Auth0-tenant** token quota (rate-limited per tenant — saturating it locks out legitimate installs) and churn **qurl-service** quota. This is defense-in-depth layered under the signed-state contract and the ALB/WAF production moat, not the only line of defense.

**One shared limiter** (not one per route) is deliberate: a per-route counter would let an attacker double the per-IP budget by alternating `/start` and `/callback` — the exact amplification the Discord reference's header comment warns against. Structure mirrors the live Discord limiter at `apps/discord/src/utils/oauth-rate-limit.js` (sliding window, per-IP slice, `maxStoreSize`/`maxRequestsPerIP` caps); it is a structural reference, not a port.

**Single-instance only:** if the bot ever scales horizontally the effective rate becomes N × the limit and the store should move to Redis (noted in the file's doc comment).

**No background sweep goroutine on purpose** — that would need lifecycle wiring into `main.go`'s shutdown drain and risks goroutine leaks in tests; opportunistic in-lock eviction keeps steady-state memory bounded for this surface's load.

## Test

New `apps/slack/internal/oauth/rate_limit_test.go` (`httptest` recorder + request, injected advanceable clock) covers every acceptance criterion:

- **under-limit** — N (< max) requests all reach a sentinel next-handler;
- **over-limit** — the max+1'th returns `429` with a non-empty positive integer `Retry-After` and does **not** reach the handler;
- **separate IPs** — two distinct IPs keep independent buckets;
- **window-expiry** — advancing the clock past the window reopens the bucket.

Plus cases locking the single-hop ALB trust (last XFF entry wins over `RemoteAddr`, limiter keys by the XFF client IP, unparseable `RemoteAddr` → `"unknown"`).

Verified in the worktree:
- `go build ./apps/slack/... ./shared/...` — clean
- `go test -race -count=1 ./apps/slack/... ./shared/...` — green (oauth package 88.6% coverage; repo total 79.4%, above the 40% floor)
- CI-pinned `golangci-lint` v2.10.1 `run ./apps/slack/...` — 0 issues
- `pre-commit run --files <changed>` — all hooks pass

### DIFF
--- a/apps/slack/internal/oauth/rate_limit.go
+++ b/apps/slack/internal/oauth/rate_limit.go
@@ -1,0 +1,226 @@
+package oauth
+
+import (
+	"math"
+	"net"
+	"net/http"
+	"strconv"
+	"strings"
+	"sync"
+	"time"
+)
+
+// Per-IP rate limiting for the public OAuth surface
+// (/oauth/qurl/start and /oauth/qurl/callback).
+//
+// Mirrors the LIVE Discord limiter at
+// apps/discord/src/utils/oauth-rate-limit.js structurally (not a direct
+// port): one sliding-window counter per client IP, shared across BOTH
+// OAuth routes so an attacker can't double the budget by alternating
+// endpoints — that amplification is exactly what the Discord reference's
+// header comment warns about.
+//
+// This is defense-in-depth, not the only line of defense: the signed-state
+// contract makes /start return 400 before any outbound Auth0 /
+// qurl-service call, and the bot's ALB / WAF is the production moat. The
+// limiter caps the CPU spent on HMAC-SHA256 state verifies and the
+// per-request allocation cost that the signed-state moat leaves unbounded.
+//
+// SCALING: single-instance only. State lives in process memory, so if this
+// bot ever runs horizontally (multiple ECS tasks behind a load balancer)
+// the effective rate becomes N × the configured limit — move the store to
+// Redis at that point. Single-task is sufficient for the per-workspace
+// install rate this surface sees.
+const (
+	// rateLimitWindow is the sliding window each IP's request count is
+	// measured over.
+	rateLimitWindow = 60 * time.Second
+	// rateLimitMaxRequests is the per-IP request budget within
+	// rateLimitWindow. The issue's explicit acceptance criterion is
+	// 10 requests / minute / IP.
+	rateLimitMaxRequests = 10
+	// maxRequestsPerIP caps the timestamps retained per IP so a single
+	// abusive IP can't grow its slice unboundedly within a window. A
+	// small multiple of the budget is plenty — anything past the budget
+	// is already being rejected, the extra slack just absorbs the append
+	// that happens before the next request trims.
+	maxRequestsPerIP = rateLimitMaxRequests * 4
+	// maxStoreSize is the hard ceiling on distinct IPs tracked at once.
+	// Under a distributed flood from many unique IPs the map can reach this
+	// cap; an arriving previously-unseen IP first triggers a bounded
+	// reclamation sweep (reclaimStaleLocked) and is only shed with a 429 if
+	// that sweep can't free a slot. Known IPs keep being served because
+	// they don't grow the map.
+	maxStoreSize = 20000
+	// reclaimScanLimit bounds the keys examined per at-cap reclamation
+	// sweep so a single arrival can't pay an O(maxStoreSize) cost. One
+	// fully-stale key freed is enough to admit the arriving IP; the cap
+	// keeps the worst-case in-lock work flat regardless of map size. Each
+	// subsequent at-cap arrival resumes the sweep, so the whole map is
+	// still reachable across requests without a background goroutine.
+	reclaimScanLimit = 1024
+)
+
+// unknownIP is the client-IP sentinel used when neither X-Forwarded-For
+// nor RemoteAddr yields a usable host. Constant because it doubles as a
+// map key and is referenced from multiple sites.
+const unknownIP = "unknown"
+
+// xForwardedForHeader is the proxy header the ALB appends the real client
+// IP to. Named once so the literal isn't repeated across clientIP and any
+// future call site.
+const xForwardedForHeader = "X-Forwarded-For"
+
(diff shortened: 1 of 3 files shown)


Kevin-layerV:
Addressed the review feedback in bbe04b81:

**Observability gap (the item you pushed on)** — fixed. The limiter now emits a `slog.Warn` on both rejection paths, mirroring the Discord reference's two `logger.warn` sites:
- over-budget: `oauth rate limiter: per-IP budget exceeded` (ip)
- at-cap shed: `oauth rate limiter: store at hard cap, shedding new IP` (ip, store_size) — the high-signal "under a sustained distributed flood" event.

Both are **throttled to ~one line per window** with a coalesced `suppressed_since` count, rather than unthrottled like the JS reference. Rationale: a control whose entire job is absorbing a flood cheaply must not become a log-amplifier under that same flood — one line per shed request at flood volume would defeat the purpose. New tests cover both paths (fire-once + the coalesced count after the interval rolls).

**XFF trust has no runtime guard** — leaving the algorithm as-is, per your note that there's nothing to change there. Confirmed the production wiring: the bot runs behind an ALB on Fargate (`apps/slack/cmd/main.go` header + the SIGTERM/ALB-drain handling), so the single-hop-ALB assumption that makes the last XFF entry trustworthy holds in prod. The doc comment on `clientIP` already states the assumption.

**Nits** — all three fixed:
- `RegisterRoutes` now resolves the clock via `cfg.now()` like the rest of the package.
- Added a one-line note to `maxStoreSize` that the store pins at cap once hit and never drains below it (intentional, bounded-memory, no background goroutine — not a leak).
- Added `TestRegisterRoutesWrapsBothRoutesWithLimiter` so a dropped `rl.middleware(...)` wrap on either route is caught.

justin-layerv (changes requested):
> Single-instance only: if the bot ever scales horizontally the effective rate becomes N × the limit and the store should move to Redis (noted in the file's doc comment).

We will have one instance per AZ at least, so this is already problematic. Consider putting the rate limiting in our internal system, if we don't already have one. Higher value.

Kevin-layerV:
Closing this after the PR-ready sweep because Justin's scaling objection is a real blocker: the bot runs with multiple tasks/AZs, while this implementation is explicitly process-local and would multiply the OAuth budget per task. Since these public OAuth endpoints only have client-IP context before a workspace exists, the replacement should be an edge/internal shared limiter rather than local bot memory. CI was green, but the design is stale for the current deployment model.

PR #645, opened 2026-06-08
feat(slack): enforce per-user mint rate limit on /qurl get

### LINKED ISSUE #400: slackdata: CheckRateLimit is a permanent stub — pick in-mem vs DDB-counter and wire enforcement
## Context

`apps/slack/internal/slackdata/rate_limit.go` defines:

```go
func (s *Store) CheckRateLimit(ctx context.Context, teamID, slackUserID string) (allow bool, retryAfter int, err error) {
    return true, 0, nil
}
```

Comment is explicit ("stub returning unconditionally allow"). Every admin verb that consults this gate is effectively un-rate-limited; the only backstop is the qurl-service API-key quota, which is workspace-wide, not per-user.

## Why this matters

Per-Slack-user mint enforcement was a load-bearing property of the pre-pivot HTTP shape (qurl-service's `/internal/v1/admin/rate-limit/check` returned `(allow, retry_after, err)` based on a Redis token bucket). The Wave 4 pivot moved policy storage to DDB but the rate-limit shape didn't follow because the production strategy wasn't picked.

A regression note from cr review on PR #231 calls this out:

> If the qurl-service API-key quota is in fact a meaningful backstop, fine; if not, this is a regression vs the pre-pivot enforcement and should be tracked as a release blocker rather than a TODO.

## Decision needed

Pick one:

1. **In-memory token bucket** per-Fargate-task. Cheap, no DDB cost. Loses state on task restart; lets a determined operator burst across restarts.
2. **DDB counter on `workspace_state`** with sliding-window. Persistent across restarts. Adds a write per admin mint; well within RCU/WCU headroom on `workspace_state` (which is otherwise read-mostly).
3. **Lambda Function URL with EventBridge throttle policy** — defers rate-limit to AWS infra. Highest cost in setup; lowest cost in code.

## Acceptance criteria

- Decision recorded in SLACK_QURL_ROLLOUT.md or a sibling design doc.
- `CheckRateLimit` returns real `(allow, retryAfter, err)` based on the chosen strategy.
- A handler-level test asserts a denied path renders the rate-limit copy and a `retry-after` hint.

## Related

- PR #231 — surfaced by Claude code review
- The `rate_limit.go` file is currently 38 lines, all stub.

Kevin-layerV:
## What
Replaces the no-op `slackdata.CheckRateLimit` stub with a real in-memory **per-Slack-user token bucket** for `/qurl get`. Each `slack_user_id` gets a budget of **30 mints/hour** (the pre-pivot enforcement value the stub's own header documented) that refills continuously at one token per (hour ÷ rate). State is held on the `Store` under a mutex, keyed by `slack_user_id`, and uses the Store's injected clock so it is deterministically testable.

On denial, `CheckRateLimit` returns the time until the next token frees up; the existing `/qurl get` caller already renders that as `Rate limit hit. Try again in …`. **No renderer or call-site change** — the signature is unchanged:

```go
func (s *Store) CheckRateLimit(ctx context.Context, slackUserID, teamID string) (allowed bool, retry time.Duration, err error)
```

The rate is a code policy (`Store.MintRatePerHour`, defaulting to 30) rather than an env knob; it is a field so the policy lives in code and tests can grant headroom to flows whose intent is orthogonal to rate limiting. **No new dependencies** (hand-rolled bucket — no `golang.org/x/time/rate`).

## Why
Closes #400: the gate was a stub that always returned `(true, 0, nil)`, so per-Slack-user mint rate limiting was **OFF in production**. A single user — or a runaway agent loop — could mint links with no in-bot ceiling, leaving qurl-service's coarser customer-level API-key quota as the only backstop.

**Design: in-memory per-Fargate-task bucket.** Tradeoffs, accepted deliberately:
- **Per-task, not global.** Slack does not route a user's slash commands to a sticky task, so with N tasks a user's effective ceiling is ≈ N×30/hr. This is an abuse-rate backstop; the customer-level API-key quota remains the hard cross-task ceiling underneath it.
- **Resets on redeploy.** A deploy hands every user a fresh full bucket. Acceptable — the goal is to blunt abuse within a task's lifetime, not to persist a rolling hourly count.

**Future upgrade path** (only if a globally-consistent cross-task limit is ever needed): a DynamoDB atomic conditional `UpdateItem` (token-count + last-refill) keyed by `slack_user_id`, at the cost of one extra write per mint. Recorded in the `rate_limit.go` header and in `apps/slack/docs/operating.md` (new "Per-user mint rate limiting" section).

The gate stays after token resolution (load-bearing: a typo'd/unauthorized alias is turned away before the gate, so it never burns quota) — that ordering and the call site were left untouched.

## Test
- `apps/slack/internal/handler_get_test.go` → `TestHandleGet_RateLimited`: drains a user's full budget via real mints (clock pinned → no refill), then asserts the next `/qurl get` replies `Rate limit hit` **and** that the upstream mint route is **not** hit on the denied call (`atomic.Int32` counter).
- `apps/slack/internal/slackdata/rate_limit_test.go`: focused unit tests for the bucket math with an advanceable clock — first-mint allow, burst-then-deny with exact retry, continuous refill, refill cap at burst, per-user isolation, and the `MintRatePerHour` override.
- Updated the `TestHandle_ConcurrentGetSharesIdempotencyKey` fixture (all 100 requests are one user) to raise `MintRatePerHour` so the idempotency-dedup property under test is not throttled by the now-real limiter.

`go build`, `go test -race ./apps/slack/... ./shared/...`, the CI-pinned golangci-lint (whole-module), and pre-commit all pass.

Closes #400

### DIFF
--- a/apps/slack/docs/operating.md
+++ b/apps/slack/docs/operating.md
@@ -38,30 +38,24 @@ not-authorized aliases are turned away before the gate, so a fat-finger
 never burns quota. When a user is over budget, `/qurl get` replies
 "Rate limit hit. Try again in …" with the time until their next mint.
 
-**Decision: in-memory token bucket, per Fargate task.** Each user has a
-bucket that refills continuously at one token per (hour ÷ rate) and is
-held in the task's memory under a mutex. This was chosen over a
-DynamoDB-backed counter because it adds **zero per-mint I/O/latency** on
-the hot path and is sufficient for the goal — blunting runaway
-loops/abuse — without billing-grade precision.
+**Decision: DynamoDB-backed fixed hourly counter.** Each
+`(slack_team_id, slack_user_id)` pair has one synthetic row in
+`workspace_mappings`, keyed with a reserved rate-limit prefix. A mint
+attempt increments the current one-hour window with an atomic conditional
+`UpdateItem`; once the count reaches the configured rate, later attempts
+are denied until the next window.
 
 Tradeoffs, accepted deliberately:
 
-- **Per-task, not global.** Slack does not route a user's slash commands
-  to a sticky task, so with N running tasks a single user's effective
-  ceiling is ≈ N × 30/hr. This is an abuse-rate backstop; qurl-service's
-  customer-level API-key quota remains the hard cross-task ceiling
-  underneath it.
-- **Resets on redeploy/restart.** A deploy hands every user a fresh full
-  bucket. Acceptable: the intent is to bound abuse within a task's
-  lifetime, not to persist a rolling hourly count across deploys.
-
-**Future upgrade path** (only if a globally-consistent cross-task limit
-is ever required): move the counter to DynamoDB via an atomic
-conditional `UpdateItem` (token-count + last-refill attributes) keyed by
-`slack_user_id` on the `channel_policies`/workspace row. That buys
-durability and cross-task consistency at the cost of one extra DDB write
-per mint.
+- **Global across tasks.** Slack slash-command routing is not sticky, so
+  the counter must be shared across every running Fargate task.
+- **Fixed window, not smoothing token bucket.** A user can spend the
+  full budget near the end of one hour and again at the start of the
+  next. The goal is a global abuse backstop, not billing-grade rolling
+  quota precision.
+- **One extra write per mint attempt.** `/qurl get` already reads DDB for
+  workspace/policy state before minting, so the additional conditional
+  write is acceptable on this path.
 
 The rate is a code-level policy (`Store.MintRatePerHour`, defaulting to
 30), **not** an environment variable — there is no operator knob to set;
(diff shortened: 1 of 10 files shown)


justin-layerv (changes requested):
We need global, not per instance.

Kevin-layerV:
### PR-ready pass: Justin global-rate-limit feedback

Reworked this from the per-task in-memory bucket Justin flagged into a shared DynamoDB fixed-window counter:

- Key is now per `(slack_team_id, slack_user_id)` under a reserved synthetic `workspace_mappings` row, so all ECS tasks share one budget.
- Enforcement uses atomic conditional `UpdateItem`; the first task in a new hour resets the row under a guard, racing tasks retry through the normal increment path.
- Updated the operator doc and stale handler/test comments so the PR no longer advertises per-instance behavior.

Local verification:

- `go test ./apps/slack/internal/slackdata`
- `go test ./apps/slack/... ./shared/...`
- `git diff --check`

justin-layerv:

PR #650, opened 2026-06-08
feat(slack): conversation-mode Events surface (read-only, dark)

justin-layerv:
> **Stacked on #648** (agent core). Base is the PR2 branch so the diff shows only this PR's changes; retarget to `main` once #648 merges.

## What

Wires the Secure Access Agent to the Slack **Events API**, behind a kill switch. The feature is **OFF in production** until `cmd/main.go` populates the new `Config` fields (`AgentLLM`, `AgentStore`, `PostMessage`) — that enablement wiring + infra (Anthropic key, DDB state table, Slack manifest scopes) lands separately (tracked issue), so this PR is dark while being fully exercised by tests.

## What lands

- **`slackdata.AgentStore`** — DDB-direct store for per-thread conversation history (optimistic-concurrency writes via a non-reserved `conv_version`) and Slack event-id dedupe (conditional put). Pending actions belong with the confirm flow, so they're deferred to that PR.
- **`agentBackend`** — adapts the qURL client + `channel_policies` to `agent.Backend`, scoping every read to what the caller can see in their channel. This is the boundary that stops the agent leaking resource existence across channels (a test asserts a resource outside the channel scope can't leak).
- **`handleEvent`** — parses `event_callback`, filters bots/edits/non-mentions/non-DMs (self-loop guard), dedupes, and dispatches a conversation turn on the existing bounded async pool (`startAsyncWorker`: panic recovery + SIGTERM drain — no `response_url`/ack coupling). Replies via `chat.postMessage`, threaded on the message root. **Always acks 200** so Slack never retries an accepted delivery.
- **Read-only:** a proposed mutation is surfaced as a preview message (the confirm flow + execution land in the next PR).

Grid-aware (`enterprise_id` partition); conversation history is JSON-serializable across turns.

## Testing

- Store: round-trip + version conflict (stale write can't clobber) + dedupe.
- Backend: channel-scope filtering (out-of-channel resource doesn't leak), alias/slug resolve, quota, nil-store graceful.
- Events path: reply posted in-thread, retry dedupe (one reply), disabled-stays-silent (still acks 200; `url_verification` still works).
- `go test -race ./apps/slack/...` green; `golangci-lint` (CI-pinned v2.10.1) clean.

## Deferred (tracked)

Production `cmd/main.go` wiring (env: `ANTHROPIC_API_KEY`, `QURL_AGENT_STATE_TABLE`, kill switch; construct the LLM, store, and `chat.postMessage` seam) + infra (state table, manifest scopes) — see the linked enablement issue. The per-workspace admin toggle + rate limits are PR6.

### DIFF
--- a/apps/slack/internal/agent/prompt.go
+++ b/apps/slack/internal/agent/prompt.go
@@ -16,6 +16,7 @@ Your job is to be the conversation on top of qURL's deterministic commands. Unde
 
 How you operate:
 - Read tools (list_resources, list_aliases, resolve_token, get_quota) are safe and run immediately. Use them freely to ground your answers in what actually exists in this channel.
+- Your reads only see what's reachable in THIS channel. Make that scope explicit when it matters ("the connectors I can see in this channel are …"), and if the user expects something you can't see, say you can only see this channel rather than implying it doesn't exist anywhere — don't present a channel-scoped answer as the whole picture.
 - Anything that protects, revokes, grants access, or changes an alias is a MUTATION. You never perform mutations yourself. You call a propose_* tool, which shows the user a confirmation card; the action only runs after they click Confirm. State plainly that you're proposing an action and that it needs confirmation.
 - If a request is ambiguous (two aliases could match, a port or environment is missing), ask ONE concise question instead of guessing or failing. The user's next message continues the thread.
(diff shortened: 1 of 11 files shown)


justin-layerv:
Enablement (cmd/main.go wiring + infra) tracked in #651.

PR #656, opened 2026-06-09
fix(slack): replace slash-command ack instead of unsupported delete_original

justin-layerv:
## Problem

The guided setup wizards — `/qurl-admin protect-url` and `/qurl-admin protect-connector` — logged an error on **every** modal open in prod:

```
WARN response_url returned non-2xx status=500 body="no_text" command=protect_url_wizard
WARN response_url delete_original failed; retrying once
```

…and the "⏳ Working on it…" ack lingered instead of clearing.

## Root cause

After opening the modal, the wizards cleared the ack by POSTing `{"delete_original": true}` to the slash command's `response_url`. Per [Slack's docs](https://docs.slack.dev/interactivity/handling-user-interaction), **`delete_original` is not supported for slash commands** — Slack ignores the field, and because the POST then carries no `text`, rejects it with `no_text` (HTTP 500). (An ephemeral ack also can't be deleted.)

## Fix

Replace the ack **in place** with `replace_original: true` + text — the exact mechanism the wizards' *error* paths already use successfully on the same `response_url` (`ErrorResponse(..., replaceOriginal: true)`). An always-present `text` field can never hit the `no_text` path, and the spinner is swapped for a brief "Opened …" confirmation rather than left stale.

- `process.go`: `deleteOriginalResponse` → `replaceOriginalResponse(message)` (retry-once preserved, renamed helpers/const).
- Both call sites updated: protect-url wizard (`handler_resource.go`) and tunnel-install wizard (`handler_tunnel.go`).

## Why the tests didn't catch it

The old tests mocked Slack as HTTP 200 and only asserted the return-false-on-500 retry path — they never inspected the posted body, and the tunnel tests literally asserted `delete_original = true` (the buggy contract). Now:

- New `TestReplaceOriginalResponsePayload` asserts the body shape: `replace_original` true, `text` present, **no `delete_original`**. Verified it **fails** against the old `{"delete_original": true}` payload.
- Tunnel tests flipped from asserting `delete_original` to asserting `replace_original` + opened-copy text + absence of `delete_original`.

## Validation

- `go test ./apps/slack/... -count=1` ✅
- `golangci-lint run --new-from-rev=origin/main ./apps/slack/...` → 0 issues ✅
- Confirmed the new regression test fails on the pre-fix payload, then passes after the fix.

Note: prod is currently pre-#613 (two delete sites); `main` is post-#613 (the two sites here). Same root cause on both.

### DIFF
--- a/apps/slack/internal/admin_handler_helpers_test.go
+++ b/apps/slack/internal/admin_handler_helpers_test.go
@@ -248,6 +248,24 @@ func parseSlackReplyBool(t *testing.T, body []byte, field string) bool {
 	return value
 }
 
+// assertWizardAckReplaced asserts a guided-wizard cleanup REPLACED its "Working
+// on it…" ack in place rather than deleting it: replace_original set,
+// delete_original NOT set (unsupported for slash commands — Slack returns
+// no_text), and the opened-copy text (wantText) present. `when` labels the open
+// path for failure messages.
+func assertWizardAckReplaced(t *testing.T, body []byte, wantText, when string) {
+	t.Helper()
+	if got := parseSlackReplyBool(t, body, "replace_original"); !got {
+		t.Fatalf("replace_original = %v, want true after %s", got, when)
+	}
+	if got := parseSlackReplyBool(t, body, "delete_original"); got {
+		t.Fatal("delete_original must not be set (unsupported for slash commands; Slack returns no_text)")
+	}
+	if text := parseSlackText(t, body); !strings.Contains(text, wantText) {
+		t.Fatalf("ack replace text = %q, want it to contain %q (after %s)", text, wantText, when)
+	}
+}
+
 // workspaceMappingHasAdmin returns true iff the workspace_mappings
 // row for `teamID` exists AND carries `slackUserID` in its
 // admin_slack_user_ids SS (set). The OAuth callback writes this row
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1480,9 +1480,10 @@ func respondJSON(w http.ResponseWriter, status int, body any) {
 // Centralized so respondSlack and the parallel writer in postResponse
 // can't drift, and so the goconst/keyword consistency stays linter-clean.
 const (
-	respFieldResponseType = "response_type"
-	respFieldText         = "text"
-	respTypeEphemeral     = "ephemeral"
+	respFieldResponseType    = "response_type"
+	respFieldText            = "text"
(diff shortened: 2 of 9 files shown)


justin-layerv:
Thanks — addressed the three notes from the latest review:

**1. Const placement asymmetry** — ✅ Fixed. Moved `exposeURLWizardOpenedMsg` to `handler_resource.go`, right above `openExposeURLWizard` (its only call site), mirroring how `tunnelInstallWizardOpenedMsg` sits next to `openTunnelInstallWizard`.

**2. Asymmetric integration coverage** — ✅ Fixed. `TestHandleExposeURLBareOpensPickerModal` now asserts the protect-url wizard's success path replaces the ack with its specific copy ("Opened the URL protection form"). To avoid two near-identical assertion blocks, I promoted the tunnel-only helper to a shared, parameterized `assertWizardAckReplaced(t, body, wantText, when)` in `admin_handler_helpers_test.go` — both wizard tests (tunnel + expose) now use it, each asserting `replace_original` + no `delete_original` + its own opened-copy text.

**3. `response_type: ephemeral` redundancy** — Keeping it intentionally for field-for-field parity with the proven `ErrorResponse` payload (and the original ack was ephemeral); harmless on a `replace_original` update, as you noted. No change.

(The earlier-commit review's nit about a `respFieldReplaceOriginal` constant was already addressed in this push — `replace_original` is now a shared constant used by both `replaceOriginalResponse` and `ErrorResponse`.)

justin-layerv:
Thanks — both observations are intentional, confirming rather than changing:

1. **Lingering "Opened…" ephemeral** — intentional and, as you note, the best available behavior: Slack offers no way to *delete* a slash-command's ephemeral ack (that's the root cause being fixed), so the wizard replaces it in place. The copy is deliberately self-explanatory so it reads fine standalone whether the user submits or cancels. Strictly better than the prior stale spinner + recurring `no_text` error.
2. **`response_type: ephemeral` in the replace payload** — intentional, for field-for-field parity with the proven `ErrorResponse` path (and the original ack was ephemeral). Harmless on a `replace_original` update, as you said. Keeping it so both cleanup payloads stay identical in shape.

No code change for these. Appreciate the review.

justin-layerv:
Round-4 review (post-rebase) re-confirms the same two intentional trade-offs already covered above — the lingering `replace_original` ephemeral (Slack can't delete a slash-command ack) and `response_type` parity with `ErrorResponse`. No new items; diff is unchanged by the rebase. Thanks.

Issue #662, opened 2026-06-09
Slack agent: mutation-execution UX (confirm card, on-behalf attribution, completion recap)


justin-layerv:
From the Slack agent-design audit (in-session tasks #5, #6, #7). These land with the planned PR4/PR5 mutation-execution flow (today mutations are propose-only as a text preview; nothing executes).

- **Confirm UI (#5):** replace the text "…isn't enabled yet" preview with an interactive Block Kit card — preview + Approve / Modify / Reject — and re-run `CheckAdmin` on confirm. Avoid confirmation fatigue (confirm only what needs it). Drives the existing mutation cores (`revokeResource`, `getWork`, `createAndExposeURLResource`/`bind*`, `buildTunnelInstall`).
- **On-behalf attribution + AI marker (#6):** label executed actions "on behalf of @user" so recipients know a human authorized but an agent performed it; mark autonomously created/unreviewed content with a visible AI-generated indicator.
- **Completion recap (#7):** proportionate summary with DIRECT LINKS to created/modified content, and explicitly name skipped steps + why ("couldn't assign @x — not a project member") rather than silently leaving fields blank.

Rationale: confirmation, accountability, and audit are the doc's consequential-action requirements; they only become relevant once the agent executes.

justin-layerv:
**UX polish from PR4a (#667) cr round 4 — post-claim error copy on the confirm card.**

The confirm flow claims the pending action *before* executing (at-most-once by design), so when a mint/revoke hits a transient upstream error after the claim, the card is replaced with terminal text and the Approve button is gone — but shared `*userError`/generic copy says "Please try again", which on a button-less card wrongly implies an in-place retry. The user must re-*ask* the agent. Tighten the post-claim core-error copy for the confirm surface to point at re-asking (e.g. "…ask me again to retry") rather than "try again". A clean fix needs per-surface error copy, since the same `userError` messages read correctly on the slash path (where "try again" = re-run the command) — so this is deferred polish, not a half-fix on the shared strings.

justin-layerv:
[...]

Issue #651, opened 2026-06-08
Enable conversation mode: cmd/main.go wiring + infra (DDB table, Anthropic key, Slack manifest)


justin-layerv:
Follow-up to the conversation-mode Events surface (#650), which is **dark** until the wiring below lands. The in-repo code (agent core #648, events surface #650) gates on `Config.AgentLLM` / `AgentStore` / `PostMessage` being non-nil; production leaves them nil today.

## In-repo wiring (`apps/slack/cmd/main.go`)
- [ ] Read `ANTHROPIC_API_KEY`; construct `agent.NewAnthropicLLM(key)` → `Config.AgentLLM`. Leave nil (dark) when unset.
- [ ] Read `QURL_AGENT_STATE_TABLE`; construct `slackdata.NewAgentStore(ddbClient, table)` → `Config.AgentStore`.
- [ ] Add a `chat.postMessage` seam (threaded, per-workspace bot token — mirror the existing `OpenView`/`PostDM` token resolution, Grid-aware) → `Config.PostMessage`.
- [ ] Read an org kill-switch env (e.g. `QURL_AGENT_DISABLED`) → `Config.AgentDisabled`.

## Infra (`qurl-integrations-infra`, private)
- [ ] Provision the `QURL_AGENT_STATE_TABLE` DynamoDB table (pk/sk schema, TTL on `ttl`) + IAM (GetItem/PutItem) for the Slack task role.
- [ ] `ANTHROPIC_API_KEY` in the task secret store + env wiring.
- [ ] Slack app manifest: subscribe to `app_mention` + `message.im`; scopes `app_mentions:read`, `im:history`, `im:read`, `chat:write`, `reactions:write` for the #661 working-on-it reaction ack.

## Rollout
Keep the per-workspace toggle default **off** and the kill switch engaged until the DPA / data-handling review clears (channel content is sent to Anthropic). The per-workspace admin toggle + rate limits ship in the toggle PR (PR6). GA flips the absent-attribute default to on.

## Done when
A sandbox workspace can @mention the agent and get a read-only reply end-to-end, with the kill switch able to silence it.

justin-layerv:
## Deferred review items to weigh before GA

From the conversation-mode PR reviews (#648, #650) — non-blocking for the dark merge, but worth a deliberate decision before the feature is enabled in production:

[...]

PR #668, opened 2026-06-09
feat(slack): extend confirm flow to set-alias / unset-alias (PR4b)

justin-layerv:
Extends the propose → Approve → execute confirm flow (#667, PR4a) to **`set-alias`** and **`unset-alias`**, behind the same `AgentConfirmEnabled` dark-launch flag (unwired in `cmd/main.go`, so off in production).

Both kinds are **admin-gated** and **direct-execute** with **benign public results** — they mirror `revoke` exactly. No `get`-style ephemeral delivery needed (an alias bind/clear isn't a credential).

### What's here
- `pendingAction` gains `Alias`/`Target` (read by the alias execute, snapshotted at propose time).
- `confirmExecutable` + `executeAgentAction` gain the two kinds; **protect** (PR4c) stays deferred — and the `confirmExecutable ↔ executeAgentAction` lockstep test now covers them.
- `set-alias` → `resolveAndBindTunnelSlugAlias` in the **click's channel** (channel-scoped, like the slash verb); `unset-alias` → new **`unbindAliasResult`**, extracted from `handleUnsetAlias` and shared with the `/qurl-admin unset-alias` slash verb (no user-visible change; clean dedup like `mapCoreError`/`requireAdminForClick` in PR4a).

### Inherits PR4a's security model (all still tested)
LLM-never-executes · admin re-check derived from the *stored* `ActionKind` (both Approve and Reject) · consume-once claim-before-execute · public-card-with-ephemeral-denials · mrkdwn-escaped summaries · read-time TTL · click-time flag re-check.

### Tests
set/unset-alias confirm orchestration (admin-gated → claim → execute → replace), the alias-fields snapshot, the lockstep pin extended to the new kinds, plus the full PR4a invariant suite.

### Deferred
protect-connector / protect-url → **PR4c** (#662) — different mechanism (Approve `OpenView`s the tunnel-install modal). The two **hard pre-enablement gates remain on #651** (get authorization; R2 public-card smoke test).

### DIFF
--- a/apps/slack/internal/handler_agent_confirm.go
+++ b/apps/slack/internal/handler_agent_confirm.go
@@ -30,6 +30,11 @@ const (
 	agentConfirmUnsupportedReply    = "I can't apply that kind of change yet."
 	agentConfirmGetDeliveredReply   = "Handled — the access link was sent privately to the approver."
 	agentConfirmFailedReply         = "Something went wrong applying that. Please try again, or use a `/qurl` command."
+	// agentConfirmInvalidAliasReply is generic ON PURPOSE: the confirm card is
+	// public, so an invalid (LLM-distilled, possibly injected) alias/target must NOT
+	// be echoed back into it — unlike the slash path, whose validation reply is
+	// ephemeral and can echo the bad token.
+	agentConfirmInvalidAliasReply = "That alias or target isn't valid — use lowercase letters, numbers, and dashes (no special characters)."
 )
 
 // pendingAction is the ephemeral snapshot persisted between proposing a mutation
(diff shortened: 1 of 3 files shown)


justin-layerv:
Thanks — the 🔴 set-alias escape gap is a real one; fixed in 6b6bbdb3.

**🔴 + 🟡 (both closed at the deeper altitude — validation):** the confirm `set-alias` case now validates `pa.Alias`/`pa.Target` through the **same grammar `parseAliasArgs` enforces** (rejecting backticks / bad charset / non-tunnel target) **before** binding. So an injected value is neither bound (closes the 🟡 data-hygiene gap) nor echoed unescaped into the public card (closes 🔴 — no out-of-grammar char can reach `resolveAndBindTunnelSlugAlias`'s code-fence echo). Crucially, on failure I surface a **generic** `agentConfirmInvalidAliasReply`, **not** the raw token — because `parseAliasArgs`'s own error message echoes the bad value, which is fine for the slash path (ephemeral reply) but would itself break the fence on the **public** card. Test: `TestConfirm_SetAliasRejectsInvalidInput` (backtick alias → generic reply, no echo, no backtick on the card). `unset-alias` was already safe (`unbindAliasResult` escapes, and an invalid alias → benign "not bound").

**Nit (`Token` field comment):** left as-is — it accurately scopes the field to get/revoke, next to `Alias`/`Target` for the alias actions.

justin-layerv:
Thanks for the thorough pass. Round-6 items are all intentional/acknowledged — dispositions:

1. **set/unset protection asymmetry** — intentional and documented in the `resolveAndBindTunnelSlugAlias` caller-contract (validation is load-bearing because the not-found copy round-trips the `$slug` the admin retypes, so it can't also escape). Good call to carry the pattern into **PR4c** so protect picks the validate-or-escape layer deliberately — noted for that PR.
2. **`confirmValidAliasBind` reconstruction coupling** — agreed it couples to the slash input format; kept as-is on purpose to reuse `parseAliasArgs` as the *single* grammar source (a value-level validator would duplicate the alias/slug grammar and is the thing that actually drifts). Accepting the trade-off per your "not worth churning now"; the lockstep + invalid-input table guard the behavior.
3. **Channel-scoping** — covered transitively by the `payload.Channel.ID == pa.ChannelID` mismatch guard (`TestConfirm_ChannelMismatchRefused`) plus the slash alias tests for bind correctness; agree the split is reasonable.

No code change this round. CI is green and `/simplify` came back clean. Treating PR4b as merge-ready (still a draft, unwired in `cmd/main.go`) — the two hard pre-enablement gates (get authorization; R2 public-card smoke test) remain on #651, protect + UX polish on #662.

PR #669, opened 2026-06-09
fix(slack): resolve set/unset-display-name by channel alias, not just slug

Kevin-layerV:
## Problem

`/qurl-admin set-display-name $<id>` and `unset-display-name $<id>` resolved the id **only** as a connector's own slug (`GET /v1/resources?slug=`). If an admin referenced a resource by a channel `$alias` whose name differs from the connector slug — e.g. an alias manually re-attached to a re-created connector — the command failed with:

> No qURL Connector with id `dashboard` was found. Run `/qurl list` to see your qURL Connector ids.

…even though `/qurl get $dashboard` and `/qurl-admin revoke $dashboard` resolve that same alias correctly, because those paths are **alias-aware** (they read `channel_policies.alias_bindings` first) while the display-name verbs were **slug-only**.

This is **not** the orphaned-alias bug fixed in #654 (which swept stale bindings when a resource is *revoked*). Here the resource is **live** — only reachable in the channel via an alias whose name ≠ slug.

## Fix

Add a channel `alias_bindings` fallback to `resolveTunnelByID`, **after** the existing slug lookup:

- **Slug stays primary.** Existing slug-targeted calls keep the efficient server-side `?slug=` filter and are *not* newly bounded by `listResourcesScanLimit`. Zero behavior change for them.
- **On a slug miss**, look up the channel alias the same way `/qurl get` does (`AdminStore.LookupChannelAlias`). qurl-service has no get-by-id, so recover the full resource via the same bounded first-page scan `/qurl list` / protect already use, matching on `resource_id`.
- **Re-assert tunnel + active** before the PATCH. A binding that points at a URL resource, a revoked resource, or one past the scan window yields a friendly `unset-alias` hint instead of a wrong/dead PATCH target.
- `unset` reverts off the resolved connector's **real slug**, correct whether the id was the slug or an alias.

Resolution stays admin-gated (unchanged) and channel-scoped (the alias is only looked up in the invoking channel) — consistent with `/qurl get` and `/qurl-admin revoke`.

## Scope — deliberately not in this PR

- `set-alias`'s source resolution has the same slug-only asymmetry, but it shares `resolveTunnelSlugAliasTarget` with `/qurl get`'s slug fallback, so changing it has a wider blast radius — left for a separate change.
- Aliases that lead with a digit or are ≤2 chars still can't be targeted: the display-name id parser uses `tunnelSlugPattern` (must start with a letter, ≥3 chars), narrower than the alias charset. Not widened here (the parser is shared with the slug path); a follow-up if it comes up.

## Testing

- `TestSetDisplayName_ResolvesByChannelAlias` — alias whose name ≠ slug resolves end-to-end and PATCHes.
- `TestUnsetDisplayName_ResolvesByChannelAlias` — same, and pins the reset value to the **real slug**.
- `TestSetDisplayName_ChannelAliasToURLResourceRejected` — alias → URL resource is refused (no PATCH), with the unset-alias hint.
- Existing display-name tests unchanged + green; full `apps/slack` suite green; `golangci-lint` v2.10.1 (CI-pinned) clean whole-module.

### DIFF
--- a/apps/slack/internal/handler_display_name.go
+++ b/apps/slack/internal/handler_display_name.go
@@ -348,13 +348,16 @@ func (h *Handler) resolveTunnelByID(ctx context.Context, log *slog.Logger, teamI
 	// the Type==Tunnel / Status==Active guard and unset's slug both still hold.
 	if channelID != "" && h.cfg.AdminStore != nil {
 		boundID, found, lookupErr := h.cfg.AdminStore.LookupChannelAlias(ctx, teamID, channelID, id)
-		if lookupErr != nil {
-			// Best-effort enhancement: log and fall through to the not-found
-			// copy rather than surfacing a store error for what may be a plain
-			// typo. The admin gate read AdminStore moments ago, so a transient
-			// failure here is unlikely.
+		switch {
+		case lookupErr != nil:
+			// channel_policies is a DIFFERENT table from the workspace_mappings the
+			// admin gate read, so it can be unavailable while the gate passes. Surface
+			// the service-unavailable signal (matching resolveTokenForGet) rather than
+			// falling through to the "no such id" copy, which would misdiagnose an
+			// outage as a typo.
 			log.Warn("display-name: channel alias lookup failed", "error", lookupErr, "team_id", teamID, "channel_id", channelID, "id", id)
-		} else if found {
+			return nil, nil, serviceUnreachableMessage
+		case found:
 			// Legacy guard, mirroring resolveTokenForGet: a pre-resource set-alias
 			// row stored a raw URL, not an `r_` id. It can't resolve to a tunnel, so
 			// refuse with the same re-bind hint `/qurl get` gives — and skip the
--- a/apps/slack/internal/handler_display_name.go
+++ b/apps/slack/internal/handler_display_name.go
@@ -355,6 +355,15 @@ func (h *Handler) resolveTunnelByID(ctx context.Context, log *slog.Logger, teamI
 			// failure here is unlikely.
 			log.Warn("display-name: channel alias lookup failed", "error", lookupErr, "team_id", teamID, "channel_id", channelID, "id", id)
 		} else if found {
+			// Legacy guard, mirroring resolveTokenForGet: a pre-resource set-alias
+			// row stored a raw URL, not an `r_` id. It can't resolve to a tunnel, so
+			// refuse with the same re-bind hint `/qurl get` gives — and skip the
+			// resource scan it could never match (which, in a >first-page workspace,
+			// would otherwise misreport it as a lookup limit).
+			if !strings.HasPrefix(boundID, "r_") {
+				log.Warn("display-name: channel alias bound to a non-resource-id target", "team_id", teamID, "channel_id", channelID, "id", id)
+				return nil, nil, legacyAliasBindingMessage(id)
+			}
 			r, msg := h.resolveActiveTunnelByResourceID(ctx, log, c, teamID, id, boundID)
 			if msg != "" {
 				return nil, nil, msg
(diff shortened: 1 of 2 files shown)


Kevin-layerV:
Thanks — addressed in ffacd0c7:

- **#1 (precedence)** — documented `resolveTunnelByID` as deliberately **slug-first** (the reverse of `/qurl get`/`revoke`'s alias-first), including the one observable divergence (a token that's both a live slug *and* an alias to a different resource). Good catch; the PR body's "consistency" framing could have misled a future reader.
- **#4b (revoked target)** — added `TestSetDisplayName_ChannelAliasToRevokedTunnelRejected`. It's the actual production scenario and guards the `Status != Active` clause independently of the URL test's `Type` clause.
- **#3 (legacy non-`r_`)** — folded a clause into the helper doc so the absent `r_`-prefix guard reads as intentional.

Skipped, with reasons:
- **#2 (scan-window copy)** — same `listResourcesScanLimit` bound as `/qurl list`/get (tracked by #590); hedging the copy on pagination would over-word the common, genuinely-dead case. Left as-is.
- **#4a (`LookupChannelAlias` error branch)** — `AdminStore` is the concrete `*slackdata.Store`, not an interface, so there's no clean seam to inject a lookup error without a fake-DDB error config — low value for a log-and-fall-through. Skipped.

Kevin-layerV:
Addressed in 44cc5a1f:

- **#1 (scan-window remediation)** — agreed, this was the sharp edge; thanks for re-flagging it. Split the first-page-miss copy on `page.HasMore`: a definitively-dead target (non-tunnel / revoked / legacy non-`r_`) keeps the `unset-alias` hint, but a miss with **more pages** now gets a non-destructive *"lookup limit, not necessarily a stale alias"* message that doesn't nudge unbinding a still-live alias.
- **#2 (coverage)** — added `TestSetDisplayName_ChannelAliasPastScanWindow` (the branch you flagged: asserts no `unset-alias` in the copy, no PATCH). Left the `LookupChannelAlias`-error and `channelID==""` branches untested per the earlier rationale (no clean injection seam for the concrete `*slackdata.Store`; low value).
- **#3 (two `ListResources` calls on the alias path)** — acknowledged and accepted: inherent to slug-first + no get-by-id, and only the rare, admin-gated alias-miss path pays the extra hop.

Kevin-layerV:
Fixed in ba207dfc — good catch, that was a real bug I introduced in the round-2 change.

- **HasMore masking a seen-but-dead target** — now track `seen` separately from the active-tunnel match. Only `!seen && HasMore` gets the soft lookup-limit copy; a seen-but-dead target (revoked / URL resource), or a complete scan with no match, gets the definitive `unset-alias` hint.
- **Test** — added `TestSetDisplayName_RevokedTargetOnFirstPageWithMorePages` (revoked target on page 1 + `has_more=true` → asserts the stale-alias hint, *not* the lookup-limit copy), and parameterized the scan-window test server to drive both the "id absent from page 1" and "id present-but-dead on page 1" cases.
- **Legacy non-`r_` interaction** — a raw-URL binding still matches no `resource_id` → `seen=false`, so a >100-resource workspace lands it on the non-destructive lookup-limit copy rather than get's re-bind hint. Non-destructive and a third-order edge; consciously leaving the `r_` guard out per the earlier thread (agreed it reads as deliberate).

Kevin-layerV:
Both optional coverage items added in 361e4143:

- `TestSetDisplayName_NoChannelResolvesBySlug` — empty `channel_id` skips the alias fallback but slug resolution still works (channel_id isn't required).
- `TestSetDisplayName_ChannelAliasLookupErrorFallsThrough` — a `channel_policies` GetItem error logs + degrades to the not-found copy (no store error surfaced, no PATCH); the admin gate reads `workspace_mappings`, so only the alias read is failed (via `fakeDDB.SetGetItemErr`).

Precedence-shadowing stays as-is — documented and deliberate, per your note. Thanks for the thorough passes.

Kevin-layerV:
Done in 7260d9d4 — added the `r_`-prefix guard mirroring `resolveTokenForGet`:

- A non-`r_` binding (a legacy raw-URL `set-alias` row) is now refused with the shared `legacyAliasBindingMessage` re-bind hint **before** the scan — fixing item 1 (no more "lookup limit" copy for legacy rows) and item 2 (skips the second `ListResources` for them).
- `TestSetDisplayName_LegacyURLBindingRefused` locks it (asserts the re-bind hint, not "lookup limit", no PATCH).

That brings the display-name path fully in line with `/qurl get`'s legacy handling. Precedence-shadowing stays documented + deliberate (your nit #1 — house style). I believe this closes everything actionable.

Kevin-layerV:
Fixed in 9d4f4f0d — you're right that "the gate just read AdminStore" didn't cover it, since the gate reads `workspace_mappings` while the alias lookup reads `channel_policies`. On a `LookupChannelAlias` error the resolver now returns `serviceUnreachableMessage` (matching `resolveTokenForGet`) instead of the misleading "no such id" — so a `channel_policies` outage reads as *"Could not reach qURL. Please try again."* Test renamed + updated to assert that (and that it doesn't misreport "no such id").

On the two remaining notes: the second `ListResources` on the alias path is inherent (no get-by-id) and is now skipped for legacy rows; the missing emoji prefix on the error copy is intentional — consistent with the file's other error messages, as you noted. Calling this converged unless you spot a blocker. Thanks for the thorough review across the rounds.

justin-layerv:

Kevin-layerV:
Reopened and rebased onto current `main` (`f3fd25ff`); head is now `b90c7684` on `fix/slack-display-name-resolve-channel-alias`.

I also added one signed follow-up commit to address the URL-resource alias UX edge: Display Name commands now reject a live URL-resource alias with connector-only copy and do **not** recommend `unset-alias`; the existing URL-resource test now pins that non-destructive behavior. PR body is updated to match.

Verification:
- Local: pinned golangci-lint config verify + lint v2.10.1, `go test -race -count=1 -coverprofile=coverage.out -covermode=atomic ./apps/slack/... ./shared/...` (83.8% total coverage), and `go vet ./apps/slack/... ./shared/...`.
- GitHub: Slack CI is green, including lint, test, govulncheck, and Docker; other repo checks are green or correctly skipped.

Current PR state: open, mergeable, checks green; remaining block is review-required branch protection.

PR #670, opened 2026-06-09
feat(slack): wire conversation-mode read-only surface (dark-by-default)

justin-layerv:
## What

Wires the **conversation-mode read-only surface** in `cmd/main.go` so the agent (#650 Events surface, #648 agent core) can go live when an operator sets the env — and stays **DARK** otherwise. This is the in-repo half of #651; infra is tracked separately (see below).

`Handler.agentEnabled()` already gates on `AgentLLM`, `AgentStore`, and `PostMessage` all being non-nil **and** `AgentDisabled` being false. Production left all of them nil, so the surface never ran. This PR constructs them from env:

| Seam | Env | Dark when |
|---|---|---|
| `AgentLLM` | `ANTHROPIC_API_KEY` → `agent.NewAnthropicLLM` | key unset |
| `AgentStore` | `QURL_AGENT_STATE_TABLE` → `slackdata.NewAgentStoreFromEnv` | table unset |
| `PostMessage` | new `chat.postMessage` seam | always wired (harmless without the other two) |
| `AgentDisabled` | `QURL_AGENT_DISABLED` kill switch | — |

## How

- **PostMessage seam** (`cmd/slack_webapi.go`): a threaded `chat.postMessage` mirroring the existing `views.open` seam — per-workspace bot token via the existing lookup, the **same Enterprise Grid org-token fallback** as `openViewWithGridFallback` (the fallback lives in the seam here because `PostMessageFunc`'s signature carries `enterpriseID`, unlike OpenView's). The response parser branches on the **`ok` field**, since `chat.postMessage` returns HTTP 200 with `ok:false` for most errors.
- **`slackdata.NewAgentStoreFromEnv`**: loads AWS config + builds the DDB client inside the `slackdata` package (mirroring `NewStore`), keeping the composition root SDK-free.
- **Kill switch fails SAFE**: a set-but-unparseable `QURL_AGENT_DISABLED` *disables* conversation mode (never silently enables).
- **One startup log line** reports surface state; the partial-config branch **names the specific missing seam** so a configured-looking deploy is never silently dark.

## Scope / boundary

- **Read-only only.** `PostMessageBlocks` + `AgentConfirmEnabled` (the propose→confirm→execute card from PR4a/PR4b) stay **dark** — they're gated separately by the get-authz + R2 hard gates on #651, not part of this PR.
- **Dark-by-default & safe to merge.** Nothing in this repo sets these env vars at runtime (verified), so merging changes prod behavior by exactly zero until an operator opts in. Per the #651 rollout, enablement waits on the DPA / data-handling review (channel content is sent to Anthropic).

## Follow-ups

- Infra (`qurl-integrations-infra`, private): DDB table + IAM, `ANTHROPIC_API_KEY` secret, Slack manifest (`app_mention` + `message.im`, scopes). _(issue link to follow)_

## Tests

`go build ./... && go test -race ./...` green; `golangci-lint v2.10.1` clean. New `cmd/` tests cover the PostMessage seam (threaded/omitempty body, token lookup, Grid fallback + no-fallback, `ok:false` Slack error, 429 + json rate-limit, HTTP error, lookup/empty-token short-circuit, default UA) and the wiring matrix (LLM/store dark-vs-wired, kill-switch incl. fail-safe, surface-state logging incl. missing-seam naming).

Part of #651.

### DIFF
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -106,22 +106,26 @@ func TestLogAgentSurfaceState(t *testing.T) {
 		name          string
 		llm           bool
 		store         bool
+		post          bool
 		killed        bool
 		wantLevel     string
 		wantSubstr    string
 		wantMissing   string // when set, asserts the "missing" attribute contains it
 		wantNoMissing bool
 	}{
-		{name: "live", llm: true, store: true, wantLevel: "INFO", wantSubstr: "LIVE", wantNoMissing: true},
-		{name: "killed", llm: true, store: true, killed: true, wantLevel: "WARN", wantSubstr: "kill switch", wantNoMissing: true},
-		{name: "fully dark", wantLevel: "INFO", wantSubstr: "no agent seams", wantNoMissing: true},
-		{name: "partial: store missing", llm: true, wantLevel: "WARN", wantSubstr: "partially configured", wantMissing: slackdata.EnvAgentStateTable},
-		{name: "partial: llm missing", store: true, wantLevel: "WARN", wantSubstr: "partially configured", wantMissing: "ANTHROPIC_API_KEY"},
+		{name: "live", llm: true, store: true, post: true, wantLevel: "INFO", wantSubstr: "LIVE", wantNoMissing: true},
+		{name: "killed", llm: true, store: true, post: true, killed: true, wantLevel: "WARN", wantSubstr: "kill switch", wantNoMissing: true},
+		{name: "fully dark", post: true, wantLevel: "INFO", wantSubstr: "no agent seams", wantNoMissing: true},
+		{name: "partial: store missing", llm: true, post: true, wantLevel: "WARN", wantSubstr: "partially configured", wantMissing: slackdata.EnvAgentStateTable},
+		{name: "partial: llm missing", store: true, post: true, wantLevel: "WARN", wantSubstr: "partially configured", wantMissing: "ANTHROPIC_API_KEY"},
+		// PostMessage is wired unconditionally today, but the LIVE claim must still
+		// require it: LLM+Store set with PostMessage nil is partial, not LIVE.
+		{name: "partial: postmessage missing", llm: true, store: true, wantLevel: "WARN", wantSubstr: "partially configured", wantMissing: "PostMessage"},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			logs := captureSlog(t)
-			logAgentSurfaceState(tc.llm, tc.store, tc.killed)
+			logAgentSurfaceState(tc.llm, tc.store, tc.post, tc.killed)
 			records := logs()
 			if len(records) != 1 {
 				t.Fatalf("got %d log records, want 1: %v", len(records), records)
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -1,0 +1,145 @@
+package main
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"log/slog"
+	"strings"
+	"testing"
+
+	"github.com/layervai/qurl-integrations/apps/slack/internal/slackdata"
+)
+
+func TestBuildAgentLLMDarkUnlessKeySet(t *testing.T) {
+	cases := []struct {
+		name    string
+		key     string
+		wantNil bool
+	}{
+		{name: "unset", key: "", wantNil: true},
+		{name: "whitespace only", key: "   ", wantNil: true},
+		{name: "present", key: "sk-ant-test", wantNil: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("ANTHROPIC_API_KEY", tc.key)
+			llm := buildAgentLLM()
+			if (llm == nil) != tc.wantNil {
+				t.Fatalf("buildAgentLLM() nil=%v, want nil=%v", llm == nil, tc.wantNil)
+			}
+		})
+	}
+}
+
+func TestBuildAgentStoreDarkWhenTableUnset(t *testing.T) {
+	t.Setenv(slackdata.EnvAgentStateTable, "")
+	if store := buildAgentStore(context.Background()); store != nil {
+		t.Fatalf("buildAgentStore() = %v, want nil when table unset", store)
+	}
+}
+
+func TestBuildAgentStoreWiredWhenTableSet(t *testing.T) {
+	// LoadDefaultConfig resolves region/credentials lazily, so this stays offline.
+	t.Setenv("AWS_REGION", "us-east-1")
+	t.Setenv(slackdata.EnvAgentStateTable, "qurl-agent-state-test")
+	store := buildAgentStore(context.Background())
+	if store == nil {
+		t.Fatal("buildAgentStore() = nil, want a store when table is set")
+	}
+	if store.TableName != "qurl-agent-state-test" {
+		t.Fatalf("store.TableName = %q, want qurl-agent-state-test", store.TableName)
+	}
+}
+
+func TestReadAgentKillSwitch(t *testing.T) {
+	cases := []struct {
+		name string
+		val  string
+		want bool
+	}{
+		{name: "unset is not disabled", val: "", want: false},
+		{name: "true disables", val: "true", want: true},
+		{name: "1 disables", val: "1", want: true},
+		{name: "false does not disable", val: "false", want: false},
+		{name: "0 does not disable", val: "0", want: false},
+		// Fail-safe: an unparseable value must DISABLE, never silently enable.
+		{name: "garbage fails safe to disabled", val: "disable", want: true},
+		{name: "yes fails safe to disabled", val: "yes", want: true},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("QURL_AGENT_DISABLED", tc.val)
+			if got := readAgentKillSwitch(); got != tc.want {
+				t.Fatalf("readAgentKillSwitch(%q) = %v, want %v", tc.val, got, tc.want)
+			}
+		})
+	}
+}
+
+// captureSlog redirects the default slog logger to a buffer for the duration of
+// the test and returns a function that yields the captured JSON log lines.
+func captureSlog(t *testing.T) func() []map[string]any {
+	t.Helper()
+	var buf bytes.Buffer
+	prev := slog.Default()
+	slog.SetDefault(slog.New(slog.NewJSONHandler(&buf, &slog.HandlerOptions{Level: slog.LevelDebug})))
+	t.Cleanup(func() { slog.SetDefault(prev) })
+	return func() []map[string]any {
+		var out []map[string]any
+		for _, line := range strings.Split(strings.TrimSpace(buf.String()), "\n") {
+			if line == "" {
+				continue
+			}
+			var rec map[string]any
+			if err := json.Unmarshal([]byte(line), &rec); err != nil {
+				t.Fatalf("unmarshal log line %q: %v", line, err)
+			}
+			out = append(out, rec)
+		}
+		return out
+	}
+}
+
+func TestLogAgentSurfaceState(t *testing.T) {
+	cases := []struct {
+		name          string
+		llm           bool
+		store         bool
+		killed        bool
+		wantLevel     string
+		wantSubstr    string
+		wantMissing   string // when set, asserts the "missing" attribute contains it
+		wantNoMissing bool
+	}{
+		{name: "live", llm: true, store: true, wantLevel: "INFO", wantSubstr: "LIVE", wantNoMissing: true},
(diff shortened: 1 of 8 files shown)


justin-layerv:
Infra follow-up filed: layervai/qurl-integrations-infra#980 (DDB state table + IAM, `ANTHROPIC_API_KEY` secret, Slack manifest). This PR is dark-by-default until that lands **and** the DPA review clears.

justin-layerv:
Thanks — round-1 dispositions:

1. **Grid fallback was silent** — fixed. The seam now emits a `slog.Warn("workspace Slack bot token missing; retrying chat.postMessage with Enterprise Grid install token", team_id, enterprise_id)` before the org-token retry, matching `openViewWithGridFallback`. No `*slog.Logger` is threaded into the seam, so it uses the default logger (as you suggested).
2. **`EnvAgentStateTable` trim asymmetry** — hardened. `NewAgentStoreFromEnv` now `strings.TrimSpace`s the table name, so a whitespace-only value is rejected as empty even if the exported constructor is called directly. Removes the guard-vs-constructor inconsistency.
3. **Over-limit drain** — leaving as-is by design: it is byte-for-byte the established `views.open` drain pattern (bounded by the 4s client timeout), and diverging here would split the two seams. Agree it is not a change request.

Pushed as `849cc34d`. Build + `go test -race` + `golangci-lint v2.10.1` all green.

justin-layerv:
Round-2 dispositions (pushed `61f6f8a5`):

1. **Orphaned `buildAdminStore` doc comment** (bug) — **fixed.** Relocated so `buildAdminStore`'s comment sits directly on its declaration again (the four new `buildAgent*`/`readAgentKillSwitch`/`logAgentSurfaceState` functions now precede it, leaving a clean additive diff). Good catch.
2. **Grid-fallback warn uses `slog.Default()`, not a request logger** — conscious accept (documented in-code). The seam is constructed at boot and `PostMessageFunc` carries no `*slog.Logger`; threading one is a delivery-path/signature change beyond this wiring PR. The warn still carries `team_id` + `enterprise_id`, which are the load-bearing keys for "which install fell back" — `event_id` correlation would be a nicety, not the primary signal.
3. **`ratelimited` on a 200 can have an empty `Retry-After`** — **added a test** (`json ratelimited without retry-after`) asserting the rate-limit sentinel still fires with an empty hint, so that path is now pinned.
4. **Body re-marshaled on the Grid retry** — **fixed.** The JSON marshal is hoisted into the outer seam (the payload is owner-independent); `postBody` now takes the pre-marshaled bytes and the retry reuses them.

justin-layerv:
Round-3 dispositions (pushed test-only commit):

1. **Untested `slackChatPostMessageResponseError` branches** — **fixed.** Added `TestSlackPostMessageFuncRejectsOversizedResponse` (body > the 64 KiB cap → "response exceeded") and `TestSlackPostMessageFuncRejectsEmptyResponseBody` (200 with no body → "empty response body"). Both branches are now pinned.
2. **Double read of `EnvAgentStateTable`** — conscious accept, as you note it's intentional + documented (the `buildAgentStore` guard avoids an AWS config load for an unset table; `NewAgentStoreFromEnv` re-reads so the exported constructor is self-contained). Flagging-only, no change.
3. **`NewAgentStoreFromEnv` doesn't thread `awsConfigFns`** — conscious accept (YAGNI). `NewStore` exposes the hook only because its tests inject a fake DDB client; the agent store's tests construct `AgentStore` directly with a fake, so no config hook is needed yet. Easy to add a variadic option if a future test wants one.

Thanks for tracing the claims (Grid parity, ok-field branching, wiring order, kill-switch direction, concurrency) — that matches the intent. Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-4 dispositions (pushed):

1. **Inverted `slackChatPostMessageTimeout` comment** — **fixed.** You're right: `agentDeliveryBudget` is 15s, *looser* than the 4s client timeout, so for a single post the 4s timeout is the binding deadline (the 15s is the outer envelope across transcript-save + post). Rewrote the comment to say exactly that.
2. **Half-renamed shared consts** — **fixed.** Completed the `OpenView`→`API` rename for the three consts the new seam shares: `defaultSlackAPIUserAgent`, `slackAPIMaxErrorSnippetBytes`, `slackAPITruncationSuffix`. (The UA default is also used by the feedback webhook, so the rename improves accuracy there too.) The views.open-*specific* consts — `slackViewsOpenURL`/`slackViewsOpenTimeout`/`slackViewsOpenResponseBodyLimit` — are intentionally left, since chat.postMessage has its own (`slackChatPostMessage*`).
3. **Oversized-but-successful response treated as an error** — conscious accept. It's byte-for-byte the views.open bounded-read behavior; `postAgentReply` only logs (no retry), so there's no double-post; and agent replies are bounded well under 64 KiB, so the false-negative is unreachable in practice. Diverging here would split the two seams for a case that can't occur.

justin-layerv:
Round-5 dispositions (pushed):

1. **Whitespace-only `QURL_AGENT_STATE_TABLE` defeated the trim** (bug) — **fixed at the root.** `NewAgentStore` now trims its `EnvAgentStateTable` fallback (`tableName = strings.TrimSpace(os.Getenv(EnvAgentStateTable))`), so `"   "` is rejected as empty for *every* caller — exactly your suggested fix. Added a regression test (`TestNewAgentStore` now sets the env to `"   "` and asserts the required-name error). My earlier `NewAgentStoreFromEnv` trim was indeed in the wrong place; the doc guarantee now holds.
2. **`ratelimited` parser asymmetry** — **added the one-liner.** `slackChatPostMessageResponseError` now documents that the `ok:false:ratelimited` branch is intentionally chat.postMessage-specific (views.open surfaces rate limits via the 429 path), with a "do not harmonize" note for future readers.
3. **Grid-fallback `slog.Warn` lacks per-event correlation** — **filed #671** (enhancement, go, priority: low) to thread a request-scoped logger into the `PostMessage` seam so the fallback warn carries `event_id`/`surface` like `openViewWithGridFallback`. Out of scope for this wiring PR (it touches `PostMessageFunc`'s signature + the delivery worker); the current warn still carries `team_id`+`enterprise_id`.

Thanks for the thorough trace (signalCtx ordering, marshal-once reuse, bounded read, fail-safe kill switch, no token logging, rename completeness, the corrected timeout comment). Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-6 dispositions (pushed):

1. **`logAgentSurfaceState` could claim LIVE without `PostMessage`** — **fixed.** It now takes `postWired` and gates the LIVE line on all three seams (LLM + Store + PostMessage), mirroring `agentEnabled()` exactly — so it can never report LIVE while the handler stays dark, even if PostMessage becomes conditional later. I deliberately kept the friendly "no agent seams configured" line gated on just the operator-set seams (LLM + Store), so a pristine non-agent deploy isn't mislabeled "partially configured" (PostMessage is wired unconditionally). Added a `partial: postmessage missing` test case.
2. **Stale cached bot token not purged on call-time auth failure** — **filed #672** (enhancement, go, priority: low). Agreed it's pre-existing (the views.open path behaves identically) and cross-seam; the agent's @mention/DM path just exercises it more. Scoped it to detect `token_revoked`/`invalid_auth`/`account_inactive` `ok:false` codes and purge via the existing `invalidateWorkspaceSlackToken`, applied to both seams.
3. **Grid fallback can chain two 4s requests** — **fixed the comment.** `slackChatPostMessageTimeout` now notes the fallback makes two sequential requests (~8s worst case), still inside `agentDeliveryBudget`.

Nits (oversized-response drain-for-reuse; `//nolint:gosec G706` on the kill-switch log) — both intentional and confirmed, no change. Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-7 dispositions (pushed):

1. **Grid-fallback timing comment overstated the worst case** — **fixed, and a great catch.** You're exactly right: `errors.Is(err, ErrSlackBotTokenNotConfigured)` is only ever true via the token *lookup* (`fmt.Errorf("...token lookup: %w", err)`), which runs before any request is built; the empty-token, HTTP, and response paths never wrap that sentinel. So when the fallback fires the first attempt made **zero** HTTP round-trips, and the org-token retry is the first and only HTTP call — no ~8s chaining. My round-6 addition was wrong; corrected the comment to say so.
2. **Fallback `slog.Warn` loses surface context** — already tracked in **#671** (thread a request-scoped logger into the PostMessage seam for `event_id`/`surface` correlation). Documented-intentional for now; the warn still carries `team_id`+`enterprise_id`.
3. **Redundant `EnvAgentStateTable` read** in `buildAgentStore` — conscious accept (deliberate: skip the AWS config load when the table is unset/dark; the env is stable across the two reads). As noted in earlier rounds.

Thanks for re-confirming the kill-switch fail-safe, the `NewAgentStore` whitespace-trim, and the `logAgentSurfaceState` three-seam gate all held. Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-8 dispositions (pushed):

1. **`logAgentSurfaceState` mislabels a *construction* failure as "missing table"** — **fixed.** When `AgentStore` is nil because the table is set but `NewAgentStoreFromEnv` failed, the surface line previously said `missing QURL_AGENT_STATE_TABLE`, contradicting the `slog.Error` `buildAgentStore` already logs. Reworded that entry to "AgentStore (QURL_AGENT_STATE_TABLE unset, or construction failed — see the logged error)" so the two lines agree.
2. **Inlined `http.Client` vs a helper** — **fixed.** Extracted `defaultSlackPostMessageClient()` to mirror `defaultSlackViewsOpenClient()`; the two seams are now symmetric and the `CheckRedirect: ErrUseLastResponse` rationale lives in one place.
3. **No client-side empty channel/text check** — conscious accept, as you note it's a reasonable trade: the handler always supplies both, and Slack returns a precise `ok:false` (`channel_not_found`/`no_text`) that the seam already surfaces. Unlike the views.open JSON-root check (which guards a programming error), a fast-fail here would only duplicate a guarantee the single caller already provides.

Thanks for the thorough "things done well" pass — confirms the kill-switch fail-safe, the ratelimited divergence note, marshal-once reuse, the three-seam LIVE gate, and the `NewAgentStore` whitespace-trim all held. Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-9 dispositions (pushed):

1. **Wasted boot work under the kill switch** — **fixed.** `buildAgentLLM` + `buildAgentStore` now run only when `QURL_AGENT_DISABLED` is off. Since the switch is read once at boot and forces the surface dark via `agentEnabled` regardless (and un-killing needs a restart), the store/LLM built under a kill were dead weight; skipping them also avoids the AWS config load + DDB client construction. `logAgentSurfaceState`'s `killed` branch already short-circuits, so the startup line is unchanged.
2. **Drain comment drift** — **fixed.** Added the keep-alive drain rationale to the chat.postMessage oversized branch so it matches its `views.open` twin.

Thanks for re-verifying the wiring (signalCtx ordering, the three-seam LIVE gate, the corrected single-round-trip Grid-fallback comment, the 4s-inside-15s budget, the kill-switch fail-safe). Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-10 dispositions (pushed, both doc-only):

1. **Grid fallback only triggers on the lookup sentinel, not a Slack `ok:false` auth error** — consistent with `openViewWithGridFallback` by design, and the seam comment already documents the trigger. The related improvement (purge the cache + surface a typed error on `token_revoked`/`invalid_auth`) is tracked in **#672**. No change.
2. **Empty-but-`nil` token lookup won't fall back** — **added the one-line note** at the empty-token guard: a `("", nil)` lookup returns a plain error (not the sentinel), so no Grid retry; fine because the real `workspaceTokenLookup` returns `ErrSlackBotTokenNotConfigured`, and a future lookup returning `("", nil)` on a miss would need the sentinel instead.
3. **Kill-switch build-skip drops boot-time store validation** — made the trade-off explicit in the comment: I'm keeping the skip (un-killing requires a restart, which rebuilds and surfaces any `QURL_AGENT_STATE_TABLE` construction error at the moment it matters); validating a store you'll never use under a kill is wasted boot work.
4. **Two response parsers share structure** — intentional, kept separate with the "do not harmonize" comment, as you note. No change.

justin-layerv:
Round-11 — terminal. All four notes are confirming/cosmetic with nothing to change:

1. **Double env read** — conscious accept (purely cosmetic; the guard-before-config-load is intentional and the env is stable). Reading once via a passed-in name would be a fine future tidy-up but isn't worth re-opening the constructor split.
2. **Kill-switch hides table misconfig until restart** — confirmed the right trade-off (documented in `main.go`); the un-kill restart surfaces any construction error at the moment it matters.
3. **LIVE wording** — accurate as-is: it says "conversation mode (read-only) is LIVE" precisely because the confirm-mode seams (`PostMessageBlocks`/`AgentConfirmEnabled`) are gated separately and stay dark.
4. **Empty-token vs sentinel** — the inline comment captures the contract a future lookup must honor; no change.

Thanks for the thorough multi-round pass. Treating this as converged: every comment across the 11 rounds is fixed, tracked (#671 seam-logger correlation, #672 token-cache purge, infra#980 provisioning), or consciously accepted with rationale. Build + `go test -race` + `golangci-lint v2.10.1` green.

PR #673, opened 2026-06-09
feat(slack): extend confirm flow to protect-url + protect-connector (PR4c)

justin-layerv:
## What

Extends the conversation-mode confirm flow to the last two mutation kinds — **protect-url** and **protect-connector** — completing the propose→Approve→execute surface. Lands **dark** (gated by `AgentConfirmEnabled`, still false), like PR4a/PR4b.

Part of #662 (mutation-execution UX) and the staged confirm-flow rollout (#651).

## protect-url — direct execute (mirrors revoke/alias)

On Approve, validate `pa.URL` + `pa.Alias` through `parseResourceExposeArgs` (the single slash grammar source), then execute the **canonical** args via `exposeURLResourceInChannel` — resolve the existing URL resource by exact target, bind the channel alias in the click's channel. Benign public result.

- Closes the validate-reconstruct-execute-raw seam (round-5 lesson): executes the parser's canonical args, not the raw LLM-distilled values.
- Invalid input → a **generic** reply (`agentConfirmInvalidProtectURLReply`), never echoing the value onto the public card. The success result echoes only the validated channel alias + backend resource alias, never the raw URL.
- Admin-gated; channel-scoped to the click's channel.

## protect-connector — modal (the genuinely different one)

On Approve, open the **same** guided tunnel-install modal the slash wizard opens, using the click's fresh `trigger_id`. The modal **submit** (`processTunnelInstall`) stays the real enforcement + bootstrap-key delivery point. The proposal is sparse, so v1 opens the blank wizard.

- **Claim-before-open is required, not just consistent:** the modal submit isn't idempotent across opens, so consume-once stops two approvers double-opening → double-minting a connector.
- **Trigger window:** `triggerReceivedAt` is captured at the click handler entry (earliest measurable) and threaded through `processAgentConfirm`, so the connector branch budgets `views.open` exactly like the slash wizard (`slackTriggerOpenViewBudgetRemaining`) — trigger-expiry becomes a graceful "ask me again" card, not a wasted Slack RPC.
- **Key-delivery privacy:** `meta.UserID` = the approving admin, so the modal's same-user-submit gate aligns the **ephemeral** key target to that admin; `meta.ResponseURL` = the card's response_url (ephemeral install = the same privacy class PR4a/PR4b already ship). `processTunnelInstall` already revokes the key if Slack doesn't confirm delivery, so the only residual is "renders in-channel" — a **hard pre-enablement smoke test** (see #651).
- `openAgentConnectorModal` skips the admin re-check (`processAgentConfirm` already did it before the claim), which also preserves trigger budget. `confirmModalRouted` is a named predicate shared by the click router and the lockstep test so they can't drift.

## Pre-enablement gates (on #651)

Two new **hard** gates filed on #651, siblings to the existing R2 + get-authz gates: **C1** connector credential-privacy smoke test (a second channel member can't see the key in-channel), **C2** connector trigger-window smoke test (with **Option B** — a fresh-trigger "Open setup form" button — noted as the escape hatch if the window proves flaky).

## Tests

`go build ./... && go test -race ./...` green; `golangci-lint v2.10.1` clean. New: protect-url on-approve (claims, binds, benign result) + invalid-input (non-http URL, backtick/bidi/missing alias → generic reply, no echo) + admin-gated + snapshot; protect-connector opens one modal with the click's trigger + `meta.UserID==approver` (+ channel/response_url) + admin-gated (no open, no claim) + trigger-expired (no open, claimed) + open-fails + OpenView-unconfigured graceful cards. The lockstep test excludes modal-routed kinds via the shared predicate; the deferred-preview gate uses a synthetic non-executable kind (every real kind is executable as of PR4c).

### DIFF
--- a/apps/slack/internal/agent/tools.go
+++ b/apps/slack/internal/agent/tools.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
+	"net/url"
 	"strconv"
 	"strings"
 )
@@ -331,6 +332,14 @@ func proposalProtectURL(f map[string]string) (*Proposal, error) {
 	target := strings.TrimSpace(f[fieldURL])
 	if target == "" {
 		return nil, errEmptyField(toolProposeProtectURL, fieldURL)
+	}
+	// Validate scheme/host at the propose layer, not just at execute — same
+	// no-dead-end-card reason as the alias check below: a malformed/non-http URL
+	// could only fail on Approve, after the claim consumes the card. Mirrors
+	// parseResourceExposeArgs's absolute-http(s) gate (the confirm execute
+	// re-validates through that grammar, but a bad URL never builds a live card).
+	if u, err := url.Parse(target); err != nil || u.Host == "" || (u.Scheme != "http" && u.Scheme != "https") {
+		return nil, fmt.Errorf("%s: %s must be an absolute http(s) URL", toolProposeProtectURL, fieldURL)
 	}
 	alias := normalizeToken(f[fieldAlias])
 	if alias == "" {
--- a/apps/slack/internal/agent/tools.go
+++ b/apps/slack/internal/agent/tools.go
@@ -5,7 +5,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"math"
-	"net/url"
 	"strconv"
 	"strings"
 )
@@ -329,24 +328,19 @@ func proposalProtectConnector(f map[string]string) (*Proposal, error) {
 }
 
 func proposalProtectURL(f map[string]string) (*Proposal, error) {
+	// Field-PRESENCE checks only (URL + alias must be provided), which steer the LLM
+	// to re-prompt. The full GRAMMAR (absolute http(s) URL, alias charset, no
+	// embedded whitespace) lives in parseResourceExposeArgs in the internal package,
+	// which this agent package can't import; the confirm layer enforces it before
+	// rendering a card (Handler.confirmDeliverable → confirmValidProtectURL, the same
+	// validator the execute path uses), so propose and execute can't drift and a
+	// grammar-invalid proposal never becomes a live Approve card.
 	target := strings.TrimSpace(f[fieldURL])
 	if target == "" {
 		return nil, errEmptyField(toolProposeProtectURL, fieldURL)
 	}
-	// Validate scheme/host at the propose layer, not just at execute — same
-	// no-dead-end-card reason as the alias check below: a malformed/non-http URL
-	// could only fail on Approve, after the claim consumes the card. Mirrors
-	// parseResourceExposeArgs's absolute-http(s) gate (the confirm execute
-	// re-validates through that grammar, but a bad URL never builds a live card).
-	if u, err := url.Parse(target); err != nil || u.Host == "" || (u.Scheme != "http" && u.Scheme != "https") {
-		return nil, fmt.Errorf("%s: %s must be an absolute http(s) URL", toolProposeProtectURL, fieldURL)
-	}
 	alias := normalizeToken(f[fieldAlias])
 	if alias == "" {
-		// Required at the propose layer: exposing a URL must bind a channel alias
-		// (exposeURLResourceInChannel + the slash `as:$alias` grammar both require
-		// it). Erroring here means an alias-less proposal — whose confirm card could
-		// only fail on Approve — is never built; the agent asks the user instead.
 		return nil, errEmptyField(toolProposeProtectURL, fieldAlias)
 	}
 	return &Proposal{
--- a/apps/slack/internal/agent/tools.go
+++ b/apps/slack/internal/agent/tools.go
@@ -118,9 +118,11 @@ func toolSpecs() []ToolSpec {
 			Description: "Propose protecting an existing URL (a reachable HTTP endpoint). Admin-gated. Does NOT execute — the user must confirm.",
 			Schema: map[string]any{
 				fieldURL:   stringProp("The target URL to protect."),
-				fieldAlias: stringProp("Suggested channel alias for the protected URL."),
+				fieldAlias: stringProp("Channel alias to bind the protected URL to (required — Slack users need a friendly name). Ask the user if not given."),
 			},
-			Required: []string{fieldURL},
+			// alias is REQUIRED: exposing a URL must bind a channel alias, so an
+			// alias-less proposal could only fail at execute — don't let one be made.
+			Required: []string{fieldURL, fieldAlias},
 		},
 	}
 }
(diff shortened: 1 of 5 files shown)


justin-layerv:
Round-1 dispositions (pushed):

1. **Connector open-error copy collapsed all causes into "ask me again"** — **fixed.** Added `agentConfirmConnectorOpenErrorReply(err)`, mirroring the slash wizard's cause distinction: `ErrSlackRateLimited` → "wait a moment, then ask again"; `auth.ErrSlackBotTokenNotConfigured` (Grid fallback couldn't cover it) → the unavailable/"use the slash command" copy (re-asking the agent won't fix a missing token); only `ErrSlackTriggerExpired`/`DeadlineExceeded`/transient → "ask me again" (which *is* the fix). Table test pins all four mappings; the distinct kinds stay in the structured log too.
2. **protect-url used `payload.Channel.ID`, connector used `pa.ChannelID`** — **fixed.** Connector now uses `payload.Channel.ID` (consistent with the other execute paths; == the proposal's channel by the mismatch guard). The modal is blank in v1, so `pa` became unused and I dropped it from `openAgentConnectorModal`.
3. **`unwrapSlackURLArg` invariant on the confirm path** — conscious accept with a doc note. Reusing `parseResourceExposeArgs` runs the LLM-distilled URL through the Slack-unwrap/unescape, whose "Slack escaped every `&`" invariant is best-effort here — but any decoding only affects the exact-target lookup (mismatch → benign "not found") and the value is never echoed, so no security impact. Keeping the single grammar source (no drift between slash + confirm) is worth more than bypassing the unwrap for this one caller; documented in `confirmValidProtectURL`.

justin-layerv:
Round-2 dispositions (pushed):

1. **Alias-less protect-url was a dead-end Approve card** (main point) — **fixed at the source.** `proposalProtectURL` now requires the channel alias: `fieldAlias` is added to the tool's `Required` set and the builder errors when it's empty. Exposing a URL fundamentally must bind a channel alias (both `exposeURLResourceInChannel` and the slash `as:$alias` grammar require it), so an alias-less proposal could only fail on Approve — and since the claim precedes execute, the card would be consumed with no retry. Now that proposal is never built; the agent asks the user for the alias instead. Propose-layer test added (`protect_url requires alias`).
2. **Connector card rendered even when OpenView was unwired** (minor) — **fixed.** New `confirmDeliverable` gates the card on executable AND (for a `confirmModalRouted` kind) `OpenView != nil`, so a protect-connector proposal on a workspace without OpenView falls back to the honest text preview rather than a burnable Approve. Same "don't render a dead-end card" principle as the protect-url fix; two delivery-gating test cases added (OpenView wired → card, unwired → preview).
3. **Double `openBudget` computation** — conscious accept, as you note it's the intentional pre-render early-out, consistent with `openTunnelInstallWizard`. No change.

justin-layerv:
Round-3 dispositions (pushed):

1. **URL not validated at the propose layer** (consistency) — **tweaked, not just noted.** You're right that the round-2 "could only fail on Approve → claim burned" argument applies identically to a malformed/non-http URL. `proposalProtectURL` now runs the same absolute-http(s) scheme/host check as `parseResourceExposeArgs` (`url.Parse` + host + http/https), so a bad URL is rejected at propose and never surfaced as a live Approve card. protect-url now fully matches the no-dead-end-card guarantee the alias check established. Added a propose-layer test (`protect_url rejects non-http url`). (The confirm execute still re-validates through the slash grammar — the propose check is the early gate, not a replacement.)
2. **Open-error copy collapses trigger-expired + deadline** — conscious accept, as you note both resolve to "ask me again" (re-asking mints a fresh trigger, which is the fix for both). The slash path's finer split ("window expired" vs "Slack didn't respond") doesn't change the user's next action here, so the confirm card keeps one reply for both; the distinct kinds stay in the structured log.
3. **Key-delivery in-channel residual** — acknowledged, gated as hard pre-enablement smoke test C1 on #651; `meta.UserID == approver` aligns the same-user-submit gate. No pre-enablement action.

justin-layerv:
Round-4 dispositions (pushed):

1. **Redundant first budget check in `openAgentConnectorModal`** — **removed.** You're right: the slash wizard's two checks bracket a *network* `urlResourceSelectOptions` fetch, but the confirm path only renders the modal in-memory between them, so the second check (immediately before `views.open`) catches the same expiry. Dropped the first; kept one pre-open check with a note on why one suffices here. `TestConfirm_ProtectConnectorTriggerExpired` still passes (the remaining check catches it).
2. **Grid-fallback path unexercised from the confirm flow** — **closed.** Added `TestConfirm_ProtectConnectorGridFallback`: workspace token missing + `payload.Enterprise.ID` set → the open retries with the org token (`owners == [T1, E1]`). Pins that the connector branch threads `enterpriseID` into `openViewWithGridFallback` (the fallback mechanism itself stays covered via the slash path).
3. **Privacy residual / `meta.UserID == approver`** — acknowledged; gated as hard pre-enablement smoke tests C1/C2 on #651, feature dark behind `AgentConfirmEnabled`. No pre-enablement action.

Also flagged the grammar-injection point: the `strings.Fields` 1-2-token cap in `parseResourceExposeArgs` is indeed what closes the door (a smuggled `as:$other` → 3 tokens → generic reply), beyond the `unwrapSlackURLArg` note — agreed that's the load-bearing guard. Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-5 dispositions (pushed):

1. **Propose validation didn't fully mirror the execute grammar** (substantive) — **fixed at the right layer, with one grammar source.** You're right that a whitespace-in-URL or out-of-charset alias (`MyDocs`, `my docs`) passed `proposalProtectURL`'s partial `url.Parse` check but dead-ends at execute (`strings.Fields` split / `validateAliasTokenForNoun`). Key constraint: the **agent package can't import `parseResourceExposeArgs`** (it's in `internal`, which imports `agent`, not vice versa), so a propose-layer check can only ever be a partial reimplementation — exactly the drift you flagged. The no-drift fix is at the confirm-deliver layer: **`confirmDeliverable` now validates a protect-url proposal through `confirmValidProtectURL` — the *same* validator execute uses** — before rendering a card; a grammar-invalid proposal falls back to the text preview instead of a burnable button. I removed the partial propose-layer scheme/host check; `proposalProtectURL` keeps only field-presence checks (URL+alias required) to steer the LLM, documented as such. Added deliver-layer tests for the whitespace-URL and out-of-charset-alias cases (the propose→approve path the seed-`pendingAction` tests don't reach).
2. **`context.DeadlineExceeded` → window-expired mapping** (minor) — no code change needed: that case hits the **default** branch of `agentConfirmConnectorOpenErrorReply` (only `ErrSlackBotTokenNotConfigured` and `ErrSlackRateLimited` are special-cased), so the reply maps to window-expired whether or not `errors.Is(DeadlineExceeded)` matches — it doesn't depend on the OpenView impl wrapping the context error. (The structured log does use `errors.Is` for observability; the real `newSlackOpenViewFunc` wraps with `%w`, so it's accurate there too.)
3. **`confirmValidProtectURL` doc comment** — thanks; left as-is.

justin-layerv:
Round-6 — terminal. All three observations are confirming/intentional with nothing to change:

1. **`confirmValidProtectURL` runs twice** (render-time `confirmDeliverable` gate + execute-time `executeAgentAction` gate) — deliberate defense-in-depth: the render gate prevents the dead-end card, the execute gate is the load-bearing security validation that also yields the canonical args. Cheap + idempotent, as you note.
2. **HTML-unescape on the LLM path** — acknowledged in the `confirmValidProtectURL` doc comment; any decode only perturbs the exact-target lookup (mismatch → benign "not found") and the value is never echoed. Keeping the comment as the rationale.
3. **C1 (key-privacy in-channel) + C2 (trigger-window)** — tracked as **hard** pre-enablement gates on #651, must clear before `AgentConfirmEnabled` flips.

Thanks for the six-round pass — the robustness gaps you surfaced (cause-specific copy, the propose→execute no-dead-end-card holes, the propose/execute grammar drift) each tightened the design. Treating this as converged: every comment is fixed or consciously accepted with rationale, CI green, `/simplify` clean.

PR #674, opened 2026-06-09
feat(slack): wire confirm-flow seams (PostMessageBlocks + AgentConfirmEnabled), dark-by-default

justin-layerv:
## What

Wires the **confirm-flow Config seams** in `cmd/main.go` so the propose→Approve→execute mutation surface (built dark across PR4a/PR4b/PR4c) can go live when an operator opts in — and stays **DARK** otherwise. This is the in-repo half of confirm-flow enablement; it mirrors #670's read-only wiring.

`Handler.agentConfirmEnabled()` is `agentEnabled() && AgentConfirmEnabled && PostMessageBlocks != nil`. Production left `PostMessageBlocks` nil and `AgentConfirmEnabled` false, so the confirm flow never ran. This PR constructs them:

| Seam | Source | Dark when |
|---|---|---|
| `PostMessageBlocks` | new `chat.postMessage`-with-blocks seam | always wired (harmless without the flag) |
| `AgentConfirmEnabled` | `QURL_AGENT_CONFIRM_ENABLED` | flag unset/false |

## How

- **Shared poster.** Extract `slackPostMessagePoster` (token lookup + Enterprise Grid fallback + response parsing) from the text PostMessage seam; `newSlackPostMessageBlocksFuncWithTokenLookup` reuses it — only the JSON body differs (`blocks` + a notification fallback `text`). No behavior change to the existing PostMessage seam (its tests still guard it); this also closes the parallel-seam reuse note from #670's review.
- **Fail-safe flag.** `AgentConfirmEnabled` from `QURL_AGENT_CONFIRM_ENABLED` via `readBoolEnvFailSafe` — a set-but-unparseable value is treated as **OFF**, so a typo can never turn mutation execution on (the kill switch's mirror: it fails to *disabled*).
- **Effective-state logging.** `logAgentSurfaceState` now takes an `agentSurfaceState` struct and reports the confirm/mutation line keyed on the **effective** predicate (read-only live AND flag AND `PostMessageBlocks`), never the raw env bool — a flag set while the surface is dark reads "set but DARK", not "LIVE" (the #670 LIVE-gate-consistency lesson, applied to the riskier flip). When confirm mode is actually live it loudly lists the hard pre-enablement gates that must clear first.

## Scope / safety

- **Dark-by-default & safe to merge.** Nothing in-repo sets `QURL_AGENT_CONFIRM_ENABLED`, so merging changes prod behavior by zero. Mutations go live only when an operator flips the flag atop a live read-only surface.
- **Gated on four HARD pre-enablement checks** (all on #651) before the flag should ever flip: **get-authz** (asker-only vs any-member), **R2** (public-card `replace_original`), **C1** (connector key renders ephemerally, never in-channel), **C2** (connector trigger-window under real latency) — plus the DPA / data-handling review.
- **No manifest change:** the Approve/Reject buttons reuse the existing `block_actions` interactivity endpoint (same as the list-revoke / tunnel-install buttons).

## Tests

`go build ./... && go test -race ./...` green; `golangci-lint v2.10.1` clean. New `cmd/` tests: the blocks seam (body carries blocks + fallback, Grid fallback threads enterpriseID, `ok:false` error), `readAgentConfirmEnabled` dark/enabled/fail-safe matrix, and the confirm-mode log keyed on effective state (live vs "set but dark" for each dark reason — killed, read-only dark, blocks unwired).

Part of #651.

### DIFF
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -188,7 +188,9 @@ func TestLogAgentSurfaceState_ConfirmMode(t *testing.T) {
 		{name: "confirm live", state: with(wired, func(s *agentSurfaceState) { s.confirmFlag = true }), wantConfirm: "CONFIRM (mutation execution) is LIVE"},
 		{name: "flag set but blocks unwired → dark", state: with(wired, func(s *agentSurfaceState) { s.confirmFlag = true; s.blocksWired = false }), wantConfirm: "confirm mode is DARK"},
 		{name: "flag set but read-only dark → dark", state: agentSurfaceState{storeWired: true, postWired: true, blocksWired: true, confirmFlag: true}, wantConfirm: "confirm mode is DARK"},
-		{name: "flag set but killed → dark", state: with(wired, func(s *agentSurfaceState) { s.confirmFlag = true; s.killed = true }), wantConfirm: "confirm mode is DARK"},
+		// Killed: only the kill-switch line; no separate confirm line (it would name the
+		// seams, not the kill switch, as the blocker).
+		{name: "flag set but killed → no confirm line", state: with(wired, func(s *agentSurfaceState) { s.confirmFlag = true; s.killed = true }), wantConfirm: ""},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -125,7 +125,8 @@ func TestLogAgentSurfaceState(t *testing.T) {
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
 			logs := captureSlog(t)
-			logAgentSurfaceState(tc.llm, tc.store, tc.post, tc.killed)
+			// confirmFlag stays false here, so only the read-only line is emitted.
+			logAgentSurfaceState(agentSurfaceState{llmWired: tc.llm, storeWired: tc.store, postWired: tc.post, killed: tc.killed})
 			records := logs()
 			if len(records) != 1 {
 				t.Fatalf("got %d log records, want 1: %v", len(records), records)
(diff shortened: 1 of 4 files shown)


justin-layerv:
Round-1 dispositions (pushed):

1. **Production blocks impl didn't disable top-level mrkdwn** — **fixed (took your first option).** `newSlackPostMessageBlocksFuncWithTokenLookup` now marshals `mrkdwn:false`, so Slack renders the fallback `text` literally — fulfilling the `PostMessageBlocksFunc` seam's defense-in-depth contract regardless of whether the caller also escaped it. This removes the dependency on every future caller remembering to escape, which matters most exactly where you flagged: a prompt-injected summary in the notification/non-block preview next to a live Approve button. The card itself already renders the summary as `plain_text`. Test asserts `mrkdwn` is explicitly `false` (pointer field, so "sent as false" ≠ "absent").
2. **Empty `blocks:[]` (no omitempty)** — conscious accept, as you note: the confirm card always carries blocks, so the empty case is unreachable in production; a test exercises it only to confirm the seam doesn't choke.
3. **Two near-identical marshal structs (text vs blocks)** — agreed, leaving separate per the "only the JSON body differs" framing; not worth a shared type.

justin-layerv:
Round-2 dispositions (pushed):

1. **Double warn when killed + flag set** — **fixed (suppress when killed).** Gated the confirm-DARK line on `!killed`. Rationale beyond cosmetics: under a kill, that line's hint ("until the read-only surface is live and PostMessageBlocks is wired") would name the *seams* as the blocker, when the actual blocker is the kill switch — so it was slightly misleading. Now the kill-switch line stands alone as the accurate cause; the un-kill restart re-reports confirm state when it's actionable. Test updated (killed + flag → only the kill-switch line).
2. **`Blocks []any` no `omitempty` (nil → `"blocks":null`)** — conscious accept, as you note: the confirm flow always populates blocks, so nil-blocks is unreachable in production; not worth an `omitempty` that would only matter for a call that never happens.

Thanks for verifying the gate consistency (byte-for-byte with `agentEnabled`/`agentConfirmEnabled`), the fail-safe poles, and that `mrkdwn:false` doesn't regress `escapeMrkdwnText`'s entity-encoding (Slack requires entity-encoding at the API level regardless of formatting). Build + `go test -race` + `golangci-lint v2.10.1` green.

justin-layerv:
Round-3 dispositions (pushed):

1. **"share this transport" overstated the sharing** — **fixed.** Reworded the poster doc: the two seams each construct their OWN poster (own `http.Client`) and reuse this transport *type*; they don't share one instance/pool. Added a note that a single shared poster is the move only if connection pooling ever matters (negligible for a low-volume bot). Took option (a) — the 15 test call sites that construct seams directly make a single-poster refactor more churn than the pooling is worth here.
2. **Two warnings on partial-config + flag set** — conscious accept; intentional and informative, as you note. Deliberately NOT suppressed like the killed case: here the confirm-DARK line's hint ("until the read-only surface is live and PostMessageBlocks is wired") accurately names the missing seam as the blocker, whereas under a kill it would misname the cause — so the two cases are treated differently on purpose.

justin-layerv:
Round-4 — terminal. All three observations are confirming/accepted with nothing to change:

1. **Two HTTP clients** — deliberate trade, already documented in the poster doc-comment (a single shared poster pools connections if `chat.postMessage` volume ever climbs; negligible for a low-volume bot).
2. **Generic warn message** — intentional: the specific env name moved to the `"env", name` structured attribute, so log-search by env still works while the message string stays shared across both flags.
3. **`channelID` threading** — confirmed correct in both seam closures.

Thanks for the four-round pass. Treating this as converged: every comment is fixed (mrkdwn:false defense-in-depth, killed-suppression, transport-type reword) or consciously accepted with rationale; CI green; `/simplify` clean. The flag flip stays held on the four HARD pre-enablement gates (get-authz, R2, C1, C2) + DPA.

PR #675, opened 2026-06-09
feat(slack): statecrawl — channel_policies reconciler for #654/#669 orphaned aliases

Kevin-layerV:
## What

Adds `apps/slack/cmd/statecrawl`, an operational reconciler for the qURL Slack bot's `channel_policies` DynamoDB table. It crawls every `(team, channel)` policy row, cross-references each referenced resource against the owning workspace's live qURL resources, and reports — and optionally repairs — the state PRs **#654** and **#669** leave us in.

This is the lever for **safely unblocking a customer ASAP** when they hit the contradictory state #654 fixes going forward: a `$alias` that reports *"already bound"* while `/qurl list` shows nothing.

## Why

- **#654** (merged) cascades `PurgeResourceFromChannel` on every revoke, so a destroyed resource no longer orphans a channel `$alias`. It does **not** backfill orphans that already exist — pre-fix revokes, or out-of-band deletes (API / SDK / MCP / CLI / token expiry) the bot never observed. `statecrawl` finds those orphans and, in a mutating run, clears them via the **same** `slackdata.Store.PurgeResourceFromChannel` verb the bot now runs, so the manual backfill is behaviorally identical to the live cascade.
- **#669** (open) makes `set/unset-display-name` resolve a channel `$alias` whose name ≠ the connector slug. `statecrawl` flags those live `name ≠ slug` bindings (informational; **never** purged).

## Structure

Deliberately mirrors qurl-service's operational CLIs (`cmd/qurl-scanner`, `cmd/qurl-bucket-backfill`):

- **`-dry-run` defaults `true`** (read-only). Mutating needs `-dry-run=false`; against a prod-looking deployment (the `-env` label, or any `prod` table name as defense-in-depth) it **also** requires `-allow-prod-purge` — a parse-time "rail" whose reject error names the flag to add.
- **`Stats`/`Snapshot`** atomic counters → a structured `statecrawl complete` summary line. `purge_errors` is ALERTABLE and forces a non-zero exit.
- **slog (JSON default, `-log-format=text`)** so every finding and every purge is an auditable, greppable record.
- Mutations only ever touch the two surfaces `PurgeResourceFromChannel` owns, and only for orphans **confirmed** dead against a fully paginated resource list. A team whose API key can't be resolved (or whose list fails) is reported `indeterminate` and never purged.

## Safety / testing — proven non-destructive without touching prod

The reconcile flow is dependency-injected (`crawl` takes the purge Store, API-key provider, and Scan client as interfaces) and exercised end-to-end with fakes:

- **Dry run issues ZERO DynamoDB mutations** while still reporting the orphan as `dry_run_would_purge`.
- **Apply** drives a *real* `slackdata.Store` over a fake DDB and asserts the orphan is cleared by exactly one `PurgeResourceFromChannel` `UpdateItem` (DELETE the dead id from `allowed_resource_ids` + REMOVE the orphaned alias key), leaving the healthy alias intact.
- **Indeterminate-never-purged** even in apply mode.
- Plus: liveness full-pagination / not-configured / list-error degradation; Scan pagination + keyless-row skipping; the prod-purge rails; classification of every finding kind.

Race-clean, ~80% statement coverage (the uncovered remainder is `main`/AWS wiring).

## Operator runbook

```sh
# 1. Dry run, scoped to the affected workspace:
go run ./apps/slack/cmd/statecrawl -env prod -team T0123ABCD
# 2. Review findings (orphan-alias / orphan-allowed-id are the purge targets).
# 3. Apply, with the prod opt-in, still scoped to one workspace:
go run ./apps/slack/cmd/statecrawl -env prod -team T0123ABCD -dry-run=false -allow-prod-purge
```

See `apps/slack/cmd/statecrawl/README.md` for full config and `jq` filters.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JoivSNDHoARMQWy55do3Gn)_

### DIFF
--- a/apps/slack/cmd/statecrawl/README.md
+++ b/apps/slack/cmd/statecrawl/README.md
@@ -1,0 +1,94 @@
+# statecrawl
+
+An operational reconciler for the qURL™ Slack bot's `channel_policies` DynamoDB
+table. It crawls every `(team, channel)` policy row and cross-references each
+referenced resource against the owning workspace's live qURL resources, then
+reports — and optionally repairs — the state that PRs **#654** and **#669**
+leave us in.
+
+## Why this exists
+
+- **#654** (merged) taught the bot to cascade `PurgeResourceFromChannel` on
+  every revoke, so destroying a resource no longer orphans a channel `$alias`.
+  It does **not** backfill orphans that already exist — pre-fix revokes, or
+  out-of-band deletes (API / SDK / MCP / CLI / token expiry) the bot never
+  observed. `statecrawl` finds those orphans (an `$alias`, or an
+  `allowed_resource_ids` member, still pointing at a revoked/deleted resource)
+  and, with `-apply`, clears them through the **same** `PurgeResourceFromChannel`
+  verb the bot now runs — so the manual backfill is behaviorally identical to
+  the live cascade.
+
+- **#669** (open) taught `set/unset-display-name` to resolve a channel `$alias`
+  whose name differs from the connector slug. `statecrawl` flags those live
+  `name ≠ slug` bindings so you can see which connectors became
+  display-name-targetable by alias. These are a **healthy, live** state and are
+  **never** purged — informational only.
+
+## Safety
+
+- **Dry run is the default** and makes zero mutations. It only reads: DynamoDB
+  `Scan`/`GetItem`, KMS `Decrypt` (for the per-workspace API key), and
+  `GET /v1/resources`.
+- `-apply` performs the orphan purge. It only ever touches the two surfaces
+  `PurgeResourceFromChannel` owns (`allowed_resource_ids` + `alias_bindings`),
+  and only for resources **confirmed** revoked/deleted against a fully paginated
+  resource list. A workspace whose API key can't be resolved, or whose resource
+  list fails to load, is reported as `indeterminate` and is **never** purged.
+
+## Configuration
+
+Each value comes from an env var (the same ones the bot uses) or its `-flag`
+override. Run the tool **once per deployment**: point the env at sandbox, run a
+dry run, review, optionally `-apply`; then repeat for prod. `-env` only stamps a
+label into the report and the apply banner — it does **not** pick tables.
+
+| Env var                         | Flag                         | Purpose                                            |
+| ------------------------------- | ---------------------------- | -------------------------------------------------- |
+| `QURL_CHANNEL_POLICIES_TABLE`   | `-channel-policies-table`    | `channel_policies` table to crawl                  |
+| `QURL_WORKSPACE_MAPPINGS_TABLE` | `-workspace-mappings-table`  | `workspace_mappings` table (constructs the Store)  |
+| `WORKSPACE_STATE_TABLE`         | `-workspace-state-table`     | per-workspace qURL API key table                   |
+| `WORKSPACE_STATE_KMS_KEY_ARN`   | `-kms-key-arn`               | CMK that envelope-encrypts the API key column      |
+| `QURL_ENDPOINT`                 | `-qurl-endpoint`             | qurl-service base URL for the liveness check       |
+|                                 | `-env`                       | report/banner label (e.g. `sandbox`, `prod`)       |
+|                                 | `-team`                      | restrict the crawl to one Slack `team_id`          |
+|                                 | `-page-limit`                | `/v1/resources` page size (default 100)            |
+|                                 | `-apply`                     | perform the purge (default: read-only dry run)     |
+
+AWS credentials and region come from the ambient environment. The role needs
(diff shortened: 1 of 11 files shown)


Kevin-layerV:
Thanks for the review — addressed all of it in `1e73d18`:

1. **Query, not Scan+filter, for `-team`** — `scanPolicyRows` now `Query`s the `slack_team_id` partition key for the scoped fast path (matching `ChannelsForResource`); the unscoped sweep keeps the `Scan`. New test asserts the single-team path issues zero `Scan` calls.
2. **Ownership invariant** — documented in `resolveLiveness`: liveness is built from the workspace's own key, so the purge is safe only while every reference is same-workspace-owned; a future cross-org-sharing feature must revisit it before it can become data loss.
3. **`isResourceID` guard in `classifyAllowedIDs`** — added, so a malformed non-`r_` SS member is never an orphan/purge target (symmetry with the alias path).
4. **`maxResourcePages` fail-safe** — `listAllResources` now errors out (→ team reported `indeterminate`, never purged) instead of looping on a server returning `HasMore=true` with a stable cursor.

Minors: `groupByTeam` buckets in one pass; README notes why `orphan_*` can exceed `purged` (dedup by `(team, channel, resource)`).

CI confirmation: verified clean locally with the CI-pinned `golangci-lint v2.10.1` and `go test -race`. The earlier `slack / lint` + `slack / required` failures were the v2.10.1 findings (hugeParam/rangeValCopy/gosec G705/misspell), fixed in `06b3b3f`; the new run should be green.

---
_Generated by [Claude Code](https://claude.ai/code)_

justin-layerv:

Kevin-layerV:
Thanks — addressed in `5649f38` (on top of the rebase onto current `main`, which also fixed the CI break: `auth.Provider` grew `SupportsDeleteAPIKey`/`DeleteAPIKey` in the interim, so the test fake needed them).

1. **Schema-constant lockstep guard (medium)** — agreed this was the important one. Added `TestSchemaConstants_MatchSlackdataWritePath`: it runs slackdata's *real* `ExposeResourceToChannel` + `BindChannelAlias` verbs over a capturing fake and asserts statecrawl's re-declared constants (`slack_team_id`, `slack_channel_id`, `allowed_resource_ids`, `alias_bindings`) appear in the DynamoDB requests slackdata actually issues. A rename in slackdata now fails CI here instead of silently making the crawl report every workspace healthy.
3. **`purgeTargets()` computed twice (nit)** — fixed; `settle` computes it once and passes it into `applyPurge`.

2. **`looksProd` `layerv.ai` substring (question)** — left as-is intentionally. It fails safe (over-restrictive: worst case a sandbox run is asked for `-allow-prod-purge`), and the sandbox tests use `layerv.xyz` / `sandbox.qurl.example`, so the prod-origin check keys off the canonical prod domain. If sandbox actually lives under `*.layerv.ai`, that's the one case to revisit — flagging for a maintainer to confirm the real sandbox endpoint, but not changing it blind since loosening it would weaken the prod rail.
4. **Sequential per-team full sweep (note)** — no change; the `Stats`/`Snapshot` shape is deliberately future-proofed for parallelism but the documented workflow is `-team`-scoped. Left as a heads-up per your note.

`go test -race` + golangci-lint v2.10.1 clean against current `main`.

---
_Generated by [Claude Code](https://claude.ai/code)_

Kevin-layerV:
Thanks — triaged; all non-blocking, and no code change this round (avoiding a cosmetic-only push that would cancel the in-flight CI and rebuild docker for wording):

- **`looksProd` endpoint double-check (#1)** — intentional, not redundant: the loop matches the `prod` substring, the final check matches `layerv.ai`. Two different signals.
- **`layerv.ai` breadth (#2)** — deliberately left broad. For a safety *rail*, over-triggering (an unexpected host asks for `-allow-prod-purge`) is the safe direction; anchoring to `api.layerv.ai` would risk *under*-triggering, which is the dangerous one. If a real non-prod host ever lives under `layerv.ai`, that's the case to revisit — flagging for a maintainer, same as the earlier review's question.
- **Sequential full-table sweep (#3)** — deferred by design; `Stats`/`Snapshot` already anticipate a parallel-segment model, and the documented workflow is `-team`-scoped.
- **"heavy" wording (#4)** — accurate to why the pointer exists (gocritic `hugeParam` flags the 96-byte struct ≥ its 80-byte threshold), leaving as-is.

Prior substantive items (schema-drift guard, `purgeTargets` recompute, the `auth.Provider` rebase fix) are in `5649f38`; CI is green locally against current `main` (`go test -race` + golangci-lint v2.10.1). PR is otherwise ready pending reviewer approval.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #676, opened 2026-06-09
feat(slack): enforce asker-only on conversation-mode get (resolves the get-authz gate)

justin-layerv:
## What

Resolves the **get-authorization** hard pre-enablement gate (one of the four on #651) by making conversation-mode `get` **asker-only**: only the member who requested an access link may Approve their own get card and receive it.

> Scope note: this resolves the **get-authz gate only**. The other three hard gates — **R2** (public-card `replace_original`), **C1** (connector key renders ephemerally), **C2** (connector trigger-window) — are live-workspace smoke tests that still gate the actual `QURL_AGENT_CONFIRM_ENABLED` flip. This is **not** "enablement unblocked."

## Why

A `get` mints a one-time access **credential**, delivered ephemerally to the clicker. Previously `get` was non-admin-gated, so **any channel member** could Approve a get card and receive the link themselves — the credential went to a random approver, not the requester. (No privilege escalation — it's channel-scoped, same as the clicker typing `/qurl get` — but the wrong person got the credential, and the asker might never receive their link.) The PR4a code comment already anticipated this: *"ephemeral delivery becomes fully correct once that enforces asker-only."*

## How

- `pendingAction` gains `Asker`; `postAgentConfirm` snapshots `env.Event.User` (the requesting member). Server-side snapshot, never read off the wire — consistent with the no-`AdminGated`-field principle.
- `askerOnly(kind)` predicate (alongside `adminGatedFor`/`confirmExecutable`/`confirmModalRouted`) keeps the authorization model one vocabulary.
- `processAgentConfirm` denies a get click when `pa.Asker == "" || payload.User.ID != pa.Asker` — **before the claim** and on **both Approve and Reject**:
  - **Before-claim is the load-bearing anti-DoS property:** gating *after* the claim would let any member consume the asker's pending get and then be denied, permanently burning the request — a DoS worse than no gate.
  - **Reject is gated too** so a non-asker can't dismiss the asker's card out from under them (an abandoned card is reaped by the 10-min TTL).
  - **`pa.Asker == ""` fails closed** (an asker-less get can never match a real clicker).
- With `clicker == asker` guaranteed, the existing ephemeral-to-clicker delivery reaches the right person and the mint actor IS the asker; the comments that flagged this gate are updated.

## Tests

`go build ./... && go test -race ./...` green; `golangci-lint v2.10.1` clean.
- asker Approve → claims + ephemeral link delivery, neutral public card;
- **non-asker Approve denied + NOT claimed → then the asker approves and succeeds** (the sequential proof that the gate sits before the claim — this is what fails if the gate ever moves below the claim);
- non-asker Reject denied + not claimed;
- the `Asker` snapshot round-trips.

Resolves the get-authz gate on #651.

### DIFF
--- a/apps/slack/internal/handler_agent_confirm.go
+++ b/apps/slack/internal/handler_agent_confirm.go
@@ -28,6 +28,7 @@ const (
 	agentConfirmExpiredReply        = "This request expired — ask me again."
 	agentConfirmAlreadyHandledReply = "That request was already handled."
 	agentConfirmAdminOnlyReply      = "That action is admin-only — ask a workspace admin to approve it."
+	agentConfirmGetNotAskerReply    = "Only the person who requested this access link can approve it. Ask me for your own link."
 	agentConfirmScopeMismatchReply  = "That request belongs to a different channel."
 	agentConfirmCanceledReply       = "Canceled — nothing was changed."
 	agentConfirmUnsupportedReply    = "I can't apply that kind of change yet."
@@ -67,6 +68,7 @@ type pendingAction struct {
 	Alias     string           `json:"alias,omitempty"`  // alias name for set/unset-alias; channel alias for protect-url
 	Target    string           `json:"target,omitempty"` // target slug for set-alias
 	URL       string           `json:"url,omitempty"`    // target URL for protect-url
+	Asker     string           `json:"asker,omitempty"`  // Slack user who requested the turn; get is asker-only (see processAgentConfirm)
 	ChannelID string           `json:"channel_id"`
 }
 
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks — round 1 is terminal (LGTM, no changes requested), and you traced the load-bearing property exactly: the gate-before-claim ordering is what the sequential test pins.

On the one non-blocking note (**in-flight get cards fail closed across the deploy**): agreed, that's the correct fail-closed behavior — any get card posted pre-deploy has no `Asker` in its stored blob, so the clicker hits `pa.Asker == ""` and is denied, bounded by the 10-min TTL. You're right the copy ("Ask me for your own link") reads slightly odd for an orphaned-own-card, but special-casing a one-deploy transient isn't worth it; I'll expect a brief spike of `non-asker click` warns right after rollout and not chase it. (And the whole surface is dark behind `AgentConfirmEnabled` anyway, so there are no live get cards to orphan until the flag flips.)

Note: `slack / docker` failed on a transient Docker Hub registry timeout during *buildx setup* (`registry-1.docker.io … context deadline exceeded`), before the image build ran — re-running it; no code involved.

PR #677, opened 2026-06-10
feat(slack): per-workspace conversation-mode toggle

justin-layerv:
## What

Adds `/qurl-admin agent on|off` so a workspace admin opts the workspace **in or out** of the qURL Secure Access Agent's conversation mode, layered on top of the existing org-level gate. Bare `/qurl-admin agent` shows the current state.

This is **PR6a** of the staged conversation-mode rollout — the per-workspace toggle mechanism. It is **dark by default**: with `QURL_AGENT_DEFAULT_ENABLED` unset (false) and no workspace opt-in, the surface behaves exactly as before this PR.

## Design

**Three-state toggle** in `workspace_mappings` (`agent_enabled`: absent / true / false):
- **absent** → falls back to the org default `Config.AgentDefaultEnabled` (env `QURL_AGENT_DEFAULT_ENABLED`; false during the staged opt-in rollout, flipped true at GA).
- **explicit false** → an opt-out that **survives the GA default flip** (the load-bearing reason for tracking absent-vs-explicit, not just a bool).
- **explicit true** → on even while the deployment default is off.

**Shared gate at both chokepoints.** `workspaceAgentEnabled` is applied at:
- **propose** (`processAgentEvent`) — *before* the dedupe marker, so a disabled workspace consumes nothing (re-enabling later doesn't swallow the prior @mention as a retry).
- **confirm-click** (`processAgentConfirm`) — *before* the claim, so it gates both Approve and Reject; a card proposed before a workspace turns the toggle off (within the ~10-min pending-action TTL) won't still execute on Approve.

It **fails closed** on a read error (don't run if we can't confirm the opt-in; never override an explicit opt-out on a transient blip). With no `AdminStore` wired, the org default governs.

## Changes

- **slackdata**: `SetAgentEnabled` / `AgentEnabledFor` (+ `boolAttr` / `readBoolPresent` helpers). `SetAgentEnabled` refuses an unbound workspace (404 → "run `/qurl setup` first").
- **handler**: `Config.AgentDefaultEnabled`, `workspaceAgentEnabled`, both gates.
- **command**: `handleAgentToggle` + `adminVerbAgent` (dispatch + help section) + `AdminActionAgentToggle` gate-audit label. Single-DDB sync admin verb (uses `adminSyncVerbBudget`, like add/remove/admins).
- **cmd**: `readAgentDefaultEnabled` (`QURL_AGENT_DEFAULT_ENABLED`, fail-safe false), wired into `Config`.

## Tests

- Store three-state incl. the **opt-out-survives-GA** case; 404 + transport-error mapping.
- `workspaceAgentEnabled` decision matrix + fail-closed-on-read-error.
- Gate wired at **both** paths: event-path gate-before-dedupe (replay proves it), confirm-path gate-before-claim.
- Command: on/off persist, status (default vs explicit), bogus-arg usage, non-admin denial (no write), no-AdminStore "not configured", help-section listing, and the (gate-unreachable, defense-in-depth) handler error branches.
- cmd env-matrix for `QURL_AGENT_DEFAULT_ENABLED` (mirrors `QURL_AGENT_CONFIRM_ENABLED`).

Verified: `go build/vet`, `go test -race ./...`, `golangci-lint v2.10.1` clean, `/simplify` clean.

## Infra follow-up

The new `agent` dispatcher verb is a niche admin verb (not advertised in Slack autocomplete). It needs adding to `hidden_verbs` in qurl-integrations-infra `envs.json` so `check-dispatcher-drift.sh` doesn't flag class-1 (routed-but-unadvertised) drift after this merges — same bucket as `protect-connector` / `protect-url` / `set-display-name`. This is a post-merge `::warning::`, not a hard failure, and not a required check on this repo. Tracking issue: _(linked in a comment below)_.

## Not in this PR

GA enablement (flipping `QURL_AGENT_DEFAULT_ENABLED` on in the deploy config) is a future enablement step. PR6b (per-team/per-user agent-turn rate limits) follows.

### DIFF
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -174,6 +174,32 @@ func TestReadAgentConfirmEnabled(t *testing.T) {
 	}
 }
 
+func TestReadAgentDefaultEnabled(t *testing.T) {
+	cases := []struct {
+		name string
+		val  string
+		want bool
+	}{
+		{name: "unset is off (staged rollout)", val: "", want: false},
+		{name: "true enables the GA default", val: "true", want: true},
+		{name: "1 enables", val: "1", want: true},
+		{name: "false stays off", val: "false", want: false},
+		{name: "0 stays off", val: "0", want: false},
+		// Fail-safe: an unparseable value must stay OFF, never silently turn the
+		// surface on for every workspace at once.
+		{name: "garbage fails safe to off", val: "enable", want: false},
+		{name: "yes fails safe to off", val: "yes", want: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("QURL_AGENT_DEFAULT_ENABLED", tc.val)
+			if got := readAgentDefaultEnabled(); got != tc.want {
+				t.Fatalf("readAgentDefaultEnabled(%q) = %v, want %v", tc.val, got, tc.want)
+			}
+		})
+	}
+}
+
 func TestLogAgentSurfaceState_ConfirmMode(t *testing.T) {
 	// The confirm/mutation line must key on the EFFECTIVE predicate
 	// (Handler.agentConfirmEnabled), never the raw flag: a flag set while the surface
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -156,6 +156,10 @@ func run() error {
 	postMessageBlocks := newSlackPostMessageBlocksFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
+	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
+	// true at GA (every workspace on unless it explicitly opted out). Fail-safe to
+	// false. The per-workspace flag itself lives in workspace_mappings (AdminStore).
+	agentDefaultEnabled := readAgentDefaultEnabled()
 	// Skip building the LLM + state store under a kill switch: it's read once at
 	// boot and forces the surface dark regardless (agentEnabled), so the store/LLM
 	// would never be used, and un-killing requires a restart anyway. This also
@@ -212,6 +216,7 @@ func run() error {
 		AgentDisabled:       agentDisabled,
 		PostMessageBlocks:   postMessageBlocks,
 		AgentConfirmEnabled: agentConfirmEnabled,
+		AgentDefaultEnabled: agentDefaultEnabled,
 	})
 
 	// Alias reads and writes must go through the same slackdata facade so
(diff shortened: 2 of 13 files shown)


justin-layerv:
Infra follow-up filed: https://github.com/layervai/qurl-integrations-infra/issues/987 — adds `agent` to `hidden_verbs` in the slack-manifest `envs.json` so `check-dispatcher-drift.sh` doesn't flag class-1 drift after this merges (same bucket as `protect-connector`/`protect-url`/`set-display-name`). Post-merge `::warning::`, not a hard failure, and not a required check on this repo.

justin-layerv:
Thanks — all three addressed in 944dd8b4:

1. **Success audit log (medium).** Added `slog.Info("agent toggle succeeded", "team_id", teamID, "user_id", userID, "enabled", enable)` next to the success replies, matching the `admin add/remove/admins` success-log pattern. (No log-assertion test — peer success logs aren't tested either, and the package-level `slog.Info` can't be captured without swapping the global default logger; the line is the deliverable.)

2. **Uniform authz (minor).** Moved the `teamID` validation + `requireAdminStoreSync` + `requireAdminSync` gate **ahead** of arg parsing. A non-admin now takes the same audited admin-only path for `agent on`, `agent`, or `agent wat` alike. `TestHandleAgentToggle_BogusArgRendersUsage` already seeds an admin, so it still asserts the usage copy; `TestHandleAgentToggle_NonAdminDenied` still asserts the denial + no write.

3. **Status read-error logging (minor).** Added `slog.Warn("agent toggle: status read failed", ...)` in `agentToggleStatus`'s error branch, closing the gap vs the two gate read-error paths that already log.

`go test -race ./...`, `golangci-lint v2.10.1`, and `/simplify` all clean. Re-running cr.

justin-layerv:
Thanks for the second pass — appreciate the verification against the integration points. On the four round-2 observations:

1. **Hot-path GetItem caching.** Agreed it's a genuine post-GA consideration (no cost today behind the dark `agentEnabled()` short-circuit). Filed as a labeled fast-follow: #678. I added a stale-direction safety note to the acceptance criteria — a stale *enabled* entry is the riskier direction for a security gate, so any cache must let an opt-out take effect promptly (write-through invalidation on `SetAgentEnabled`, or a short-enough TTL). Deferring rather than speculatively caching a dark surface.

2. **Event-path fail-closed → silence to an enabled workspace.** Intended posture — identical to the existing `MarkEventSeen` failure path, and erring toward silence over running an unconfirmed gate is the right call for a security toggle. Keeping as-is.

3. **Confirm-path "expired" copy on a transient read error.** Defensible as noted: re-asking is the correct user action on a transient blip, and the card TTLs out anyway. A dedicated "try again" copy + branch would add surface area for a rare case where the outcome (re-click/re-ask) is the same. Keeping the generic copy.

4. **Status/set double-read of the workspace row.** Negligible on the cold admin path (one extra point read per admin command, not per turn) — not restructuring, agreed.

Both already-tracked follow-ups remain open: infra `hidden_verbs` sync (qurl-integrations-infra#987) and the schema-comment mirror (no migration; DDB is schemaless for non-key attrs).

PR #679, opened 2026-06-10
feat(slack): admin-set default link expiry per resource in the /qurl list Edit modal

Kevin-layerV:
## What

Adds a **Default link expiry** dropdown to the `/qurl list` → Edit modal so qURL bot admins (and the owner) can set, per resource, how long links minted with `/qurl get` stay redeemable. The reply's "link expires in …" suffix renders the effective value. Links remain one-time use.

- Built-in default is unchanged: **1 minute**, used whenever no override is stored.
- Options: 1 minute (default) / 30 minutes / 1 hour / 6 hours / 24 hours / 7 days — the non-default values mirror the Discord bot's expiry choices (`apps/discord` `EXPIRY_LABELS`) for cross-integration consistency.
- Applies to all three mint surfaces (slash command, Create-qURL button, conversation-mode confirm) — they share `getWork`.

## Why store it bot-side

qurl-service has no per-resource default-TTL field — only the per-mint `expires_in` on qURL creation (the resource's `session_duration_cap` is a different knob: it caps post-redemption session length, not link expiry). The Slack bot is the only consumer of this setting, so it lives in the bot's own DynamoDB rather than adding an upstream API surface: a `resource_default_ttls` `Map<resource_id, duration>` attribute on the existing `workspace_mappings` row, following the `alias_bindings` map pattern. No table-schema or IAM change (same table, existing `GetItem`/`UpdateItem` grants).

## Safety posture

- **Admin-gated**: the dropdown rides the existing Edit modal, whose submission re-checks `CheckAdmin` before any write.
- **Changed-only writes**: the dropdown only writes when it moved off the value the modal opened with (same diff posture as Display Name), so a degraded pre-fill (transient read failure) can never silently clear a real override.
- **Fail-safe reads**: a store read failure — or a stored value outside the option set — falls back to the built-in 1 minute default at mint time and never reaches the wire.
- **No phantom rows**: writes are conditioned on the workspace row existing (`workspace_not_bound` otherwise), so the lazy map seed can't materialize a row that would break `BindWorkspace`'s first-claim condition. Clears are idempotent.
- Selecting the default option **clears** the stored entry, keeping absence-of-entry the single representation of "default".
- qurl-service still enforces the plan's max expiry at mint time (free tier caps at 3 days, so a 7-day override on a free workspace surfaces the server's error at mint).

## Tests

- Store contract: set/get/overwrite/clear round-trip, per-resource isolation, clear idempotence, unbound-workspace refusal without phantom rows.
- Modal: dropdown options + pre-fill (stored, none, unrecognized), `private_metadata` baseline carry.
- Submission: set, reset, untouched-no-write (asserted via DDB update hook), unlisted-value field error, absent-block (in-flight pre-deploy view) reads as unchanged.
- Mint: stored override on the wire + reply suffix, built-in default without override, junk stored value falls back to default.

### DIFF
--- a/apps/slack/internal/handler_edit.go
+++ b/apps/slack/internal/handler_edit.go
@@ -107,6 +107,7 @@ func (h *Handler) handleListEditClick(w http.ResponseWriter, payload *interactio
 		// process shutdown cancels them coherently.
 		enumCtx, enumCancel := context.WithTimeout(h.baseCtx, adminGateBudget)
 		meta.ExposedChannels = h.exposedChannelsForEdit(enumCtx, log, &meta)
+		meta.DefaultTTL = h.defaultTTLForEdit(enumCtx, log, meta.TeamID, meta.ResourceID)
 		enumCancel()
 
 		view, err := TunnelEditModal(&meta, snapshot.DisplayName, snapshot.Aliases)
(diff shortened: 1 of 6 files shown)


Kevin-layerV:
Addressing the review threads:

- **Extra `GetItem` per mint — accepted deliberately.** It's projected to the single map entry and eventually consistent, and the mint path is already several round-trips (resolve → rate-limit → key lookup → mint) behind a per-user rate limit, so one more cheap read doesn't move the needle. A `(team, resource)` cache would add invalidation complexity for no observed need.
- **Over-plan override at mint time** — improved in 53002dfd: a 400 mint rejection while an override is active now names the override and ends with "Please contact your Slack admin for assistance" instead of the generic retry copy (which would loop the user on a failure only an admin can fix). Non-overridden 400s keep the existing copy; `mapMintError` still logs the upstream code/detail first. The bot can't pre-validate at set time because it can't see the workspace plan.
- **Nil-guard contract on the mint-path read** — already covered by the 6bcc3f9d refactor: both call sites go through `storedLinkExpiry`, which nil-checks `AdminStore` explicitly.
- The `SetResourceDefaultTTL.clear` op-name suffix is intentional (distinguishes the two UpdateItem shapes in logs), and the double option-slice build per modal render stays as-is — 6 tiny maps per open.

Kevin-layerV:
On the latest review's two open threads:

- **"Worth double-checking" invariant — confirmed, it holds.** An override is never stored as the default value: `parseEditLinkExpiry` maps the default option to `""` (clear), and `linkExpiryOptions`' doc pins absence-of-entry as the single representation of "default". Even a hand-edited row storing `"1m"` is benign — it's a recognized option equal to the built-in default, so the mint behaves identically and the over-plan branch correctly doesn't fire (1 minute can't exceed any plan cap).
- **Folding the TTL read into the API-key read isn't available** — the key lookup goes through shared/auth's DDBProvider against the workspace-state table, not `workspace_mappings`, so there's no same-row read to piggyback on. The extra projected, eventually-consistent `GetItem` per mint stands as the accepted cost (per the earlier thread).

Items 2–4 (two UpdateItems on a rare admin write, documented pre-fill staleness window, the deliberate mapMintError-first ordering) are as-designed per the review's own read.

PR #680, opened 2026-06-10
feat(slack): per-user and per-team agent-turn rate limits

justin-layerv:
## What

Caps how many conversation-mode **turns** a single member, and a whole workspace, can drive per rolling hour — a cost backstop on LLM spend. This is **PR6b** of the staged rollout (PR6a was the per-workspace toggle).

Enforced in `processAgentEvent` **after dedupe** (count unique messages, not redeliveries) and **before the turn runs** (the LLM is what we're capping). Confirm-clicks (`processAgentConfirm`) are deliberately **not** limited: they're consume-once, admin-gated, and carry no LLM cost.

## Design

- **Fails OPEN.** This is a cost backstop, not a security gate, so a transient counter error logs and *allows* the turn rather than dropping a legitimate member — the opposite of the fail-closed workspace/dedupe gates directly above it.
- **Per-user bumped first**, then per-team, so one member spamming can't inflate the shared per-team counter for everyone else.
- **`BumpTurnCount` = atomic `UpdateItem ADD`** on a fixed-window counter, not read-modify-write. The per-team counter is a single hot item shared by every member, so only a strict atomic increment holds the cap under concurrent turns — exactly when a cost backstop matters. (A leaky RMW would fail precisely when a workspace is busy, the case it exists to protect.)
- **`pk = teamID`** (a workspace), *not* the enterprise-else-team event partition used for conversation/dedupe — a per-workspace cost cap must not collapse into one shared bucket across an Enterprise Grid. The sk is window-keyed so each window is a fresh, TTL-reaped item (no reset write).

## Config

- `Config.AgentMaxTurnsPerUserPerHour` / `AgentMaxTurnsPerTeamPerHour`; env `QURL_AGENT_MAX_TURNS_PER_USER_PER_HOUR` / `..._TEAM_PER_HOUR`.
- Conservative **non-zero defaults (30 / 300)** so a GA-live agent has a backstop even if the operator never sets the env; an explicit **`0` disables** the cap (honored as a sentinel, distinct from absent). Malformed/negative fails safe to the default.

## Known tradeoffs (accepted)

- **Per-team counter is a single hot DDB item** written once per turn. Turns are human-paced (low write rate); intentionally *not* sharded — that would be over-engineering for this volume.
- **A team-denied turn still consumed a per-user increment** (user is bumped first). Documented on `agentTurnLimited`; both counters only inflate within the window and reset when it rolls — a non-issue for a backstop.
- **Fixed-window burst:** up to ~2× the cap across a window boundary. Standard fixed-window property; acceptable for a backstop (sliding window would be more machinery than the goal warrants).
- **The rate-limited *reply* is intentionally not rate-limited.** It's a `chat.postMessage` (no LLM cost — the thing being capped), posted on every over-cap turn so a capped member isn't met with silence. Whether to throttle it to once/window for Slack-API/channel-noise reasons is a deferred maintainer decision: #681.
- **Per-node clock skew at window boundaries.** The window sk is keyed on each node's `now().Truncate(window)`, so skewed clocks near an hour boundary can split a workspace's count across two items (each under-counting) — inherent to per-node fixed windows, compounding the ~2× burst above. Acceptable for a backstop; counter inflation within a window is likewise harmless (only ever compared `> limit`, then TTL-reaped).

## Tests

- Store boundary (`stubDDB`): atomic-ADD request shape (key, `ADD turn_count :one SET ttl = :ttl`, `UPDATED_NEW`), response parse, validation guards (400), transport error (503).
- Gate (real `AgentStore` over the in-memory fake): per-user cap, per-team cap across distinct users, **user-first containment** (an over-cap member doesn't bump the team counter), **fail-open** on a counter error, disabled (0/0) runs every turn and writes no counter items.
- cmd env-matrix (unset→default, explicit, `0`=unlimited, malformed/negative→default).

The handler fake `memAgentDDB.UpdateItem` (previously a no-op stub) now actually applies the ADD — a no-op would make these tests vacuously pass.

`go test -race ./...`, `golangci-lint v2.10.1`, and `/simplify` all clean.

## Live-smoke item (fakes can't verify)

Two items to confirm on a live workspace alongside the other conversation-mode smoke checks:
- Real DynamoDB `ADD` returns the **post-increment** value **atomically** — the tests assert the request shape + parse a canned `UPDATED_NEW`, but the atomic return-value semantics are a (correct-per-DDB-docs) assumption a fake can't verify.
- On an Enterprise Grid install, events carry a populated `team_id` so the per-workspace cap engages — an empty `team_id` makes `BumpTurnCount` 400 → fail-open → the cap silently never applies (consistent with how `workspaceAgentEnabled` keys on `team_id`).

## Not in this PR / no infra change

No infra change needed — the counter attrs are schemaless in the existing `agent_state` table and reuse its DynamoDB TTL. GA enablement (flipping the agent on) and any cap tuning are operator/deploy-config steps.

### DIFF
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -200,6 +200,37 @@ func TestReadAgentDefaultEnabled(t *testing.T) {
 	}
 }
 
+func TestReadAgentMaxTurns(t *testing.T) {
+	cases := []struct {
+		name string
+		env  string
+		val  string
+		read func() int
+		want int
+	}{
+		{"per-user unset → default backstop", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "", readAgentMaxTurnsPerUser, defaultAgentMaxTurnsPerUser},
+		{"per-user explicit", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "5", readAgentMaxTurnsPerUser, 5},
+		// An explicit 0 is honored as "unlimited" — distinct from absent (which gets
+		// the conservative default).
+		{"per-user 0 = unlimited", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "0", readAgentMaxTurnsPerUser, 0},
+		// Fail-safe: a typo'd or negative cap falls back to the backstop, never to
+		// "unlimited" by accident.
+		{"per-user malformed → default", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "lots", readAgentMaxTurnsPerUser, defaultAgentMaxTurnsPerUser},
+		{"per-user negative → default", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "-3", readAgentMaxTurnsPerUser, defaultAgentMaxTurnsPerUser},
+		{"per-team unset → default backstop", "QURL_AGENT_MAX_TURNS_PER_TEAM_HOUR", "", readAgentMaxTurnsPerTeam, defaultAgentMaxTurnsPerTeam},
+		{"per-team explicit", "QURL_AGENT_MAX_TURNS_PER_TEAM_HOUR", "150", readAgentMaxTurnsPerTeam, 150},
+		{"per-team 0 = unlimited", "QURL_AGENT_MAX_TURNS_PER_TEAM_HOUR", "0", readAgentMaxTurnsPerTeam, 0},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv(tc.env, tc.val)
+			if got := tc.read(); got != tc.want {
+				t.Fatalf("%s: %s=%q → %d, want %d", tc.name, tc.env, tc.val, got, tc.want)
+			}
+		})
+	}
+}
+
 func TestLogAgentSurfaceState_ConfirmMode(t *testing.T) {
 	// The confirm/mutation line must key on the EFFECTIVE predicate
 	// (Handler.agentConfirmEnabled), never the raw flag: a flag set while the surface
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -208,18 +208,18 @@ func TestReadAgentMaxTurns(t *testing.T) {
 		read func() int
 		want int
 	}{
-		{"per-user unset → default backstop", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "", readAgentMaxTurnsPerUser, defaultAgentMaxTurnsPerUser},
-		{"per-user explicit", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "5", readAgentMaxTurnsPerUser, 5},
+		{"per-user unset → default backstop", "QURL_AGENT_MAX_TURNS_PER_USER_PER_HOUR", "", readAgentMaxTurnsPerUser, defaultAgentMaxTurnsPerUser},
+		{"per-user explicit", "QURL_AGENT_MAX_TURNS_PER_USER_PER_HOUR", "5", readAgentMaxTurnsPerUser, 5},
 		// An explicit 0 is honored as "unlimited" — distinct from absent (which gets
 		// the conservative default).
-		{"per-user 0 = unlimited", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "0", readAgentMaxTurnsPerUser, 0},
+		{"per-user 0 = unlimited", "QURL_AGENT_MAX_TURNS_PER_USER_PER_HOUR", "0", readAgentMaxTurnsPerUser, 0},
 		// Fail-safe: a typo'd or negative cap falls back to the backstop, never to
 		// "unlimited" by accident.
-		{"per-user malformed → default", "QURL_AGENT_MAX_TURNS_PER_USER_HOUR", "lots", readAgentMaxTurnsPerUser, defaultAgentMaxTurnsPerUser},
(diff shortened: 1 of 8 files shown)


justin-layerv:
Thanks for the thorough pass. On the four observations + nits:

1. **Empty `env.TeamID` → cap silently never applies.** Agreed it's consistent with `workspaceAgentEnabled` (also keys on `env.TeamID`), and Slack populates `team_id` on every event including Grid. Added a line to the live-smoke checklist (PR body) to confirm `team_id` is present on a Grid install so the per-workspace cap engages. No code change — failing open on a (not-expected) empty team_id is the correct backstop posture.

2. **Empty `env.Event.User` → `user#` bucket collapse.** This is unreachable: `shouldDispatchAgentEvent` (the only gate before `processAgentEvent` — sole production caller is `handleAgentEvent`) rejects `e.User == ""`. Documented the invariant inline at the gate (ab75a313) rather than adding a guard — the function already relies on that same filter for `BotID`/`Subtype`, so a redundant runtime check would be dead code inconsistent with the surrounding style.

3. **Two sequential DDB writes before the LLM.** Confirmed — correctly ordered for the user-first containment property (not parallelizable without losing it), human-paced, and already noted under "Known tradeoffs" in the PR body. No change.

4. **Cap semantics (N allowed, N+1 blocked).** Confirmed intentional and matches the 30/300 "turns per hour" operator framing; `TestAgentTurnLimit_PerUser` locks it.

Nits (uniform reply privacy posture, `readIntEnvFailSafe` 0-sentinel) — noted, thanks. Re-running cr on the doc-comment commit.

justin-layerv:
Thanks — addressed both documentation nudges in 426a6c10:

1. **Attempt-counting (transient-failed turns still count).** Added to the gate comment: the count is of turn *attempts*, not answered turns, so the cap is "N attempts/hour" — the right unit for a cost backstop since the LLM round-trip is the spend whether or not it produced a usable answer.
2. **Fail-open weakest under DDB stress.** Documented on `overTurnLimit` as the conscious tradeoff: DDB throttling is also when a busy workspace racks up cost, but dropping a legitimate member's turn on a transient blip is worse for a backstop than briefly running uncapped.

3. **Two sequential DDB writes per turn** — confirmed; post-ack, human-paced, and already under "Known tradeoffs". No change.
4. **`scope` log embeds the Slack user id (`user#U…`)** — intended and consistent with the admin-action log sites (which log `user_id`); Slack user ids are opaque workspace-scoped handles, not PII. Keeping it (the scope is the meaningful unit for a rate-limit log line).

Both correctness confirmations (off-by-one, concurrency/containment, idempotent ttl SET) match the intent — appreciate the trace. Re-running cr on the doc commit.

justin-layerv:
Thanks for the careful third pass. On the four observations:

1. **Uncapped rate-limited reply.** Deferred to a maintainer decision — filed #681. The reply carries no LLM cost (the thing being capped), and silently dropping a capped member's repeat mention reproduces the silent-drop UX PR6a's cr flagged; the advisor explicitly weighed and rejected a suppress-repeats marker as "negligible gain", so #680 ships always-reply. The issue records both positions + the cheap `count == limit+1` implementation if a maintainer wants to throttle. Not unilaterally flipping an advisor-endorsed design on an optional nit.
2. **Team-cap reply wording (fixed, dd3d4625).** Rephrased to the neutral *"Conversation mode is at its limit for now …"* — accurate for both the per-user and per-team caps (doesn't blame an innocent member on a busy workspace), still uniform so it doesn't leak which cap fired.
3. **Two `now()` calls → adjacent-window edge.** Accepted as cosmetic: it requires a turn to cross the exact one-second hour boundary *between* two sequential millisecond-apart DDB writes, and it self-corrects next turn. Threading a shared timestamp through `BumpTurnCount` would add permanent API surface for a never-happens edge — not worth it.
4. **Fake-rejection blast radius.** Agreed, and thanks for verifying `BumpTurnCount` is the only `UpdateItem` path on `AgentStore` (aliases/policies/workspace ride different stores/fakes).

Off-by-one, concurrency/containment, and idempotent-ttl confirmations all match intent. Re-running cr on the wording commit.

justin-layerv:
Thanks — on the three round-4 points:

1. **Env/field naming asymmetry (fixed, 7eda07d5).** Renamed the env vars to `QURL_AGENT_MAX_TURNS_PER_USER_PER_HOUR` / `QURL_AGENT_MAX_TURNS_PER_TEAM_PER_HOUR` so both surfaces read "per X per hour" and match the `AgentMaxTurnsPer{User,Team}PerHour` fields. They're new + undeployed, so the rename is free (no migration). PR body updated to match.
2. **Two writes per limited turn** — confirmed; documented under "Known tradeoffs" and on the gate. No change.
3. **Env vars not in a deploy doc** — agreed it's consistent with the repo (the existing `QURL_AGENT_*` vars aren't documented in-repo either, so not a regression). Filed #682 to surface all five `QURL_AGENT_*` vars + defaults (incl. the 30/300 caps) in the deploy runbook before GA, with a note to review the defaults for the expected GA workload.

Appreciate the correctness re-trace (off-by-one, per-scope fail-open composition, TTL math). Re-running cr on the rename.

justin-layerv:
Thanks for the fifth pass and the independent re-verification. All three are documentation acknowledgements — added to the PR's "Known tradeoffs":

1. **Limited-reply post is intentionally uncapped.** Noted explicitly: the reply is a `chat.postMessage` (no LLM cost — the capped resource), posted on every over-cap turn so a capped member isn't met with silence. Whether to throttle it for Slack-API/channel-noise reasons is the deferred maintainer decision in #681.
2. **Per-node clock skew at boundaries.** Added to the tradeoffs alongside the ~2× burst — inherent to per-node fixed windows, acceptable for a backstop.
3. **Unbounded counter inflation within a window.** Confirmed intended — only ever compared `> limit`, then TTL-reaped; folded into the same note.

Both live-smoke items (atomic DDB `ADD` return; Grid `team_id` populated so the cap engages rather than 400→fail-open) remain flagged in the PR body for the live workspace check. No code change this round.

Issue #661, opened 2026-06-09
Slack agent: immediate ack/status indicator on the Events path


justin-layerv:
From the Slack agent-design audit (in-session task #4). **Highest-value UX win.**

Today the conversation surface posts nothing until the final reply (and `saveAgentHistory` runs before `postAgentReply`, adding latency) — the user waits in silence. Best practice: show an immediate lightweight acknowledgment right after input, and progressive status for multi-step work.

**Preferred approach:** `reactions.add` a glanceable emoji (e.g. 👀 / ⏳) to the user's message when the turn starts, clearing/replacing it when the reply posts — the doc calls out emoji reactions as the low-noise ack. Alternative: post a "_thinking…_" placeholder and `chat.update` it to the final reply (needs `PostMessage` to return the message `ts` + an update port).

**Why not done on the dark PR (#650):** needs a new Slack-client port (reactions or message-update) and the `reactions:write` scope, both of which land with the `cmd/main.go` enablement wiring — see #651. Implement alongside enablement so the scope + client method travel together. Keep it behind the existing kill switch.

PR #683, opened 2026-06-10
feat(slack): agent "working on it" reaction ack on the Events path

justin-layerv:
## What

The conversation agent posts nothing until the final reply, so a user who @mentions or DMs it waits in silence. This adds a glanceable 👀 reaction on the **triggering message** the moment the agent commits to a turn, removed when the turn ends — the Slack agent-design "low-noise ack." Resolves **#661** (the audit's *highest-value UX win*).

## Design

- **Seam:** a 2-method `ReactionPort` (`Add`/`Remove`) on `Config`, **nil-able** so the ack is simply absent until wired. Behind the kill switch via `agentEnabled()`, like the rest of the surface.
- **Lifecycle** (`handler_agent.go`): `addAgentAck` reacts 👀 **synchronously** once the turn is committed — *after* the disabled/dedupe/rate-limit gates, so only turns that actually run get acked — and a **deferred** `clearAgentAck` removes it on **every** exit (reply, error, panic).
  - The add **stays synchronous** so the deferred clear can't race an in-flight `reactions.add` and strand the 👀 (a goroutine'd add is the trap — commented in code).
  - The clear runs on a **fresh ctx off `baseCtx`** because the turn ctx is spent by defer time — the same spent-ctx lesson as `postAgentReply`.
- **Best-effort:** a reaction failure logs and is ignored; it never fails the turn, and `nil` Reactions is a zero-cost no-op.
- **cmd seam:** generalize the chat.postMessage poster into `slackWebAPIPoster` (op + injected response parser), reused by `reactions.add`/`reactions.remove` with the **same per-workspace token lookup + Enterprise Grid fallback**. The reactions parser treats the idempotent no-ops (`already_reacted` / `no_reaction`) as success. **chat.postMessage is byte-identical** (op=`chat.postMessage`, no benign codes — its existing tests guard it).

## Accepted failure modes (best-effort ack)

- **Process death between add and the deferred clear** (OOM/SIGKILL/redeploy) leaves a stuck 👀 — dedup already committed, so there's no retry. Cosmetic, acceptable for a best-effort ack.
- **A failed best-effort `reactions.remove`** leaves a stale 👀 next to the posted reply. Logged, not retried.

## Tests

- Handler (`handler_agent_ack_test.go`): added on success and **cleared on success / turn-error / panic**; best-effort (a failing reaction still posts the reply); nil-seam no-op; **not** added for a rate-limited turn (the ack sits after the gate).
- cmd seam (`slack_reactions_test.go`): request shape (URL routing, `{channel,timestamp,name}` body, bearer token), Grid fallback, benign-code tolerance, real-error surfacing.

`go test -race ./...`, `golangci-lint v2.10.1`, and `/simplify` all clean.

## Dark by default / infra

The seam is wired but **inert** until the agent surface is live, and `reactions.add`/`.remove` need the **`reactions:write`** scope in the Slack manifest to actually land. That scope (plus the rest of the sandbox agent enablement) lands in a separate **infra** PR — this PR is safe to merge ahead of it (a missing scope just means the best-effort ack no-ops).

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -154,6 +154,10 @@ func run() error {
 	// the slash-command modals.
 	postMessage := newSlackPostMessageFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
 	postMessageBlocks := newSlackPostMessageBlocksFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
+	// reactions.add/remove seam for the agent's best-effort "working on it" ack. Always
+	// wired (same token lookup as the post seams); inert until the agent surface is live
+	// and needs the reactions:write scope in the Slack manifest to actually land.
+	agentReactions := newSlackReactionPortWithTokenLookup(workspaceTokenLookup, userAgent, slackReactionsAddURL, slackReactionsRemoveURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
@@ -224,6 +228,7 @@ func run() error {
 		AgentDefaultEnabled:         agentDefaultEnabled,
 		AgentMaxTurnsPerUserPerHour: agentMaxTurnsPerUser,
 		AgentMaxTurnsPerTeamPerHour: agentMaxTurnsPerTeam,
+		Reactions:                   agentReactions,
 	})
 
 	// Alias reads and writes must go through the same slackdata facade so
--- a/apps/slack/cmd/slack_postmessage_test.go
+++ b/apps/slack/cmd/slack_postmessage_test.go
@@ -220,7 +220,7 @@ func TestSlackPostMessageFuncRejectsOversizedResponse(t *testing.T) {
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
 		// Write more than the bounded-read limit so the seam trips its cap.
 		_, _ = w.Write([]byte(`{"ok":true,"padding":"`))
-		_, _ = w.Write([]byte(strings.Repeat("a", slackChatPostMessageResponseBodyLimit+1024)))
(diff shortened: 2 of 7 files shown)


justin-layerv:
Thanks — addressed both in 9125a692:

**Synchronous-add latency (design point).** Keeping the add synchronous, but addressing your actual concern — the *magnitude* — directly: tightened `agentAckTimeout` 4s → **2s**, decoupled from the chat.postMessage budget. A cosmetic "working on it" ack that hasn't landed in ~2s is already too late to feel responsive, so giving up (no 👀; the deferred remove then hits `no_reaction` → benign, turn proceeds) is the right best-effort behavior. That's a zero-concurrency, synchronous shrink of the worst case you flagged.

The async variant you sketched (goroutine add + deferred clear *joins* the add) is safe and correct — but a second per-turn goroutine off `baseCtx` lands in the SIGTERM-drain / `WaitTimeout` / baseCtx-not-Background discipline (the #650 spent-ctx lesson) and is hard to test deterministically, which is a lot of reasoning surface for a best-effort cosmetic ack. The benefit is also narrow: a Slack-degraded `reactions.add` usually coincides with a degraded `chat.postMessage` reply, so the turn is slow regardless. Filed as a tracked follow-up: **#684**.

**Nit (http.Client comment).** Reworded — the shared client is safe for the concurrent turns the async pool may run, not "one call at a time".

Both correctness confirmations (defer-ordering-on-panic, spent-ctx, failed-add → benign-clear, byte-identical chat.postMessage) match the intent — appreciate the trace. Re-running cr on the changes.

justin-layerv:
Thanks for the second pass. Both watch-items are accounted for:

1. **Synchronous add on the critical path** — agreed this is the one spot the feature can add latency. Round-1 tightened the budget to 2s; the real fix (async add + clear-joins-add, taking it off the critical path entirely) is tracked in **#684** so it won't get lost. Good catch on the Grid-fallback sharing the single 2s budget across two `postOnce` calls — that's in scope for #684's design (the budget should cover the whole add, fallback included).
2. **Shared http.Client asymmetry** (reaction posters share one; the post seams each build their own) — documented at both sites, harmless at this volume; leaving as-is, just noting it's deliberate.

No code changes this round. CI: `slack/test` green, `slack/docker` (arm64) still building.

PR #686, opened 2026-06-10
fix(slack): reload-merge-retry agent transcript on save conflict (#666)

### LINKED ISSUE #666: Slack agent: reload-and-retry once on conversation-save version conflict
Follow-up from the #650 cr (commit 604063ad round).

Two DIFFERENT messages in the same thread arriving close together both pass dedupe (distinct ts), both `loadAgentHistory` at the same `version`, both run, and one `SaveConversation` hits `ErrConversationConflict` — logged and dropped. **Both still post replies**, but the losing turn's exchange silently vanishes from the persisted transcript, so the next turn's context is missing a reply the user already saw ("why did it forget what I just said").

Acceptable for v1 and consistent with the `saveAgentHistory` doc comment, but worth closing: on `ErrConversationConflict`, reload the latest persisted history + version, re-append this turn's exchange, and retry the save once (bounded). Care needed since the two turns diverged from the same base — define the merge (append-loser-after-winner is the likely intent) and cap retries to avoid a livelock under sustained same-thread concurrency.

Low priority; the surface is dark and same-thread concurrent distinct messages are uncommon.

justin-layerv:
## What

Closes #666.

Two different messages in the same thread arriving close together both pass dedupe (distinct `ts`), both `loadAgentHistory` at the same version, both run, and one `SaveConversation` loses the optimistic-concurrency race (`ErrConversationConflict`). Previously the losing turn was **logged and dropped** — both still posted replies, but the loser's exchange silently vanished from the persisted transcript, so the next turn loaded a thread missing a reply the user already saw ("why did it forget what I just said").

## Change

On `ErrConversationConflict`, instead of dropping: reload the winner's transcript, graft just this turn's **delta** on top (append-loser-after-winner, the resolution #666 calls for), re-trim for well-formedness, and retry the save **exactly once**. Neither turn is lost.

The delta is `newHistory[len(loadedHistory):]` — `a.Run` returns the loaded history as an exact prefix (it copies history, then appends this turn's messages), so the suffix is precisely this turn's exchange. That invariant is now **pinned by a test** (`agent.TestRun_AppendsToPriorHistoryWithoutMutatingInput`); if `a.Run` ever stops being pure-append, the test fails loudly rather than letting the delta logic corrupt transcripts.

The grafted reply was computed against the pre-conflict context, so it may be slightly stale relative to the winner's turn — structurally valid, semantically best-effort, and strictly better than dropping.

### Guards
- **Reload drops, never grafts, on a hard load error or a corrupt blob** (`reloadForMerge`, deliberately diverging from `loadAgentHistory`'s overwrite-on-corrupt): the winner's turn is never clobbered, and the delta never lands on a garbage base.
- **The merge seam is rejected if the delta head is an orphaned `tool_results` message** — defense-in-depth against a future non-pure-append `a.Run`. A malformed merged transcript would poison *every* future turn on the thread, and a short merged transcript is never trimmed, so this can't rely on `trimAgentHistory` to repair it.
- **Exactly one retry** — an unbounded reload race would pin the delivery worker.

### Structure
Extracts `persistBoundedHistory` (trim + byte-guard + marshal + save, returns the conflict sentinel — called on both the first save and the retry) and `reloadForMerge`. The common **no-conflict path is unchanged**: one trim + one `PutItem`.

## On transcript well-formedness (the merge seam)

The merge joins the winner's turn-end to this turn's delta. When the winner's turn ended at a proposal or the iteration cap, it ends in a `user{ToolResults}` message, and the delta opens with the `user{Text}` message `a.Run` prepends — i.e. a **user→user adjacency**. This is *not* new to #666: the shipped proposal-followup flow already persists exactly this shape (`agent.go` ends those turns at `user{ToolResults}`; the next turn appends `user{Text}`). The merge produces the identical adjacency, so it is well-formed by the same construction that makes sequential turns well-formed. `assertTranscriptWellFormed` therefore checks tool_use/tool_result pairing + a clean head, and deliberately does **not** assert role alternation.

(While confirming this I noted that `toSDKMessages` doesn't coalesce consecutive same-role messages — a *pre-existing* property of the proposal/cap persistence, independent of this PR. Flagged separately for investigation; not in scope here.)

## Tests

- `TestSaveAgentHistory_ConflictMergesAndRetries` — the core proof: a concurrent winner forces a conflict; the turn reloads, merges (winner's proposal turn + this turn's tool-using turn), retries once, and the stored transcript is winner+delta in order and **well-formed** across the seam. Exactly 2 writes.
- `TestSaveAgentHistory_ReloadFailureDropsTurn` — hard load error on reload → drop, winner's blob intact, no merged retry.
- `TestSaveAgentHistory_CorruptReloadDropsTurn` — undecodable winner blob → drop, blob not clobbered.
- `TestSaveAgentHistory_MalformedDeltaHeadDropsTurn` — delta opening with `tool_results` → drop (the seam guard).
- `TestSaveAgentHistory_RetriesAtMostOnce` — double conflict → exactly 2 writes, never loops.

`memAgentDDB` is upgraded to faithfully enforce the version CAS (it previously ignored the `OR conv_version = :ev` condition), mirroring `agentFakeDDB.evalCond`, so the race is driven for real at the handler level; plus a forced-conflict counter for the double-conflict case (a passive CAS fake can't produce a second conflict deterministically).

## Verification

- `go build ./... && go vet ./... && go test -race ./...` — green (full suite; the faithful CAS doesn't perturb the async multi-turn tests)
- `/simplify`: clean (4 cleanup agents — the split, per-test setup, delta-param passing, and assertion structure all reviewed as justified; adopted the one defense-in-depth suggestion, the delta-head seam guard)

Dark surface — no behavior change when the agent is disabled.

### DIFF
--- a/apps/slack/internal/agent/agent_test.go
+++ b/apps/slack/internal/agent/agent_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"encoding/json"
 	"errors"
+	"reflect"
 	"strings"
 	"testing"
 )
@@ -373,6 +374,23 @@ func TestRun_AppendsToPriorHistoryWithoutMutatingInput(t *testing.T) {
 	}
 	if len(history) <= priorLen {
 		t.Fatalf("expected history to grow; got %d", len(history))
+	}
+
+	// EXACT-PREFIX invariant: handler_agent.go's saveAgentHistory computes this
+	// turn's delta as history[len(prior):] and, on a concurrent-save conflict,
+	// re-applies just that delta onto the winner's transcript. That is correct only
+	// if Run returns the input history UNCHANGED as a prefix. If Run ever rewrites
+	// or reorders prior history, this assertion must fail loudly — otherwise the
+	// delta logic would silently corrupt persisted transcripts.
+	for i := range prior {
+		if !reflect.DeepEqual(history[i], prior[i]) {
+			t.Fatalf("Run did not preserve prior history as an exact prefix at index %d: got %+v want %+v", i, history[i], prior[i])
+		}
+	}
+	// The delta begins with the user message Run prepends — the clean turn boundary
+	// the conflict-merge grafts onto the winner's transcript.
+	if d := history[priorLen]; d.Role != roleUser || d.Text != "follow-up" {
+		t.Fatalf("delta does not begin with the user message: %+v", d)
 	}
 }
 
--- a/apps/slack/internal/handler_agent.go
+++ b/apps/slack/internal/handler_agent.go
@@ -418,7 +418,17 @@ func (h *Handler) processAgentEvent(ctx context.Context, log *slog.Logger, env *
 	// Save before posting: the transcript must be durably consistent before the
 	// user can fire a follow-up turn against it. The post is the slower,
 	// user-visible step, so this trades a little reply latency for that ordering.
-	h.saveAgentHistory(log, partition, threadKey, newHistory, version)
+	//
+	// a.Run returns the loaded `history` as an exact prefix of newHistory (it
+	// copies history, then appends this turn's messages — pinned by
+	// agent.TestRun_AppendsToPriorHistoryWithoutMutatingInput). So this turn's
+	// delta is the suffix; saveAgentHistory re-applies just it onto the winner's
+	// transcript if a concurrent turn won the save race.
+	var delta []agent.Message
+	if len(newHistory) >= len(history) {
+		delta = newHistory[len(history):]
+	}
+	h.saveAgentHistory(log, partition, threadKey, newHistory, delta, version)
 
 	// Deliver: an interactive confirm card for an executable proposal once the
 	// confirm flow is enabled, else the text reply/preview (merged #650 behavior).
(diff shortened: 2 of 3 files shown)


justin-layerv:
Thanks for the deep review. Addressed the one substantive item in `0b16d808`:

**Residual-risk (runtime exact-prefix):** Fixed. Added `agentRunPreservedPrefix` — the call site now verifies `newHistory` begins with `history` **element-for-element**, not just `len(newHistory) >= len(history)`. On mismatch it leaves `delta` nil and logs an error: the first save still persists `newHistory` normally, only the conflict-merge is skipped (degrades to the pre-#666 drop, never a wrong-suffix graft). It's slightly stronger than the suggested last-message check — it catches a rewrite of *any* prefix element, not just the boundary — and it's cheap (≤ `maxPersistedMessages` comparisons, once per turn, post-LLM). `TestAgentRunPreservedPrefix` covers the rewritten-prefix-but-longer case the length check would have missed.

**Nits (no code change, confirming intent):**
- `len(delta) == 0` log line: agreed it's a near-dead path; it now also catches the prefix-violation drop, and the comment already points to the call-site guard. Left as-is.
- `reloadForMerge` empty-blob → `(nil, version, true)`: intentional. Unreachable in the conflict path (a conflict implies a present higher-version row) and harmless if it ever fired.
- `toSDKMessages` non-coalescing: agreed it's strictly pre-existing (the merge reproduces the proposal-followup adjacency that already ships). Filed as a separate investigation to confirm the pinned SDK accepts consecutive same-role messages — not in scope here.

PR #689, opened 2026-06-10
feat(slack): resolve channel names for the agent system prompt (#659)

### LINKED ISSUE #659: Slack agent: populate TurnContext.ChannelName for the system prompt
Follow-up from the #650 conversation-mode review (cr round-4, note #2).

`agent.TurnContext` models a `ChannelName` field (used by `describeChannel` to render `#name (Cxxxx)` in the system prompt), but `processAgentEvent` never populates it — the Slack Events payload carries only the channel **id**, not its name. `describeChannel` falls back to the bare channel id, so this is a graceful, deliberate gap, not a bug.

**To populate it** the backend would need a `conversations.info` lookup per turn (or a cached channel-name map), which also means adding `channels:read` / `groups:read` to the Slack manifest scopes. That cost is why it is deferred rather than done inline.

**Scope**
- Resolve the channel name (cached; invalidate on rename) and set `TurnContext.ChannelName`.
- Add the required manifest scope(s) and document them alongside the #651 enablement wiring.
- Keep the `describeChannel` id-fallback for DMs / lookup failures.

Low priority: the model already receives the channel id; the name is a readability nicety.

justin-layerv:
## What

Closes #659.

The agent's system prompt renders the channel via `describeChannel` as `#name (Cxxxx)`, but `processAgentEvent` never populated `TurnContext.ChannelName` (the Events payload carries only the id) — so it always fell back to the bare id. This resolves the name.

## Change

- **New seam** `Config.ResolveChannelName` (`ResolveChannelNameFunc`) → `newSlackResolveChannelNameFuncWithTokenLookup` in cmd: a `conversations.info` read on the per-workspace bot token with the Enterprise Grid org-token fallback (same token resolution as the PostMessage / Reactions seams). Nil leaves `ChannelName` empty — `describeChannel` falls back to the id — so the agent works without the scope.
- **Per-Handler TTL cache** (`channelNameCache`, 30 min): the per-turn lookup costs at most one `conversations.info` per channel per TTL. **Failures are negative-cached**, so a workspace lacking `channels:read`/`groups:read` falls back to the id for the TTL instead of re-hitting Slack every turn.
- **Bounded + best-effort**: a 3 s resolve timeout so a slow `conversations.info` can't stall the turn; **DMs are skipped** (no name); any error leaves `ChannelName` empty and is logged at debug — the turn never fails. Resolved names are **truncated to 80 chars** (Slack's max) as a defensive bound on the prompt input (Slack channel names are already lowercase-alnum/hyphen/underscore + ≤80, so meaningful prompt injection isn't expressible — the cap is belt-and-suspenders).

## Degrades gracefully — infra dependency

The seam is **wired now** and falls back to the channel id until the manifest grants the read scopes. Tracked in **layervai/qurl-integrations-infra#1002** (add `channels:read` + `groups:read`). When those land + the app reinstalls, the agent picks up names automatically — no code change.

## Why this shape (per design review)

- Resolve in the handler layer (before building `TurnContext`), not in the agent backend — the backend is Slack-agnostic and shouldn't own a Slack seam.
- Cache on the Handler (process lifetime), keyed `teamID:channelID` — names resolve once per channel per TTL across all threads.
- Blocking-with-timeout (not async-warm): simpler to reason about; the resolve is on the async turn worker (off the user's HTTP-200 path), bounded by 3 s, and rare (cache miss only).
- A single TTL with negative-caching (no rename-event subscription, no dual-TTL) — TTL staleness is acceptable for a readability nicety.

## Tests

- `TestChannelNameCache` — fresh hit, negative entry (empty name) is still a hit, TTL expiry, nil-receiver safety.
- `TestResolveChannelName` — nil seam → empty; empty channel id → no seam call; success cached (one seam call); error negative-cached (one call); over-long name truncated.
- `TestResolveChannelNameSeam_OKAndError` / `_GridFallback` (cmd) — `conversations.info` parse (`ok`→name, `missing_scope`→error→fallback) + the Grid org-token retry.
- `TestHandleEvent_AgentResolvesChannelNameSkippingDMs` — end-to-end: a channel `@mention` resolves its channel; a DM does not.

## Verification

- `/simplify`: clean (4 cleanup agents — the cache, the seam's Grid-fallback duplication, the negative-cache, and the handler-layer altitude all reviewed as correct)

Dark surface — no behavior change when the agent is disabled; and when the seam is nil/unscoped, behavior is identical to today (bare channel id).

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -158,6 +158,10 @@ func run() error {
 	// wired (same token lookup as the post seams); inert until the agent surface is live
 	// and needs the reactions:write scope in the Slack manifest to actually land.
 	agentReactions := newSlackReactionPortWithTokenLookup(workspaceTokenLookup, userAgent, slackReactionsAddURL, slackReactionsRemoveURL, nil)
+	// conversations.info seam so the agent's system prompt can name the channel
+	// ("#general (C123)"). Always wired (same token lookup); degrades to the bare
+	// channel id until the channels:read / groups:read scopes are in the manifest.
+	agentResolveChannelName := newSlackResolveChannelNameFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackConversationsInfoURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
@@ -229,6 +233,7 @@ func run() error {
 		AgentMaxTurnsPerUserPerHour: agentMaxTurnsPerUser,
 		AgentMaxTurnsPerTeamPerHour: agentMaxTurnsPerTeam,
 		Reactions:                   agentReactions,
+		ResolveChannelName:          agentResolveChannelName,
 	})
 
 	// Alias reads and writes must go through the same slackdata facade so
--- a/apps/slack/cmd/slack_resolve_channel_test.go
+++ b/apps/slack/cmd/slack_resolve_channel_test.go
@@ -1,0 +1,72 @@
+package main
+
+// Tests for the conversations.info channel-name seam (#659): a successful name
+// read, an ok:false (e.g. missing_scope) surfaced as an error so the agent falls
+// back to the channel id, and the Enterprise Grid org-token fallback shared with
+// the chat.postMessage poster.
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"strings"
+	"testing"
+
+	"github.com/layervai/qurl-integrations/shared/auth"
+)
+
+// channelInfoServer replies with a named channel for "C_ok" and missing_scope for
+// anything else, and records the bearer token + channel of each request.
+func channelInfoServer(t *testing.T) (srv *httptest.Server, auths *[]string) {
+	t.Helper()
+	var captured []string
+	srv = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		captured = append(captured, r.Header.Get("Authorization"))
+		if r.URL.Query().Get("channel") == "C_ok" {
+			_, _ = w.Write([]byte(`{"ok":true,"channel":{"id":"C_ok","name":"general"}}`))
+			return
+		}
+		_, _ = w.Write([]byte(`{"ok":false,"error":"missing_scope"}`))
+	}))
+	t.Cleanup(srv.Close)
+	return srv, &captured
+}
+
+func TestResolveChannelNameSeam_OKAndError(t *testing.T) {
+	srv, auths := channelInfoServer(t)
+	resolve := newSlackResolveChannelNameFuncWithTokenLookup(staticTokenLookup("xoxb-test"), "qurl-slack/test", srv.URL, srv.Client())
+
+	name, err := resolve(context.Background(), "T1", "", "C_ok")
+	if err != nil || name != "general" {
+		t.Fatalf("ok resolve: name=%q err=%v", name, err)
+	}
+	if (*auths)[0] != testBearerXoxb {
+		t.Errorf("auth = %q, want %q", (*auths)[0], testBearerXoxb)
+	}
+	// ok:false (missing the channels:read scope, a DM, not_found, ...) surfaces as an
+	// error → the caller treats it as "no name" and falls back to the channel id.
+	if _, err := resolve(context.Background(), "T1", "", "C_x"); err == nil || !strings.Contains(err.Error(), "missing_scope") {
+		t.Fatalf("expected missing_scope error, got %v", err)
+	}
+}
+
+func TestResolveChannelNameSeam_GridFallback(t *testing.T) {
+	srv, _ := channelInfoServer(t)
+	var owners []string
+	lookup := func(_ context.Context, ownerID string) (string, error) {
+		owners = append(owners, ownerID)
+		if ownerID == "T1" {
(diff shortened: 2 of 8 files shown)


justin-layerv:
Ran a 4-agent `/simplify` pass after rebasing onto main (now includes #687/#688). **Simplification, efficiency, and altitude axes: clean** (altitude reviewed as correct on placement, seam shape, cache depth, the truncation-vs-escaping call, and deferring the manifest scope to infra).

The **reuse** axis flagged one thing: the `conversations.info` seam hand-rolls the token-lookup + Enterprise Grid fallback that `slackWebAPIPoster` has. It couldn't reuse the poster directly (poster POSTs + discards the body; this GETs + parses it), and the fallback is **pre-existing** — `post` and `openView` already duplicate it — so a real dedup is a cross-seam refactor of shipped code, out of scope for a channel-name feature. **Deferred to #690** (extract a shared Grid-fallback combinator across the three seams), labeled `enhancement`/`go`/`priority: low`.

justin-layerv:
Thanks — addressed all three minor items (comment clarifications; no behavior change):

1. **Transient errors share the 30-min negative TTL** — documented at the negative-cache `put`: a transient timeout/5xx suppresses the name for up to the TTL exactly like a permanent `missing_scope`. Kept the single TTL deliberately (per the design review, a dual transient-vs-structural TTL is over-investment for a readability nicety); noted a shorter transient TTL as the refinement if it ever feels stale.
2. **No eviction of expired entries** — documented on `channelNameCache`: expired entries are overwritten on the next resolve but never reaped, so the map is bounded by the distinct channels the bot is mentioned in (small in practice; no sweeper).
3. **DM-skip vs mpim/group-DM** — clarified in the population comment: `shouldDispatchAgentEvent` only dispatches `app_mention` (channels) and `message.im` (1:1 DMs), filtering everything else **including mpim/group DMs**, so skipping `channel_type=="im"` leaves just real channels to resolve — mpim never reaches the resolve.

Separately, the `/simplify` reuse item (Grid-fallback dup across the Slack seams) is deferred to #690 (labeled).

justin-layerv:
Round 2 addressed (all comment-only, no behavior change):

1. **No singleflight on miss** — documented on `channelNameCache`: "at most one per channelNameTTL" now reads "once the entry exists," and notes concurrent turns for the same uncached channel aren't deduped (a burst can issue a few before the first populates) — bounded, fine for a nicety.
2. **mpim/group-DM precision** — corrected the population comment: `shouldDispatchAgentEvent` admits `app_mention` (any channel type) + `message.im`; the `im` skip covers 1:1 DMs, and an `app_mention` in a group DM (mpim) still resolves here but `conversations.info` returns no usable name → negative-cached, harmless. (Kept the behavior; per the design review I'm not adding an mpim-specific skip for a harmless, negative-cached one-off.)
3. **Overlapping 3s/4s bounds** — documented: the 3s ctx is the BINDING deadline; the seam's HTTP client's 4s is the looser outer bound.
4. **Body-limit drain duplication** — same cross-seam-transport cleanup as the Grid-fallback dup; folded into #690.

Security/`QueryEscape`: agreed it's safe today (channel id is trusted base32 from the Events payload) — already noted in the seam comment; left as-is.

justin-layerv:
Round 3 — took the main point as a real fix (not another comment):

1. **Transient failures no longer get the long negative-cache (the main point).** `resolveChannelName` now caches only a *definitive answer*: an answered `ok:false` (missing_scope, not_found) or a raw transport error takes the 30-min TTL, but a `context.DeadlineExceeded` / `context.Canceled` (the 3s budget tripping on a slow Slack, or a canceled turn) is **not** cached — it uses the id this turn and re-attempts next turn. So "one slow response = 30 min of degraded prompt" is gone; the load-bearing `missing_scope` → long-cache path (stops a scope-less workspace hammering Slack) is intact. New test `TestResolveChannelName` case asserts a transient ctx error re-attempts each call (not cached). Doc comments updated to say exactly what's cached vs not.
2. **`channel=` interpolation comment** — replaced the thin "avoids a net/url shadow" justification with the real reason: the id is a signature-verified Slack object id (C/G/D + base32) whose charset can't contain a query-reserved character, so raw interpolation equals escaping. Kept it unescaped (a no-op for this id); a future untrusted source would need escaping.
3. **Cache key `teamID:channelID`** — agreed it's fine (channel ids are globally unique, the prefix is harmless/clearer); no change.

Build + `-race` + lint green.

justin-layerv:
Round 4 — considered all three; declining each with reasoning (none blocking, no code change warranted):

1. **Transport error on the long TTL (low).** Considered the seam-sentinel to classify answered-vs-transport across the cmd↔internal boundary — declining it as over-investment for a best-effort readability nicety. The ctx carve-out (`DeadlineExceeded`/`Canceled`) already covers the dominant failure: the deliberately-tight 3s budget tripping when Slack is slow. A raw transport blip is rare, the resolve is best-effort + debug-logged, and the next turn re-attempts once the entry expires. (Fair point that "stable" slightly overstates a transport error — the intent is "we carve out only the two ctx cases"; the documented behavior is correct.)
2. **Raw `channel=` interpolation (very low).** Declining — the id is a signature-verified Slack object id (C/G/D + base32) whose charset can't contain a query-reserved char, so raw interpolation is provably equivalent to escaping; re-adding `net/url` for a no-op isn't worth it. The comment now states that real reason.
3. **Grid-fallback breadcrumb (very low).** Declining — a `slog.Warn` per Grid-fallback *resolve* would be log noise disproportionate to a best-effort nicety (unlike the post seam, where a mis-attributed message is operationally significant; a channel-name read isn't), and resolve failures are already debug-logged. Easy follow-up if an operator ever wants it.

Thanks for the four thorough rounds — the transient-cache catch in particular made this materially better.

Issue #665, opened 2026-06-09
Slack agent DECISION: raw Events API vs Agents/Assistants container


justin-layerv:
From the Slack agent-design audit (in-session tasks #11, #10). **Needs a product decision before GA.**

We built on the **raw Events API** (`@mention` + `message.im`). Slack's **Agents & Assistants** platform feature natively unlocks several best-practice patterns we otherwise have to hand-roll or skip:
- the split-pane **assistant container** UI;
- `assistant_thread_started` / `assistant_thread_context_changed` events;
- `assistant.threads.setStatus` (the "is thinking…" status — see the ack issue);
- `assistant.threads.setSuggestedPrompts` (**first-run suggested prompts, task #10** — 2–4 short contextual prompts; we have none today);
- `assistant.threads.setTitle`.

Requires `assistant:write` + the AI-app toggle in the manifest. **Not mutually exclusive** — a common pattern is the Assistant container for the 1:1/DM surface + `@mention` in channels.

Decision: stay on raw Events (and hand-roll ack/status + a posted suggested-prompts help message), or adopt the Assistant container for DMs. Gates suggested prompts + native status, so decide before enabling DMs.

justin-layerv:
## Recommendation: **hybrid — adopt the Assistant container for DMs (read/ask surface), keep `@mention` for channels.** Defer implementation to a post-sandbox-dogfood PR.

This is an engineering recommendation; the product owner ratifies before GA. It's **additive**, not a rework.

### What changed since this issue was filed
This is no longer a greenfield "decide before enabling DMs" choice — **#992 already shipped `message.im` + `im:history`/`im:read`**, so the sandbox dogfood is *already* running raw-events DMs. So the real question is whether to **upgrade** that raw DM surface to the Assistant container.

### Current Slack facts (verified, June 2026)
- **Additive + coexists.** The agent/assistant container (split-pane / top-bar / Chat tab) is a separate 1:1 surface; channel `app_mention` is unchanged — *"@-mentions in channels can happen like they do today; whether you support this or not is up to you."* Adopting it does **not** touch our channel handling.
- **Works on raw HTTP Events** (no Socket Mode / Bolt). Manifest: enable **Agents & AI Apps** (auto-adds `assistant:write`) + subscribe `assistant_thread_started`, `assistant_thread_context_changed` (we already have `message.im`).
- **`assistant_thread.context` carries `channel_id`/`team_id`/`enterprise_id`** = "the channel the user is currently viewing."
[...]

PR #691, opened 2026-06-10
feat(slack): assistant-container first-run prompts + title (#665)

justin-layerv:
## What

First slice of the **Slack Assistants-container adoption** (#665 — direction ratified). When a user opens the agent's assistant pane, Slack delivers `assistant_thread_started`; this handles it by setting the thread **title** + suggested **starter prompts** via a new `assistant.threads.*` seam.

> **Scope note:** this is the **cosmetic first-run win**. The *load-bearing* reason for adopting the container — DM reads scoped via `assistant_thread.context.channel_id` — is a later slice. This slice models that `context.channel_id` (documents intent) but doesn't use it yet.

## Design

- **Additive + UX-only.** The `@mention` / `message.im` turn flow is untouched. `handleAgentEvent` branches `assistant_thread_started` **before** the turn-dispatch filter into a no-LLM, no-turn handler. The 200 ack is already sent by `handleEvent`.
- **Layered gating.** Global `agentEnabled()` (in `handleAgentEvent`) → nil-seam / malformed-event guards → **per-workspace `workspaceAgentEnabled()` inside the worker**. The "Agents & AI Apps" toggle is app-level, so once infra lands the pane appears for *every* install; gating the worker on the same per-workspace opt-in the turn path uses means a workspace not opted into conversation mode gets **no** prompts (a clicked prompt's turn would be dropped by that same gate, so live-looking-but-dead prompts are worse than none).
- **Best-effort.** Scheduled on the async pool off `h.baseCtx`; the two calls are independent (prompts first — the higher-value affordance — so they land even if `ctx` expires before the title call), and a failure is logged, never surfaced.
- **New seam** `Config.AssistantThreads` (`AssistantThreadsPort`: `SetTitle` + `SetSuggestedPrompts`) — two posters over the shared `slackWebAPIPoster` (token lookup + Grid fallback + parse), mirroring the reactions port. Wired in `cmd/main.go`. Request shapes verified against Slack's `assistant.threads.*` docs (`channel_id`, `thread_ts`, `title`; `prompts:[{title, message}]`).
- **Dark + graceful.** `assistant_thread_started` only arrives once the manifest toggle + `assistant:write` scope are set; a nil seam (or unset manifest) is a no-op. Tracked in **layervai/qurl-integrations-infra#1004**. When that lands + the app reinstalls, opening the pane shows the title + prompts automatically — no code change.

The v1 starter prompts are **DM-safe** (capability/how-to, answerable without a channel read) because the pane is a 1:1 DM with no channel scope until the context-scoping slice — a channel-read prompt would resolve against the empty DM and read as broken:

```
"What can you do?"            → "What can you help me with?"
"How do I get access?"       → "How do I request access to a connector?"
"How do I protect a connector?"
```

Channel-aware prompts (using `assistant_thread.context.channel_id`) land with the later DM-scoping slice.

## Tests (6)

- `TestHandleEvent_AssistantThreadStarted` — end-to-end: the event routes to `SetTitle` + `SetSuggestedPrompts` with the right channel/thread + the static title/prompts.
- `TestAssistantThreadStarted_WorkspaceDisabledSkipped` — a workspace not opted into conversation mode (`AgentDefaultEnabled` off) gets no seam calls.
- `TestAssistantThreadStarted_NilSeamIsNoOp` — unwired seam → accepted + silently dropped (no panic).
- `TestAssistantThreadStarted_MissingThreadFieldsSkipped` — malformed `assistant_thread` (no channel/thread) → skipped, no seam call.
- `TestAssistantStarterPrompts` — 2–4 prompts, each with a label + message; title set.
- `TestSlackAssistantThreadsPort_RequestShapes` (cmd) — `setTitle` posts `{channel_id, thread_ts, title}`; `setSuggestedPrompts` posts `{channel_id, thread_ts, prompts:[{title, message}]}`; bearer token + URL routing.

## Verification

- `/simplify`: clean — the port (mirrors `ReactionPort`), the early dispatch branch, the handler/worker split, the layered gating, and the best-effort sequential 2-call posture all reviewed as correct.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -162,6 +162,11 @@ func run() error {
 	// ("#general (C123)"). Always wired (same token lookup); degrades to the bare
 	// channel id until the channels:read / groups:read scopes are in the manifest.
 	agentResolveChannelName := newSlackResolveChannelNameFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackConversationsInfoURL, nil)
+	// assistant.threads.* seam for the Assistants-container first-run UX (title +
+	// suggested prompts). Always wired (same token lookup); inert until the "Agents &
+	// AI Apps" manifest toggle + assistant:write scope are set (the events don't
+	// arrive, and the calls would no-op, until then).
+	agentAssistantThreads := newSlackAssistantThreadsPortWithTokenLookup(workspaceTokenLookup, userAgent, slackAssistantSetTitleURL, slackAssistantSetSuggestedPromptsURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
@@ -234,6 +239,7 @@ func run() error {
 		AgentMaxTurnsPerTeamPerHour: agentMaxTurnsPerTeam,
 		Reactions:                   agentReactions,
 		ResolveChannelName:          agentResolveChannelName,
+		AssistantThreads:            agentAssistantThreads,
 	})
 
 	// Alias reads and writes must go through the same slackdata facade so
--- a/apps/slack/cmd/slack_assistant_test.go
+++ b/apps/slack/cmd/slack_assistant_test.go
@@ -1,0 +1,67 @@
+package main
+
+// Tests for the assistant.threads.* seam (container Slice 1): request shape (URL
+// routing, {channel_id, thread_ts, title} for setTitle and {…, prompts:[{title,
+// message}]} for setSuggestedPrompts, bearer token).
+
+import (
+	"context"
+	"encoding/json"
+	"io"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/layervai/qurl-integrations/apps/slack/internal"
+)
+
+const testAssistantThreadTS = "100.1"
+
+func TestSlackAssistantThreadsPort_RequestShapes(t *testing.T) {
+	type captured struct {
+		path, auth string
+		body       map[string]any
+	}
+	var got []captured
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		raw, _ := io.ReadAll(r.Body)
+		var b map[string]any
+		_ = json.Unmarshal(raw, &b)
+		got = append(got, captured{path: r.URL.Path, auth: r.Header.Get("Authorization"), body: b})
+		_, _ = w.Write([]byte(`{"ok":true}`))
+	}))
+	t.Cleanup(srv.Close)
+
+	port := newSlackAssistantThreadsPortWithTokenLookup(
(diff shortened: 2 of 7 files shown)


justin-layerv:
Thanks — addressed everything, plus the CI fail:

- **CI / Validate PR Title**: fixed — the title started uppercase ("Assistants…"); the workflow requires `^[a-z]`. Retitled to start lowercase.
- **#1 (gate divergence) — fixed.** `setAssistantFirstRun` now gates on `workspaceAgentEnabled(ctx, log, teamID)` before setting anything — the same per-workspace opt-in the turn path uses (cheap `workspace_mappings` GetItem, on the async `baseCtx`). So once the app-level manifest toggle lands, a workspace that hasn't opted into conversation mode gets **no** prompts (a clicked prompt's turn would be dropped by that same gate). New test `TestAssistantThreadStarted_WorkspaceDisabledSkipped`.
- **#2 (\"this channel\" prompts) — fixed.** Softened the v1 prompts to DM-safe capability/how-to starters ("What can you do?", "How do I get access?", "How do I protect a connector?") the agent answers without a channel read. The pane is a 1:1 DM with no channel scope until the `context.channel_id` slice, so channel-read prompts would have resolved against the empty DM — those land with that later slice.
- **Nit (ordering comment) — fixed.** Corrected the reasoning: the two calls are independent (a failure in one never suppresses the other regardless of order); prompts go first only so they land if `ctx` expires before the title call.
- **Nit (unused `assistantThreadContext`)**: confirmed — `golangci-lint v2.10.1` is clean; staticcheck doesn't flag unused struct fields, so it's not relying on the json tags. Intentional (documents the next slice's intent).

Separately, verified the `assistant.threads.setTitle` / `setSuggestedPrompts` request shapes against Slack's docs (`channel_id`, `thread_ts`, `prompts:[{title, message}]`, optional `title`) — they match the seam.

justin-layerv:
Thanks — corrected the description (it was written before the round-1 prompt/gate changes and went stale):

- **Starter prompts** in the body now match the code (the DM-safe `"What can you do?"` / `"How do I get access?"` / `"How do I protect a connector?"`), with the rationale (1:1 DM, no channel scope until the context-scoping slice). Removed the contradicting `"List this channel's aliases"`.
- **Design** section now documents the layered gating (global `agentEnabled` → guards → per-workspace `workspaceAgentEnabled`).
- **Tests** list updated to 6 (added `TestAssistantThreadStarted_WorkspaceDisabledSkipped`).

Nits, confirmed:
- **Token lookup per verb** — yes, intentional: it mirrors the reactions/post seams exactly (each poster owns its token resolution + Grid fallback). Fine for this best-effort, rare (pane-open) path; if the `assistant.threads.*` surface grows (e.g. `setStatus` per turn), a shared per-event token resolution would be the optimization — noted for that slice.
- **`assistantThreadContext.EnterpriseID` / `assistantThread.UserID` unused** — deliberate (modeling the full event shape for the context-scoping slice), and `golangci-lint v2.10.1` is clean (staticcheck doesn't flag unused struct fields).

No code change in this round — the description was the only correction.

Issue #693, opened 2026-06-11
slack agent: make working-on-it acks surface-exclusive (pane setStatus vs channel reaction) post-#1004


justin-layerv:
## Context

Container Slice 2 (#665) added the native assistant-pane "thinking…" status
(`assistant.threads.setStatus`) on `message.im` (pane) turns, as the pane-native
counterpart to the 👀 reaction ack (#661). It is deliberately **additive**: the 👀
reaction still fires for every turn, including `im` turns.

This is the only dark-safe option today, because the code cannot tell from the event
whether the "Agents & AI Apps" pane is enabled for the install (infra
qurl-integrations-infra#1004). Pre-#1004 the app's DM is a plain DM where `setStatus`
has no assistant thread to scope to (so it fails, and the 👀 is the working ack);
post-#1004 the DM *is* the pane.

Keeping both acks additive has three interim costs:

1. **Double-ack in the pane** (post-#1004): if reactions render in the assistant pane,
   an `im` turn shows BOTH a 👀 reaction and the native "thinking…" status.
2. **Per-era log noise**: `setStatus` fails on every pre-#1004 `im` turn (logged at
   Debug to keep it quiet). Symmetrically, post-#1004 the reaction add may fail in the
   pane and would Warn on every pane turn (`addAgentAck` logs at Warn).
3. **Doubled critical-path round-trip**: an `im` turn now makes two serial best-effort
   round-trips before the LLM (reaction add, then `setStatus`), each capped at
   `agentAckTimeout` (2s) — worst case ~4s of cosmetic pre-LLM delay vs ~2s. Negligible
   against the 90s turn budget, but real.

## Proposed change

Once post-#1004 pane-reaction behavior is known, make the two acks **surface-exclusive**:
- assistant pane (`message.im`) turn → native `setStatus` only.
- channel (`app_mention`) turn → 👀 reaction only.

This single change resolves all three costs: no double-ack (one indicator per surface),
no per-era log noise (each ack only runs where it works), and no doubled round-trip (an
`im` turn does one ack, not two). Relates to #684 (taking the ack round-trips off the
turn's critical path) and the progressive-status track (#663).

## Blocked on / verify first (qurl-integrations-infra#1004 smoke test)

Before flipping to exclusive, the #1004 pre-enablement smoke test must confirm, against
the live pane:
- `setStatus` **shows** the "thinking…" indicator while a pane turn runs, and
[...]

Issue #663, opened 2026-06-09
Slack agent: progressive status / plan blocks / response streaming


justin-layerv:
From the Slack agent-design audit (in-session task #8). Today a multi-tool turn shows nothing until the final reply.

- Progressive status for multi-step turns ("Searching workspace…" → "Found 3…").
- Collapsible **plan blocks** where the agent makes decisions (one short phrase per step), visually secondary to the final output.
- **Stream** long-form responses (summaries/drafts/analysis) so users start reading immediately; keep short structured replies as single messages.

Larger effort: needs Block Kit plan rendering and/or `chat.update` streaming, and interacts with the Events-API-vs-Assistant-container decision (see the decision issue). Sequence after the ack/status indicator.

justin-layerv:
**PR2 scope (response-streaming delivery) — pinned acceptance criteria**

PR1 (#700) landed the agent-package streaming core (`streamingLLM` + `anthropicLLM.StreamComplete` + `WithStreamSink` + the `complete()` router), dark/test-exercised. PR2 wires the Slack delivery + turn path. Capturing the contract here so it isn't lost:

- **Slack seam**: an `AgentStreamPort` (Start/Append/Stop) over `chat.startStream` / `chat.appendStream` / `chat.stopStream`, mirroring the `AssistantThreadsPort`/`ReactionPort` per-team-token + Enterprise-Grid-fallback poster pattern in `cmd/slack_webapi.go`. `chat.startStream` returns the stream `ts` (which `appendStream`/`stopStream` need), so it needs a response-parsing poster variant — the existing `slackWebAPIPoster`s are fire-and-forget.
- **Coalescing / rate limit**: `chat.appendStream` is Tier-4 (~100/min); buffer text deltas and flush on a size/time threshold rather than per token. The Go path has no Slack `streamer` helper (Python/Node SDK only), so the buffer is hand-rolled; also eyeball how `appendStream` renders markdown spans split across appends (`**bo`+`ld**`).
- **No-double-post invariant**: once `startStream` succeeds, the existing `postAgentReply` text fallback must NOT also fire. On mid-stream `appendStream` failure or ctx-cancel (turn timeout), `stopStream` to finalize the partial — never a second post.
[...]

PR #694, opened 2026-06-11
feat(slack): assistant-pane thinking status on agent turns (#665)

justin-layerv:
## What

Container **Slice 2** of the Slack Assistants-container adoption (#665). On a DM (assistant-pane) turn the agent now shows Slack's **native "thinking…" status** in the pane via a new `assistant.threads.setStatus` seam — the pane-native counterpart to the 👀 reaction ack (#661), and a step toward the progressive-status track (#663).

> **Scope:** UX-only and dark. The pane exists only once the "Agents & AI Apps" toggle + `assistant:write` scope land (infra **layervai/qurl-integrations-infra#1004**); until then the seam no-ops / fails harmlessly. No behavior change to the @mention/DM turn flow.

## Design

- **Where.** `setAgentThinkingStatus` runs in `processAgentEvent` right after the reaction ack and **before the LLM call**, only past the disabled/dedupe/rate-limit gates — so the status marks only turns that actually run (same as the 👀).
- **im-only.** Fires only for `message.im` (pane) turns; `app_mention` is a channel with no assistant thread to scope `setStatus` to.
- **Synchronous, no paired clear.** Set on the turn's critical path (like `addAgentAck`, sharing `agentAckTimeout`) so it's visible while the turn runs. Slack **auto-clears** the status when the agent posts its reply (every turn exit posts one), with a 2-min server-side timeout as backstop — so there's no deferred clear. The `thread_ts` **must** equal the reply's for the auto-clear to fire; both derive from `agentEventRootTS(&env.Event)` (one pure call), and a test pins the equality.
  - Sync-vs-async was weighed: an async set that landed after a fast turn's reply would strand the status until the 2-min timeout, so sync is correct (and robust even if the auto-clear assumption is wrong).
- **Additive, not exclusive.** The 👀 reaction ack still fires for every turn (no change to #661). The code can't detect from the event whether the pane is enabled (#1004), so it can't safely *swap* reaction→status for `im` — pre-#1004 the DM is a plain DM where `setStatus` fails and the 👀 is the only working ack (it's live in sandbox). Making the acks **surface-exclusive** once post-#1004 pane behavior is known is tracked in **#693** (which also removes the interim double-ack, the per-era log noise, and the doubled critical-path round-trip).
- **Dark + best-effort.** Nil seam = no-op; a failure never fails the turn. `setStatus` fails on every pre-#1004 `im` turn (no assistant thread yet), so it logs at **Debug, not Warn**, to avoid per-turn noise.
- **New seam verb** `AssistantThreadsPort.SetStatus` — a third poster over the shared `slackWebAPIPoster` (token lookup + Grid fallback + parse), mirroring `setTitle`/`setSuggestedPrompts`. Wired in `cmd/main.go`. Request shape (`{channel_id, thread_ts, status}`) verified against the `assistant.threads.setStatus` docs.

## Tests

- `TestAgentStatus_SetForPaneTurnOnReplyThread` — an `im` turn sets one status with the pane channel/thread, status `"is thinking…"`, **on the same thread the reply posts to** (the auto-clear precondition).
- `TestAgentStatus_NotSetForAppMention` — a channel `app_mention` turn sets **no** status.
- `TestAgentStatus_NilSeamIsNoOp` — unwired seam: the `im` turn still runs and replies, no panic.
- `TestAgentStatus_BestEffortDoesNotFailTurn` — `setStatus` returns an error (the pre-#1004 reality) → the turn still posts its reply.
- `TestSlackAssistantThreadsPort_RequestShapes` (cmd) — extended: `setStatus` posts `{channel_id, thread_ts, status}` to the right URL with the bearer token.

## Verification

- `/simplify`: clean after two comment-dedup cleanups (call-site comment no longer restates the function doc; const comment trimmed). The reuse / altitude / efficiency reviews confirmed the seam extension, the `im` gate, the sibling-funcs split (correct interim altitude), and the structural `thread_ts` coupling. The one efficiency note (the additive `im` double round-trip) is deferred to **#693**, which removes it.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -162,11 +162,11 @@ func run() error {
 	// ("#general (C123)"). Always wired (same token lookup); degrades to the bare
 	// channel id until the channels:read / groups:read scopes are in the manifest.
 	agentResolveChannelName := newSlackResolveChannelNameFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackConversationsInfoURL, nil)
-	// assistant.threads.* seam for the Assistants-container first-run UX (title +
-	// suggested prompts). Always wired (same token lookup); inert until the "Agents &
-	// AI Apps" manifest toggle + assistant:write scope are set (the events don't
-	// arrive, and the calls would no-op, until then).
-	agentAssistantThreads := newSlackAssistantThreadsPortWithTokenLookup(workspaceTokenLookup, userAgent, slackAssistantSetTitleURL, slackAssistantSetSuggestedPromptsURL, nil)
+	// assistant.threads.* seam for the Assistants-container UX (first-run title +
+	// suggested prompts, and the per-turn "thinking…" status). Always wired (same token
+	// lookup); inert until the "Agents & AI Apps" manifest toggle + assistant:write scope
+	// are set (the events don't arrive, and the calls would no-op, until then).
+	agentAssistantThreads := newSlackAssistantThreadsPortWithTokenLookup(workspaceTokenLookup, userAgent, slackAssistantSetTitleURL, slackAssistantSetSuggestedPromptsURL, slackAssistantSetStatusURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
--- a/apps/slack/cmd/slack_assistant_test.go
+++ b/apps/slack/cmd/slack_assistant_test.go
@@ -1,8 +1,8 @@
 package main
 
-// Tests for the assistant.threads.* seam (container Slice 1): request shape (URL
-// routing, {channel_id, thread_ts, title} for setTitle and {…, prompts:[{title,
-// message}]} for setSuggestedPrompts, bearer token).
+// Tests for the assistant.threads.* seam (container Slices 1–2): request shape (URL
+// routing, {channel_id, thread_ts, title} for setTitle, {…, prompts:[{title, message}]}
+// for setSuggestedPrompts, {…, status} for setStatus, bearer token).
 
 import (
 	"context"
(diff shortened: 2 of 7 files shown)


justin-layerv:
Thanks — addressed all three notes (pushed in `4b917864`):

1. **Critical-path latency framing.** Clarified the `setAgentThinkingStatus` doc so `agentAckTimeout` reads as a *per-call round-trip cap, NOT a deadline shared with the reaction ack* — and called out that an `im` turn where both seams hang can add up to `2×agentAckTimeout` before the LLM, with that additive cost tracked in #693. (The fix itself stays deferred; making the acks surface-exclusive in #693 removes the second round-trip entirely.)

2. **Debug→Warn post-#1004.** Good catch — Debug is right today (every pre-enablement `im` turn fails, so Warn-per-turn would be noise), but it would hide a *genuine* failure once the pane is live. Added an explicit acceptance criterion to #693: post-enablement, a pane-turn `setStatus` failure logs at **Warn**, not Debug. (#693 comment: https://github.com/layervai/qurl-integrations/issues/693#issuecomment-4676315083)

3. **`im`-only vs mpim.** Confirmed consistent: `setAgentThinkingStatus` and the channel-name-resolution skip both key on `ChannelType == "im"`, so an `app_mention` in a group DM (mpim) correctly gets no status (no assistant thread to scope to).

justin-layerv:
Addressed all three (pushed in `e9722ce4`):

**Latency confirmation (substantive).** Confirmed it fast-fails, not hangs: `setAgentThinkingStatus` runs the call under the 2s `agentAckTimeout` ctx, and the poster's HTTP client is independently bounded at `slackChatPostMessageTimeout` (4s) — so the call is double-bounded and the ctx (2s) cancels first. `assistant.threads.setStatus` on a non-existent assistant thread returns a *fast structured Slack error* (synchronous arg validation), not a slow path, so pre-#1004 each `im` turn pays one cheap throwaway round-trip, never a 2s stall. Added that to the doc comment so a future reader doesn't have to re-derive it. (Real-world exposure is also tiny: in prod the agent is dark via `AgentDefaultEnabled=false` so no turns run, and the only place it's live — sandbox — is exactly where #1004 lands, so the pre-enablement window there is short.)

**Nit — doc line break.** Fixed the "That" strand → "The auto-clear only fires when the reply lands on the SAME thread…".

**Nit — render convention.** Added to the `agentThinkingStatus` const comment: Slack renders it as `<app> is thinking…`.

justin-layerv:
Thanks — both are observations, dispositioned:

1. **Additive double round-trip / pre-#1004 wasted leg.** Acknowledged; deferred to #693 (surface-exclusive acks remove the second leg). The PR body + the `setAgentThinkingStatus` comment both carry the cost note, and the Debug log level keeps the pre-enablement era quiet — agreed it's the right call there.

2. **Unicode ellipsis in `"is thinking…"`.** Intentional, keeping it. It follows the repo's house typography (em-dashes and `…` already appear in the agent's strings/comments and elsewhere in `internal/`), and since Slack renders the status verbatim it's functionally identical to the doc's `"is thinking..."`. Trivially reversible if we ever want to match the vendor example exactly.

That covers all cr feedback across the three rounds — no outstanding requested changes. CI is green on `e9722ce4`.

PR #695, opened 2026-06-11
feat(slack): persist assistant-pane context channel (#665)

justin-layerv:
## What

Container **Slice 3a** of the Assistants-container adoption (#665) — the foundation for scoping the assistant pane's reads to the channel a user opened it from. The pane DM has no channel of its own, and Slack delivers the source channel (`assistant_thread.context.channel_id`) **only on the container events**, not on the `message.im` turns — so it must be **persisted per pane thread** for a later turn to use.

> **Scope split:** this PR is **persist-only and dark** — nothing reads the context yet. The read-side (override `TurnContext.ChannelID` so the agent reads the opener's channel, behind a membership/authz gate) is the follow-up slice **3b**. Keeping them apart isolates the read-path/authz change from the plumbing.

## Design

- **Store.** New `AgentStore.PutThreadContext` / `GetThreadContext` over an `actx#<thread_key>` item — mirrors the pending-action methods (unconditional **last-write-wins** Put, read-time TTL enforcement, the same validation / `stringAttr` / `readNumber` helpers). The complete layer ships now (`Get` is exercised by 3b).
- **Partition = team id, not the conversation partition.** Context is *written* on `assistant_thread_started` / `_context_changed` and *read* on the `message.im` turn — three event types — and only the team id is guaranteed identical across them on Grid (the enterprise field can vary by event type). The thread key is globally unique, so org-grain partitioning would buy nothing. (Same reasoning that partitions pending-actions on team id.)
- **TTL reuses `conversationTTL`** so the context lives exactly as long as the conversation it scopes. (3b refreshes it per-turn so it tracks conversation activity.)
- **Handlers.** Persist on `assistant_thread_started` (alongside the first-run UX) and on a **new** `assistant_thread_context_changed` handler (the user switching the channel they're viewing while the pane is open, last-write-wins). The shared scaffolding (validate, per-workspace opt-in gate, async schedule off `h.baseCtx`) is factored into `scheduleAssistantContainerEvent`. The persist is **decoupled from the `AssistantThreads` seam** — it depends on the `AgentStore`, so context persists even where the seam is nil — and gated on the same per-workspace opt-in as a turn. Best-effort: a store error is logged, never surfaced.
- **Key alignment.** `agentThreadKey(channel, threadRoot)` factors the `channel:thread_root` key; the container-event **write** key and the turn **read** key (`agentEventThreadKey`, now delegating to it) can't drift.

## Tests

- Store (`agentstate_test.go`): round-trip + TTL, last-write-wins overwrite, read-time expiry, validation.
- Handlers (`handler_agent_context_test.go`): `started` persists context **and** runs first-run UX; `context_changed` overwrites it with no re-UX; persists **even with a nil seam** (proves the decoupling); workspace-disabled skips both; missing `context.channel_id` skips the persist but keeps the UX.

## Verification

- `/simplify`: clean after applying its findings — extracted the shared `scheduleAssistantContainerEvent` (deduped the two handlers), trimmed a duplicated rationale comment, dropped a structurally-guaranteed key-alignment test (3b's write→read roundtrip covers it for real), and fixed three stale doc comments (the `assistant_thread` object now arrives on both container events; `Context` is now used). The reuse / efficiency / altitude reviews affirmed the team-id partition, the `actx#` item type, and the `conversationTTL` reuse.

**Dark** until infra **layervai/qurl-integrations-infra#1004** (AI-app toggle + `assistant:write`). 3a also adds an infra dependency: the **`assistant_thread_context_changed`** event must be subscribed in the manifest — noted on #1004.

### DIFF
--- a/apps/slack/internal/handler_agent.go
+++ b/apps/slack/internal/handler_agent.go
@@ -17,10 +17,11 @@ import (
 
 // Slack Events API event types this handler reacts to.
 const (
-	slackEventTypeAppMention             = "app_mention"
-	slackEventTypeMessage                = "message"
-	slackEventTypeAssistantThreadStarted = "assistant_thread_started"
-	slackChannelTypeIM                   = "im"
+	slackEventTypeAppMention                    = "app_mention"
+	slackEventTypeMessage                       = "message"
+	slackEventTypeAssistantThreadStarted        = "assistant_thread_started"
+	slackEventTypeAssistantThreadContextChanged = "assistant_thread_context_changed"
+	slackChannelTypeIM                          = "im"
 )
 
 // agentProposalPreviewPrefix prefixes a proposed-mutation reply while
@@ -83,7 +84,8 @@ type slackEventEnvelope struct {
 }
 
 // slackInnerEvent is the inner `event` object for app_mention / message events,
-// plus the assistant_thread object on assistant_thread_started.
+// plus the assistant_thread object on the container events (assistant_thread_started
+// and assistant_thread_context_changed).
 type slackInnerEvent struct {
 	Type        string `json:"type"`
 	User        string `json:"user"`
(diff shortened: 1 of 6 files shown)


justin-layerv:
Thanks — all three are non-blocking and dispositioned:

1. **TTL refresh is 3b's job (tracked).** Correct that 3a *alone* has the gap — there's no turn read path here to refresh on. The per-turn refresh lands with the read-side in **3b** (the `PutThreadContext` doc already notes "the turn path refreshes it like SaveConversation refreshes the transcript"), and it's explicit on 3b's checklist so it can't slip silently. Dark until then, so no live exposure.

2. **`context_changed` DDB amplification — leaving it.** Agreed it's acceptable (gated + best-effort + off the ack path). An "unchanged" guard would cost *more*: the event doesn't carry the prior channel, so skipping a redundant write needs a GetItem first — a read to maybe save a write, net-negative in the common case. And Slack only fires the event on an actual change, so same-channel back-to-backs barely happen.

3. **nil-seam now schedules a worker — intended.** That's the decoupling: persist depends on the AgentStore, not the AssistantThreads seam, so the worker must run to persist even when the seam is nil. Pinned by `TestAssistantContext_PersistsEvenWithNilSeam`. (In production the seam is always wired, so this only shifts the nil-seam/test path.)

All CI green on `9d3ce1fd`; no code changes from this round.

Issue #697, opened 2026-06-11
slack agent: unify channelNameCache + channelMembershipCache into one generic TTL cache


justin-layerv:
Surfaced by /simplify during Slice 3b.

`channelMembershipCache` (handler_agent_membership.go) and `channelNameCache` (handler_agent_channel.go) are structurally identical TTL caches — same `mu/ttl/now/m` struct, same `Entry{value, expiresAt}`, same `new*Cache`/`get`/`put`/nil-receiver handling, same "only cache definitive answers" discipline. They differ only in the stored value type (`string` vs `bool`).

A generic `ttlCache[V any]` (`get(key) (V, bool)` / `put(key string, v V)`) would collapse both.

**Deliberately deferred from 3b** because:
- `apps/slack/internal` + `cmd` currently use **zero generics** — introducing the first one to dedupe ~40 lines is a net cost against the house style (cf. the `slackWebAPIPoster` doc: parallel-but-distinct seams are the intended design).
- The fix necessarily edits the already-shipped `channelNameCache` (#659), out of scope for one slice.

A judgment-call cleanup, not a clear win — filed for visibility. Reasonable outcomes: do it if/when the codebase adopts generics elsewhere, or close as WONTFIX if the no-generics style holds.

## Acceptance criteria
- [ ] A single generic TTL cache backs both channel-name and channel-membership memoization, OR a decision to keep them separate is recorded here.

justin-layerv:
Candidate scope from the #698 (Slice 3b) cr — optional membership-cache tunings to weigh alongside the unification:
- **Asymmetric TTL**: a shorter TTL on `member==true` (bounds the security-relevant revocation lag — a removed member keeping scope) and a longer one on `member==false` (the safe-direction usability cache). 30m symmetric is already justified as acceptable for the low-to-moderate disclosure, so this is a tightening, not a fix.
- **Singleflight** around the membership seam: concurrent first turns for the same `(channel, user)` before the cache fills each issue their own bounded `conversations.members` scan (bounded by the per-turn rate cap today).

PR #698, opened 2026-06-11
feat(slack): membership-gated assistant-pane read scope (#665)

justin-layerv:
## What

Container **Slice 3b** — the consume side of read-scoping Slice 3 (3a, #695, persisted the pane context). On an assistant-pane (`message.im`) turn, scope the agent's reads to the channel the user opened the pane from (the persisted `context.channel_id`) by overriding `TurnContext.ChannelID` — **only when the user is a confirmed member of that channel**. This is the load-bearing payoff of the whole container: pane Q&A ("what can I reach here?") now answers against the opener's channel instead of the empty DM.

## The security model (why the membership gate)

The agent's read backends (`AllowedResourceIDsForChannel`, `GetChannelPolicy`, `LookupChannelAlias`) key **only on the channel** — no per-user check — so whoever's channel lands in the `TurnContext` sees that channel's full qURL alias→resource topology. And `context.channel_id` can be a channel the user is only **previewing** (a public channel they haven't joined). Without a gate, a non-member could enumerate a channel's access topology through the pane.

So the membership check **is the authorization** for pane-scoped reads, and it **fails closed**: a nil seam, an error, a timeout, no stored context, or a non-member all fall back to the **un-scoped DM**. `resolveChannelMembership` collapses every uncertain path to "not a member."

- **Mutations are unaffected** (verified + regression-pinned). The confirm flow anchors to `env.Event.Channel` (propose) and the click's channel (execute), **never** `TurnContext.ChannelID` — so the override widens *reads* only, never the action target. A new test runs a scoped pane turn to a proposal and asserts the pending action anchors to the **DM**, not the context channel.
- **Bounded, best-effort determination.** The `conversations.members` scan is bounded to **2 pages** — a member of a very large channel beyond that bound reads as *not-confirmed* and gets the un-scoped pane. **This degradation is intentional**: it's in the safe direction (deny, never grant), and the disclosure it guards (a workspace member learning a public channel's topology) is low-to-moderate, so an exhaustive scan isn't warranted.
- **Revocation lag** is an accepted, documented tradeoff: a `(channel, user)` decision is cached for `channelMembershipTTL` (30m), so a removed member keeps a scoped pane for up to one TTL (standard for cached auth; strictly weaker than a never-member, who is denied immediately).

## Design

- **Seam** `Config.ChannelMembership` (`ChannelMembershipFunc`, mirrors `ResolveChannelName`): cmd `conversations.members` bounded 2-page scan (`fetchConversationsMembersPage` + the Grid-aware token-lookup wrapper) → internal cache + fail-closed resolver. **Reuses the `channels:read` / `groups:read` scopes** (no new scope; private channels self-handle — `conversations.members` errors unless the bot is in the channel → fail-closed → no scope, and private ⟹ the user is a member anyway).
- **Override site** (`processAgentEvent`): for an `im` turn, `paneContextChannel` does `GetThreadContext` → membership gate → on a confirmed member, set the operating channel + **refresh the context TTL** so channel-awareness tracks the conversation's activity (3a deferred this to here). "Operate on" (reads) is separated from "post to" (the reply still lands in the DM).
- **Efficiency:** a nil seam short-circuits the `GetThreadContext` entirely; `app_mention` turns pay nothing; membership is cache-consulted before any network call.

## Tests

- Internal gate (`handler_agent_membership_test.go`): member → scopes to the context channel; non-member / check-error / no-context / nil-seam → un-scoped DM (no topology leak); cache policy (definitive answers cached, ctx-timeout not); per-turn TTL refresh; and **the operate-vs-post security pin** (scoped proposal → pending action anchored to the DM).
- cmd seam (`slack_membership_test.go`): request shape (channel + limit + bearer), member detection, cursor paging, the **2-page bound** (stops rather than following the cursor forever), and `ok:false` → error.

## Verification

- `/simplify`: applied its findings — added the operate-vs-post regression pin (its top recommendation), reframed the gate doc (it's the *primary* pane-read boundary, not mere defense-in-depth), gave the cmd seam keep-alive-drain parity with its sibling, and trimmed a duplicated doc. The reuse review confirmed the parallel cache/resolver/seam are the codebase's deliberate parallel-but-distinct style (zero generics); the cache-unification option is tracked in **#697**.

**Dark** until infra **qurl-integrations-infra#1004** (AI-app toggle + `assistant:write` + the `assistant_thread_*` events). The read scopes are the same ones tracked for #659.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -162,6 +162,12 @@ func run() error {
 	// ("#general (C123)"). Always wired (same token lookup); degrades to the bare
 	// channel id until the channels:read / groups:read scopes are in the manifest.
 	agentResolveChannelName := newSlackResolveChannelNameFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackConversationsInfoURL, nil)
+	// conversations.members seam: gates whether an assistant-pane turn may scope its reads
+	// to the channel the user opened the pane from (only a confirmed member's pane is
+	// scoped). Always wired (same token lookup + channels:read / groups:read scopes as the
+	// name resolver); fail-closed, so until the scopes are in the manifest it answers
+	// missing_scope and the pane stays on the un-scoped DM.
+	agentChannelMembership := newSlackChannelMembershipFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackConversationsMembersURL, nil)
 	// assistant.threads.* seam for the Assistants-container UX (first-run title +
 	// suggested prompts, and the per-turn "thinking…" status). Always wired (same token
 	// lookup); inert until the "Agents & AI Apps" manifest toggle + assistant:write scope
@@ -239,6 +245,7 @@ func run() error {
 		AgentMaxTurnsPerTeamPerHour: agentMaxTurnsPerTeam,
 		Reactions:                   agentReactions,
 		ResolveChannelName:          agentResolveChannelName,
+		ChannelMembership:           agentChannelMembership,
 		AssistantThreads:            agentAssistantThreads,
 	})
 
--- a/apps/slack/cmd/slack_membership_test.go
+++ b/apps/slack/cmd/slack_membership_test.go
@@ -1,0 +1,138 @@
+package main
+
+// Tests for the conversations.members membership seam (container Slice 3b): request shape
+// (channel + limit + bearer), member detection, bounded-scan paging (follows next_cursor up
+// to maxMembershipPages then stops), and Slack ok:false → error.
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"testing"
+
+	"github.com/layervai/qurl-integrations/shared/auth"
+)
+
+type membershipReq struct{ query, auth string }
+
+// membershipServer serves the given JSON bodies in order (clamped to the last) and records
+// each request's query + Authorization header.
+func membershipServer(t *testing.T, bodies ...string) (*httptest.Server, *[]membershipReq) {
+	t.Helper()
+	var got []membershipReq
+	i := 0
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		got = append(got, membershipReq{query: r.URL.RawQuery, auth: r.Header.Get("Authorization")})
+		body := bodies[len(bodies)-1]
+		if i < len(bodies) {
+			body = bodies[i]
+		}
+		i++
+		_, _ = w.Write([]byte(body))
+	}))
+	t.Cleanup(srv.Close)
+	return srv, &got
+}
+
+func TestSlackChannelMembership_FoundAndRequestShape(t *testing.T) {
+	srv, got := membershipServer(t, `{"ok":true,"members":["U1","U2","U3"]}`)
+	fn := newSlackChannelMembershipFuncWithTokenLookup(staticTokenLookup("xoxb-test"), "qurl-slack/test", srv.URL, srv.Client())
+
+	member, err := fn(context.Background(), "T1", "", "C9", "U2")
+	if err != nil || !member {
+		t.Fatalf("U2 is in the channel: member=%v err=%v", member, err)
+	}
+	if len(*got) != 1 {
+		t.Fatalf("a hit on the first page must not page further, got %d requests", len(*got))
+	}
+	req := (*got)[0]
+	if req.auth != testBearerXoxb {
+		t.Errorf("auth = %q", req.auth)
+	}
+	q, _ := url.ParseQuery(req.query)
+	if q.Get("channel") != "C9" || q.Get("limit") != "1000" {
+		t.Errorf("request query = %q (want channel=C9 limit=1000)", req.query)
+	}
+}
+
+func TestSlackChannelMembership_NotFound(t *testing.T) {
+	srv, _ := membershipServer(t, `{"ok":true,"members":["U1","U3"]}`)
+	fn := newSlackChannelMembershipFuncWithTokenLookup(staticTokenLookup("xoxb-test"), "qurl-slack/test", srv.URL, srv.Client())
+
+	member, err := fn(context.Background(), "T1", "", "C9", "U2")
+	if err != nil || member {
+		t.Fatalf("U2 absent + no next page → not a member: member=%v err=%v", member, err)
+	}
+}
+
+func TestSlackChannelMembership_PagingFindsOnSecondPage(t *testing.T) {
+	srv, got := membershipServer(t,
+		`{"ok":true,"members":["U1"],"response_metadata":{"next_cursor":"CURSOR2"}}`,
+		`{"ok":true,"members":["U2"]}`,
+	)
+	fn := newSlackChannelMembershipFuncWithTokenLookup(staticTokenLookup("xoxb-test"), "qurl-slack/test", srv.URL, srv.Client())
+
+	member, err := fn(context.Background(), "T1", "", "C9", "U2")
+	if err != nil || !member {
+		t.Fatalf("U2 on page 2 → member: member=%v err=%v", member, err)
+	}
+	if len(*got) != 2 {
+		t.Fatalf("want 2 requests (paged once), got %d", len(*got))
+	}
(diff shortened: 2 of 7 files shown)


justin-layerv:
Thanks — addressed (pushed in `c86a328a`):

**Transient negative-cache (the one to weigh).** Fixed: `resolveChannelMembership` now caches **only a definitive seam answer** (member true/false, no error); an error of *any* kind — transient 5xx, network blip, or a stable `missing_scope` — is no longer cached, so it re-checks next turn. I chose "don't cache any error" over distinguishing answered-vs-transient because it's simpler and strictly safe (never locks a member out on *any* error). The cost — re-hitting Slack on a stable `missing_scope` — is bounded by the per-turn rate cap + the 3s budget, and rare (the read scopes are present once the pane is enabled), which is the right tradeoff for an access gate. Doc + the cache test updated to match.

**Coverage gaps.**
- Added `TestPaneScope_AppMentionSkipsMembershipCheck` — an `app_mention` turn (with a context row present under its thread key) runs **zero** membership checks (the im-type gate keeps the channel path out).
- Added `TestSlackChannelMembership_GridFallback` — mirrors the conversations.info Grid test: workspace token unconfigured → retry on the org token, asserts owners `[T1, E1]`.
- The transient-vs-answered test is now subsumed: `TestResolveChannelMembership_Caching` asserts an error is **not** cached (re-checked), covering both.

**Nits.**
- Cache key omits `enterpriseID` — confirmed intentional, consistent with `channelNames` (team/channel/user ids are globally unique).
- The TTL-refresh `PutThreadContext` on the turn ctx — left as-is: it's consistent with the turn's other per-turn DDB calls (`loadAgentHistory`, `MarkEventSeen`) which also run on the turn ctx, and the turn ctx + the DDB client's own timeout bound it.

CI green on `c86a328a`.

justin-layerv:
Both reconciled (pushed in `14fb214e`, doc-only):

**1. Stale header sentence.** Rewrote the `resolveChannelMembership` header caching paragraph to match the code: *only a definitive seam answer (member true/false, no error) is cached — an error of any kind is not, so it re-checks next turn.* The two halves of the func now agree. Good catch — that sentence was indeed copied from `resolveChannelName` and not updated for membership's deliberate no-negative-cache choice.

**2. Interface contract vs impl.** Reconciled the `ChannelMembershipFunc` doc to match the implementation: a Slack **`ok:false`** now surfaces as an **error** (moved from the `(false, nil)` clause to the error clause); `(false, nil)` is returned only when the bounded scan genuinely doesn't find the user. And this is the deliberate design, not just doc-fitting: returning an error (rather than `(false, nil)`) on `ok:false` is what keeps a **transient** `ratelimited`/`internal_error` from being cached as "not a member" — the resolver caches only no-error answers, so all `ok:false` (stable or transient) correctly re-check next turn.

**Non-blocking notes.** Added the one-line `enterpriseID`-omission note at the cache key. The unbounded cache and per-turn TTL-refresh write are both intentional + already documented (left as-is).

CI green on `14fb214e`.

justin-layerv:
Added both doc sentences (pushed in `61d3512d`, doc-only):

- **Cache cardinality.** Noted that the `(channel, user)` key space is larger than `channelNameCache`'s (channel only), that it has no active reaper like its sibling, but stays bounded because it's keyed by *actual pane usage* (members who open a pane from a channel), not the team×channel×user cartesian product — with a pointer that a reaper or the #697 unification can revisit if very large workspaces need it.
- **Join lag.** The TTL doc now spells out the lag in BOTH directions: removal (keeps scope — the security-relevant revocation lag) AND join (a cached `false` delays a newly-joined member's scoped pane — benign, fails safe, usability-only).

The remaining items are confirmed intentional, no action: the redundant nil-check (skips the store read when the seam is absent), `membershipPageLimit = 1000` (Slack's practical max), the serial-on-hot-path scan (bounded + cached after the first turn), and the stable-`missing_scope` per-turn re-hit (moot until #1004 delivers pane events; bounded by the per-turn rate cap post-enablement).

CI green on `61d3512d`.

justin-layerv:
Good catch on the coverage framing (pushed in `6d1ac944`):

**`conversations.members` page-size honesty.** You're right that the "vast majority" framing silently assumed Slack honors `limit=1000`. Softened the comments: coverage is now described as *up to maxMembershipPages × the **effective** per-page size*, explicitly noting Slack may return fewer than requested (≤200 recommended; pages can be short even when not exhausted) — so no assumption of 2×1000. The degradation stays fail-closed/safe regardless. I can't do the empirical check here (no live Slack, and the pane is dark until #1004), so I gated it: added a pre-enablement smoke-test item on qurl-integrations-infra#1004 to confirm the real page size against a large channel and tune `maxMembershipPages` / `membershipPageLimit` if it's small. (Kept `limit=1000` — requesting the max can only help where honored.)

**Nit — refresh-on-attempt.** Added a half-line: the TTL refresh runs on the *attempt* (before the turn), so an active thread keeps channel-awareness even if the turn then fails transiently — it tracks activity, not success.

Everything else (cache-key `enterpriseID` omission, the bounded-scan request-count assertion as the load-bearing one, serial-on-hot-path, error-not-cached) confirmed as intended.

CI green on `6d1ac944`.

justin-layerv:
Addressed (pushed in `2eee9775`, doc-only):

**Item 2 — beyond-bound member's cached `false`.** Good distinction. Reworked the `channelMembershipTTL` doc so the lag note covers it explicitly: a cached `false` for a REAL member — whether they joined after it was cached *or* they sit beyond the bounded scan (`membershipPageLimit × maxMembershipPages`) — keeps them DM-scoped for up to one TTL, re-checked only when the entry expires. Called out the beyond-bound case as the most likely degradation in a very large channel, and contrasted it with errors (which are never cached). So the framing no longer leans only on the errors-arent-cached case.

**Item 1 — effective page size.** It's gated as a pre-enablement smoke-test item on qurl-integrations-infra#1004 (the `maxMembershipPages` comment points there): confirm the real `conversations.members` page size against a large channel and tune the bound if it's small. Agreed the failure mode is silent, which is exactly why it's a go-live gate rather than left implicit.

**Item 3 — latency placement.** Confirmed fine: the membership round-trip is on the async turn goroutine (past the ack/dedupe gate), so it never touches Slack's 3s event-ack window; bounded to 3s + cached per (channel,user). No change.

CI green on `2eee9775`.

justin-layerv:
Addressed (pushed in `2c63fd16`, doc-only):

**Private channel, bot absent.** Added the one-liner to `resolveChannelMembership`: the fail-closed default also denies a LEGITIMATE member of a private channel the bot isn't in (`conversations.members` errors there) — acceptable, since that channel's qURL data is likely empty without the bot present anyway. Now stated alongside the public-preview non-member case.

The rest are confirmed, no action:
- **Effective page size** — gated on qurl-integrations-infra#1004 as the pre-enablement measurement; agreed it's *the* knob to tune against a large channel before flipping on.
- **Comment density** — conscious choice, consistent with the sibling `ResolveChannelName` seam's house style and the security-critical nature of this gate.
- **Per-turn TTL-refresh write** — intentional (tracks activity, mirrors `SaveConversation`), documented.
- **ctx-timeout not tested distinctly** — agreed it collapses into the covered generic-error branch; not worth a dedicated test.

Thanks for the thorough security review across the rounds — the gate, the fail-closed matrix, and the operate-vs-post invariant are all the better for it. CI green on `2c63fd16`.

justin-layerv:
Thanks — this round is all optional tuning / defer-able, no code change:

- **Asymmetric TTL** (short on `member==true` to bound revocation lag, longer on `false`) — a nice tightening; since 30m symmetric is already justified for the low-to-moderate disclosure, I noted it as candidate scope on the cache-revisit issue (#697) rather than expand this slice.
- **Singleflight** around the seam — also noted on #697 (the natural home you pointed at); bounded by the per-turn rate cap today.
- **Cold-cache cost on the critical path** — acknowledged; bounded (one GetThreadContext + ≤2 × 3s scans, cached thereafter), and on the async turn goroutine, so off the ack path.
- **No reaper** — acknowledged in-code + tracked in #697.

That settles the review across the rounds — the gate, the fail-closed matrix (incl. beyond-bound + private-bot-absent), and the operate-vs-post pin are all in place, and the remaining ideas are tuning tracked in #697. CI green on `2c63fd16`.

PR #699, opened 2026-06-11
feat(slack): channel-aware assistant-pane starter prompts (#665)

justin-layerv:
## What

Container **Slice 3c** — restores the **channel-aware first-run starter prompts** Slice 1 deliberately deferred (it shipped generic DM-safe placeholders because the pane had no channel scope yet). Now that pane reads are channel-scoped + membership-gated (3a/3b), when a user opens the pane from a channel whose name resolves, the starter prompts **name it** — *"What can I reach in #general?"* — instead of the generic *"How do I request access to a connector?"*. This finishes the container's first-run experience.

## Design

- **Leak-free — no membership gate at prompt time.** The prompt only **names** the channel the user is *already viewing* (the `context.channel_id` they opened the pane from), so it can't surface anything they don't already see. The scoped **answer** stays membership-gated at turn time (Slice 3b's `paneContextChannel`) — a non-member who clicks the prompt gets the un-scoped DM, never the channel's topology. Naming a channel ≠ enumerating it.
- **Chooser at the seam.** `setAssistantFirstRun` → `assistantStarterPromptsFor`: if the context channel's name resolves → `channelAwareStarterPrompts(name)`, else the generic `assistantStarterPrompts` fallback. One branch covers both "no context channel" and "name didn't resolve" — `resolveChannelName` returns `""` (without hitting Slack) for an empty id.
- **Reuses the `resolveChannelName` seam** (#659): a cached, negative-cached, 3s-bounded `conversations.info` lookup — at most one cached call per pane open, off the ack path, best-effort, and it **warms the same cache the first scoped turn reads** (no duplicate fetch). The channel name is Slack-trusted + charset-constrained + length-capped, so interpolating it is safe.
- **Titles short + channel-neutral** (Slack truncates the label); the **message** carries `#<channel>`.

## Tests

- `TestAssistantPrompts_ChannelAwareWhenContextResolves` — context channel resolves → prompts name `#general`.
- `TestAssistantPrompts_GenericWhenContextNameUnresolved` — name doesn't resolve (e.g. no `channels:read`) → generic set, no channel named.
- `TestAssistantPrompts_GenericWhenNoContextChannel` — no context channel → generic set, and the resolve is skipped (no seam call).

## Verification

- `/simplify`: clean — reuse (correctly reuses `resolveChannelName` + the test helpers), efficiency (the lookup warms the turn-path cache), and altitude (the leak-free reasoning verified sound) all confirmed; folded a redundant empty-context guard into the single resolve branch per its one finding.

**Dark** until infra **qurl-integrations-infra#1004** (the `assistant_thread_started` event + the `channels:read` scope the name resolve needs — already reused from #659).

### DIFF
--- a/apps/slack/internal/handler_agent_assistant.go
+++ b/apps/slack/internal/handler_agent_assistant.go
@@ -9,17 +9,36 @@ import (
 // thread (assistant.threads.setTitle) — the product's user-facing agent name.
 const assistantThreadTitle = "qURL Secure Access Agent"
 
-// assistantStarterPrompts are the first-run suggested prompts shown when a user
-// opens the assistant pane (assistant.threads.setSuggestedPrompts). Static for v1
-// and deliberately DM-SAFE: the pane is a 1:1 DM with the agent, which has no
-// channel scope until the context-scoping slice (assistant_thread.context.channel_id),
-// so a channel-read prompt ("what can I reach here?") would resolve against the empty
-// DM and read as broken. v1 uses capability/how-to starters the agent answers without
-// a channel read; channel-aware prompts land with the context.channel_id slice.
+// assistantStarterPrompts are the DM-SAFE fallback first-run prompts: capability/how-to
+// starters the agent answers without a channel read. They're used when the pane has no
+// context channel (opened from a DM or the App Home) or its name can't be resolved (no
+// channels:read scope). When the context channel does resolve, channelAwareStarterPrompts
+// names it instead (see assistantStarterPromptsFor).
 var assistantStarterPrompts = []SuggestedPrompt{
 	{Title: "What can you do?", Message: "What can you help me with?"},
 	{Title: "How do I get access?", Message: "How do I request access to a connector?"},
 	{Title: "How do I protect a connector?", Message: "How do I protect a connector?"},
+}
+
+// channelAwareStarterPrompts names the channel the user opened the pane from, so the
+// starters leverage the pane's channel scope ("What can I reach in #general?"). Leak-free:
+// the prompt only NAMES a channel the user is already viewing (channelName came from the
+// context.channel_id they opened the pane from); the scoped ANSWER stays membership-gated at
+// turn time (paneContextChannel). Titles stay short + channel-neutral (Slack truncates the
+// label); the message carries the channel name. Interpolating channelName is safe: Slack
+// normalizes channel names to a constrained charset (lowercase/digits/hyphens/underscores)
+// before conversations.info returns it, and it renders here as plain Slack text (not
+// interpreted) — the same trust the turn-path system prompt already places in the name. The
+// generic "What can you do?" capability starter is kept (last) so a brand-new user opening
+// the pane from a channel isn't worse off than from a DM.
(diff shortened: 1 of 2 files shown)


justin-layerv:
All three addressed in one push (`a6c022ac`):

1. **Charset provenance.** Reworded the `channelAwareStarterPrompts` doc — the constrained charset is *Slack's* normalization of channel names (before `conversations.info` returns them), not something this code enforces; and it renders as plain Slack text, not interpreted. So the safety attribution is now accurate (same trust the turn-path system prompt already places in the name).

2. **Capability starter kept.** Good call — added `"What can you do?"` as the 4th channel-aware prompt (Slack allows up to 4), so a brand-new user opening the pane *from a channel* gets the channel-aware starters AND the onboarding one, no worse off than from a DM. Pinned by `TestAssistantPrompts_ChannelAwareWhenContextResolves` (asserts the capability starter is retained).

3. **Test gap.** `TestAssistantPrompts_GenericWhenContextNameUnresolved` now asserts `res.resolved("C9")` — proving the generic fallback came from an *attempted* (empty) resolve, not an accidental short-circuit, mirroring the no-context test's negative assertion.

CI green on `a6c022ac`.

justin-layerv:
Applied the test strengthening (`064c44ac`): `TestAssistantPrompts_ChannelAwareWhenContextResolves` now asserts `len(got) == 4`, that **both** channel-aware messages (index 0 "reach" and index 1 "get access") name `#general`, and that the capability starter is last — so a future edit that drops a prompt or stops interpolating the second message is caught.

On the **4-prompt ceiling**: acknowledged — the set deliberately sits at Slack's documented max for `setSuggestedPrompts`, zero headroom. The `len == 4` assertion now also serves as a tripwire if anyone adds a 5th without dropping one; the seam passes the slice straight through and a reject-at-limit would surface as a logged best-effort error (never fails the pane). Flagged for the next editor, as you noted.

CI green on `064c44ac`.

justin-layerv:
Thanks — all three are optional and dispositioned (no code change; the slice is converged):

1. **Cache-warming test.** Declining as disproportionate: it's an efficiency property, and the optimization is *structural* — the pane-open resolve and the first scoped turn use the identical cache key (`teamID:context.channel_id`), which you verified against the source. A full pane-open→turn integration test to pin an efficiency-only property isn't worth the surface; the key alignment is the guard.

2. **2–4 bound on the channel-aware set.** It's already guarded, just split across tests: `TestAssistantPrompts_ChannelAwareWhenContextResolves` asserts `len(got) == 4`, so a 5th prompt fails *that* test (the ≥2 floor is trivially met at 4). Consolidating both sets into `TestAssistantStarterPrompts`'s bound check is a fine future tidy-up but not worth a cycle on a converged PR.

3. **Comment density / three-way duplication.** The in-code `channelAwareStarterPrompts` doc is the load-bearing home for the leak-free reasoning (a future editor of that function needs it there); the commit/PR copies are review-time framing. Keeping the in-code one is the right call.

That settles all cr feedback across the three rounds — round 1's three nits applied (charset provenance, capability starter kept, attempted-resolve assertion), round 2's test strengthening applied, round 3 dispositioned. CI green on `064c44ac`; no outstanding requested changes.

PR #700, opened 2026-06-11
feat(slack): token-stream the agent reply in the loop core (#663)

justin-layerv:
## What

**Response-streaming PR1** — the agent-package **core** for token-streaming the agent's reply so it renders token-by-token instead of posted whole. The user chose this (true token streaming of the final reply) over progress/plan-block streaming. **Dark**: nothing wires a sink yet; the Slack delivery seam (`chat.startStream`/`appendStream`/`stopStream`) + turn-path wiring land in **PR2** (`assistant:write`-gated).

## Design

- **`streamingLLM` — an optional extension to the `LLM` port** (`StreamComplete`), asserted at the seam in a `complete()` router (like `io.WriterTo`). Every existing non-streaming `LLM` fake stays valid: with no sink, or an LLM that doesn't implement the extension, `Run` takes the unchanged `Complete` path. Zero ripple to the existing loop/tests.
- **`anthropicLLM.StreamComplete`** issues the *same* request as `Complete` (shared `buildParams`) over `Messages.NewStreaming`, accumulates the SSE events into the full `Message` (shared `fromSDKMessage`), and taps each text delta to an `onText` sink. A streamed turn and a non-streamed turn produce **byte-identical** text, tool calls, and usage — only delivery timing differs.
- **`WithStreamSink` (per-turn)** wires the sink; the handler builds a fresh `Agent` per turn, so the sink never outlives its turn.
- **`streamTextDelta`** reads the *flattened* `MessageStreamEventUnion` fields (`Type` + `Delta.Text`) the SDK already populates at decode time, avoiding the per-token JSON re-parse the typed `.AsAny()` accessors perform on the hot path.

### Contract for PR2 (pinned by tests)

A `propose_*` round's **narration streams to the sink, yet `Run` still returns the `Proposal`** (never a `Reply`): the text streams before the `tool_use` block arrives and the stream is append-only, so there's no suppressing it — the agent "speaks," then the confirm card follows. A **no-narration** propose round streams nothing, so PR2's lazy stream-start (open on first delta) opens no stream at all. The common Q&A case is clean because the system prompt leads with the answer, not preamble.

## Tests

- `TestRun_Streaming_ForwardsFinalReplyDeltas` — deltas reassemble to the reply; usage preserved through the streaming path.
- `TestRun_Streaming_ProposeRoundStreamsNarrationThenProposal` — **the defining contract**: narration streams AND `Run` returns the `Proposal`, no `Reply`.
- `TestRun_Streaming_NoNarrationPropose_EmitsNoDeltas` — lazy-start friendly: nothing streamed.
- `TestRun_Streaming_MultiRound_StreamsTerminalReply` — both rounds take the stream path; sink holds exactly the terminal reply.
- `TestRun_NonStreamingLLM_WithSink_FallsBackToComplete` / `TestRun_StreamingLLM_NoSink_UsesComplete` — streaming activates only when a sink is set *and* the LLM supports it.
- `TestStreamTextDelta_ExtractsTextDeltasOnly` — extraction pinned against real wire JSON, incl. a tool-args (`input_json`) delta yielding `""` (the fake-LLM tests bypass the SDK decode, so this is verified directly, mirroring `TestToSDKMessages_*`).

## Verification

- `/simplify`: 4 angles. Applied — replaced the per-token double-`.AsAny()` re-parse with the flattened-field tap (efficiency + simplification) and closed its test-coverage gap; trimmed a duplicated doc invariant. Skipped (noted): unifying the streaming/non-streaming test fakes (would break the fallback test's premise + the path counters) and threading the sink through `Run` (ripples all callers; the per-turn `Agent` construction + Option mirror `WithMaxIterations`).

Part of #663 (response streaming). Follow-up PR2 wires the Slack streaming seam + turn path.

### DIFF
--- a/apps/slack/internal/agent/agent.go
+++ b/apps/slack/internal/agent/agent.go
@@ -294,7 +294,7 @@ func (a *Agent) Run(ctx context.Context, tc *TurnContext, history []Message, use
 
 	var usage Usage
 	for range a.maxIterations {
-		resp, err := a.complete(ctx, &Request{SystemStable: systemPreamble, SystemPerTurn: perTurn, Tools: tools, Messages: msgs})
+		resp, err := a.roundTrip(ctx, &Request{SystemStable: systemPreamble, SystemPerTurn: perTurn, Tools: tools, Messages: msgs})
 		if err != nil {
 			return Result{Usage: usage}, msgs, fmt.Errorf("agent: llm complete: %w", err)
 		}
@@ -348,10 +348,10 @@ func (a *Agent) Run(ctx context.Context, tc *TurnContext, history []Message, use
 	return Result{Reply: iterationCapMessage, Usage: usage}, msgs, nil
 }
 
-// complete runs one model round-trip, routing to the streaming path when a per-turn
+// roundTrip runs one model round-trip, routing to the streaming path when a per-turn
 // sink is set and the LLM implements [streamingLLM], else the plain [LLM.Complete].
 // Centralizing the choice here keeps [Run]'s loop identical for both paths.
-func (a *Agent) complete(ctx context.Context, req *Request) (Response, error) {
+func (a *Agent) roundTrip(ctx context.Context, req *Request) (Response, error) {
 	if a.streamSink != nil {
 		if s, ok := a.llm.(streamingLLM); ok {
 			return s.StreamComplete(ctx, req, a.streamSink)
--- a/apps/slack/internal/agent/agent.go
+++ b/apps/slack/internal/agent/agent.go
@@ -164,6 +164,18 @@ type LLM interface {
 	Complete(ctx context.Context, req *Request) (Response, error)
 }
 
+// streamingLLM is the optional streaming extension to [LLM]. When the configured
+// LLM implements it AND a per-turn stream sink is set ([WithStreamSink]), [Run]
+// forwards the assistant's text deltas to the sink as the model generates them,
+// instead of waiting for the whole round-trip. It's a separate interface (asserted
+// at the seam, like io.WriterTo) so every existing non-streaming [LLM] fake stays
+// valid — they simply take the [LLM.Complete] path. onText receives only non-empty
+// text deltas; the full [Response] (text, tool calls, usage) is still returned when
+// the round-trip ends, identical to [LLM.Complete].
+type streamingLLM interface {
+	StreamComplete(ctx context.Context, req *Request, onText func(delta string)) (Response, error)
+}
+
 // Backend is the port to qURL read operations, each scoped by the
 // implementation to what the caller (in [TurnContext]) may see in their channel.
 // Every method returns a model-readable summary string (the tool_result the LLM
@@ -213,6 +225,10 @@ type Agent struct {
 	llm           LLM
 	backend       Backend
 	maxIterations int
+	// streamSink, when non-nil and the LLM implements [streamingLLM], receives the
+	// assistant's text deltas as they stream. Per-turn: the handler constructs a
+	// fresh Agent for each turn, so the sink never outlives the turn it serves.
+	streamSink func(delta string)
 }
 
 // Option configures an [Agent].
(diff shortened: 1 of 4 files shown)


justin-layerv:
Thanks for the careful review. Addressed both non-affirmation items:

1. **Real-SDK usage accumulation** (the byte-identical-usage gap): closed in fc1ea4ec with `TestStreamAccumulate_MergesUsageAndContent`. It drives the SDK's `Accumulate` directly with a real `message_start` → text deltas → `message_delta` sequence and asserts `fromSDKMessage` merges input/cache tokens (from `message_start`) with the final output tokens (from `message_delta`) and reassembles the text + stop reason — no HTTP mock needed. The streaming path's logic is now unit-verified piece-by-piece (`streamTextDelta` against wire JSON + this).

2. **Partial-deltas-then-error semantics**: out of PR1's scope by design (no sink is wired, so it can't manifest here), but it's a real PR2 contract. Captured as a pinned acceptance criterion on the streaming tracking issue #663 — PR2 will decide it (finalize the partial via `chat.stopStream`, never a second post) and pin a test, the streaming analogue of this PR's propose-narration contract.

justin-layerv:
Thanks — addressed round 2 (594d7c1b):

1. **Pin the "every round's text" contract for an intermediate narrating round** (your only real ask): added `TestRun_Streaming_IntermediateNarrationThenReply_StreamsBoth`. It scripts a narrating `list_resources` round followed by a text-only reply round and asserts the sink holds both rounds' text concatenated (`narration + reply`) while `Result.Reply` stays the terminal round's text — locking exactly the behavior PR2's finalization must reason about, and guarding against a future refactor that streams only the terminal round.

2. **Discriminator literal nit**: resolved by removing the discriminator entirely — `streamTextDelta` now reads the flattened `Delta.Text` directly. It's non-empty only for a text delta (the SDK fills a different flattened field for input_json/stop/thinking, and non-delta events carry no text), and `TestStreamTextDelta`'s wire-JSON cases pin that — so there's no longer a literal to drift on an SDK rename.

3. **`complete()` near-collision nit**: renamed the round-trip router to `roundTrip()`.

4. **Partial-deltas-then-error** (forward-looking): tracked as a pinned acceptance criterion on the streaming epic #663 (finalize the partial via `chat.stopStream`, never a second post; pin a test) — PR2 will decide it explicitly, the streaming analogue of this PR's propose contract.

justin-layerv:
Thanks — all three doc points addressed in 116c2bf2 (doc-only, no logic change):

1. **Inverse sink/Reply divergence**: documented on `WithStreamSink` — `Result.Reply` can hold text the sink never saw (the iteration-cap message + empty-reply fallback are synthesized by `Run`, not streamed), so the PR2 finalizer must reconcile against `Result` rather than assume finalized == concatenated deltas in either direction. Good catch — that's the one that would've bitten PR2.

2. **Partial-then-error**: documented on `WithStreamSink` — deltas already delivered are not rolled back on a later stream error; the caller owns how a partially-streamed-then-failed turn finalizes. (Also pinned as a PR2 acceptance criterion on #663.)

3. **Pointer rationale nit**: corrected — it avoids the *second* copy of the union per call (`stream.Current` already copied it once into the loop variable), not per-token copying wholesale.

4. **Per-round type assertion**: left as-is per your note (negligible; hoisting would muddy the centralized routing in `roundTrip`).

justin-layerv:
Thanks — addressed in 6bffdf28:

1. **Misleading error context**: `Run` now wraps a round-trip failure as `agent: llm round-trip: %w` (was `llm complete`), so a streaming failure no longer reads the self-contradictory `llm complete: anthropic messages stream: …`.

2. **Per-iteration type assertion**: hoisted — `New` detects the `streamingLLM` capability once into `Agent.streamLLM`, so `roundTrip` is now a single nil check per round.

3. **streamTextDelta SDK coupling / thinking**: noted as a forward item. `buildParams` disables thinking and `TestStreamTextDelta` pins the must-be-empty cases, so it's safe today; the `streamTextDelta` doc flags that each delta type fills a different flattened field, which is the spot to re-check if a later change enables thinking or new content blocks.

4. **Mid-stream-error contract test**: added `TestRun_Streaming_ErrorAfterPartialDeltas_RetainsDeltas` — a stream that emits text then fails surfaces the error from `Run` while the sink keeps the partial deltas, locking the documented no-rollback contract (PR2 owns the finalization built on it).

justin-layerv:
Thanks for the thorough pass. Dispositions on the three non-blocking notes (all "fine to leave", no code change):

1. **`onText == nil` branch unreachable through `Run`**: intentional port-level contract — `streamingLLM.StreamComplete` documents that a nil `onText` makes it "just a streaming `Complete`", so a future direct caller can use it that way even though `roundTrip` always passes a non-nil sink. Exercising the nil form needs an HTTP/SSE mock of the real Anthropic endpoint (the fake-LLM loop tests don't reach `anthropicLLM.StreamComplete`), which isn't worth standing up for PR1 — leaving the defensiveness in.

2. **`message_delta` stop-reason shape**: agreed it's the most likely SDK-shape foot-gun; `TestStreamTextDelta` keeps an explicit `message_delta` -> `""` row in the table precisely so an SDK bump that flattened it differently would fail loudly rather than leak. Will keep that case if the union shape changes.

3. **PR2 contract in three places** (PR body / `WithStreamSink` doc / test names): noted — PR2 will treat the `WithStreamSink` doc as the source of truth and reconcile the others when it lands.

Appreciate the de-risking framing for PR2's finalization.

justin-layerv:
Thanks — round 6 is all forward-looking / non-regression, no PR1 changes:

1. **`streamTextDelta` SDK coupling**: agreed — `TestStreamTextDelta` is the mitigation and stays green-or-explain on SDK bumps.

2. **Synchronous sink back-pressure** (good catch): a slow sink stalling `stream.Next()` is exactly why PR2 coalesces rather than `chat.appendStream`-per-delta. Captured the back-pressure rationale explicitly on #663 (buffer + flush on size/time, decouple from the stream-read goroutine if a Slack round-trip would otherwise back-pressure the model stream) — not just the Tier-4 rate limit.

3. **Usage dropped on stream error**: agreed it's symmetric with `Complete` (returns no usage on error too), so not a regression; the under-report is bounded to a failed turn. Noted on #663 next to the partial-error finalization.

Appreciate the thorough review across the rounds.

Issue #701, opened 2026-06-11
slack agent: audit protect-connector actions at the modal-submit enforcement point


justin-layerv:
## Gap

The App Home review surface (#664) records executed agent mutations via `recordAgentAudit`, hooked at the confirm-execute site (`executeAgentAction` in `handler_agent_confirm.go`). Five of the six mutation kinds — get, revoke, set-alias, unset-alias, protect-url — flow through `executeAgentAction` and are audited.

**protect-connector does not.** An agent-proposed `protect_connector` routes through `openAgentConnectorModal` → the modal submit (`processTunnelInstall`), which `return`s from `processAgentConfirm` *before* the `recordAgentAudit` call. So an agent-confirmed protect-connector never lands in the actor's App Home review list.

The `auditTargetFor` and `agentHomeActionLabel` switches already carry defensive `protect_connector` arms (currently unreachable for real audit data).

## Fix

Record an `AuditEntry` at the real protect-connector enforcement point (`processTunnelInstall`, the modal submit) when the install was **agent-initiated**, distinguishing it from a slash-command `/qurl-admin protect-connector` (which is not an agent action and must NOT be audited). The agent-vs-slash provenance is available via the modal's `private_metadata` (the agent flow seeds it with the pending-action context).

## Acceptance

- An agent-confirmed protect-connector appears in the approver's App Home review list (kind = protect_connector, target = the connector identity).
- A slash-command-initiated protect-connector does **not** appear (not an agent action).
- Per-viewer scope preserved (keyed on the modal submitter).

Issue #664, opened 2026-06-09
Slack agent: App Home review/audit surface for agent actions


justin-layerv:
From the Slack agent-design audit (in-session task #9). The doc calls an audit trail / review surface foundational: users need a way to see what the agent did on their behalf, especially for async/bulk actions.

Scope: publish an App Home view listing recent agent actions (proposals shown, mutations executed, by whom/when, with links) backed by a persisted action log. Pairs naturally with mutation execution (PR4/PR5) — that's when there are actions worth auditing. Requires `app_home` + a Home-tab publish path.

PR #702, opened 2026-06-11
feat(slack): add App Home review surface for agent actions (#664)

justin-layerv:
## What

The **App Home review surface** (#664): on `app_home_opened`, the agent publishes the viewer's **own** recent confirmed actions to the Slack App Home tab — a personal "what have I done through the Secure Access Agent" log. Two commits: the audit **store + write hook** (capture), then the **view + seam** (surface).

## Design

- **Per-viewer scope — forced by channel-scope, not a preference.** A workspace-wide view would aggregate actions across channels a viewer isn't a member of (a back-door channel-scope leak, in the one codebase that defends channel scope everywhere). So the store keys each entry `audit#<user_id>#<unix_nanos>` and `ListAuditEntries` does a per-user `begins_with` query — a viewer only ever sees actions they performed. (Workspace/compliance audit is a heavier, separate feature — not this.)
- **Store** (`slackdata/agentaudit.go`): `AuditEntry` + `PutAuditEntry` (point write) / `ListAuditEntries` (DDB `Query`, `begins_with`, `ScanIndexForward=false` → newest-first, `Limit`, read-time TTL filter, corrupt-payload skip). New `audit#` prefix on the existing single-table store; TTL'd (14d default).
- **Write hook**: `recordAgentAudit` in `processAgentConfirm`, gated on `res.attributed` (only an actually-executed mutation core), best-effort (a nil store / write error never affects the click), and run **after** the card swap so the `PutItem` adds no card-swap latency. Keyed to the **approver** who ran it.
- **View**: `handleAppHomeOpened` (home-tab + non-empty-user gated, async off `baseCtx` behind the per-workspace opt-in) → `ListAuditEntries` → `buildAgentHomeView` → `views.publish` (new `AppHomePublishFunc` seam, mirroring the chat/assistant/reactions posters: per-team token + Grid fallback).
- **Honest labels**: the action label is **neutral** (the action attempted, not a success claim), so a failed action doesn't read as a success. Each entry shows action + target + channel + time + reason; the stored `Outcome` (the formatted public card text) is **not** echoed in the summary — escaping its intentional backticks for safety would render it degraded, and it repeats the target shown cleanly above — so a clean per-result success/failure line is the **#704** follow-up.

## Security — the Home view is a public-echo surface

`Target` / `Outcome` / `Reason` are partly LLM-distilled / core-generated, so they're escaped with the **same primitives the confirm card uses** (`escapeMrkdwnCode` for the backticked target, `escapeMrkdwnText` for free text). A stored alias named `*ignore*` or one carrying bidi/zero-width chars renders inert — pinned by `TestAgentHomeEntryText_EscapesInjectedEcho`.

## Tests

- Store: per-viewer scope (with a `U1`/`U12` prefix-collision decoy — the `#` delimiter is the security boundary), newest-first, limit, read-time TTL, corrupt-skip, query-error, validation.
- View: empty-state, neutral labels + target + Outcome + channel mention; **echo escaping**.
- Handler: home-tab publishes the viewer's actions; a non-home tab / empty user don't; the write-hook round-trip; `memAgentDDB.Query` implemented.

## Verification

- `/simplify`: 4 angles. Applied — render the captured `Outcome` + neutralize success-asserting labels (a failed action no longer reads as a success); defer the audit write off the card-swap path; pre-size the view slice. Skipped (noted): the twin `Query` test fakes (cross-package, can't share cheaply) and the `slackChannelMention` near-miss (its install-specific fallback is wrong here).

## Deferred (labeled issue)

- **#701** — protect-connector routes through the modal (not `executeAgentAction`), so it isn't audited yet; tracked with acceptance criteria.

**Dark** until enablement: the manifest needs the **App Home tab** feature + the **`app_home_opened`** event subscription (both fold into the sandbox-enablement PR). Until then `app_home_opened` never arrives and the `views.publish` seam no-ops.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -173,6 +173,11 @@ func run() error {
 	// lookup); inert until the "Agents & AI Apps" manifest toggle + assistant:write scope
 	// are set (the events don't arrive, and the calls would no-op, until then).
 	agentAssistantThreads := newSlackAssistantThreadsPortWithTokenLookup(workspaceTokenLookup, userAgent, slackAssistantSetTitleURL, slackAssistantSetSuggestedPromptsURL, slackAssistantSetStatusURL, nil)
+	// views.publish seam for the App Home review surface (a user's own recent confirmed
+	// actions). Always wired (same token lookup); inert until the manifest's App Home tab
+	// feature + app_home_opened subscription are set — the event doesn't arrive, and the
+	// view would no-op, until then.
+	agentAppHomePublish := newSlackAppHomePublishFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackViewsPublishURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
@@ -247,6 +252,7 @@ func run() error {
 		ResolveChannelName:          agentResolveChannelName,
 		ChannelMembership:           agentChannelMembership,
 		AssistantThreads:            agentAssistantThreads,
+		AppHomePublish:              agentAppHomePublish,
 	})
 
 	// Alias reads and writes must go through the same slackdata facade so
--- a/apps/slack/cmd/slack_webapi.go
+++ b/apps/slack/cmd/slack_webapi.go
@@ -402,6 +402,41 @@ func newSlackPostMessageBlocksFuncWithTokenLookup(lookup slackBotTokenLookup, us
 	}
 }
 
+const slackViewsPublishURL = "https://slack.com/api/views.publish"
+
+// slackViewsPublishResponseError preserves the views.publish error shape via the shared
+// mapper (no benign codes).
+func slackViewsPublishResponseError(statusCode int, header http.Header, raw []byte) error {
+	return slackWebAPIResponseError("views.publish", nil, statusCode, header, raw)
+}
+
+// slackHomeView is the {"type":"home", "blocks":[...]} view object views.publish wraps
+// the App Home content in.
+type slackHomeView struct {
+	Type   string `json:"type"`
+	Blocks []any  `json:"blocks"`
+}
+
+// newSlackAppHomePublishFuncWithTokenLookup builds the [internal.AppHomePublishFunc]
+// seam: a views.publish that sets a user's App Home tab to the agent's review surface
+// (their own recent confirmed actions). It shares the poster (token lookup + Grid
+// fallback + response parsing) with the chat seams — only the JSON body differs. No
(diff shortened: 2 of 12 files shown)


justin-layerv:
Thanks — addressed in the latest push:

1. **Channel mention now guarded**: `agentHomeEntryText` validates `e.Channel` against `slackChannelIDPattern` and omits the mention segment for a malformed id, so the surface is uniformly defended like the escaped `Target`/`Outcome`/`Reason` around it. (Used the pattern directly rather than `slackChannelMention`, whose install-specific fallback copy is wrong here.)

2. **Pre-size fix**: `buildAgentHomeView` now seeds `4+len(entries)` so the empty-state path's 4th block doesn't force a realloc.

3. **Limit-before-filter**: leaving as-is per your note — with the 14d TTL on recent actions a short window of expired/corrupt items in the newest-20 is unlikely, and a "fill to N" guarantee would need over-fetch + post-filter trim. Documented on `ListAuditEntries`. Happy to file a follow-up if you'd prefer the guarantee.

justin-layerv:
Thanks — addressed in the latest push:

1. **Outcome echo (good catch on the degraded mrkdwn)**: agreed the escaping is the correct security call and shouldn't be relaxed, so the cleanest fix is to **not echo the formatted `Outcome` in the summary view** — its intentional backticks render as the `ˊ` substitute when escaped, and it largely repeats the target shown cleanly above. `Outcome` stays captured in the record; a clean per-result line (with success/failure, de-degraded) needs the action core to hand back a structured result, tracked as **#704**. The entry now shows label + target + channel + time + reason.

2. **Limit-before-filter**: documented on `ListAuditEntries` now (Limit caps the SCANNED newest-N before the read-time TTL/decode filter, so a window heavy with expired/corrupt items can return fewer than `limit` — fine for a recent-actions view, not a fill-to-N guarantee).

3. **Publish on every `app_home_opened`**: leaving as-is per your note — bounded by user action + the async pool, and `views.publish` is idempotent.

justin-layerv:
Thanks — fixed the one substantive item:

1. **PR description / Outcome mismatch** (good catch): the "Honest labels" section was stale — it described the round-1 behavior where the Outcome was rendered. Updated the PR body to match what ships: honesty comes from the **neutral labels** alone (a failed action reads as "Revoke … attempted", never a false "Revoked"), the summary shows action + target + channel + time + reason, and the formatted `Outcome` is **not** echoed (clean per-result success/failure is #704). The code + comments were already accurate; only the description lagged.

2. **Outcome write-only until #704**: agreed — deliberate seam for the follow-up, confirmed non-sensitive (the get credential is `ephemeralText`, the stored Outcome is the neutral card constant). Leaving it captured for #704.

3. **app_home_opened debounce / Limit-before-filter / twin Query fakes**: all accepted as noted (bounded by user action + async pool; documented; cross-package and not cheaply shareable).

Appreciate the three thorough rounds.

Issue #706, opened 2026-06-11
slack agent: stream the reply in public channels (app_mention), not just the pane


justin-layerv:
## Background

The streaming-delivery PR wires native reply streaming (`chat.startStream`/`appendStream`/`stopStream`) only for **pane (DM / `message.im`)** turns — `newAgentReplyStreamer` returns nil for anything but an im turn, mirroring `setAgentThinkingStatus`. An `app_mention` turn (channel) still posts its reply as one message.

`chat.startStream` accepts `recipient_team_id` + `recipient_user_id` for streaming into a channel, so a channel reply could stream too.

## Ask

Extend `newAgentReplyStreamer` to engage for `app_mention` (channel) turns: stream into the channel thread with the recipient params, and verify the channel-streaming UX + the finalization contracts (no-double-post, proposal-card-still-posts, partial-error) hold the same as the pane path. Keep the pane path unchanged.

Out of scope unless trivial: a slow-sink back-pressure decoupling (the sink is currently called synchronously from the model-stream goroutine; coalescing bounds it, but a channel firehose may warrant draining off-goroutine).

justin-layerv:
Note from #707 (pane streaming): `slackAgentStreamPort.StartStream` currently sends `recipient_team_id = teamID` (the caller's workspace). That's correct for same-workspace pane DMs, but when this is extended to channels / Enterprise Grid where the recipient can be in a **different** team, `recipient_team_id` must be the recipient's team, not the owning workspace's `teamID`. There's a code comment at the `StartStream` marshal pointing here.

PR #707, opened 2026-06-11
feat(slack): stream the agent reply into the assistant pane natively

justin-layerv:
## Summary

Streams the conversation agent's final reply into the Slack **assistant pane** token-by-token,
so a user reading a long answer (a summary, a draft, an analysis) starts reading immediately
instead of waiting for the whole turn to land as one message. Builds directly on the streaming
**core** (#700, merged) — the `WithStreamSink` seam — and wires it to Slack's native streaming
methods (`chat.startStream` / `chat.appendStream` / `chat.stopStream`).

Advances **#663** (the *stream long-form responses* bullet). The other two bullets of #663 —
progressive multi-step status and collapsible plan blocks — are out of scope here and remain open.

Kept **dark/inert** like the rest of the agent surface: the seam is nil-safe and only engages
once the turn path has a non-nil `AgentStream` port **and** the turn is a pane (DM) turn. No
behavior changes until enablement.

## Scope

- [x] `apps/slack/`

## Changes

**Part 1 — the Slack seam (commit 1):**
- `slackWebAPIPoster` refactor: `postOnce` returns the response body, a shared `gridPost` holds
  the Enterprise-Grid org-token fallback, and `post` (unchanged signature — every existing seam
  is unaffected) discards the body. `chat.startStream` reads the stream `ts` straight off
  `gridPost`'s returned body.
- `AgentStreamPort` Config seam + `slackAgentStreamPort` over `chat.startStream` (parses the
  stream `ts`) / `appendStream` / `stopStream`, sharing the per-team token lookup, the Grid
  fallback, and one `http.Client` across the three posters. Wired in `cmd/main.go`.

**Part 2 — the turn-path wiring + finalization (commit 2):**
- `agentReplyStreamer` (`handler_agent_stream.go`): the `WithStreamSink` callback for a pane
  turn. Lazily opens `chat.startStream` on the first non-empty delta, coalesces deltas
  (`agentStreamFlushBytes = 48`) to bound `appendStream` call rate (Slack streaming is Tier-4),
  and finalizes exactly once.
- `processAgentEvent` finalization — the **no-double-post invariant**:
  - a streamed reply is delivered *by the stream* (the caller skips the normal post);
  - a **proposal** still posts its confirm card (the streamed text was only the agent's narration);
  - an **error** with a live stream finalizes the partial instead of double-posting an error over it;
  - a **synthesized** reply (the iteration-cap / empty-fallback that the deltas never carried) is
    reconciled — appended once, only if the stream didn't already carry it — before stop.
- Finalization runs on a **fresh delivery budget off `h.baseCtx`** (not the possibly-spent turn
  ctx), and a `start`/`append` failure mid-flight marks the stream **broken** so the caller posts
  the full reply rather than leaving a truncated bubble. Both mirror the sibling delivery steps
  (`saveAgentHistory` / `postAgentReply`).
- A `startStream` failure, or a turn that streams no deltas, falls back to the normal posted reply.
- The "thinking…" pane status auto-clears on the streamed message (same thread), consistent with
  the non-streaming path.

**Pane-only**, mirroring `setAgentThinkingStatus`: native streaming targets the assistant pane,
so an `app_mention` in a channel keeps the non-streaming post path. Channel streaming via
`recipient_*` is a follow-up — **#706**.

## Test Plan

- [x] `make check` (fmt / vet / lint / test-race) green locally for `apps/slack` — including
  `golangci-lint` at the CI-pinned **v2.10.1**.
- [x] New tests added: `handler_agent_stream_test.go` pins every finalization branch — no-deltas,
  coalesced normal reply, synthetic-reply reconcile (appended, not doubled), proposal (stops but
  caller posts the card), error-after-deltas, error-before-deltas, `startStream` failure fallback,
  and a **mid-stream `appendStream` failure** falling back to the posted reply.
- [x] For Slack-impacting changes: branch is current with `main` and will be kept green.
- [ ] Manual testing: deferred to sandbox enablement (the seam is dark until `AgentStream` is wired
  and the `assistant:write` scope is live — already present in the sandbox manifest).

## Prod / Enablement Notes

- Native streaming requires the **`assistant:write`** scope, already added to the sandbox Slack
  manifest (qurl-integrations-infra #1018, merged). No new infra change is required to land this PR.
- The port stays unwired in any environment where `AgentStream` is nil, so this is safe to merge
  ahead of broader enablement.

## Related Issues

Advances #663. Follow-up: #706 (channel streaming via `recipient_*`).

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -178,6 +178,11 @@ func run() error {
 	// feature + app_home_opened subscription are set — the event doesn't arrive, and the
 	// view would no-op, until then.
 	agentAppHomePublish := newSlackAppHomePublishFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackViewsPublishURL, nil)
+	// chat.startStream/appendStream/stopStream seam for native AI-app reply streaming in
+	// the assistant pane. Always wired (same token lookup + Grid fallback); inert until the
+	// assistant:write scope + the pane (Agents & AI Apps) are enabled, and only engaged for
+	// a pane turn — otherwise the agent posts the reply normally.
+	agentStream := newSlackAgentStreamPortWithTokenLookup(workspaceTokenLookup, userAgent, slackChatStartStreamURL, slackChatAppendStreamURL, slackChatStopStreamURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
@@ -253,6 +258,7 @@ func run() error {
 		ChannelMembership:           agentChannelMembership,
 		AssistantThreads:            agentAssistantThreads,
 		AppHomePublish:              agentAppHomePublish,
+		AgentStream:                 agentStream,
 	})
 
 	// Alias reads and writes must go through the same slackdata facade so
--- a/apps/slack/cmd/slack_webapi.go
+++ b/apps/slack/cmd/slack_webapi.go
@@ -293,11 +293,13 @@ func newSlackWebAPIPoster(lookup slackBotTokenLookup, userAgent, url, op string,
 }
 
 // postOnce resolves one owner's bot token (workspace team or, on the Grid fallback,
-// the enterprise org) and POSTs the already-marshaled body.
-func (p *slackWebAPIPoster) postOnce(ctx context.Context, ownerID string, body []byte) error {
+// the enterprise org) and POSTs the already-marshaled body, returning the (size-bounded)
+// response body alongside the parsed error. Most seams ignore the body (via post); only
+// chat.startStream reads it, for the stream ts (see slackAgentStreamPort.StartStream).
+func (p *slackWebAPIPoster) postOnce(ctx context.Context, ownerID string, body []byte) ([]byte, error) {
 	token, err := p.lookup(ctx, ownerID)
 	if err != nil {
-		return fmt.Errorf("%s token lookup: %w", p.op, err)
+		return nil, fmt.Errorf("%s token lookup: %w", p.op, err)
 	}
 	token = strings.TrimSpace(token)
 	if token == "" {
@@ -305,11 +307,11 @@ func (p *slackWebAPIPoster) postOnce(ctx context.Context, ownerID string, body [
 		// trigger the Grid fallback. That's fine: the real workspaceTokenLookup returns
 		// ErrSlackBotTokenNotConfigured (which does fall back), never an empty-but-nil
 		// token. A future lookup returning ("", nil) on a miss would need the sentinel.
-		return fmt.Errorf("%s token lookup: empty token", p.op)
+		return nil, fmt.Errorf("%s token lookup: empty token", p.op)
 	}
 	req, err := http.NewRequestWithContext(ctx, http.MethodPost, p.url, bytes.NewReader(body))
 	if err != nil {
-		return fmt.Errorf("%s request build: %w", p.op, err)
+		return nil, fmt.Errorf("%s request build: %w", p.op, err)
 	}
 	req.Header.Set("Authorization", "Bearer "+token)
 	req.Header.Set("Content-Type", "application/json")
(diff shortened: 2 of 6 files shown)


justin-layerv:
Thanks for the thorough pass (both rounds). Pushed `97290016` addressing everything; dispositions below.

**Fixed in code:**

1. **`finalizeError` broken-symmetry (medium).** Right call — it was asymmetric. `finalizeError` now returns `!s.broken`, so a turn that errors *after* the stream already broke falls back to a posted error instead of suppressing it (the user no longer reads a truncated partial as the final answer). Mirrors `finalizeReply`'s broken path. Added `TestAgentStreamer_ErrorAfterBroken_CallerPostsError` and updated both caller comments in `handler_agent.go`.

2. **`StartStream` collapsed two failure modes (nit).** Split: a JSON-unmarshal failure now returns `fmt.Errorf("chat.startStream: decode response: %w", err)` (keeps the parse error), and only a present-but-empty `ts` returns the "no stream ts" sentinel.

3. **`Contains` reconcile coupling (low).** Kept `Contains` deliberately and documented why: `result.Reply` is `TrimSpace`'d while `streamed` is not, so a stricter `HasSuffix` would false-negative on trailing whitespace and **double-post**; `Contains`' worst case is the benign reverse (skipping a reply the user already saw stream). Comment added at the reconcile site.

**Investigated + documented + tracked:**

4. **Dialect mismatch `markdown_text` (stream) vs mrkdwn `text` (fallback) (medium).** Confirmed the agent is constrained to *"light Slack-compatible markdown — short bullets, backticks"* (`agent/prompt.go`), a subset where the two render compatibly (backticks identical; `- ` bullets list-render under `markdown_text`). The cr's worst cases (`<url|label>`, `*bold*`) require the agent to violate its prompt. Added a dialect note to the streamer doc and filed **#708** to verify rendering parity in the live client at enablement (it's dark now, so can't eyeball today) and to keep the prompt constraint aligned with the dialect.

5. **Double-message on a mid-stream append failure (19:39 nit-1).** Agreed it's a cosmetic wart once enabled, but unavoidable in-stream (the stream is already broken, so the remainder can't be appended). Filed **#709** to decide overwrite/delete/marker vs accept at enablement. The fallback degrades safely today (the full answer is always delivered).

**Acknowledged, no change:**

6. **`RecipientTeamID` for Grid divergence (nit).** Correct for the same-workspace pane DMs this serves today; added a code note pointing the channel/Grid case to **#706** (where recipient and owner team can diverge).
7. **Synchronous I/O in `onDelta` (low/perf).** Inherent to the `WithStreamSink` callback contract (the loop calls the sink synchronously); the 48-byte coalesce bounds the call rate. Acceptable for native streaming, as you noted.
8. **Coverage of the synthetic-skip-when-broken guard (19:39 nit-4).** Now pinned by `TestAgentStreamer_AppendFailureAtFinalize_FallsBackToPost` (a sub-threshold reply whose only append — the finalize tail — fails).

Build/vet/`-race`/`golangci-lint` v2.10.1 all green locally.

justin-layerv:
Round 2 — thanks, pushed two refinements:

- **`streamed` recorded before append confirmation.** Fixed: `flush` now writes `streamed` only *after* a successful `AppendStream`, so the buffer never claims text that didn't reach Slack. Removes the latent trap (no future `!broken` read can see phantom content). No behavior change today (the reconcile is `!broken`-gated), just honest semantics.
- **Synchronous Slack I/O on the loop goroutine — explicit answer.** Deliberate for v1. The `WithStreamSink` contract is in-order/single-goroutine, so the POSTs are synchronous by construction; the 48-byte coalesce bounds the call rate, and the backpressure only *slows the reveal* — the upstream LLM stream buffers, nothing is dropped. A buffered async hand-off (one ordered worker draining a channel) is the optimization if the interleave proves to matter once enabled, but I'd rather not add concurrency to the finalization state machine while the feature is dark and unmeasured. Documented the rationale on `onDelta` and added a backpressure-measurement task to #708.

Already-tracked items, no code change: the broken-stream dup-content UX (#709), `recipient_team_id` Grid divergence (#706), the `Contains` heuristic (deliberate, documented), and UTF-8-across-deltas / dialect parity (#708). Build/vet/`-race`/lint v2.10.1 green locally.

justin-layerv:
Round 3 — thanks, and thanks for the explicit correctness spot-checks (the `broken ⟹ pending empty` invariant, `finalizeError` symmetry, `baseCtx` finalize, and the `Contains`-vs-`HasSuffix` rationale all match the intent). No code change this round — the three observations are enablement-time follow-ups, now tracked:

1. **429 mid-reply → truncated partial + full repost.** The concrete trigger you named is captured on #709, linked to the coalesce-threshold tuning on #708 (a larger window reduces both the 429 risk and this fallback's frequency).
2. **Proposal narration + confirm-disabled → two messages.** Reachable only if streaming is enabled while confirm is off (sandbox enablement runs confirm ON, so it doesn't manifest there). Added to the #708 flag-matrix validation.
3. **Streamed bubble can carry more than `Result.Reply`.** The documented stream-only-has-more divergence; added to the #708 parity check.

`make check` (fmt/vet/lint/test-race) is green locally at golangci-lint v2.10.1. Holding here for merge — not merging (the author merges).

justin-layerv:
Force-push `efd09745` → `f97e5e1f` is **doc-only**: fixed a stale comment in `postOnce` that still referenced `postReadingBody` (the no-op wrapper removed during /simplify — `chat.startStream` now reads the body via `gridPost` directly). No code change; same verified diff content otherwise.

PR #711, opened 2026-06-12
feat(slack): answer follow-up replies in agent channel threads (flag-gated)

justin-layerv:
## Summary

Adds natural **thread continuity** to the channel agent: once the agent has joined a channel thread (via an `@mention`), a plain reply in that same thread continues the conversation — no re-`@mention` per message. Reported pain: "if I reply to the agent in a thread it started, it doesn't respond unless I tag it again."

**Dark by default** (`QURL_AGENT_CHANNEL_FOLLOWUPS`, fail-safe off). It is **not** the channel-message firehose — see the bound below.

## Scope

- [x] `apps/slack/`

## Changes

- **`shouldDispatchAgentEvent`** now takes a `channelFollowupsEnabled` flag. A channel message is admitted only when the flag is on **and** it's a thread **reply**; a top-level channel message is never admitted. `@mentions`/DMs are unchanged.
- **`agentChannelFollowupDropped`** (new) is the own-thread gate, run in `processAgentEvent` **before dedupe/ack**: a channel thread reply proceeds only if the agent already has a saved transcript for that thread (it joined via an `@mention`). Otherwise it's dropped — consuming no dedupe marker, no `👀`, and on a lookup error staying **silent** rather than posting into unrelated chatter. `@mentions`/DMs skip the gate.
- **`Config.AgentChannelFollowups`** + `agentChannelFollowupsEnabled()` (gated on `agentEnabled`), wired from `QURL_AGENT_CHANNEL_FOLLOWUPS` in `cmd/main.go` (fail-safe off).

### Why it's bounded (the data-handling note)

Slack has no "only messages in my threads" event, so enabling this requires the manifest to subscribe `message.channels`/`message.groups` with `channels:history`/`groups:history` — the bot then **receives every message in channels it's a member of**, and only acts on its own threads (dropping the rest on arrival, before any LLM/storage). That's a real data-handling expansion, so the manifest change is a **separate infra PR** and enabling it needs the data-handling review + a per-workspace re-OAuth. This PR is the (dark) code only.

One behavior note: in an open agent thread it answers **any** member's reply, not just the original asker (the thread is already visible to the channel). Mutations still pass the existing confirm-card + asker-only authz, so that boundary is unchanged — only who can continue the *conversation*.

## Test Plan

- [x] `make check` (fmt / vet / lint v2.10.1 / test-race) green for `apps/slack`.
- [x] `TestShouldDispatchAgentEvent` extended: thread reply admitted only with the flag on; top-level channel message never admitted; `@mention`/DM unaffected by the flag.
- [x] `TestAgentChannelFollowupDropped` (new) pins the own-thread gate: joined thread continues; unknown thread / other channel dropped; `@mention` and DM never dropped.
- [ ] Manual: deferred to enablement (dark until the manifest + flag are on).

## Related Issues

Conversational-UX follow-up to the agent rollout (#663-era). The enabling manifest change (events + `*:history` scopes) ships as a separate `qurl-integrations-infra` PR, gated on the data-handling review.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -185,6 +185,7 @@ func run() error {
 	agentStream := newSlackAgentStreamPortWithTokenLookup(workspaceTokenLookup, userAgent, slackChatStartStreamURL, slackChatAppendStreamURL, slackChatStopStreamURL, nil)
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
+	agentChannelFollowups := readAgentChannelFollowups()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
 	// true at GA (every workspace on unless it explicitly opted out). Fail-safe to
 	// false. The per-workspace flag itself lives in workspace_mappings (AdminStore).
@@ -250,6 +251,7 @@ func run() error {
 		AgentDisabled:               agentDisabled,
 		PostMessageBlocks:           postMessageBlocks,
 		AgentConfirmEnabled:         agentConfirmEnabled,
+		AgentChannelFollowups:       agentChannelFollowups,
 		AgentDefaultEnabled:         agentDefaultEnabled,
 		AgentMaxTurnsPerUserPerHour: agentMaxTurnsPerUser,
 		AgentMaxTurnsPerTeamPerHour: agentMaxTurnsPerTeam,
@@ -1072,6 +1074,16 @@ func readAgentKillSwitch() bool {
 // surface is live and PostMessageBlocks is wired (Handler.agentConfirmEnabled).
 func readAgentConfirmEnabled() bool {
 	return readBoolEnvFailSafe("QURL_AGENT_CONFIRM_ENABLED", false, false)
+}
+
+// readAgentChannelFollowups reads QURL_AGENT_CHANNEL_FOLLOWUPS — the flag that lets
+// the agent answer non-@mention thread replies in channel threads it already joined
+// (Handler.agentChannelFollowupsEnabled). Absent → off. FAILS SAFE to off: enabling
+// it means the manifest subscribes message.channels/groups, so the bot then receives
+// every message in channels it's a member of — a data-handling expansion that a typo
+// must never turn on. Only takes effect once the read-only surface is live.
+func readAgentChannelFollowups() bool {
+	return readBoolEnvFailSafe("QURL_AGENT_CHANNEL_FOLLOWUPS", false, false)
 }
 
 // readAgentDefaultEnabled reads QURL_AGENT_DEFAULT_ENABLED — the per-workspace
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -357,6 +357,15 @@ type Config struct {
 	// ship enabled while the confirm flow stays staged (dark → beta).
 	AgentConfirmEnabled bool
 
+	// AgentChannelFollowups gates whether the agent answers non-@mention thread
+	// replies in channel threads it already joined (a follow-up without a re-@mention;
+	// see shouldDispatchAgentEvent). False (the default) keeps channels @mention-per-
+	// turn. True requires the manifest to subscribe message.channels/groups with
+	// channels:history/groups:history — so the bot then RECEIVES every message in
+	// channels it's a member of (and only acts on its own threads). That's a
+	// data-handling expansion: enable only after the review + a workspace re-OAuth.
+	AgentChannelFollowups bool
+
 	// AgentDefaultEnabled is the per-workspace conversation-mode default for a
 	// workspace that hasn't set the toggle (`/qurl-admin agent on|off`, stored in
 	// workspace_mappings via AdminStore). False during the staged per-workspace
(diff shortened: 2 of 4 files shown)


justin-layerv:
Round 1 — thanks; dispositions below (+ ran /simplify, clean on all four angles).

**Fixed now:**
- **#3 (test gap, fail-closed branch).** Added: `TestAgentChannelFollowupDropped` now injects a `LoadConversation` error (`memAgentDDB.getErr`) and asserts the follow-up is dropped — pinning the "lookup error → stay silent, don't post into unrelated chatter" path the PR calls out.

**Deferred with a tracked, labeled issue (qurl-integrations #712):**
- **#1 (firehose pool/RCU at scale)** + **#2 (double `LoadConversation`).** These are real but inherently enablement-time + need load data: the own-thread filter can't run synchronously (the webhook ack path has no store access), and the obvious pre-pool in-memory cache has a cross-task correctness gap (a thread joined by task A can have a reply delivered to task B, whose local cache lacks it → it would drop a *valid* reply). The robust fix (isolate firehose admission from the main turn pool) should be load-tested at enablement, not built speculatively on dark code. Filed **#712** (priority: medium, to land before broad enablement), added a code reference at the gate, and the monitoring step is in the enablement PR (#1040). This matches your "address or consciously accept with monitoring before enablement" framing.
- **#2** is folded into #712 (the read-consolidation is part of the same pre-pool work; minor + low-volume on its own — only the accepted-follow-up path, not the firehose).

**Documented:**
- **#4 (trimmed/expired transcript → silent drop of a real follow-up).** Fail-closed and acceptable; adding it to the enablement notes in #1040 so it isn't later reported as a bug (re-`@mention` resumes the thread).

Build/vet/`-race`/lint v2.10.1 green; dark by default.

justin-layerv:
Round 2 — thanks; the two actionable items addressed:

- **Note 1 (corrupt transcript → silent drop).** Good catch — `loadAgentHistory` reports a corrupt/undecodable blob as no history, so `len(history)==0` also covers a thread the agent *joined* but whose blob can't be read. Reworded the doc comment from "no saved transcript" to "no **decodable** transcript" (spelling out the empty/undecodable case explicitly, distinct from the `err != nil` DDB-down path), and added a test (`SaveConversation` a corrupt blob → the follow-up drops) to lock that intent in.
- **Notes 2/3 (firehose read + double read).** Tracked in #712, gated before broad enablement. Agreed on the sequencing dependency — #712 should land before the manifest/flag enablement (#1040); that's now an explicit acceptance criterion on #712 and a monitoring step in #1040's runbook. Also noted your point that the read only fires for workspaces already opted into conversation mode (the gate runs after `workspaceAgentEnabled`), which softens the interim.
- **Note 4 ("answers any member's reply").** On the record as a deliberate semantic widening; mutation authz (confirm card + asker-only) unchanged.

Build/vet/`-race`/lint green; dark by default.

justin-layerv:
Round 3 — all three notes addressed:
- **Flattened** the nested conditional in `shouldDispatchAgentEvent` to a single `&&` (nit 2).
- **Trimmed** the `agentChannelFollowupDropped` doc comment: the #712 scaling roadmap moved out of the function contract, leaving a one-line pointer (nit 3).
- **Double transcript read (nit 1)** stays deferred to #712, now framed explicitly as an *enablement prerequisite* (not just an optimization) — `enhancement,go,pr-polish-followup,priority: medium`, referenced from the code + the manifest-enablement PR, with a note added to the issue.

`go build/vet/test -race` green, golangci-lint v2.10.1 clean, `/simplify` clean (4 lenses). Head `447db465`.

justin-layerv:
Round 4 — addressed:
- **#1 thread_broadcast** — added a comment at the subtype guard noting it excludes `thread_broadcast` (a thread reply the user also broadcasts to the channel) as an accepted v1 bound, and filed **#714** (`enhancement,go,pr-polish-followup,priority: low`) to handle it before broad enablement.
- **#2 transient DDB error → silent drop** & **#3 fast follow-up before first persist** — keeping the fail-closed/silent behavior by design: in a channel thread we can't *confirm* is the agent's own (lookup errored, or the opening transcript isn't saved yet), posting would risk answering into unrelated channel chatter. The asymmetry vs the @mention error path is intentional — an @mention is an unambiguous address; a bare thread reply isn't. Documented in the function contract.
- **#4 double transcript read** — already tracked in **#712** (now marked an enablement prerequisite).

`go build/vet/test -race` green, golangci-lint v2.10.1 clean, `/simplify` clean. Head `b8607097`.

justin-layerv:
Round 5 — approve-in-spirit, converged. Every note is already tracked or by-design, no new code:
- **Firehose unbounded read** (gate runs before the per-user/team rate limiter) & **double transcript read** — the load-bearing **enablement prerequisite**, tracked in **#712**. Explicitly *not* a merge blocker (the surface ships dark behind `QURL_AGENT_CHANNEL_FOLLOWUPS`). Keeping the "thread the loaded history through" interim win deferred to #712 too — it's one holistic fix (pool isolation + both reads) and a partial version now would couple the gate's load to the turn path.
- **Implicit flag coupling** (`agentChannelFollowupDropped`→`isAgentChannelFollowup` doesn't re-check the flag) — correct under the single dispatch path; documented in the contract, called out to watch if a second path ever appears.
- **Silent stop on a corrupted live thread** — deliberate fail-closed; documented as a known support signature.
- **`thread_broadcast` dropped** — **#714**.

claude-review=success on `b8607097`, CI green, `/simplify` clean. Merge-ready (no push this round — nothing actionable).

PR #713, opened 2026-06-12
test(slack): pin consecutive same-role merge in agent transcript

justin-layerv:
## What

While reviewing the conversation-agent transcript for well-formedness, I traced a sequence the shipped proposal-followup flow produces and confirmed it is **valid** — then pinned that reliance with a comment + test. **No functional change.**

## The transcript shape

`toSDKMessages` ([llm.go](apps/slack/internal/agent/llm.go)) appends exactly one `anthropic.MessageParam` per domain `Message` — it does **not** coalesce consecutive same-role messages.

A turn that ends in a proposal ([agent.go](apps/slack/internal/agent/agent.go) propose path) or hits the iteration cap is persisted ending in a `user{tool_results}` message. The next turn, `a.Run` prepends a fresh `user{Text}`. So a follow-up @mention in the same thread builds this request:

```
assistant{tool_use} , user{tool_result} , user{text}
```

i.e. **two consecutive `user`-role messages** sent to the Messages API.

## Question investigated: accept-and-merge, or roles-must-alternate 400?

**Answer: the Messages API combines consecutive same-role turns into one — accepted, no 400.** Evidence, against the pinned SDK (`anthropic-sdk-go v1.46.0`, root `go.mod`):

- The doc on the exact field `buildParams` populates — `MessageNewParams.Messages` (`message.go:9264-9270`) — states: *"Consecutive `user` or `assistant` turns in your request will be combined into a single turn."*
- The pinned SDK contains **zero** role-alternation validation (no "must alternate" string anywhere) — it serializes and POSTs; the merge is server-side.
- The *merged* user turn is itself well-formed: `tool_result` leads it, the transcript still opens with a `user` turn, and every `tool_use` keeps its matching `tool_result`.

So this is **pre-existing and correct** — not a latent bug. (The "consecutive same-role → 400" guidance some references still carry reflects the early-2023 strict-alternation behavior, since relaxed; the version-pinned SDK contract is authoritative here.)

## Change

- **Comment** in `toSDKMessages` documenting that the consecutive same-role pair is intentional and *why it is safe* (the API merge + merged-turn validity), so it isn't re-opened as a bug.
- **`TestToSDKMessages_KeepsConsecutiveUserTurnsAfterProposal`** over the full cross-turn shape `[user{text}, assistant{tool_use}, user{tool_results}, user{text}]`, asserting one `MessageParam` per domain message (no coalescing), both trailing messages are `user`-role, and both turns' content survives — so a future edit can't "fix" the pair by merging it in the translation layer (which would be wrong).

## Verification

- `go test ./apps/slack/internal/agent/` — pass (incl. the new test)
- `go vet` / `go build ./apps/slack/...` — clean
- `golangci-lint run --new-from-rev=HEAD` — **0 issues** introduced by this diff

### DIFF
--- a/apps/slack/internal/agent/llm.go
+++ b/apps/slack/internal/agent/llm.go
@@ -157,17 +157,15 @@ func toSDKTools(specs []ToolSpec) []anthropic.ToolUnionParam {
 // preserving tool_use and tool_result blocks so the model sees a valid
 // transcript across turns.
 //
-// It emits exactly one MessageParam per domain Message and deliberately does NOT
-// coalesce consecutive same-role messages. A turn that ends in a proposal or hits
-// the iteration cap is persisted ending in a user{tool_results} message
-// ([Agent.Run]); the next turn prepends a fresh user{Text}, so a follow-up
-// @mention in the same thread yields two consecutive user-role messages here
-// (assistant{tool_use}, user{tool_result}, user{text}). That is well-formed, not a
-// roles-must-alternate bug: the Messages API combines consecutive same-role turns
-// into one (see MessageNewParams.Messages in the pinned anthropic-sdk-go), and the
-// merged user turn is itself valid — tool_result leads it, the transcript still
-// opens with a user turn, and every tool_use keeps its matching tool_result.
-// Pinned by TestToSDKMessages_KeepsConsecutiveUserTurnsAfterProposal.
+// It emits one MessageParam per domain Message and never coalesces consecutive
+// same-role messages. A proposal or iteration-cap turn is persisted ending in a
+// user{tool_results} message, and the next turn prepends a fresh user{Text}
+// ([Agent.Run]) — so a same-thread follow-up yields two consecutive user-role
+// messages (assistant{tool_use}, user{tool_result}, user{text}). That is
+// intentional, not a roles-must-alternate bug: the Messages API combines
+// consecutive same-role turns into one well-formed turn (MessageNewParams.Messages,
+// pinned anthropic-sdk-go v1.46.0). Pinned by
+// TestToSDKMessages_KeepsConsecutiveUserTurnsAfterProposal.
 func toSDKMessages(msgs []Message) []anthropic.MessageParam {
 	out := make([]anthropic.MessageParam, 0, len(msgs))
 	for i := range msgs {
--- a/apps/slack/internal/agent/llm.go
+++ b/apps/slack/internal/agent/llm.go
@@ -156,6 +156,18 @@ func toSDKTools(specs []ToolSpec) []anthropic.ToolUnionParam {
 // toSDKMessages rebuilds the SDK message history from domain messages,
 // preserving tool_use and tool_result blocks so the model sees a valid
 // transcript across turns.
+//
+// It emits exactly one MessageParam per domain Message and deliberately does NOT
+// coalesce consecutive same-role messages. A turn that ends in a proposal or hits
+// the iteration cap is persisted ending in a user{tool_results} message
+// ([Agent.Run]); the next turn prepends a fresh user{Text}, so a follow-up
+// @mention in the same thread yields two consecutive user-role messages here
+// (assistant{tool_use}, user{tool_result}, user{text}). That is well-formed, not a
+// roles-must-alternate bug: the Messages API combines consecutive same-role turns
+// into one (see MessageNewParams.Messages in the pinned anthropic-sdk-go), and the
+// merged user turn is itself valid — tool_result leads it, the transcript still
+// opens with a user turn, and every tool_use keeps its matching tool_result.
+// Pinned by TestToSDKMessages_KeepsConsecutiveUserTurnsAfterProposal.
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks for the review. Addressed all three minor/optional items in `b68eec97` (no functional change):

1. **Comment length** — trimmed the `toSDKMessages` doc comment: dropped the merged-turn enumeration and anchored the contract to the exact pinned SDK version (`anthropic-sdk-go v1.46.0`). Also de-duplicated the rationale — `toSDKMessages` is now the single source, and the test doc comment references it instead of restating the `MessageNewParams.Messages` citation.
2. **Test didn't independently pin "tool_use keeps its tool_result"** — added a pairing assertion: the propose `tool_use` and its ack `tool_result` now must both survive carrying the same id (via a `proposeID` const, which also keeps the id below goconst's threshold). Removed the over-claim from the production comment so it no longer asserts more than is pinned.
3. **Durability of the server-side-merge claim** — made the durability anchor explicit by naming `v1.46.0` in the comment, since the guarantee rests on the pinned SDK + server contract (as you noted, a unit test can't exercise the server merge).

Also ran `/simplify` (reuse / simplification / efficiency / altitude) — the only actionable finding was the cross-location comment duplication, now resolved; altitude review confirmed the seam-level placement is correct (a producer-side change in `Run` would break the #666 save-conflict-merge delta invariant).

justin-layerv:
Round 2 — fixed the substantive nit in `d1384663`. You're right that `strings.Contains(wire, proposeID)` was non-discriminating: a regression dropping the `tool_result` while keeping the `tool_use` (or vice-versa) would have left the id present and the test green, so it didn't pin the pairing the comment promised.

Replaced the wire-marshal with decoded-structure assertions:
- `params[1].Content[0].OfToolUse.ID == proposeID` **and** `params[2].Content[0].OfToolResult.ToolUseID == proposeID` — the call and its result must carry the **same** id (true pairing).
- the `tool_result` carries the propose ack on its text block, and `params[3]` is the follow-up user text.

This also removes the coupling to the SDK's JSON serialization you flagged. The structural checks above it (`len(params) == len(history)`, both trailing roles `user`) remain as the load-bearing no-coalescing guard.

justin-layerv:
Round 3 — fixed the doc-comment wording in the latest commit. You're right that `toSDKMessages` emits **zero** params for an empty assistant turn (`len(block) > 0` skip, pinned by `TestToSDKMessages_SkipsEmptyAssistantTurn`), so "one MessageParam per domain Message" was overstated.

- Production comment now reads "emits **at most one** MessageParam per domain Message (empty assistant turns are dropped — see `TestToSDKMessages_SkipsEmptyAssistantTurn`) and **never merges distinct messages**, so consecutive same-role messages stay separate." That frames the actual invariant (no merging) rather than implying a strict 1:1.
- Added a note in the test that its `len(params) == len(history)` check holds because the fixture has no empty (droppable) turns — so the two invariants no longer appear to conflict.

Comment-only change; no executable diff.

justin-layerv:
Round 4 — addressed both optional-polish notes in the latest commit:

1. **Doc-comment length** — trimmed the `toSDKMessages` comment: dropped the turn-by-turn persistence narration and lean on `[Agent.Run]` for where the consecutive-user pair arises. Kept the load-bearing parts you named — the invariant ("at most one param per message, never merge same-role") plus the server-merge rationale + the `v1.46.0` anchor.
2. **Index assumptions** — added `len(params[i].Content) == 1` guards before the `Content[0]` accesses, so the index-0 intent is explicit and fails loudly with a clear message if a fixture change (e.g. giving the assistant turn `Text`) shifts `tool_use` off index 0.

3. **Scope** — agreed; the PR body already states this unit test pins only the translation layer's no-coalescing and that the "no 400" guarantee rests on the pinned SDK/server contract, not anything the test executes. No change needed there.

Thanks for the thorough passes.

justin-layerv:
Round 5 — thanks for the LGTM. Both discussion points are deliberate; capturing the rationale here so it's on record:

1. **Server-side merge vs. client-side merge.** This is an intentional choice, and you're right that the test now actively prevents the alternative. The investigation established that the Messages API already combines consecutive same-role turns (documented on `MessageNewParams.Messages`, pinned `anthropic-sdk-go v1.46.0`), so the scope was to *document + pin that reliance* rather than add client-side coalescing. Client-side merge in `toSDKMessages` is viable (it's pure translation — it wouldn't touch the persisted `[]Message` or the #666 save-conflict delta invariant), but it would add normalization code to a shipped function to handle a case the server already handles correctly, with no behavior change. Keeping `toSDKMessages` a faithful 1:1 of the domain transcript (and letting the server merge) is the leaner boundary. If we ever want to decouple from that server contract, the test name and comment make the switch obvious — that's the point of pinning it.

2. **Speculative test scope.** Agreed it guards a hypothetical future "fix" rather than a current violation. Given the invariant is genuinely non-obvious (a future reader could plausibly "correct" the consecutive-user pair in the translation layer and silently rely on/break the server merge), pinning it with the rationale inline is worth the small cost.

No code change for this round (LGTM, no changes required).

Issue #708, opened 2026-06-11
slack agent: verify streamed (markdown_text) vs posted (mrkdwn) reply rendering parity at enablement


justin-layerv:
Surfaced by cr on #707 (native reply streaming into the assistant pane).

## Context

The streamed reply path (`chat.appendStream`/`startStream`, `handler_agent_stream.go`) sends **`markdown_text`** (standard Markdown). The non-streaming fallback (`postAgentReply` → `chat.postMessage` `text`, the merged #650 behavior) renders **Slack mrkdwn**. The two dialects differ for links (`[label](url)` vs `<url|label>`), emphasis (`**bold**` vs `*bold*`), and mentions (`<@U>`/`<#C>`).

Today this is **low-risk** because the agent is constrained to *"light Slack-compatible markdown only: short bullets, backticks for resource ids and aliases"* (`internal/agent/prompt.go`), a subset where the two render compatibly:
- backticks render identically under both;
- `- ` bullets list-render under `markdown_text` and show as literal dashes under mrkdwn (cosmetic, arguably nicer when streamed);
- the agent is *not* instructed to emit mrkdwn links/mentions/`*bold*`.

The streaming surface is **dark** until sandbox enablement, so the actual Slack-client rendering can't be eyeballed yet.

## Tasks (at enablement)

- [ ] In sandbox, compare a streamed reply vs the same text posted on the fallback path (force a `startStream`/`appendStream` failure) — confirm bullets, backticks, code blocks, and any inline links render acceptably the same.
- [ ] If they diverge unacceptably, normalize one way — candidates: (a) switch `postAgentReply` to `chat.postMessage` `markdown_text` so both paths share the standard-Markdown dialect (watch the deliberate no-escape semantics the non-streaming path relies on), or (b) keep both and accept the cosmetic bullet difference.
- [ ] Keep the `agent/prompt.go` "light Slack-compatible markdown" constraint aligned with the streaming dialect: if the prompt ever loosens to allow mrkdwn links/mentions, the streamed path must normalize them first.

## Acceptance

Documented confirmation that streamed and posted replies render equivalently for the agent's actual output vocabulary, or a normalization landed. Cross-reference the code note in `handler_agent_stream.go`.

justin-layerv:
Scope addition from cr round 2 on #707: also measure **streaming backpressure** at enablement.

[...]

PR #717, opened 2026-06-12
fix(slack): render agent answers as markdown_text so channel replies format correctly

justin-layerv:
## Problem

The conversation-mode Secure Access Agent's free-text answer was posted via `chat.postMessage`'s `text` field, which Slack renders as **mrkdwn** (single `*` bold, `_` italic, no headings, `<url|label>` links). But LLMs emit **standard Markdown** (`**bold**`, `## headings`, `[label](url)`), so channel replies showed literal `**` and broken links.

The streaming pane already renders correctly because `chat.appendStream` takes `markdown_text` (standard Markdown). The same answer string was being fed to two parsers with **different dialects** — only the non-streaming channel path was wrong.

## Fix

Post the agent's own answer through a new `markdown_text` seam (`PostMarkdownMessage`) so a channel reply renders through Slack's Markdown parser **identically to the streaming pane**. No hand-rolled mrkdwn converter — Slack does the conversion. Standard Markdown is now the single canonical dialect for both delivery paths.

- Only the agent's **free-text answer** routes through `markdown_text`.
- The LLM-distilled **proposal preview stays escaped mrkdwn** on the `text` seam (injection defense — it must not be re-parsed as Markdown). The confirm card is unchanged.
- The seam **falls back to `PostMessage`** when unwired (not part of the `agentEnabled` gate), so a turn still delivers.
- Realigns the prompt's `OUTPUT` guidance (standard Markdown, not "Slack-compatible") and the stale streaming dialect comment (both paths now `markdown_text`).

## Why `markdown_text` (not a converter)

`chat.postMessage` accepts `markdown_text` as a field distinct from the `mrkdwn` boolean. Slack's [Markdown block docs](https://docs.slack.dev/reference/block-kit/blocks/markdown-block/) confirm this dialect is **standard CommonMark** (`**bold**`, `# heading`, `[label](url)`) and is purpose-built for LLM output so "Slack [does] the translating." Letting Slack's own parser convert gives exact parity with the streaming pane and avoids owning a fragile Markdown→mrkdwn converter (bold/italic collisions, code-span protection, etc.).

> **Reviewer note:** this relies on `markdown_text` rendering standard Markdown — confirmed via the [chat.postMessage method reference](https://docs.slack.dev/reference/methods/chat.postMessage/) (lists `markdown_text` = "message text formatted in markdown") and the [Markdown-block docs](https://docs.slack.dev/reference/block-kit/blocks/markdown-block/), and it is the same field the shipped streaming pane uses. Live-workspace verification (render of `**bold**`/bullets, push-notification/accessibility fallback for a `markdown_text`-only body) and the masked-link threat-model sign-off are tracked in **#720** for the pre-enablement pass (#708) — none block this flag-gated, dark-by-default change.

## Tests

- `cmd`: `markdown_text` builder posts the body verbatim in the `markdown_text` field and **omits** `text` (Slack rejects the pairing) + omits empty `thread_ts`.
- `internal`: `deliverAgentResult` routes a free-text answer to the `markdown_text` seam (body verbatim) and a proposal preview to the escaped mrkdwn seam; markdown seam falls back to text when unwired.
- Full `apps/slack/...` suite green; `golangci-lint` adds zero new findings vs. `main`.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -153,6 +153,9 @@ func run() error {
 	// Both PostMessage seams share the per-workspace token lookup + Grid fallback with
 	// the slash-command modals.
 	postMessage := newSlackPostMessageFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
+	// markdown_text seam for the agent's free-text answer, so a channel reply renders
+	// standard Markdown the same way the streaming pane does (see PostMarkdownMessage).
+	postMarkdownMessage := newSlackPostMarkdownMessageFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
 	postMessageBlocks := newSlackPostMessageBlocksFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
 	// reactions.add/remove seam for the agent's best-effort "working on it" ack. Always
 	// wired (same token lookup as the post seams); inert until the agent surface is live
@@ -248,6 +251,7 @@ func run() error {
 		AgentLLM:                    agentLLM,
 		AgentStore:                  agentStore,
 		PostMessage:                 postMessage,
+		PostMarkdownMessage:         postMarkdownMessage,
 		AgentDisabled:               agentDisabled,
 		PostMessageBlocks:           postMessageBlocks,
 		AgentConfirmEnabled:         agentConfirmEnabled,
--- a/apps/slack/cmd/slack_postmessage_test.go
+++ b/apps/slack/cmd/slack_postmessage_test.go
@@ -383,3 +383,62 @@ func TestSlackPostMessageBlocksFuncSurfacesSlackError(t *testing.T) {
 		t.Fatalf("error = %v, want channel_not_found", err)
 	}
 }
+
+// mdTestChannel is the channel id the markdown_text builder tests post to. A
+// shared const keeps the literal out of goconst's repeated-string count.
+const mdTestChannel = "C_chan"
+
+func TestSlackPostMarkdownMessageFuncPostsMarkdownTextField(t *testing.T) {
+	t.Parallel()
+	var rawBody string
+	var gotBody struct {
+		Channel      string `json:"channel"`
+		ThreadTS     string `json:"thread_ts"`
+		MarkdownText string `json:"markdown_text"`
+		Text         string `json:"text"`
+	}
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		raw, _ := io.ReadAll(r.Body)
+		rawBody = string(raw)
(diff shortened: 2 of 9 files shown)


justin-layerv:
Thanks — addressing the review feedback.

**Masked-link rendering surface (the one ask for an explicit decision):** Accepting the rich rendering for this change, deliberately. Rationale:
- It's **not a new surface**: the streaming pane (#707) already renders the agent's answer via `markdown_text`, so clickable/masked links already render there. This PR makes channel replies *consistent* with the pane rather than introducing a new dialect.
- The **injection-prone distilled** surfaces (proposal summaries) stay escaped mrkdwn; only the model's own conversational prose renders rich, and the system prompt already forbids the agent inventing links.
- Bare URLs autolink in **both** dialects, so the only newly-activated property is masked links — and the right lever for that (if the threat model says "no") is **dialect-level** hardening applied to *both* the channel and pane paths, not a channel-only transform that would re-diverge them.

I've made this an explicit, documented decision rather than an incidental side effect: a code comment now records it at the routing point (`deliverAgentResult`), and the pre-enablement security ratification is tracked in **#720**.

**Notification/accessibility fallback** (markdown_text-only body) and **live render of `**bold**`/bullets**: both are live-workspace verification items I can't exercise in CI; folded into **#720** to be confirmed before GA enablement (the surface is dark / pre-enablement per #708).

**12k length cap** and **`PostMessageFunc` type reuse**: noted, no action — replies are prompt-constrained short, and the builder names the param `markdownText` locally.

Latest push (`866a041f`) also folds in `/simplify` cleanups: inlined the `agentReplyMarkdown` predicate into `deliverAgentResult`, extracted a shared `deliverAgentText` delivery helper, and de-duplicated the routing test. CI is green (lint/test/docker).

justin-layerv:
Thanks for the thorough re-review — addressed the LGTM round's notes in `b730239a`:

- **`agentReplyText`'s unreachable `Reply` passthrough (note 3):** fixed now rather than deferred. Refocused the function to its actual job — the escaped proposal preview, else the error fallback — and removed the dead `return result.Reply` branch and its stale "deliberately NOT escaped / agent's own answer" comment (that path is now `markdown_text` via `deliverAgentResult`). `TestAgentReplyText` is updated to pin the new contract (a non-proposal result renders the error fallback; the answer-renders-verbatim invariant lives in `TestDeliverAgentResult_RoutesByDialect`). The split is now self-documenting with no production-dead code.
- **`text` + `markdown_text` mutual exclusivity (note 2):** good catch — if Slack accepts `text` purely as the notification fallback alongside `markdown_text`, that's the cheap fix for the a11y gap. Captured as an explicit open question on #720 (item 2) to settle against the live API before choosing the fallback approach.
- **Notification/a11y fallback (note 1):** tracked in #720 as a real pre-enablement gap, not just render parity.

CI is green on `b730239a` (lint/test/docker). The masked-link threat-model sign-off and the live-render/notification verification remain tracked in #720 for the pre-enablement pass (#708).

PR #718, opened 2026-06-12
feat(slack): isolate channel follow-up firehose to a separate admission pool

### LINKED ISSUE #712: slack agent: pre-pool admission filter for channel thread follow-ups (avoid firehose pool/RCU pressure)
Surfaced by cr on #711 (channel thread-reply continuity). To address **before broad enablement** of `QURL_AGENT_CHANNEL_FOLLOWUPS` (the manifest/flag enablement is qurl-integrations-infra #1040).

## Problem

With the flag on + `message.channels`/`groups` subscribed, the bot receives **every** message in channels it's a member of. Any channel thread reply with non-empty text passes the synchronous `shouldDispatchAgentEvent` and is dispatched to the **shared** async pool (`h.sem`, `process.go`), where `agentChannelFollowupDropped` does a DynamoDB `LoadConversation` **before** dropping non-agent-thread replies.

So in a busy channel, unrelated thread chatter can:
- **(a)** burn RCU on a per-reply transcript read, and
- **(b)** compete for / saturate the same pool that `@mention`/DM turns, slash commands, and interactions use — surfacing as `"async pool saturated — dropping event"` for *legitimate* turns.

The own-thread filter can't run synchronously (the webhook ack path has no store access), so admission necessarily precedes filtering.

## Why it's deferred (not fixed in #711)

- The obvious fix — an in-memory LRU of known agent threads, checked pre-pool — has a **cross-task correctness gap**: Slack delivers each event to one task via the ALB, but a thread joined by task A (its transcript is in shared DDB) can have a reply delivered to task B, whose local cache wouldn't have it → B would drop a *valid* reply. So a per-process negative cache can't safely drop.
- The robust fix is a real change (e.g. a **separate small bounded semaphore** for the gate/admission checks so the firehose can't starve the main turn pool, or a positive-cache fast-path + DDB confirm) that should be **load-tested at enablement**, not built speculatively on dark code.

## Tasks

- [ ] Isolate firehose admission load from the main turn pool (separate bounded pool for the gate check, or equivalent) so non-agent channel replies can't cause `@mention`/DM saturation.
- [ ] Eliminate the double `LoadConversation` on the accepted-follow-up path (`agentChannelFollowupDropped` reads, then `processAgentEvent` reads again at the turn) — thread the loaded `history`/`version` through, or fold the gate into the single load. (cr #711 item 2; minor on its own — low-volume — but addressed holistically here.)
- [ ] Monitor the async-pool saturation metric during the staged sandbox enablement (#1040) and re-evaluate before any prod enablement.

## Acceptance

A busy channel the bot is in does not degrade `@mention`/DM latency or cause pool-saturation drops for legitimate turns, validated under representative load before broad enablement.

justin-layerv:
## Summary

Implements **#712** — the firehose-isolation prerequisite before `QURL_AGENT_CHANNEL_FOLLOWUPS` can be enabled at scale. With the flag on and `message.channels`/`groups` subscribed, the bot receives **every** message in channels it's a member of. Before this PR, each channel thread reply rode the **shared** async pool (`h.sem`) and did a DynamoDB `LoadConversation` (the own-thread gate) before non-agent replies were dropped — so a busy channel could (a) burn RCU on per-reply reads and (b) saturate the pool that `@mention`/DM/slash/interaction work shares, surfacing as `"async pool saturated"` drops for *legitimate* turns.

Two changes:

1. **Separate bounded pool for channel follow-ups.** Follow-ups now dispatch to `h.followupSem` (a distinct pool); `@mention`/DM stay on `h.sem`. A busy channel's firehose can only saturate the follow-up pool — the main pool is fully isolated. The follow-up pool's saturation is logged **distinctly** (`agent: follow-up admission pool saturated — dropping channel reply`) so it can be monitored separately during enablement.

2. **Eliminate the double read.** The gate (`agentChannelFollowupDropped`) now returns the transcript it loaded (`*loadedHistory`); `resolveTurnHistory` reuses it, so an accepted follow-up reads DynamoDB **once**, not twice. `@mention`/DM are unchanged (they skip the gate and load at the turn).

`runOnPool(sem, …)` was extracted in `process.go` so both pools share the same spawn / `wg` / panic-recover / timeout logic and drain together on shutdown; `readPoolSizeEnv` is shared by both pool-size env readers.

## Design note (why the simple option)

The issue floats two robust fixes: a separate semaphore for *just the gate check* (admitted turns hop back to the main pool), or routing the *whole* follow-up turn to its own pool. This PR takes the **latter** — it meets the acceptance criterion (the main pool is protected) with far less surface, and avoids adding new untested behavior (a two-hop dispatch, admitted-then-dropped, dedupe-after-admission) to a code path that is still **dark**. The follow-up pool is sized **generously** (`QURL_SLACK_MAX_CONCURRENT_FOLLOWUP_ASYNC`, default = the main pool's 50) — main-pool isolation holds at *any* size, so it's sized to avoid head-of-line between the pool's fast gate reads and its 90s turns. If the distinct saturation metric shows the follow-up pool starving legitimate follow-ups under real load, that's the evidence-backed trigger to split the gate check onto its own smaller pool (issue task 3).

## Type of Change

- [x] New feature (the dark code path; flag enablement is separate, infra #1040)

## Testing

- `go build/vet`, full `-race` suite, and golangci-lint v2.10.1 all green.
- **`TestAgentEventFollowupPoolIsolation`** pins the isolation both directions: sizing each pool to 1 and holding one pool's slot, the *other* path still runs a full (fake-LLM) turn and posts — a saturated follow-up pool drops follow-ups but not `@mention`s, and a saturated main pool drops `@mention`s but not follow-ups. Deterministic under `-race`.
- `TestAgentChannelFollowupDropped` extended to assert the gate returns the preloaded transcript on an admitted follow-up (the one-read path) and `nil` for `@mention`/DM.

## ⚠ Load validation is pending enablement

This is the dark code path. The acceptance criterion ("a busy channel does not degrade `@mention`/DM latency or cause pool-saturation drops, **validated under representative load**") **cannot** be confirmed until the staged sandbox enablement actually runs the bot in a live, busy channel. This PR lands the mechanism + the distinct metric to watch; "validated under load" is explicitly *not yet proven*. Issue task 3 (monitor the follow-up-pool saturation metric during sandbox enablement) remains open.

## Prod Rollout Tasks

- [x] Confirmed this PR has no prod rollout tasks — additive to a dark code path; no Terraform / live-resource change. Enablement (the flag + manifest scopes) is tracked in infra #1040.

Closes #712.


justin-layerv:
Round 1 — all four addressed (none blocking; thanks for the careful read):

1. **Env var documented** — added a `QURL_SLACK_MAX_CONCURRENT_FOLLOWUP_ASYNC` row to `docs/operating.md` beside its sibling, with the "watch the distinct follow-up saturation log + tune at enablement" guidance.
2. **Wider optimistic-lock window** — noted in the `loadedHistory` doc-comment: reusing the gate's read snapshots `version` slightly earlier (before dedupe/ack), so the read→save window is marginally wider; a concurrent save is then no worse than a version conflict, which `saveAgentHistory` already resolves by reload-and-merge.
3. **Symmetric single saturation log** — the `@mention`/DM path now dispatches via `runOnPool(h.sem, …)` directly with its own one-line log, matching the follow-up branch. `startAsyncWorkerWithTimeout` keeps its inner log for `runAsync`/slash callers (which have no caller-side log), so no observability regression — each agent path now logs its drop exactly once.
4. **Consistency nit** — the routing branch now reads `&envCopy.Event` (the value the work closure captures), so the two can't diverge.

build/vet/`-race`/lint v2.10.1 green. `/simplify`: the full 4-lens pass ran clean on the main diff; this round is reductive (a doc row, a comment, a method swap), self-reviewed. Head `b76a7081`.

justin-layerv:
Round 2 — no blocking bugs; all four observations addressed:

1. **Wider version window** — documented in the `loadedHistory` comment last round; `saveAgentHistory`'s reload-and-merge on conflict makes it no worse than before. A conscious, documented tradeoff.
2. **Gate reads on dropped firehose can still saturate the follow-up pool** — correct, and by design for Option B (the whole follow-up turn rides `h.followupSem`). Now visibly tracked in **#719** — the gate-on-its-own-pool (Option-A) escalation, triggered by the distinct metric at enablement.
3. **Doubled worst-case concurrency** — added to #719: verify the DynamoDB / HTTP client connection pools are sized for the higher ceiling (main 50 + follow-up 50) at enablement, so the new pool doesn't trade pool-saturation drops for connection-pool contention.
4. **Explicit shutdown-drain test** — dispositioned: the shared-`h.wg` drain is already exercised by `TestAgentEventFollowupPoolIsolation` (each pool's work runs under `t.Cleanup(h.Wait)` with a post assertion — a wg leak would hang the test); a dedicated test would near-duplicate it, and you rated it low-value.

No code change this round (every item is an accepted tradeoff or an enablement-time operational check tracked in #719), so head stays `b76a7081`, `claude-review` green. This PR is the code half of #712 (tasks 1+2); #719 carries task 3.

PR #722, opened 2026-06-12
fix(slack): alias the reserved word `ttl` in the agent rate-limit counter

justin-layerv:
## Summary

`BumpTurnCount` (the agent turn-rate counter) wrote `SET ttl = :ttl` in its DynamoDB `UpdateExpression` with **no** `ExpressionAttributeNames`. `ttl` is a DynamoDB reserved word, so DynamoDB rejected **every** `UpdateItem` with a 400 `ValidationException` ("Attribute name is a reserved keyword; reserved keyword: ttl").

The turn-rate gate is **fail-open** — it logs a `Warn` and allows the turn when the counter errors (a deliberate availability choice for a cost backstop). So a 400 on *every* call silently disabled the per-user and per-team agent turn caps **entirely**, not just during a real counter outage.

This aliases the attribute as `#ttl` via `ExpressionAttributeNames`, restoring the cap. The (intentional) fail-open policy is unchanged — it just no longer fires on a code bug.

## Scope

- [x] `apps/slack/`

## Changes

- **`internal/slackdata/agentstate.go`** — alias `ttl` → `#ttl` in `BumpTurnCount`'s `UpdateExpression` via `ExpressionAttributeNames`, with a comment on the reserved-word hazard. The `PutItem`-based markers (`putMarkerIfAbsent`) set `ttl` as a raw item-map key, which is fine — only *expressions* reserve it, and `ttl` is the only reserved-word collision in the package, so no other call site is affected.
- **`internal/slackdata/agentstate_ratelimit_test.go`** — add `TestBumpTurnCount_TTLReservedWordIsAliased`, which pins the alias invariant (the names map maps `#ttl`→`ttl`, and the expression `SET`s via the alias). Relax the overlapping `ttl` assertion in `TestBumpTurnCount_AtomicAddRequest` so the alias invariant lives in one named test.
- **`internal/handler_agent_test.go`** — update `memAgentDDB.UpdateItem`'s exact-match expectation to the aliased shape. Both in-memory fakes (store-boundary `stubDDB`, handler-boundary `memAgentDDB`) had pinned the buggy `SET ttl = :ttl` string, which is why the bug shipped green.

## Test Plan

- [x] `make check` (`fmt vet lint test-race`) green locally — `golangci-lint v2.10.1` reports 0 issues; `go test ./internal/... ./cmd/... -race` passes
- [x] New tests added for new functionality — the regression test was confirmed **RED** against the buggy bare-`ttl` expression (empty names map + `SET ttl = :ttl`) before the fix, then **GREEN** after
- [x] Manual testing completed — bug surfaced bringing the agent up live in sandbox; the counter now succeeds (UPDATED_NEW returns the running count)

## Related Issues

Follow-up hardening for the live Slack Secure Access Agent turn-cap gate. Found while bringing the agent up in sandbox.

### DIFF
--- a/apps/slack/internal/handler_agent_test.go
+++ b/apps/slack/internal/handler_agent_test.go
@@ -342,16 +342,17 @@ func memEvalSaveCond(cond string, existing map[string]ddbtypes.AttributeValue, p
 }
 
 // UpdateItem fakes only the one shape BumpTurnCount emits — "ADD turn_count :one SET
-// ttl = :ttl" — applying the number ADD and the SET, then returning UPDATED_NEW. A
-// no-op stub would make the rate-limit tests vacuously pass (count always 0), so it
-// actually mutates; anything other than that exact shape errors loudly.
+// #ttl = :ttl" (ttl is a DynamoDB reserved word, so it's aliased via #ttl) — applying
+// the number ADD and the SET, then returning UPDATED_NEW. A no-op stub would make the
+// rate-limit tests vacuously pass (count always 0), so it actually mutates; anything
+// other than that exact shape errors loudly.
 func (f *memAgentDDB) UpdateItem(_ context.Context, in *dynamodb.UpdateItemInput, _ ...func(*dynamodb.Options)) (*dynamodb.UpdateItemOutput, error) {
 	f.mu.Lock()
 	defer f.mu.Unlock()
 	if f.updateErr != nil {
 		return nil, f.updateErr
 	}
-	if expr := aws.ToString(in.UpdateExpression); expr != "ADD turn_count :one SET ttl = :ttl" {
+	if expr := aws.ToString(in.UpdateExpression); expr != "ADD turn_count :one SET #ttl = :ttl" {
 		return nil, fmt.Errorf("memAgentDDB.UpdateItem: unsupported expression %q", expr)
 	}
 	k := memKey(in.Key)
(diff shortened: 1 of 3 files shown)


justin-layerv:
**Deferred (out of scope for this fix), tracked with acceptance criteria:**

- [layervai/qurl-integrations-infra#1055](https://github.com/layervai/qurl-integrations-infra/issues/1055) — no alarm on the agent turn-rate counter's **fail-open** path. This bug was silent precisely because the gate fails open by design; the `Warn` it logged every turn was the signal that should have caught it, but nothing watches it. The issue proposes a CloudWatch alarm on that fail-open rate (sandbox + prod).

**For reviewers:**

- The regression test (`TestBumpTurnCount_TTLReservedWordIsAliased`) was verified to **fail** against the buggy bare-`ttl` expression (empty names map + `SET ttl = :ttl`) before the fix, then pass after — it cannot pass against the bug it guards.
- `/simplify` ran clean on reuse / efficiency / altitude. One "redundant assertion" suggestion was intentionally kept: the alias-map check (`ExpressionAttributeNames["#ttl"] == ttl`) and the `SET #ttl = :ttl` substring check guard **distinct** failure modes (a missing/wrong names map vs. a bare reserved word), and the `stubDDB` test double doesn't model the DynamoDB SDK's runtime name↔expression coupling that would otherwise let one subsume the other.

justin-layerv:
Thanks for the thorough review and the independent re-verification (including the `conv_version`-vs-reserved-`version` check — exactly the adjacent collision worth ruling out).

On the non-blocking item — the fakes not modeling reserved words, so a *new* reserved word in a *new* expression would be caught nowhere — I've taken you up on the offer and filed it rather than expanding this surgical fix: **#723**. It captures your three options (adopt the AWS `expression` builder, a dynamodb-local integration smoke test, or a reserved-word lint helper) with acceptance criteria. Out of scope here, and the only current collision (`ttl`) is fixed and pinned in this PR. Labeled `enhancement` / `go` / `priority: low`.

PR #724, opened 2026-06-13
fix(slack): deliver the agent's one-time link in-conversation, not via an invisible ephemeral

justin-layerv:
## Summary

On Approve of a conversation-mode `get`, the agent's one-time access link was delivered as a **response_url ephemeral**. Slack **does not render ephemerals in a 1:1 DM** — it accepts the POST with a 2xx and shows nothing — so a user who asked the agent for a link **in a DM** got the "sent privately to the approver" card but **never saw the link** (the reported bug).

This delivers the sensitive output **in the same conversation the user approved in**, surface-aware:

- **1:1 DM** (`D`-prefix channel): post a **normal message** — the DM is already private (just the user and the app) — into the **card's own thread** (snapshotted on the pending action at propose time), so it lands in the **assistant pane** the user is looking at, not adrift in the DM timeline.
- **channel / group DM**: keep the response_url **ephemeral**, scoped to the clicker, so a one-time link never enters shared history.

Reuses the already-wired `PostMessage` seam — no new plumbing.

### Also: the card no longer lies on failure

The get case previously returned the "sent privately" **success** copy *unconditionally*, even when the mint itself failed — which is what masked this bug. Now:

- **mint / resolve failure** → a neutral failure card; the token-bearing detail is delivered privately and is **never echoed** on the public card.
- **successful mint, delivery failed** → "couldn't deliver here", not a false success.

## Scope

- [x] `apps/slack/`

## Changes

- `internal/handler_agent_confirm.go`:
  - `pendingAction.ThreadTS` — snapshot the card's thread at propose time so the link is delivered back into it.
  - `isDirectMessageChannel` (D-prefix) + `deliverConfirmPrivate` — surface-aware private delivery (in-thread DM message vs in-channel ephemeral), reporting whether it succeeded.
  - `executeAgentAction` get case branches on the mint outcome (neutral failure card + private detail vs success card + link).
  - `finalizeConfirmedAction` — extracted from `processAgentConfirm` (execute → deliver → card + delivery-failure correction → swap → audit); keeps the click orchestration under the complexity budget.
- `internal/handler_agent_confirm_test.go`: link routing by surface (DM → in-thread `PostMessage`, channel → ephemeral), the DM path end-to-end, propose-time thread capture, mint-failure honesty + no-leak, and delivery-failure detection.

## Test Plan

- [x] New tests added; the DM in-thread delivery, surface routing, and no-leak invariants are pinned
- [ ] **Manual testing pending deploy** — in-DM ephemeral non-rendering is *live Slack behavior*, so this is unit-tested + reasoned but cannot be verified pre-merge (the sandbox runs deployed `main`). I'll confirm the link renders in the assistant pane once it deploys.

## Related

Found bringing the live Slack Secure Access Agent up in sandbox: a user asked for an access link in the assistant DM, approved, and never saw the link.

### DIFF
--- a/apps/slack/internal/handler_agent_confirm.go
+++ b/apps/slack/internal/handler_agent_confirm.go
@@ -35,7 +35,14 @@ const (
 	agentConfirmCanceledReply       = "Canceled — nothing was changed."
 	agentConfirmUnsupportedReply    = "I can't apply that kind of change yet."
 	agentConfirmGetDeliveredReply   = "Handled — the access link was sent privately to the approver."
-	agentConfirmFailedReply         = "Something went wrong applying that. Please try again, or use a `/qurl` command."
+	// agentConfirmGetFailedReply is the get's neutral FAILURE card (mint/resolve failed).
+	// It must NOT echo the LLM-distilled token onto the public card and must NOT claim
+	// "sent privately" — the detailed reason is delivered privately instead.
+	agentConfirmGetFailedReply = "Couldn't generate the access link. Check the details I sent you, or ask me again."
+	// agentConfirmGetDeliveryFailedReply covers a SUCCESSFUL mint whose private delivery
+	// failed: the card must not claim success when the user received nothing.
+	agentConfirmGetDeliveryFailedReply = "Generated the access link, but couldn't deliver it here. Ask me again."
+	agentConfirmFailedReply            = "Something went wrong applying that. Please try again, or use a `/qurl` command."
 	// agentAttributionAgentName names the actor in an executed action's attribution
 	// footer — the product's user-facing agent name, never "bot".
 	agentAttributionAgentName = "qURL Secure Access Agent"
@@ -75,6 +82,11 @@ type pendingAction struct {
 	URL       string           `json:"url,omitempty"`    // target URL for protect-url
 	Asker     string           `json:"asker,omitempty"`  // Slack user who requested the turn; get is asker-only (see processAgentConfirm)
 	ChannelID string           `json:"channel_id"`
+	// ThreadTS is the thread the confirm card was posted into (empty for a top-level
+	// card). The get's private delivery posts the link back into THIS thread so it
+	// lands where the user is looking — the assistant pane is a threaded view, so a
+	// top-level post would miss it. See deliverConfirmPrivate.
+	ThreadTS string `json:"thread_ts,omitempty"`
 }
 
 // confirmValidAliasBind validates a bare (no leading "$") alias + target — as the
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks for the thorough review — all addressed in bfcc14ee.

**1. Test gap (the card-swap composition).** Extracted the override into a pure `composeConfirmCard(res, delivered, asker, approver)` and added `TestComposeConfirmCard`, which directly pins the case you flagged: a successful mint (`agentConfirmGetDeliveredReply` + link) with `delivered=false` → `agentConfirmGetDeliveryFailedReply`, asserting the card **never echoes the link** — plus the rest of the matrix (delivered success, failed-get-unchanged, non-get attributed, unattributed byte-exact). The harness's `qurlBackendServer` implements only resource creation, not the qURL-mint endpoint, so a live successful get isn't reachable in-test; the pure function tests the exact composition, and `deliverConfirmPrivate` returning `false` is covered separately — together they cover the wiring.

**2. Group DMs (mpim).** Good catch — the agent does reach a group DM via channel-followups. A normal post there would **leak** the link to other members, so the ephemeral path is the only non-leaking option (and ephemerals do render in multi-party conversations, unlike the 1:1 IM). If one somehow didn't render in an mpim the link would be hidden there, not leaked. Documented as an explicit boundary in `deliverConfirmPrivate` and filed #725 to verify/limit it.

**3. DM persistence tradeoff.** Made explicit in the `deliverConfirmPrivate` doc: in a 1:1 DM the link persists in that user's history, acceptable for a one-time-use credential scoped to the same user (it burns on first redemption regardless).

**4. Misleading failure copy.** Fixed — `agentConfirmGetFailedReply` is now self-contained ("Couldn't generate the access link — check the request and ask me again."), so it no longer promises a detail message the double-fault case may not have delivered.

**5. Audit records `res.cardText`.** Agreed it's intentional: the audit reflects the **mutation** outcome (the mint happened or it didn't), not the private delivery. Noted in the comment.

justin-layerv:
Round 2 confirmations:

**1. One-time-use is unconditional, and the admit window is 1 minute.** `getWork` hardcodes `OneTimeUse: true` (`handler_get.go:385` — there is no `once` flag) and `ExpiresIn: resourceLinkExpiry` = `"1m"`. So a link sitting in DM history is dead within 1 minute of minting **or** on first redemption, whichever comes first — the durable-secret window is ≤1 min and only for an *unredeemed* link. That's a hard guarantee, not a convention, and it's what the persistence tradeoff rests on. The optional `chat.delete`-after-TTL is therefore hygiene (sweeping a dead link out of history) rather than closing a live window; happy to file it if you'd like it tracked, but the 1-minute expiry largely neutralizes the retention concern.

**2. E2e success-path DM test** — filed #726 to extend the harness (`qurlBackendServer` has no mint endpoint today) and add the mint -> in-thread `PostMessage` assertion. The composition + no-leak invariants are unit-pinned in the meantime.

**3. Empty `ThreadTS` -> top-level DM post** — agreed, benign: the agent always threads its card, so `ThreadTS` is non-empty in practice, and a top-level DM post is still visible. The "lands where the user is looking" guarantee only softens in that non-occurring fallback.

**4. Group DM (#725)** — tracked; the failure direction there is hide-not-leak, the safe one.

Issue #725, opened 2026-06-13
Verify (or bound) agent get-link delivery in group DMs (mpim)


justin-layerv:
## Context

[#724](https://github.com/layervai/qurl-integrations/pull/724) delivers a conversation-mode `get`'s one-time access link in the conversation the user approved in: a normal in-thread message in a 1:1 DM (where Slack doesn't render ephemerals), an ephemeral in channels and group DMs (where a normal post would leak the link to other members).

For a **group DM (mpim, `G`-prefix)** the ephemeral path is the only *non-leaking* option — a normal post would expose the one-time link to every member. But #724's premise is that Slack won't render an ephemeral in a 1:1 DM; whether ephemerals reliably render in **multi-party DMs** is an assumption (they do in channels). If they don't render in an mpim, the link would be invisible there (the same class as the bug #724 fixes), with no safe normal-post fallback.

The agent reaches a group DM only via channel-followups on a threaded mpim message, so this is a rare, low-priority surface.

## Options

- Confirm ephemerals render in a group DM (likely yes — multi-party, like a channel).
- If they do **not**: the agent can't privately deliver a link inside a group DM without leaking. Decide between (a) refuse `get` in a group DM with a clear message ("ask me in a channel or a DM"), or (b) deliver to the approver's 1:1 app-DM (a context switch, but the only private channel).

## Acceptance

A test/confirmation that the ephemeral renders in a group DM, **or** an explicit user-facing boundary (refuse / redirect) so the link is never silently invisible there.

PR #727, opened 2026-06-13
fix(slack): deliver the channel agent link via chat.postEphemeral so the card-replace can't overwrite it

justin-layerv:
## Summary

When the agent's conversation-mode `get` is **approved in a channel/thread**, the one-time access link was delivered as a **response_url ephemeral**, and then the public card was replaced on the **same response_url** — so the card-replace **overwrote the link** (it "flashed, then disappeared", per a live sandbox repro). Only `get` posts both an ephemeral *and* a card-replace, so only `get` hit it; the 1:1-DM path (a normal `PostMessage`, already decoupled) was unaffected.

This adds a **`chat.postEphemeral` seam** and delivers the channel/group link through it: a **standalone ephemeral**, scoped to the clicker and threaded to the card, with **zero shared state** with the response_url — so the card-replace physically can't touch it. The 1:1-DM branch stays on `PostMessage` (ephemerals don't render in a DM). A nil/erroring seam returns `false`, feeding the existing card-downgrade.

`chat.postEphemeral` is the canonical standalone ephemeral, independent of the interaction response — **correct by construction**, not a guess about the response_url collision mechanic.

## Scope

- [x] `apps/slack/`

## Changes

- **`internal/handler.go`** — `PostEphemeralFunc` type + `Config.PostEphemeral` field.
- **`cmd/slack_webapi.go` + `cmd/main.go`** — wire `chat.postEphemeral`, sharing the token-lookup / Grid-fallback / `ok:false` poster (`newSlackWebAPIPoster`) with `chat.postMessage`.
- **`internal/handler_agent_confirm.go`** — `deliverConfirmPrivate`'s channel branch now uses `PostEphemeral`; dropped the now-unused `responseURL` param (both branches are response_url-free).
- **Tests** — channel routing now asserts `chat.postEphemeral` (+ nil/error → card-downgrade); the cmd seam shape (`user` / `thread_ts` omitempty / `ok:false`).

## Test Plan

- [x] New tests added; channel-surface routing + the nil/error → card-downgrade are pinned
- [ ] **Manual testing pending deploy** — "ephemerals don't render in a 1:1 DM" / response_url-ephemeral rendering is *live Slack behavior* the unit tests can't model (they assert request shape). Verified-pending-deploy; I'll confirm the channel link renders post-deploy.

## Related

Found bringing the live Slack agent up in sandbox (follow-on to #722, #724): a `get` approved in a channel thread showed the link flash and immediately get overwritten by the "Handled" card.

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -154,6 +154,9 @@ func run() error {
 	// Both PostMessage seams share the per-workspace token lookup + Grid fallback with
 	// the slash-command modals.
 	postMessage := newSlackPostMessageFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
+	// chat.postEphemeral seam: delivers a get's one-time link privately in a channel as a
+	// standalone ephemeral (the response_url ephemeral collides with the card-replace).
+	postEphemeral := newSlackPostEphemeralFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostEphemeralURL, nil)
 	// markdown_text seam for the agent's free-text answer, so a channel reply renders
 	// standard Markdown the same way the streaming pane does (see PostMarkdownMessage).
 	postMarkdownMessage := newSlackPostMarkdownMessageFuncWithTokenLookup(workspaceTokenLookup, userAgent, slackChatPostMessageURL, nil)
@@ -253,6 +256,7 @@ func run() error {
 		AgentLLM:                    agentLLM,
 		AgentStore:                  agentStore,
 		PostMessage:                 postMessage,
+		PostEphemeral:               postEphemeral,
 		PostMarkdownMessage:         postMarkdownMessage,
 		AgentDisabled:               agentDisabled,
 		PostMessageBlocks:           postMessageBlocks,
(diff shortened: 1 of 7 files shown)


justin-layerv:
Thanks — all three non-blocking observations confirmed; no code changes:

**1. `PostEphemeral` outside the `agentEnabled()` gate** — intentional. It's needed only for the narrow channel-`get` delivery, so gating the *whole* agent on it would over-darken (every non-channel-get path would go dark for a seam it doesn't use). `cmd/main.go` co-wires `PostMessage` + `PostEphemeral`, so they're always both present in prod. And the degradation isn't silent: a nil/erroring seam logs a `Warn` **and** the card downgrades to "couldn't deliver here" — the user sees a failure, not a false success. Happy to add a cmd startup guard (`PostMessage` wired ⇒ `PostEphemeral` wired) for belt-and-suspenders if you'd prefer, but it's a deliberate-deletion-only hazard today.

**2. Threaded-ephemeral rendering** — agreed, that's the post-deploy confirmation. The reassuring part holds: the clicker just interacted with the card in-thread, so they have the thread open when the ephemeral lands. #725's group-DM boundary stays hide-not-leak.

**3. `user_not_in_channel`** — unreachable on the happy path (the clicker is a member by definition); handled as a downgrade purely defensively.

PR #730, opened 2026-06-13
fix(slack): name resources in the agent list by their channel alias

### LINKED ISSUE #729: list_resources shows channel-aliased resources without their alias (bare "- (url)")
## Context

Surfaced in review of [#728](https://github.com/layervai/qurl-integrations/pull/728) (which strips internal resource ids from agent tool output). `formatResourceLine` renders a resource by its `r.Alias` (qurl-service alias) or `r.Slug`, then type. A resource with **neither** renders a bare `- (url)` with no user-facing identifier.

This is the common protect-url case: the agent's URL upsert does **not** set a qurl-service alias ("Slack binds the channel alias separately"), and URLs have no slug — so the resource's **channel alias** (e.g. `$jptest4370`) lives in Slack's `alias_bindings`, not on `r.Alias`. `list_resources` therefore shows `- (url)` for it: the model has no handle to name it and the user can't tell which URL it is.

(Before #728 the line was `- (url, r_ws_…)` — the opaque id was the only "identifier", useless to users anyway, so #728 didn't regress *actionability*; it just made the empty line visible.)

## Fix

In the channel resource list builder (`ListResources` → `collectChannelResources` → `formatResourceLine`), join each resource's id against the channel's `alias_bindings` (rid → channel alias) and render the **channel alias** as the label when the resource has no qurl-service alias/slug. The channel alias map is already loaded for scope (`AllowedResourceIDsForChannel`), so the join is local.

## Acceptance

A channel-aliased URL resource appears in `list_resources` as `- $channelalias …` (its channel handle), never a bare `- (url)`.

Low priority — the resource is still reachable via its channel alias in `get`/etc.; this is the list *display* being uninformative, not a functional break.

justin-layerv:
## Summary

The conversation agent's `list_resources` rendered channel-protected URLs as a **handle-less `- (url)`**. `formatResourceLine` labeled resources by their **intrinsic** qURL alias or slug — but an agent-protected URL carries **neither**: its handle is the **channel-alias binding** (Slack-side, in `channel_policies.alias_bindings`), not on the qURL resource. So the model listed a resource it couldn't name.

This **joins the channel's alias bindings** (memoized per turn like the existing `channelAllowed` read, and now shared with `list_aliases`) and prefers the channel alias as the resource label: **`channelAlias > intrinsic alias > slug`**. For the agent-protected-URL case this surfaces `$channelalias`; it's purely additive otherwise. As a bonus it **reconciles the divergence** the #728 reviewer flagged — `list_aliases` showed the channel alias while `list_resources` showed the intrinsic one; now both prefer the channel alias.

A resource with **no channel alias AND no intrinsic alias/slug** (a rare direct resource-id policy row) still renders handle-less — there's no token to show; accepted as-is.

## Scope

- [x] `apps/slack/`

## Changes

- `internal/handler_agent_backend.go`:
  - new `channelPolicy` `sync.Once` memo (mirrors `channelAllowed` / `channelResources`) for `GetChannelPolicy`;
  - `ListResources` builds a `rid→channelAlias` map and passes it into `formatResourceLine`;
  - `formatResourceLine` gains a `channelAlias` param (top precedence) and stays a pure render helper — the caller owns the memoized reads;
  - `ListAliases` now reads via the shared memo (dedups the read in a both-tools turn).
- `internal/handler_agent_backend_test.go`:
  - **load-bearing** `TestAgentBackend_ListResources_ShowsChannelAliasForUnaliasedResource` — a resource with no intrinsic alias/slug, asserting `$channelAlias` appears (verified to render a bare `- (url)` *without* the join, so a regression can't pass green);
  - `TestFormatResourceLine` gains a channel-alias-precedence case.

## Test Plan

- [x] The join test was confirmed **RED** without the join, then green with it (the old fixtures hid it — the alias coincided with the intrinsic one)
- [x] **Delivery-confirmable** (like #728): the rendered output is asserted in tests; post-deploy I'll spot-check that `list_resources` shows `$alias` for a channel-protected URL

## Related

Closes #729 (surfaced reviewing #728). Follow-on to the live-agent fixes #722 / #724 / #727 / #728.

### DIFF
--- a/apps/slack/internal/handler_agent_backend.go
+++ b/apps/slack/internal/handler_agent_backend.go
@@ -149,15 +149,19 @@ func (b *agentBackend) ListResources(ctx context.Context, tc *agent.TurnContext)
 		return "No resources are protected in this channel yet.", nil
 	}
 	// Channel alias bindings name resources that carry no intrinsic alias/slug (e.g.
-	// agent-protected URLs). Build rid -> first bound channel alias.
+	// agent-protected URLs). Build rid -> channel alias. alias_bindings is map[alias]rid,
+	// so a resource can have several; GetChannelPolicy ranges that Go map (randomized
+	// order), so pick the lexicographically smallest deterministically — otherwise the
+	// label for a multi-bound resource would flip turn-to-turn.
 	entries, err := b.channelPolicy(ctx, tc)
 	if err != nil {
 		return b.fail("list resources: aliases", err)
 	}
 	channelAlias := make(map[string]string, len(entries))
 	for i := range entries {
-		if _, ok := channelAlias[entries[i].ResourceID]; !ok {
-			channelAlias[entries[i].ResourceID] = entries[i].Alias
+		rid, alias := entries[i].ResourceID, entries[i].Alias
+		if cur, ok := channelAlias[rid]; !ok || alias < cur {
+			channelAlias[rid] = alias
 		}
 	}
 	lines := make([]string, 0, len(resources))
--- a/apps/slack/internal/handler_agent_backend.go
+++ b/apps/slack/internal/handler_agent_backend.go
@@ -58,6 +58,16 @@ type agentBackend struct {
 	resourcesOnce sync.Once
 	resources     []client.Resource
 	resourcesErr  error
+
+	// Per-turn memo of the channel's alias bindings (GetChannelPolicy), shared by
+	// list_aliases and list_resources. list_resources joins it (rid -> $channelAlias) to
+	// name resources that carry no intrinsic alias/slug — e.g. an agent-protected URL,
+	// whose handle lives here, not on the qURL resource. Mirrors allowedOnce: same
+	// channel_policies row, a different projection (one extra GetItem/turn, negligible
+	// against the workspace resource scan).
+	policyOnce    sync.Once
+	policyEntries []slackdata.PolicyEntry
+	policyErr     error
 }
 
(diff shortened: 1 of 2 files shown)


justin-layerv:
**cr round 2 — disposition**

Round 1's one actionable nit (non-deterministic alias on multi-binding) is fixed in c66058e7: smallest-alias-on-collision + corrected the misleading "first" comment, with a flake-test (`...PicksSmallestAliasOnMultipleBindings`) verified RED against the old first-wins loop (~half of 30 runs picked `$zzz`) and green with the deterministic choice.

The three round-2 notes are all non-blocking and I'm keeping the code as-is, per your own recommendations:

1. **Two GetItems on the same row / eventual-consistency skew** — harmless here: a binding present in the allowed-union but absent from the policy read just falls back to slug/intrinsic. Consistent with the acknowledged "one extra GetItem/turn, negligible" altitude.
2. **`channelPolicy` error now fails the whole listing** — keeping the fail-fast; it matches the documented `resourcesOnce` posture, and `channelAllowed` already read the same row a few lines earlier so a failure here is a rare transient. (You noted you wouldn't change it either.)
3. **Lexicographic tiebreak** — correct given the data shape: `PolicyEntry.CreatedAt` is row-level (one `createdAt` copied to every entry), so per-alias recency isn't available. Smallest-alias is the only stable deterministic option.

`/simplify` re-run on the two-commit branch came back clean across all four angles. CI green (docker image build finishing).

PR #734, opened 2026-06-13
fix(slack): continue thread_broadcast agent follow-ups

### LINKED ISSUE #714: slack agent: continue channel thread follow-ups sent as thread_broadcast
## Context
Spun off from the #711 round-4 review (channel thread-reply continuity).

`shouldDispatchAgentEvent` drops any message with a subtype (`e.Subtype != ""`). That guard is correct for `message_changed`/`message_deleted`/`channel_join`/`bot_message`/etc., but it also excludes **`thread_broadcast`** — the subtype Slack sets when a user replies *in a thread* with "also send to channel" checked. Such a reply is a legitimate in-thread follow-up to the agent, but today it won't continue the conversation (the user has to @-mention the agent again).

This is an accepted **v1 bound** — channel follow-ups ship dark behind `QURL_AGENT_CHANNEL_FOLLOWUPS`, and broadcasts are a narrow case. Tracking it so it isn't lost.

## Scope / acceptance criteria
- Admit `subtype == "thread_broadcast"` through `shouldDispatchAgentEvent` (keep dropping all other subtypes) when channel follow-ups are enabled and it's a thread reply.
- Confirm the downstream `agentChannelFollowupDropped` "is it the agent's own thread?" gate and the turn path behave correctly for a broadcast event (it carries `thread_ts` + channel; the agent's reply should be a normal threaded reply, not itself a broadcast).
- Add a `TestShouldDispatchAgentEvent` case: `thread_broadcast` in the agent's thread is admitted (flag on), still dropped (flag off); other subtypes stay dropped.
- Land before broad enablement, alongside #712.

## Repro
With follow-ups enabled: @-mention the agent in a channel → it replies in-thread → reply to it using "also send to channel" → the agent does not respond (the broadcast reply is dropped at the subtype guard).

justin-layerv:
## Summary
- allow Slack `thread_broadcast` message subtypes through the agent dispatch gate only when channel follow-ups are enabled and the event is a channel thread reply
- keep all other subtyped messages dropped, including subtyped DMs and non-thread channel messages
- add coverage for dispatch gating plus a full handler path that replies normally in the existing thread

## Business relevance
This preserves Slack agent conversation continuity when a user replies in-thread and also broadcasts the reply to the channel, reducing confusing silent drops without widening the agent to unrelated channel chatter.

## Validation
- `go test -run 'TestShouldDispatchAgentEvent|TestHandleEvent_AgentRepliesToThreadBroadcastFollowup' ./apps/slack/internal`
- `go test -count=1 ./...`
- `go vet ./...`
- `git diff --check`

Closes #714

### DIFF
--- a/apps/slack/internal/handler_agent.go
+++ b/apps/slack/internal/handler_agent.go
@@ -23,6 +23,7 @@ const (
 	slackEventTypeAssistantThreadContextChanged = "assistant_thread_context_changed"
 	slackEventTypeAppHomeOpened                 = "app_home_opened"
 	slackChannelTypeIM                          = "im"
+	slackMessageSubtypeThreadBroadcast          = "thread_broadcast"
 )
 
 // agentProposalPreviewPrefix prefixes a proposed-mutation reply while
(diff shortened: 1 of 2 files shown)


justin-layerv:
Addressed Claude review: updated the shouldDispatchAgentEvent doc comment to call out the thread_broadcast carve-out. I’m intentionally keeping the repeated subtype checks explicit across app_mention, DM, and channel-message paths so each surface stays visibly fail-closed.

justin-layerv:
Confirmed the non-blocking Slack event-shape note against Slack's current docs: thread_broadcast is the message subtype Slack sends when a thread reply is broadcast to the channel, and the older reply_broadcast subtype is no longer served. The docs describe thread_broadcast as a pointer/reference to the actual thread, which matches this PR's single-event handling assumption. Sources: https://docs.slack.dev/reference/events/message/thread_broadcast and https://docs.slack.dev/reference/events/message/reply_broadcast

PR #736, opened 2026-06-13
feat(slack): stream agent replies in channel mentions

### LINKED ISSUE #706: slack agent: stream the reply in public channels (app_mention), not just the pane

justin-layerv:
## Summary
- Stream Slack agent replies for channel `app_mention` turns using the existing native stream finalizer.
- Pass a named `AgentStreamStart` request into `chat.startStream` so bot-token team and recipient team stay distinct for shared-channel/Grid cases.
- Add turn-level coverage for channel streaming: no double post, proposal card still posts, and partial stream errors finalize without a fallback double post.

## Business Relevance
This gives channel users the same progressive agent response UX as the assistant pane, reducing perceived latency in public Slack conversations while preserving existing pane behavior and safe fallbacks.

## Validation
- `go test ./...`
- `go vet ./...`
- `golangci-lint run --new-from-rev=origin/main --timeout=5m ./...`
- `make build-slack build-cli`

Closes #706

### DIFF
--- a/apps/slack/cmd/slack_webapi.go
+++ b/apps/slack/cmd/slack_webapi.go
@@ -527,21 +527,25 @@ func newSlackAgentStreamPortWithTokenLookup(lookup slackBotTokenLookup, userAgen
 	}
 }
 
-func (p *slackAgentStreamPort) StartStream(ctx context.Context, teamID, enterpriseID, channelID, threadTS, recipientUserID string) (string, error) {
-	// RecipientTeamID is the caller's workspace team — correct for the pane (DM) turns this
-	// serves today, where the recipient and the owning workspace are the same. The channel /
-	// Enterprise-Grid streaming follow-up (#706) can stream to a recipient in a DIFFERENT team,
-	// where recipient_team_id must be the recipient's team, not teamID — revisit this there.
+func (p *slackAgentStreamPort) StartStream(ctx context.Context, start *internal.AgentStreamStart) (string, error) {
+	if start == nil {
+		return "", errors.New("chat.startStream: missing start request")
+	}
 	body, err := json.Marshal(struct {
 		Channel         string `json:"channel"`
 		ThreadTS        string `json:"thread_ts,omitempty"`
 		RecipientTeamID string `json:"recipient_team_id,omitempty"`
 		RecipientUserID string `json:"recipient_user_id,omitempty"`
-	}{Channel: channelID, ThreadTS: threadTS, RecipientTeamID: teamID, RecipientUserID: recipientUserID})
+	}{
+		Channel:         start.ChannelID,
+		ThreadTS:        start.ThreadTS,
+		RecipientTeamID: start.RecipientTeamID,
+		RecipientUserID: start.RecipientUserID,
+	})
 	if err != nil {
 		return "", fmt.Errorf("chat.startStream request marshal: %w", err)
 	}
-	raw, err := p.start.gridPost(ctx, teamID, enterpriseID, body)
+	raw, err := p.start.gridPost(ctx, start.TeamID, start.EnterpriseID, body)
 	if err != nil {
 		return "", err
 	}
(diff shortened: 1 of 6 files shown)


justin-layerv:
@claude Please run a /simplify pass on this PR. Focus only on whether the implementation and tests can be simplified without losing behavior or coverage. Include nits/optional simplifications; if there are none, please say that the simplify pass is clean.

justin-layerv:
Addressed the review feedback in 701066a:

- Added `source_team` as a second shared-channel recipient-team hint after `user_team`, before falling back to the event `team_id`.
- Added explicit tests for `source_team`, same-workspace `team_id` fallback, and the no-recipient-team posted-path fallback.
- Updated the stale call-site comment in `processAgentEvent` so it no longer says streaming is pane-only.

For the verification question: Slack public `app_mention` docs say app_mention events can be treated similarly to message events, but their basic examples do not enumerate shared-channel extras. This patch now supports both shared-channel team fields (`user_team` and `source_team`) and keeps the same posted-reply fallback when Slack does not provide any recipient team.

justin-layerv:
@claude Please run a fresh /simplify pass on the current PR head (701066a). Focus only on whether the implementation and tests can be simplified without losing behavior or coverage. Include nits/optional simplifications; if there are none, please say that the simplify pass is clean.

justin-layerv:
@claude Please run a fresh /simplify pass and code review on the current PR head (704d58c). Please include optional/nit feedback if anything remains; I am treating those as in scope for this PR.

justin-layerv:
@claude Please run a fresh /simplify pass and code review on the current PR head (70ba3a3a). I addressed the previous code nits that fit the repo lint rules: split the helper fallback test from the streamer fallback test, added the user_team-over-source_team precedence assertion, documented the IM routing short-circuit, and replaced the structured "tunnel" fixture with client.ResourceTypeTunnel to satisfy goconst. I intentionally kept AgentStreamStart as a pointer because this is a six-string request object and passing it by value conflicts with the repo lint direction for heavy params; I also kept the test constants because inlining them made golangci-lint fail with goconst. Current local validation is green: go test ./..., go test -race -count=1 ./..., go vet ./..., golangci-lint run --new-from-rev=origin/main --timeout=5m ./..., and make build-slack build-cli. Please include optional/nit feedback if anything remains; I am treating those as in scope.

justin-layerv:
@claude Please run a fresh /simplify pass and code review on the rebased current PR head (b7cfcb5a). I addressed the prior current-head nits that fit the repo lint rules: split the helper fallback test from the streamer fallback test, added the user_team-over-source_team precedence assertion, documented the IM routing short-circuit, and kept lint-required test constants instead of inlining. AgentStreamStart remains a pointer because passing this six-string request object by value conflicts with the repo lint direction for heavy params. Current local validation on the rebased branch is green: go test ./..., go test -race -count=1 ./..., go vet ./..., golangci-lint run --new-from-rev=origin/main --timeout=5m ./..., and make build-slack build-cli. Please include optional/nit feedback if anything remains; I am treating those as in scope.

justin-layerv:
Tracked the only remaining external/manual verification from the review thread in #754: live Slack Connect / Enterprise Grid shared-channel smoke for app_mention streaming. This is not repo-local code feedback; current PR behavior is covered by tests and degrades to the posted-reply path if Slack rejects a recipient pairing.

PR #737, opened 2026-06-13
fix(slack): make agent acks surface-exclusive

### LINKED ISSUE #693: slack agent: make working-on-it acks surface-exclusive (pane setStatus vs channel reaction) post-#1004

justin-layerv:
## Summary

Make the Slack agent's working-on-it indicator surface-exclusive for the assistant-pane rollout, behind an explicit fail-safe rollout flag. Default behavior stays additive before the Slack pane is enabled; setting `QURL_AGENT_SURFACE_EXCLUSIVE_ACKS=true` after the #1004 smoke gate makes pane (`message.im`) turns use native `assistant.threads.setStatus` only, while channel turns keep the reaction ack only.

Business relevance: this gives Slack users one clear progress indicator per surface once the pane is live, removes the pane double-ack path, and avoids spending two cosmetic round trips before the agent starts answering, without risking indicator loss or Warn-per-turn noise during the manifest rollout window.

## Scope

- [x] `apps/slack/`
- [ ] `apps/teams/`
- [ ] `apps/discord/`
- [ ] `apps/cli/`
- [ ] `apps/zapier/`

## Changes

- Add `QURL_AGENT_SURFACE_EXCLUSIVE_ACKS` / `Config.AgentSurfaceExclusiveAcks` as the post-pane rollout switch.
- Default/pre-pane mode: `message.im` turns keep the existing reaction fallback and attempt native `setStatus` best-effort with Debug-level failures.
- Exclusive/post-pane mode: `message.im` turns use native `setStatus` only; setStatus failures log at Warn.
- Channel turns always use the reaction add/remove path and never attempt `setStatus`.
- Add a startup Warn for the exclusive-acks misconfig where `AssistantThreads` is unwired and the kill switch is not active.
- Extend tests to pin default additive fallback, exclusive pane status-only behavior, channel reaction-only behavior, Warn-level failures, nil-seam/default fallback behavior, panic cleanup/no-reaction paths, and the env flag fail-safe.

## Test Plan

- [x] `go test ./apps/slack/internal -run 'TestAgent(Status|Ack)'`
- [x] `go test ./apps/slack/cmd -run 'TestReadAgentSurfaceExclusiveAcks'`
- [x] `go test ./...`
- [x] `go vet ./...`
- [x] `go build ./...`
- [x] `golangci-lint v2.10.1 run --timeout=5m ./...` — `0 issues`
- [x] `go tool govulncheck ./...` — no called vulnerabilities
- [x] `go test -race -count=1 -coverprofile=coverage.out -covermode=atomic ./apps/slack/... ./shared/...` — total coverage 81.0%

## Related Issues

Closes #693.

Go-live note: layervai/qurl-integrations-infra#1004 remains the live Slack manifest/smoke gate. After that smoke confirms the pane status shows and auto-clears, infra should set `QURL_AGENT_SURFACE_EXCLUSIVE_ACKS=true` to activate the exclusive pane behavior. Once exclusive mode is enabled, persistent `assistant.threads.setStatus` failures will log Warn per pane turn and there is intentionally no reaction fallback, so treat Warns there as status-path breakage to investigate.

### DIFF
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -184,6 +184,7 @@ func TestLogAgentSurfaceState_ExclusiveAcksWarnsWhenAssistantThreadsUnwired(t *t
 		{name: "exclusive with assistant threads", state: wired},
 		{name: "exclusive without assistant threads", state: with(wired, func(s *agentSurfaceState) { s.assistantThreadsWired = false }), wantWarn: true},
 		{name: "default without assistant threads", state: with(wired, func(s *agentSurfaceState) { s.exclusiveAcksFlag = false; s.assistantThreadsWired = false })},
+		{name: "killed suppresses exclusive warning", state: with(wired, func(s *agentSurfaceState) { s.killed = true; s.assistantThreadsWired = false })},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -174,6 +174,31 @@ func TestReadAgentConfirmEnabled(t *testing.T) {
 	}
 }
 
+func TestReadAgentSurfaceExclusiveAcks(t *testing.T) {
+	cases := []struct {
+		name string
+		val  string
+		want bool
+	}{
+		{name: "unset is off until pane smoke", val: "", want: false},
+		{name: "true enables", val: "true", want: true},
+		{name: "1 enables", val: "1", want: true},
+		{name: "false stays off", val: "false", want: false},
+		{name: "0 stays off", val: "0", want: false},
+		// Fail-safe: a typo must keep the pre-pane reaction fallback, not remove it.
+		{name: "garbage fails safe to off", val: "enable", want: false},
+		{name: "yes fails safe to off", val: "yes", want: false},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			t.Setenv("QURL_AGENT_SURFACE_EXCLUSIVE_ACKS", tc.val)
+			if got := readAgentSurfaceExclusiveAcks(); got != tc.want {
+				t.Fatalf("readAgentSurfaceExclusiveAcks(%q) = %v, want %v", tc.val, got, tc.want)
+			}
+		})
+	}
+}
+
 func TestReadAgentDefaultEnabled(t *testing.T) {
 	cases := []struct {
 		name string
--- a/apps/slack/cmd/agent_wiring_test.go
+++ b/apps/slack/cmd/agent_wiring_test.go
@@ -174,6 +174,38 @@ func TestReadAgentConfirmEnabled(t *testing.T) {
 	}
 }
 
+func TestLogAgentSurfaceState_ExclusiveAcksWarnsWhenAssistantThreadsUnwired(t *testing.T) {
+	wired := agentSurfaceState{llmWired: true, storeWired: true, postWired: true, assistantThreadsWired: true, exclusiveAcksFlag: true}
+	cases := []struct {
+		name     string
+		state    agentSurfaceState
+		wantWarn bool
+	}{
+		{name: "exclusive with assistant threads", state: wired},
+		{name: "exclusive without assistant threads", state: with(wired, func(s *agentSurfaceState) { s.assistantThreadsWired = false }), wantWarn: true},
+		{name: "default without assistant threads", state: with(wired, func(s *agentSurfaceState) { s.exclusiveAcksFlag = false; s.assistantThreadsWired = false })},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			logs := captureSlog(t)
+			logAgentSurfaceState(tc.state)
+			var found bool
+			for _, rec := range logs() {
+				msg, _ := rec["msg"].(string)
+				if strings.Contains(msg, "QURL_AGENT_SURFACE_EXCLUSIVE_ACKS is set but AssistantThreads is not wired") {
+					found = true
+					if rec["level"] != "WARN" {
+						t.Fatalf("warning level = %v, want WARN", rec["level"])
+					}
+				}
+			}
+			if found != tc.wantWarn {
+				t.Fatalf("exclusive ack wiring warning found = %v, want %v; records=%v", found, tc.wantWarn, logs())
+			}
+		})
+	}
+}
+
 func TestReadAgentSurfaceExclusiveAcks(t *testing.T) {
 	cases := []struct {
 		name string
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -194,6 +194,7 @@ func run() error {
 	agentDisabled := readAgentKillSwitch()
 	agentConfirmEnabled := readAgentConfirmEnabled()
 	agentChannelFollowups := readAgentChannelFollowups()
+	agentSurfaceExclusiveAcks := readAgentSurfaceExclusiveAcks()
 	// Per-workspace toggle default: false during the staged opt-in rollout, flipped
 	// true at GA (every workspace on unless it explicitly opted out). Fail-safe to
 	// false. The per-workspace flag itself lives in workspace_mappings (AdminStore).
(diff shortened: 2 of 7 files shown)


justin-layerv:
@claude /simplify

justin-layerv:
Addressed the first review round in `a4e741a`:

- Added `QURL_AGENT_SURFACE_EXCLUSIVE_ACKS` / `Config.AgentSurfaceExclusiveAcks` so the app can deploy before infra#1004 without removing the pre-pane DM reaction fallback or logging Warn on every pre-pane `im` turn.
- Default mode now preserves the additive behavior: reaction fallback + best-effort `setStatus`, with `setStatus` failures at Debug.
- Exclusive mode is the post-#1004 path: pane turns use `setStatus` only, and failures log at Warn.
- Added tests for the default fallback, exclusive pane status-only behavior, channel reaction-only behavior, Warn-vs-Debug failure logging, and the env flag fail-safe.
- Took the /simplify point about avoiding the closure-return helper: `startAgentWorkingAck` now returns only whether a reaction needs deferred cleanup. I kept that helper because it owns the rollout flag decision and keeps `processAgentEvent` under the repo's gocyclo threshold; I also extracted the existing operating-channel selection out of the large turn function.

Re-ran local validation after the change: focused tests, `go test ./...`, `go vet ./...`, `go build ./...`, pinned `golangci-lint v2.10.1` (0 issues), `go tool govulncheck ./...`, and Slack race/coverage (81.0%).

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the round-2 Claude notes in `c9d0706`:

- Restored the `agentAckTimeout` comment's pre-pane/default-mode caveat: `im` turns can still attempt both the reaction fallback and `setStatus` until `QURL_AGENT_SURFACE_EXCLUSIVE_ACKS=true` flips after the infra#1004 smoke gate. The PR closes #693 by landing the guarded mechanism; prod cost reduction activates when infra flips the explicit post-smoke flag.
- Clarified `startAgentWorkingAck`'s return doc to say the reaction add was attempted, matching the best-effort/no_reaction-tolerant cleanup behavior.
- Re-ran `go test ./apps/slack/internal -run 'TestAgent(Status|Ack)'` after the comment cleanup.

The nil `AssistantThreads` + exclusive flag behavior is intentionally fail-safe for turn delivery but indicator-less under misconfiguration; the PR body keeps the flag tied to post-smoke infra enablement so that should not be set before the native pane seam is live.

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the latest `/simplify`/review feedback in `c05f3cab`:

- Restored panic-safety ordering for default pane turns: reaction cleanup is now registered before attempting the native pane `setStatus`, so a status-path panic cannot strand the fallback reaction.
- Tightened `TestAgentStatus_BestEffortDoesNotFailTurn` to assert exclusive mode does not fall back to reactions when `setStatus` fails.
- Added `TestAgentStatus_DefaultPaneReactionClearedOnStatusPanic` to pin the cleanup invariant directly.
- Re-ran `go test ./apps/slack/internal -run 'TestAgent(Status|Ack)'` and `go vet ./apps/slack/...`.

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the latest optional `/simplify` nit in `af35fa99`:

- Removed the redundant outer `ChannelType == im` guard and now rely on `setAgentThinkingStatus`'s internal channel/nil-seam guard, which keeps the safer local invariant in the helper while simplifying the call site.
- Re-ran `go test ./apps/slack/internal -run 'TestAgent(Status|Ack)'`, `go vet ./apps/slack/...`, and `go tool govulncheck ./...`.

justin-layerv:
@claude /simplify

justin-layerv:
@claude /simplify

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the latest optional review items in `0bf5feba`:

- Added a startup Warn if `QURL_AGENT_SURFACE_EXCLUSIVE_ACKS` is set while `AssistantThreads` is not wired, so the intentionally indicator-less exclusive path is loud under misconfiguration.
- Added an exclusive-mode setStatus panic test: no reaction fallback is touched, and the recover path still posts `agentErrorReply`.
- Tightened the Debug/Warn status-log assertions to match the status failure message instead of treating any unrelated Warn as a failure.

Re-ran full local validation after the change: focused Slack tests, `go test ./...`, `go vet ./...`, `go build ./...`, pinned `golangci-lint v2.10.1` (0 issues), `go tool govulncheck ./...`, and Slack/shared race coverage.

justin-layerv:
@claude /simplify

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the latest Claude Review optional nits in `a66d979d`:

- Made `setAgentThinkingStatus` error branches symmetric with an explicit return after the exclusive Warn path.
- Replaced the remaining new `👀` prose in this PR path with `eyes reaction` wording.

Re-ran full local validation: focused Slack tests, `go test ./...`, `go vet ./...`, `go build ./...`, pinned `golangci-lint v2.10.1` (0 issues), `go tool govulncheck ./...`, and Slack/shared race coverage.

justin-layerv:
@claude /simplify

justin-layerv:
Addressed the latest `/simplify` low-severity/optional items in `f834eee4`:

- Suppressed the exclusive-ack wiring Warn while the global kill switch is active, matching the existing confirm-mode boot-log pattern.
- Added default-mode status-thread auto-clear coverage.
- Added default nil-AssistantThreads coverage proving the reaction fallback remains active before the exclusive rollout flips.
- Tightened the `startAgentReactionAck` doc comment to describe cleanup registration rather than implying a successful reaction add.

Rebased onto current `origin/main` and re-ran full local validation: focused Slack tests, `go test ./...`, `go vet ./...`, `go build ./...`, pinned `golangci-lint v2.10.1` (0 issues), `go tool govulncheck ./...`, and Slack/shared race coverage.

justin-layerv:
@claude /simplify

PR #745, opened 2026-06-13
fix(discord): parse qurl webhook owner from raw body

### LINKED ISSUE #492: [discord-bot] Receiver should parse rawBody directly for owner_id extraction (defense-in-depth)
## Context

\`apps/discord/src/routes/qurl-webhook.js\` extracts \`owner_id\` from \`req.body\` (Express-parsed) before HMAC verification. Today this is safe because the \`rawBodyJson\` middleware (\`apps/discord/src/server.js\`) is mounted before the router with a hard \`limit: '1mb'\` cap.

The routing dependency is a footgun for a future \`server.js\` refactor that re-orders mounts: a fall-through to the default Express JSON parser would lose the size cap and let large unbounded payloads through the pre-HMAC parse.

## Acceptance criteria

- Receiver extracts \`owner_id\` by re-parsing \`req.rawBody.toString()\` directly (still bounded by the 1MB middleware cap).
- Test pinning that the receiver doesn't depend on \`req.body\` having been pre-parsed.
- Comment in \`server.js\` notes that the 1MB cap belongs to the receiver's threat model, not just the middleware's.

## Why deferred from #485

Reviewer flagged as smaller-item (\"today it's safe\"). Defense-in-depth, not blocking. Filed at reviewer's suggestion.

justin-layerv:
## Summary
- Re-parse `req.rawBody.toString()` in the qURL webhook verifier and use that parsed payload for owner-secret routing and event handling.
- Document that the `/webhooks` 1mb raw-body cap is part of the receiver threat model.
- Add a regression test proving the receiver still succeeds when `req.body` is clobbered before the router.

## Business relevance
This reduces webhook receiver fragility around future Express middleware refactors, keeping Discord view/expiry webhooks bounded and correctly authenticated before trust is granted.

## Validation
- `npm ci`
- `npx jest --runTestsByPath tests/qurl-webhook.test.js tests/qurl-webhook-expired.test.js tests/server.test.js --coverage=false`
- `npm run lint`
- `npm test -- --ci --silent`
- `npm audit --audit-level=high --omit=dev`
- `docker build -t discord-bot-test .`

Closes #492

### DIFF
--- a/apps/discord/src/routes/qurl-webhook.js
+++ b/apps/discord/src/routes/qurl-webhook.js
@@ -105,14 +105,29 @@ function verifyAndResolve(req) {
   // SECURITY: pre-HMAC parse depth/size MUST stay bounded by the
   // rawBodyJson middleware cap (server.js, `limit: '1mb'`). Loosening
   // that cap or adding extended type coercion here widens the
-  // pre-trust window — req.body is attacker-controlled JSON until
-  // the HMAC check below succeeds.
+  // pre-trust window — this parsed body is attacker-controlled JSON
+  // until the HMAC check below succeeds.
   // The lookup itself grants no trust: a forged owner_id either
   // misses the cache (→ 401) or hits a real secret the attacker
   // can't forge against (→ 401).
-  const ownerId = req.body && typeof req.body.owner_id === 'string' ? req.body.owner_id : null;
+  // Under today's express.json wiring, malformed JSON is rejected
+  // before this route runs; this catch is forward defense for raw-body
+  // parser refactors and parser-differential cases.
+  // Deliberately do not strip a UTF-8 BOM: qurl-service's Go JSON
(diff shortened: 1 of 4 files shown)


justin-layerv:
Confirmed the latest non-blocking review observations on head ab3e79ab:

- The redundant parse is deliberate: `express.json` keeps the current 1mb cap and malformed-JSON rejection before the router, while the route reparses `rawBody` so owner routing and event handling use the signed bytes.
- `rawBody.toString()` is intentionally UTF-8 only; qurl-service emits UTF-8 JSON and alternate charsets/BOM-prefixed bodies are outside the accepted wire contract.
- The malformed-JSON test intentionally allows current 500 vs future 400 so it pins the parser-boundary behavior without blocking a tighter global parser-error handler later.

No follow-up issue needed from these confirmations; the review did not request a code change.

PR #748, opened 2026-06-13
fix(slack): escape display names in mrkdwn replies

### LINKED ISSUE #532: Defense-in-depth: escape Slack mrkdwn in /qurl list resource descriptions
Follow-up from PR #530 (cr review, "track separately" — pre-existing, not a regression).

## Observation
`formatTunnelListLine` (and the legacy renderer before it) appends `r.Description` verbatim into the `/qurl list` ephemeral payload. A description containing Slack control sequences — `<!channel>`, `<!here>`, `<@U…>` — renders as a real Slack token in the message.

## Impact (low)
- The list reply is `response_type: ephemeral`, so only the invoking user sees it and `<!channel>`/`<!here>` do **not** broadcast a notification to others. The effect is a rendered (cosmetic) mention in the invoker's own view.
- Resource descriptions today are set by the API/SDK and the `/qurl tunnel install` flow (e.g. "Slack tunnel install for <slug>"), not arbitrary end-user free-text in Slack, which further limits exposure.

Not a regression (the pre-PR renderer had the same behavior) and not exploitable for cross-user notification, but worth hardening if descriptions ever become broadly user-editable.

## Proposed fix
Run `r.Description` (and any other server-sourced free-text rendered into Slack payloads) through an mrkdwn-escape helper (escape `&`, `<`, `>` per Slack's escaping rules) before interpolation. Apply consistently across the Slack app's user-facing renderers.

## Acceptance criteria
- A resource description containing `<!channel>` / `<@U…>` renders as inert text in `/qurl list`.
- A shared escape helper is used at all server-text→Slack-payload sites in the Slack app.

justin-layerv:
## Summary
- Escape resource descriptions/display names at Slack mrkdwn render boundaries for `/qurl aliases`, tunnel install confirmations, and display-name update confirmations.
- Keep stored and PATCHed description values raw; escaping happens only when building Slack payload text.
- Add regression coverage for `/qurl list`, `/qurl aliases`, tunnel install, and display-name success copy with Slack control sequences.

## Business relevance
Closes #532. This prevents server-sourced qURL resource descriptions/display names from rendering Slack control sequences such as `<!channel>` or `<@U...>` as live mentions or formatting in user/admin Slack replies, keeping the Slack integration predictable as descriptions become more broadly editable.

## Test plan
- `go test ./apps/slack/...`
- `go test -count=1 ./...`
- `go build ./...`
- `go test -race -count=1 -coverprofile=coverage.out -covermode=atomic ./apps/slack/... ./shared/...` (total coverage: 81.0%)
- `go tool govulncheck ./...`
- `golangci-lint run --timeout=5m --allow-parallel-runners --new-from-rev=origin/main ./apps/slack/... ./shared/...` (0 issues)

### DIFF
--- a/apps/slack/internal/handler_aliases.go
+++ b/apps/slack/internal/handler_aliases.go
@@ -180,7 +180,7 @@ func groupsNeedResolution(groups []aliasGroup) bool {
 // (see the call site for why the list path, not the per-id path, is the
 // reliable slug source). Every resource on the page is indexed —
 // intentionally NOT filtered to type=tunnel like /qurl list — so a
-// legacy URL binding still resolves its target_url and renders the
+// legacy URL binding still resolves its target_url and renders the escaped
 // "<url> (legacy URL) → $alias" line.
 //
 // Best-effort: a fetch failure yields (nil, false) — a nil-map lookup
@@ -304,7 +304,7 @@ func formatAliasGroupLine(target, slug, description string, aliases []string) st
 			left += " — " + escapeMrkdwnText(description)
 		}
 	case target != "":
-		left = target + " (legacy URL)"
+		left = escapeMrkdwnText(target) + " (legacy URL)"
 	default:
 		// No slug resolved — never fall back to the opaque resource_id.
 		// Show the channel aliases alone so the row still renders and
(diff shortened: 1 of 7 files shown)


justin-layerv:
Tracked the second Claude pass's agent-context observation in #755. I’m keeping #748 focused on deterministic Slack mrkdwn payload renderers; the agent path feeds model context and posts free-text replies through markdown_text, so it needs its own read-to-reply fixture rather than a blind mrkdwn escape.

PR #757, opened 2026-06-13
perf(shared): cache workspace API key lookups

### LINKED ISSUE #263: Cache APIKey() lookups to drop KMS Decrypt per slash-command
## Summary

`shared/auth/DDBProvider.APIKey` is called on every `/qurl create` / `/qurl list` slash command (see `apps/slack/internal/handler.go:authenticatedClient`). Each call costs:

- One DDB `GetItem` (eventually consistent since PR #254 dropped `ConsistentRead`)
- One KMS `Decrypt` round-trip (~30-50ms p99)

For a steady-state workspace this is fine. But:
1. KMS `Decrypt` is rate-limited per CMK (~5,500 req/s shared across the account).
2. The cost compounds with the qURL client's `WithRetry(2)` policy on transient 5xx.

## Acceptance criteria

- Add an in-memory `sync.Map[teamID] -> {apiKey string, expiresAt time.Time}` cache to `DDBProvider`.
- TTL: 5 minutes is a reasonable starting point — long enough to drop the typical session, short enough that a rotation propagates within human-perceptible "try again" tolerance.
- Cache `MISS` falls through to the existing DDB+KMS path.
- `SetAPIKey` invalidates the cache entry (workspace key rotated).
- `DeleteAPIKey` invalidates the cache entry (workspace disconnected).
- Document the TTL trade-off in a code comment.

## Why this is a follow-up

PR #254 (`feat/slack-oauth`) introduced the DDB-backed key store. The caching is independent of that work, the slash-command surface is still low-traffic on the pilot workspaces, and the original cr review #1 already flagged it as a "consider an in-memory cache" suggestion rather than a merge blocker.

## Notes

- Watch out for `errors.Is(err, ErrWorkspaceNotConfigured)` — negative-cache that too with a short TTL or skip caching it (a fresh install would otherwise wait the TTL for the bot to notice the key).
- No external integrations need to change.

Surfaced by /pr-ready loop on PR #254.

justin-layerv:
## Summary

Cache successful `DDBProvider.APIKey` lookups in memory for five minutes so steady-state Slack slash commands avoid repeating one DynamoDB `GetItem` plus one KMS `Decrypt` per command.

Business relevance: this reduces latency and lowers shared CMK decrypt pressure for active workspaces while keeping key rotations and disconnects visible within a short operator/customer retry window.

## Scope

- [ ] `apps/slack/`
- [ ] `apps/teams/`
- [ ] `apps/discord/`
- [ ] `apps/cli/`
- [ ] `apps/zapier/`
- [x] `shared/` (triggers tests for ALL apps — coordinate with maintainers)

## Changes

- Adds a mutex-guarded per-workspace API key cache with a five minute TTL and inline comments documenting the rotation/disconnect trade-off.
- Caches only successful decrypts; not-configured/missing-key results still fall through to storage so fresh installs are noticed immediately.
- Coalesces concurrent cold/expired lookups per workspace so a burst shares one DDB+KMS fill.
- Uses generation-guarded invalidation so older in-flight reads cannot repopulate a rotated or deleted key.
- Retries healthy coalesced waiters when the owner lookup is canceled, so one timed-out slash-command request does not fan out a canceled result to other live requests.
- Seeds the cache after successful `SetAPIKey` writes so eventually-consistent post-write reads cannot pin an old key.
- Evicts after successful `DeleteAPIKey` calls and forces strongly consistent refills for one TTL so the deleting process cannot re-cache a revoked key from an eventually-consistent post-delete read.
- Adds tests for cache hits, expiry refresh, sweep cleanup, non-cached misses, concurrent coalescing, set/delete invalidation, stale in-flight reads, owner/waiter cancellation and retry, post-delete strong refill, and panic cleanup.

## Test Plan

- [x] `make check` passes locally
- [x] New tests added for new functionality
- [x] Manual testing not applicable; shared provider cache covered by unit/race tests and CI

Local validation:

- `go test -count=1 ./shared/auth`
- `go test -count=1 ./...`
- `go vet ./...`
- `GOLANGCI_LINT_CACHE=/tmp/qurl-integrations-golangci-cache go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --timeout=5m ./...`
- `PATH=/tmp/qurl-integrations-bin:$PATH GOLANGCI_LINT_CACHE=/tmp/qurl-integrations-golangci-cache make check`

Note: local `golangci-lint` is v2.12.2, while CI pins v2.10.1. The exact pinned version above passes with `0 issues`.

## Related Issues

Closes #263

Follow-ups: #765 tracks shared TTL/singleflight helper extraction; #766 tracks cross-instance invalidation if rotation/disconnect needs to be stricter than the accepted TTL window.

### DIFF
--- a/shared/auth/ddb_provider.go
+++ b/shared/auth/ddb_provider.go
@@ -140,13 +140,13 @@ type DDBProvider struct {
 	TableName string
 	Encryptor FieldEncryptor
 
-	// Cache successful APIKey lookups at the DDB/decrypt boundary, following
-	// the Slack bot-token cache pattern: per-workspace in-flight fills,
-	// generation-guarded invalidation, and inline expiry sweeps. The cache is
-	// process-local, so sibling instances can keep a decrypted key, including a
-	// revoked key, until the five-minute TTL from #263 expires; stricter
-	// cross-instance invalidation is tracked in #766. The decrypted key remains
-	// in heap memory for that TTL as the accepted KMS/latency trade-off.
+	// Cache successful APIKey lookups at the DDB/decrypt boundary with
+	// per-workspace in-flight fills, generation-guarded invalidation, and inline
+	// expiry sweeps. The cache is process-local, so sibling instances can keep a
+	// decrypted key, including a revoked key, until the five-minute TTL from #263
+	// expires; stricter cross-instance invalidation is tracked in #766. The
+	// decrypted key remains in heap memory for that TTL as the accepted
+	// KMS/latency trade-off.
 	// SetAPIKey seeds this process after a successful write. DeleteAPIKey evicts
 	// this process and forces strongly-consistent refills for one TTL so a
 	// same-process eventually-consistent post-delete read cannot re-cache the
@@ -157,8 +157,8 @@ type DDBProvider struct {
 	// generations. Generation entries are retained after invalidation even after
 	// an in-flight slot is deleted: that slot deletion detaches the old owner,
 	// which may still finish later, so resetting the generation to zero could let
-	// it cache an old key. The map only grows for workspaces that rotate or
-	// disconnect in this process, matching the sibling Slack token lookup.
+	// it cache an old key. The map grows for each workspace seeded or
+	// invalidated in this process and is retained for the process lifetime.
 	apiKeyCacheMu         sync.Mutex
 	apiKeyLookupInFlight  map[string]*apiKeyLookupCall
 	apiKeyCacheGeneration map[string]uint64
@@ -461,6 +461,9 @@ func (p *DDBProvider) finishAPIKeyLookup(workspaceID string, call *apiKeyLookupC
 	if result.err == nil && p.apiKeyCacheGeneration[workspaceID] == generation {
 		p.cacheAPIKey(workspaceID, result.apiKey, now)
 	}
+	// Waiters already attached to this fill receive its result even if a
+	// concurrent SetAPIKey/DeleteAPIKey invalidated the generation. The stale
+	// value is not cached, so only those already in-flight callers can observe it.
 	call.apiKeyLookupResult = result
 	if p.apiKeyLookupInFlight[workspaceID] == call {
 		delete(p.apiKeyLookupInFlight, workspaceID)
--- a/shared/auth/ddb_provider.go
+++ b/shared/auth/ddb_provider.go
@@ -136,11 +136,15 @@ type DDBProvider struct {
 	TableName string
 	Encryptor FieldEncryptor
 
-	// Cache successful APIKey lookups for a short window so steady-state slash
-	// commands do not spend one DDB read plus one KMS Decrypt each time. Five
-	// minutes smooths normal command sessions while still letting key rotations
-	// propagate within a human "try again" interval.
-	apiKeyCache sync.Map // map[string]cachedAPIKey
+	// Cache successful APIKey lookups at the DDB/decrypt boundary so every
+	// long-lived consumer avoids repeat KMS Decrypt calls during normal slash
+	// command bursts. The cache is process-local: rotation or disconnect on one
+	// serving instance cannot evict sibling instances, so they may continue using
+	// a previously decrypted key until this TTL expires. Five minutes keeps that
+	// cross-instance staleness bounded while still collapsing the common
+	// multi-command session; Slack bot-token caching stays shorter because it
+	// gates reinstall detection rather than qURL API retry pressure.
+	apiKeyCache sync.Map // map[string]*cachedAPIKey
 
 	// Now is injected so tests can pin the wall clock for configured_at /
 	// updated_at assertions without poking package-global state. Defaults
@@ -271,11 +275,10 @@ func (p *DDBProvider) APIKey(ctx context.Context, workspaceID string) (string, e
 
 	now := p.nowOrDefault()
 	if cached, ok := p.apiKeyCache.Load(workspaceID); ok {
-		entry, ok := cached.(cachedAPIKey)
+		entry, ok := cached.(*cachedAPIKey)
 		if ok && now.Before(entry.expiresAt) {
 			return entry.apiKey, nil
 		}
-		p.apiKeyCache.Delete(workspaceID)
 	}
 
 	// Eventually-consistent read is correct here: the per-workspace key
@@ -316,11 +319,22 @@ func (p *DDBProvider) APIKey(ctx context.Context, workspaceID string) (string, e
 		return "", errors.New("DDBProvider.APIKey: decrypted plaintext is empty")
 	}
 	apiKey := string(pt)
-	p.apiKeyCache.Store(workspaceID, cachedAPIKey{
+	p.cacheAPIKey(workspaceID, apiKey, now)
+	return apiKey, nil
+}
+
+func (p *DDBProvider) cacheAPIKey(workspaceID, apiKey string, now time.Time) {
+	entry := &cachedAPIKey{
 		apiKey:    apiKey,
 		expiresAt: now.Add(apiKeyCacheTTL),
+	}
+	p.apiKeyCache.Store(workspaceID, entry)
+	time.AfterFunc(apiKeyCacheTTL, func() {
+		cached, ok := p.apiKeyCache.Load(workspaceID)
+		if ok && cached == entry {
+			p.apiKeyCache.Delete(workspaceID)
+		}
 	})
-	return apiKey, nil
 }
 
 // SlackBotToken looks up the per-workspace Slack bot token captured during
--- a/shared/auth/ddb_provider.go
+++ b/shared/auth/ddb_provider.go
@@ -38,6 +38,7 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -89,6 +90,8 @@ const (
 	EnvWorkspaceStateKMSKeyARN = "WORKSPACE_STATE_KMS_KEY_ARN"
 )
 
+const apiKeyCacheTTL = 5 * time.Minute
+
 // ErrWorkspaceNotConfigured is the sentinel returned by APIKey when the
 // workspace has no row in the workspace_state table (i.e. the admin has
 // not yet completed /oauth/qurl/start → /callback). The Slack handler
@@ -133,10 +136,21 @@ type DDBProvider struct {
 	TableName string
 	Encryptor FieldEncryptor
 
+	// Cache successful APIKey lookups for a short window so steady-state slash
+	// commands do not spend one DDB read plus one KMS Decrypt each time. Five
+	// minutes smooths normal command sessions while still letting key rotations
+	// propagate within a human "try again" interval.
+	apiKeyCache sync.Map // map[string]cachedAPIKey
+
 	// Now is injected so tests can pin the wall clock for configured_at /
 	// updated_at assertions without poking package-global state. Defaults
 	// to time.Now.
 	Now func() time.Time
+}
+
+type cachedAPIKey struct {
+	apiKey    string
+	expiresAt time.Time
 }
 
 // SlackBotTokenInstall is the workspace-scoped Slack OAuth material persisted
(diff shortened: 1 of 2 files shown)


justin-layerv:
Addressed Claude review round 1 on the latest head:\n\n- documented the process-local/cross-instance staleness window for rotation and disconnect\n- added the TTL rationale versus the shorter Slack bot-token cache\n- kept the cache at the DDB/decrypt boundary intentionally for long-lived consumers\n- switched cache entries to exact-entry pointers and scheduled TTL cleanup so plaintext entries do not linger indefinitely after TTL\n- removed the expired-path delete that could race with a fresh store\n\nRe-running /simplify and waiting for fresh CI + Claude review.

justin-layerv:
Addressed Claude review round 2 and rebased onto current main:\n\n- added per-workspace generation tracking so SetAPIKey/DeleteAPIKey invalidation cannot be undone by an older in-flight APIKey read\n- added per-workspace in-flight coalescing so concurrent cold/expired lookups share one DDB+KMS fetch\n- added race-tested coverage for both the stale-read invalidation sequence and concurrent cold-fill coalescing\n- reran full make check locally with the CI-pinned golangci-lint version\n\nRe-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Review follow-up from the latest Claude pass:

- Replaced the per-cache-fill `time.AfterFunc` cleanup with a minute-gated inline sweep under `apiKeyCacheMu`, using the injected `Now` clock and matching the Slack token lookup pattern.
- Added a sweep-focused unit test so expired entries are reclaimed without a background timer.
- Documented the revoke/rotation visibility trade-off, the pointer to `newWorkspaceSlackTokenLookupWithInvalidation`, and why generation entries are retained after invalidation.
- Added a waiter-context comment for the owner-cancellation trade-off.

Local validation on the new head: `make check` passes.

justin-layerv:
Review follow-up from the latest Claude pass:

- Replaced the now-redundant `sync.Map` with a plain `map[string]*cachedAPIKey` guarded by `apiKeyCacheMu`, matching the Slack token cache pattern after the in-flight/generation coordination was added.
- Hardened the panic path by marking the owner lookup as finished before `finishAPIKeyLookup`, preventing a panic in finish from attempting a second close.
- Kept the 5-minute TTL because issue #263 explicitly accepts that starting cross-instance staleness window; the code comment now says that directly and names the revoked-key visibility trade-off.

Local validation on the new head: `make check` passes.

justin-layerv:
Review follow-up from the latest Claude pass:

- Seed `apiKeyCache` after successful `SetAPIKey` writes, while bumping generation and clearing in-flight reads, so an eventually-consistent post-write `GetItem` cannot pin the old key for the TTL.
- Added regression coverage for SetAPIKey serving the seeded key even when DDB would still return the old value.
- Added single-flight behavior tests for waiter cancellation and panic cleanup.
- Updated the PR description to remove the stale `sync.Map` wording and describe the mutex-guarded map, coalescing, generation guard, SetAPIKey seeding, and expanded tests.
- Rebased onto current `origin/main`; local `make check` passes on the rebased head.

justin-layerv:
Final Claude-note follow-up:

- Documented the residual eventual-consistency case and plaintext heap-residency trade-off in the API-key cache comment.
- Adjusted SetAPIKey cache seeding to use the injected clock directly for cache expiry while keeping the DDB timestamp UTC/RFC3339.
- Added explicit tests proving DDB transport errors and decrypt errors are not cached.
- Created follow-up issue #765 for extracting the duplicated TTL/singleflight/generation cache machinery into a shared helper.

Local validation on this head: `make check` passes.

justin-layerv:
Final latest-review follow-up:

- Added a DDBProvider struct note that the provider owns mutex-protected cache state and must be used by pointer / not copied after first use.
- Documented the exact SetAPIKey/APIKey raw-key round trip so cache seeding and DDB decrypt stay byte-consistent.
- Created follow-up #766 for cross-instance API key cache invalidation if rotation/disconnect needs stricter semantics than the issue #263 TTL window.
- Updated the PR body to list both deferred follow-ups (#765 and #766).

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Final coverage follow-up from the latest Claude pass:

- Added a cache-hit/canceled-context regression test documenting the intentional fast-path behavior.
- Added a coalesced waiter error test proving an owner DDB failure is shared with attached waiters and does not populate the cache.
- Rebased onto current `origin/main`; local `make check` passes on the rebased head.

This should close the remaining optional coverage gaps Claude called out.

justin-layerv:
Last Claude-note coverage follow-up:

- Added the owner-context-cancellation waiter test Claude suggested: a live coalesced waiter now explicitly observes the owner cancellation error, and the canceled fill is asserted not to populate the cache.

Local validation on this head before push: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude-note follow-up:

- Clarified the same-instance post-DeleteAPIKey stale-refill window in the cache comment, not just sibling-instance staleness.
- Stamped successful cache fills/seeds at completion time so the effective TTL mirrors the sibling Slack token cache instead of starting before DDB/KMS/write latency.

Local validation on this head before push: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Follow-up issue cleanup from the latest Claude pass:

- Updated #766 to cover both revocation stale windows: sibling pods that do not receive local invalidation, and same-instance post-DeleteAPIKey eventual-consistency refills that can pin a revoked key for one TTL.
- Added explicit acceptance criteria for cross-instance invalidation / no-cache cooldown / strongly-consistent post-delete refill options and coverage expectations.

No code change after the green final head.

justin-layerv:
Rebased onto current `origin/main` after GitHub reported the branch behind again.

Local validation on the rebased head: `make check` passes.

justin-layerv:
Latest Claude-note follow-up on the new head:

- Closed the same-instance post-DeleteAPIKey stale-refill window by forcing strongly consistent refills for one TTL after local delete invalidation.
- Added coverage for delete-vs-in-flight stale fills and post-delete strong refills so a revoked key is not re-cached from an eventually-consistent read.
- Mirrored the Slack cache cleanup nits: expired entries are removed on read miss, strong-read markers are swept, SetAPIKey seed uses the existing timestamp, and private seed/invalidate paths guard blank workspace IDs.
- Updated #766 to narrow the remaining deferral to cross-instance invalidation / slash-command backstop behavior.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude discussion-point follow-up:

- Changed `DDBProvider.APIKey` so a healthy coalesced waiter retries when the owner lookup fails only because the owner context was canceled/deadlined. That keeps one timed-out slash-command request from fanning out a canceled result to other live requests.
- Added API-level coverage proving a waiter succeeds after owner cancellation, plus the lower-level retry predicate coverage.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Final nit follow-up from Claude latest pass:

- `DDBProvider.APIKey` now honors an already-canceled caller context before returning a cached key, preserving the pre-cache cancellation contract while still avoiding extra DDB/KMS work.
- Updated the cache-hit test to assert canceled callers get `context.Canceled` and no additional `GetItem` occurs.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude review follow-up:

- Added the requested inline comment at `shouldRetryAPIKeyLookupAfterSharedError` explaining the DDB-brownout trade-off: a lower-layer `context.Canceled`/`DeadlineExceeded` can trigger sequential retries by healthy waiters, but each retry re-enters singleflight as a new owner, so it is bounded and not a fan-out spike.
- Re-ran `go test -race -count=1 ./shared/auth` and `make check`; both pass on this head.

Security/maintainer sign-off request for @Kevin-layerV and @vikramlayerv: issue #263 explicitly set a 5-minute TTL as the starting rotation propagation tolerance, and this PR documents that sibling instances may keep a decrypted/revoked key in heap and grant access until TTL. Please explicitly confirm that this 5-minute cross-instance revoked-key window is accepted for this PR; #766 tracks stricter cross-instance invalidation if we need a tighter revocation guarantee.

justin-layerv:
Latest nit follow-up:

- Added coverage for the `context.DeadlineExceeded` branch in `shouldRetryAPIKeyLookupAfterSharedError`, including the no-retry case when the waiter context is already canceled.
- Added coverage that blank private cache mutation guards in `invalidateAPIKeyCache` and `seedAPIKeyCache` are no-ops.
- Updated #765 so the shared-helper follow-up also evaluates the single-global-mutex versus keyed/sharded-locking trade-off.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude follow-up:

- Removed the dead `now` parameter from `fetchAndFinishAPIKeyLookup`.
- Rephrased the shared-provider doc comment so it describes the Slack bot-token cache pattern instead of naming an unexported function in another package.
- Removed the timing sleep from the owner-cancellation retry test.
- Verified `apps/slack/cmd/main.go` wires one shared `*DDBProvider`: `run()` constructs `ddbProvider`, passes it to the slash handler via `AuthProvider`, and passes the same pointer to `buildOAuthConfig`, whose callback calls `cfg.Provider.SetAPIKey`.
- Tightened the generation-retention comment to make clear why it cannot be reclaimed just because the in-flight map entry is gone; invalidation detaches the old owner and it may still finish later.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude follow-up:

- Trimmed the large `DDBProvider` cache field comment down to the durable invariants: process-local TTL staleness, plaintext heap residency, seed-on-write, delete eviction, and post-delete strong refill.
- Updated #765 to explicitly carry both the global-lock/sharding decision and the generation-retention/reclaim policy, including the detached-owner safety constraint.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude follow-up:

- Removed the misleading Slack-cache cross-references from the shared provider comments.
- Corrected the generation-growth comment: entries are created by seed or invalidation in this process and retained for the process lifetime.
- Added the explicit note that already-attached waiters can receive a pre-rotation/pre-delete value once, but that value is not cached after the generation changes.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude follow-up:

- Added a one-retry cap for shared context-error waiter retries. Healthy waiters still escape a canceled/deadlined owner once, but persistent lower-layer context-like errors cannot spin on a healthy caller.
- Updated retry predicate coverage to assert `context.DeadlineExceeded` retries only before the retry limit and does not retry when the waiter context is canceled.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Latest Claude nit follow-up:

- Wrapped early/waiter context cancellation errors as `DDBProvider.APIKey: %w` so all APIKey error paths have consistent log context while preserving `errors.Is` behavior.

Local validation on this head: `go test -race -count=1 ./shared/auth` and `make check` pass.

justin-layerv:
Final technical status on head `91e19ac9`:

- All CI checks are green, including `slack / required`, CodeQL, dependency checks, TruffleHog, shared all-app tests, Discord checks, Slack lint/test/govulncheck/docker.
- Latest Claude review is complete and non-blocking; no inline PR review comments are open.
- Latest local validation before push: `go test -race -count=1 ./shared/auth` and `make check` pass.
- Follow-ups are filed and labeled: #765 for shared TTL/singleflight helper extraction, including lock/generation policy; #766 for stricter cross-instance revocation invalidation if the accepted TTL window needs tightening.

Remaining human gate: branch protection still reports `REVIEW_REQUIRED`. Please review/approve and explicitly confirm the #263 5-minute cross-instance revoked-key window is accepted for this PR.

PR #760, opened 2026-06-13
fix(slack): split tunnel bootstrap key delivery

### LINKED ISSUE #515: Split tunnel bootstrap key delivery from install instructions
## Context
`/qurl tunnel install` currently renders the bootstrap API key in the ephemeral Slack response that also contains the install instructions. The install flow warns admins to treat the message as secret, and the key expires, but Slack workspaces may retain/export admin-readable message content.

## Scope
Design and implement a separate bootstrap-key delivery path for tunnel installs, likely using the Slack DM surface once production PostDM/bot-token plumbing is available. Docker, Docker Compose, and Kubernetes flows can prompt for a key supplied out-of-band; ECS/Fargate may need secret-manager-specific copy or a DM-delivered key.

## Acceptance criteria
- The main install-instructions message no longer contains the plaintext bootstrap key for environments that can consume an out-of-band secret.
- If DM delivery is used, failures fail closed before minting or revoke the freshly minted key.
- ECS/Fargate has explicit guidance for storing the key in Secrets Manager without requiring the instructions message to carry the key.
- Tests cover successful secret delivery and failed secret-delivery cleanup.
- Customer-facing copy clearly distinguishes install instructions from the temporary secret.

justin-layerv:
## Summary
- Split qURL Connector tunnel install bootstrap-key delivery from the main Slack install instructions
- Deliver the temporary bootstrap key by Slack DM using `conversations.open` plus `chat.postMessage`, with the existing per-workspace bot-token lookup and Enterprise Grid fallback
- Require `commands,chat:write,im:write` for new/reinstalled Slack app tokens so the bot can open 1:1 DMs before sending `dm:true` links and connector bootstrap keys
- Fail closed before minting when DM delivery is not wired, and revoke freshly minted keys if DM or install-instruction delivery is not confirmed
- Update ECS/Fargate copy to point admins at AWS Secrets Manager without including the key in the install instructions

## Rollout note
Existing Slack installs must reinstall or reauthorize before this build is rolled out so Slack issues a token with `chat:write` and `im:write`. Without that pre-rollout reinstall, guided connector setup and `/qurl get dm:true` fail closed with the reinstall prompt until the workspace updates its Slack app token.

## Security tradeoff
The bootstrap key is no longer present in channel-visible install instructions, which removes the broad-audience leak path this issue targets. The key is now sent as a normal 1:1 Slack DM, so it can persist in Slack history until the admin deletes it or retention policy removes it. This is intentional and mitigated by one-hour TTL, revoke-on-unconfirmed-delivery, single-use connector bootstrap semantics, and DM copy telling the admin to delete the message when workspace retention allows.

## Business relevance
This reduces secret exposure in Slack-visible install instructions while preserving the guided connector install flow for admins. It also gives operators deterministic cleanup behavior when Slack cannot confirm delivery, so short-lived bootstrap keys do not remain usable after a failed install attempt.

## Validation
- `go test -count=1 ./...`
- `go vet ./...`
- `GOLANGCI_LINT_CACHE=$(mktemp -d) go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --timeout=5m ./...`

Closes #515


justin-layerv:
Addressed the Claude review notes in 7d94fa01: documented the intentional DM-before-instructions ordering, kept the helper nil guard with rationale, moved the discard-notice DM onto the handler base context, and added test coverage for the discard DM content.

justin-layerv:
Addressed the second Claude review in ee41dc8f: agent protect-connector modals now carry payload.Enterprise.ID, the inline key-target comment says DM instead of ephemeral, and tests assert both agent modal metadata and modal-submit DM enterpriseID propagation.

justin-layerv:
Addressed the third Claude review in 1879c002: renamed the DM helper to the neutral postTunnelInstallDM, added explicit discard-DM failure logging, and made the alias-remint test assert the reminted bootstrap key is delivered by DM. The Enterprise Grid live-smoke item is tracked as #762 with security/enhancement/priority: low labels.

justin-layerv:
Addressed the latest scope note in 293e7b82: Slack install defaults and docs now require commands,chat:write,im:write so fresh workspace/org installs have the DM-capable token needed by bootstrap-key delivery. Local validation stayed green: go test -count=1 ./..., go vet ./..., and golangci-lint v2.10.1.

justin-layerv:
Rebased onto current main and resolved the tunnel install test overlap by preserving response_url retry behavior while keeping bootstrap keys DM-only. Also cleaned up stale key-delivery comments. Local validation on 0465ac3a: go test -count=1 ./..., go vet ./..., and golangci-lint v2.10.1.

justin-layerv:
Addressed the latest Claude review in bbf5f62e: chat.postMessage missing_scope now wraps a typed Slack missing-scope sentinel, tunnel bootstrap DM failures caused by missing scopes point admins to the Slack reinstall link after revoking the key, handler_get trims enterpriseID consistently, and the double discard-notice redundancy is documented. Local validation stayed green: go test -count=1 ./..., go vet ./..., and golangci-lint v2.10.1.

justin-layerv:
Addressed the latest Claude review in 17b06ef4: SLACK_BOT_SCOPES overrides are now additive, so legacy overrides like commands keep required defaults instead of aborting startup; tests cover the legacy override and views:write-only cases. I also narrowed required install scopes to commands,chat:write after checking Slack's current chat.postMessage docs: the method requires chat:write and supports passing a user ID as channel to open the 1:1 DM when needed, while im:write is for conversations.open, which this path does not call. Local validation stayed green: go test -count=1 ./..., go vet ./..., and golangci-lint v2.10.1.

justin-layerv:
Addressed the latest Claude consistency nit in 20f09412: /qurl get dm:true now uses the same Slack reinstall guidance when PostDM fails with ErrSlackMissingScope, without leaking the minted link into the channel response. Also updated follow-up issue #762 with the chat:write-only first-DM live-smoke acceptance addendum. Local validation stayed green: go test -count=1 ./..., go vet ./..., and golangci-lint v2.10.1.

justin-layerv:
Addressed the latest Claude DM-history note in 4818d567: the bootstrap-key DM now tells admins to delete the DM from Slack history when workspace retention policy allows, and the unreachable helper fallback branches were removed. The chat:write-only live validation remains tracked in #762 because it requires a real Slack workspace. Local validation stayed green: go test -count=1 ./..., go vet ./..., and golangci-lint v2.10.1.

justin-layerv:
Addressed the latest Claude pass on head 678eabc1:

- Confirmed the response_url false-negative behavior is intentional: once the secret DM is delivered, install instructions still need confirmed delivery before the bootstrap key remains valid. If Slack does not confirm the instruction post, we revoke the short-lived key and send best-effort discard notices to avoid a valid secret being separated from trusted install context.
- Added the operator migration note to `apps/slack/docs/operating.md`: before deploying builds that depend on newly required Slack scopes, send affected workspaces through the reinstall link so guided connector setup does not fail closed on day one.
- Updated #762 so the live-only follow-up covers both Enterprise Grid org-token fallback and `chat:write`-only first-DM validation for a user who has never messaged the bot.
- The `SLACK_BOT_SCOPES` union behavior remains intentional: required defaults stay load-bearing, while operators can add scopes without accidentally removing command or DM delivery.

justin-layerv:
Addressed the latest Claude findings on head 2e30d31b:

- Fixed the revoked-key idempotency replay risk. Tunnel bootstrap idempotency keys now use the exact install attempt timestamp instead of a 25-minute bucket, so a typed retry after a DM-delivery revoke does not replay the just-revoked key.
- Added `TestTunnelInstallRetryAfterDMRevokeUsesFreshIdempotencyKey` to simulate the rollout case: first attempt mints then revokes on `missing_scope`; second attempt succeeds with a different idempotency key and a fresh bootstrap key.
- Kept same-attempt stability covered by `TestTunnelBootstrapIdempotencyKeyUsesExactAttemptTime`, so duplicate submissions from the same modal attempt still share an idempotency key.
- For the `chat:write` first-DM concern: code remains on `chat.postMessage(channel=<user_id>)` with `commands,chat:write` based on Slack's documented method behavior; the real-workspace smoke for a never-DM'd user is tracked in #762 because it cannot be validated hermetically here.

Local validation on this head:
- `go test -count=1 ./...`
- `go vet ./...`
- `GOLANGCI_LINT_CACHE=$(mktemp -d) go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --timeout=5m ./...` (`0 issues`)

justin-layerv:
Addressed the latest idempotency follow-up on head 86690217:

- Typed `protect-connector` now keys bootstrap idempotency by Slack `trigger_id` when present. That preserves dedupe for a Slack HTTP retry of the same slash command while still giving a human re-run after a DM-delivery revoke a fresh attempt/key.
- Modal submissions now key by Slack `view.id` with the modal-created timestamp only as a fallback. The helper comment explicitly documents the same-modal retry tradeoff: modal retries dedupe to avoid double-minting, and `processTunnelInstall` revokes any key whose delivery is not confirmed.
- The retry-after-DM-revoke regression remains covered, and the idempotency unit test now also fences same-trigger retry stability.

Local validation on this head:
- `go test -count=1 ./...`
- `go vet ./...`
- `GOLANGCI_LINT_CACHE=$(mktemp -d) go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --timeout=5m ./...` (`0 issues`)

justin-layerv:
Addressed the DM-delivery gating concern on head 23ed44b9:

- Production `PostDM` no longer assumes `chat.postMessage(channel=<user_id>)` can auto-open a first DM. It now explicitly calls `conversations.open(users=<user_id>)`, parses the returned DM channel id, then calls `chat.postMessage(channel=<dm_channel_id>)`.
- Added `im:write` to `DefaultBotScopes()` and updated config tests, Slack install tests, operating docs, PR rollout notes, and #762. Required scopes are now `commands,chat:write,im:write`.
- Added DM transport tests for open-then-post with Enterprise Grid fallback and for `conversations.open` `missing_scope` wrapping `ErrSlackMissingScope`, so existing reinstall guidance still triggers.

Local validation on this head:
- `go test -count=1 ./...`
- `go vet ./...`
- `GOLANGCI_LINT_CACHE=$(mktemp -d) go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --timeout=5m ./...` (`0 issues`)

justin-layerv:
Addressed Claude's latest pre-merge asks on head 4e6e06a7:

- Restored the load-bearing upstream idempotency contract note in `tunnelBootstrapIdempotencyKey`, including the qurl-service#775 tracking reference.
- Added the modal retry asymmetry directly to the helper comment: same-view retries intentionally dedupe to avoid double-minting; if a retry follows a DM-failure revoke, replayed plaintext would already be dead and a fresh human retry gets a new modal/view id.
- Updated the PR body with an explicit security tradeoff section: moving the key out of channel-visible instructions reduces the broad-audience leak path but makes the secret a durable 1:1 DM until deletion/retention; mitigations are TTL, revoke-on-unconfirmed-delivery, one-time bootstrap semantics, and delete-the-DM copy.

Validation after the comment-only code change:
- `go test -count=1 ./apps/slack/internal -run 'TestTunnelBootstrapIdempotencyKeyUsesExactAttemptTime|TestTunnelInstallRetryAfterDMRevokeUsesFreshIdempotencyKey'`

The prior full local gate on the same code path remains green:
- `go test -count=1 ./...`
- `go vet ./...`
- `GOLANGCI_LINT_CACHE=$(mktemp -d) go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --timeout=5m ./...` (`0 issues`)

PR #775, opened 2026-06-13
fix(shared): validate API key cache hits

### LINKED ISSUE #766: Bound cross-instance API key cache revocation staleness
## Summary

PR #757 adds the issue #263 `DDBProvider.APIKey` cache with a 5-minute TTL. That TTL is explicitly accepted in #263 for steady-state KMS/latency reduction.

The same-instance post-`DeleteAPIKey` stale-refill case is handled in PR #757 now: after local delete invalidation, the deleting process uses strongly consistent refills for one TTL, and tests assert it does not re-cache a revoked key from an eventually-consistent post-delete read.

The remaining revocation/rotation staleness window is cross-instance:

- sibling pods do not observe another pod's local `SetAPIKey` / `DeleteAPIKey` invalidation;
- those siblings may continue serving a previously decrypted key until their local TTL expires;
- the hot slash-command path sends that cached key directly to qurl-service, so a stale/revoked key can surface as a user-visible 401 until the TTL rolls over.

## Why

This is acceptable for PR #757 because #263 explicitly chose the 5-minute propagation window and the PR documents the trade-off inline. If disconnect/rotation becomes a stricter security action, we should avoid relying only on process-local TTL expiry for revoked qURL API keys.

## Acceptance criteria

- Add a cross-instance invalidation path for qURL API key cache entries, or replace the process-local-only cache with an equivalent mechanism that bounds sibling-process stale reads.
- Preserve the existing no-negative-cache behavior for fresh installs and transient decrypt/DDB failures.
- Keep steady-state slash-command KMS Decrypt pressure reduced relative to the pre-#757 implementation.
- Add tests or an integration plan that proves one instance does not serve a revoked/rotated key after another instance invalidates it.
- Consider whether the slash-command path should invalidate/retry on qurl-service auth failures as an additional backstop, and document the chosen behavior.

Surfaced by Claude review on PR #757.

justin-layerv:
## Summary
- Validate cached QURL API keys against DynamoDB encrypted key material before returning cached plaintext.
- Evict and strongly refill the local cache when sibling rotations/deletes change the stored material; local Set/Delete still seed or invalidate this process immediately.
- Coalesce same-workspace/same-token validation reads and fail open on validation transport errors so a DDB blip does not turn a warm cache into a Slack outage.
- Add bounded recheck handling for validation races and focused regression coverage for rotation, delete, validation outage, context-error retry, validation coalescing, and local writer races.

## Business relevance
This fixes #766 by keeping the KMS-saving API key cache from serving stale keys for the full TTL after another instance rotates or deletes a workspace key. That preserves the cost/latency win from caching decrypted keys while restoring prompt revocation behavior for Slack command traffic.

## Risk notes
- Warm cache hits now perform one strongly-consistent projected DynamoDB `GetItem`; this intentionally keeps KMS Decrypt off the hot path while trading back some DDB RCU/latency to make sibling rotations/deletes visible immediately when DDB is reachable.
- Validation transport errors serve the cached key until TTL rather than failing Slack commands; this preserves availability during a DDB blip and matches the pre-existing TTL-bounded revocation window when DDB cannot be reached.

## Validation
- `go test -count=1 ./shared/auth`
- `go test -race -count=1 ./shared/auth`
- `go test -count=1 ./...`
- `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`

### DIFF
--- a/internal/ttlcache/cache.go
+++ b/internal/ttlcache/cache.go
@@ -169,6 +169,38 @@ func (c *Cache[V]) InvalidateWith(key string, underLock func()) {
 	}
 }
 
+// InvalidateIfWith invalidates key only when predicate accepts the currently
+// cached result. It returns false when key is missing or predicate rejects it.
+func (c *Cache[V]) InvalidateIfWith(key string, predicate func(Result[V]) bool, underLock func()) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.initLocked()
+	cached, ok := c.entries[key]
+	if !ok || predicate == nil || !predicate(cached.result) {
+		return false
+	}
+	c.evictLocked(key, cached)
+	delete(c.inFlight, key)
+	c.generation[key]++
+	if underLock != nil {
+		underLock()
+	}
+	return true
+}
+
+// CachedResultMatches reports whether key currently has a cached result that
+// predicate accepts. It does not sweep expiry; callers should use it only to
+// compare against an entry they just obtained from GetOrStart.
+func (c *Cache[V]) CachedResultMatches(key string, predicate func(Result[V]) bool) bool {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	c.initLocked()
+	cached, ok := c.entries[key]
+	return ok && predicate != nil && predicate(cached.result)
+}
+
 // Seed replaces key with a cached result and detaches any in-flight owner.
 func (c *Cache[V]) Seed(key string, result Result[V], ttl time.Duration, at time.Time) {
 	c.SeedWith(key, result, ttl, at, nil)
--- a/shared/auth/ddb_provider.go
+++ b/shared/auth/ddb_provider.go
@@ -31,6 +31,8 @@ import (
 	"crypto/aes"
 	"crypto/cipher"
 	"crypto/rand"
+	"crypto/sha256"
+	"encoding/binary"
 	"errors"
 	"fmt"
 	"io"
@@ -93,9 +95,14 @@ const (
 )
 
 const (
-	apiKeyCacheTTL                         = 5 * time.Minute
-	apiKeyCacheSweepEvery                  = time.Minute
-	apiKeySharedContextErrorRetryLimit int = 1
+	apiKeyCacheTTL                     = 5 * time.Minute
+	apiKeyCacheSweepEvery              = time.Minute
+	apiKeySharedContextErrorRetryLimit = 1
+	apiKeyValidationRecheckLimit       = 3
+
+	apiKeyValidationProjectionKey        = "#api_key"
+	apiKeyValidationProjectionDataKey    = "#data_key"
+	apiKeyValidationProjectionExpression = apiKeyValidationProjectionKey + ", " + apiKeyValidationProjectionDataKey
 )
 
 // ErrWorkspaceNotConfigured is the sentinel returned by APIKey when the
@@ -144,21 +151,26 @@ type DDBProvider struct {
 	Encryptor FieldEncryptor
 
 	// Cache successful APIKey lookups at the DDB/decrypt boundary with the
-	// shared TTL singleflight helper. The helper intentionally uses one mutex
-	// for cache maps, in-flight fills, generation invalidation, and the
-	// strong-read sidecar below: fills run outside the lock, so unrelated
-	// workspaces only serialize the short map operations instead of DDB/KMS.
-	//
-	// The cache is process-local, so sibling instances can keep a decrypted key,
-	// including a revoked key, until the five-minute TTL from #263 expires;
-	// stricter cross-instance invalidation is tracked in #766. The decrypted key
-	// remains in heap memory for that TTL as the accepted KMS/latency trade-off.
+	// shared TTL singleflight helper. Cache hits still perform a strongly
+	// consistent DDB projection read of the encrypted key material before
+	// returning plaintext; that trades steady-state DDB-read avoidance for
+	// bounded cross-instance rotation/revocation staleness without putting KMS
+	// Decrypt back on the steady-state slash-command path. Definitive
+	// changed/deleted validation results evict the cache; validation transport
+	// errors fall back to the cached key so a DDB blip does not become a full
+	// Slack outage. The decrypted key remains in heap memory for the TTL as the
+	// accepted KMS/latency trade-off. Validation reads are strongly consistent
+	// so sibling rotations/deletes take effect immediately when DDB is reachable
+	// rather than after eventual-read replication lag.
+	// The validation token is intentionally tied to encrypted key material, so a
+	// future same-plaintext rewrap would invalidate warm entries and re-run KMS
+	// decrypts even though the API key value did not change.
 	// SetAPIKey seeds this process after a successful write. DeleteAPIKey evicts
 	// this process and forces strongly-consistent refills for one TTL so a
 	// same-process eventually-consistent post-delete read cannot re-cache the
 	// revoked key.
 	apiKeyCacheOnce   sync.Once
-	apiKeyLookupCache *ttlcache.Cache[string]
+	apiKeyLookupCache *ttlcache.Cache[cachedAPIKey]
 
 	// Protected by apiKeyLookupCache's mutex through ttlcache hooks. Generation
 	// entries are retained by the helper after invalidation even after an
@@ -168,6 +180,7 @@ type DDBProvider struct {
 	// seeded or invalidated in this process and is retained for the process
 	// lifetime.
 	apiKeyStrongReadUntil map[string]time.Time
+	apiKeyValidationCalls map[string]*apiKeyValidationCall
 
 	// Now is injected so tests can pin the wall clock for configured_at /
 	// updated_at assertions without poking package-global state. Defaults
@@ -175,17 +188,30 @@ type DDBProvider struct {
 	Now func() time.Time
 }
 
+type cachedAPIKey struct {
+	apiKey     string
+	cacheToken apiKeyCacheToken
+}
+
 type apiKeyLookupStart struct {
-	apiKey string
+	apiKey     string
+	cacheToken apiKeyCacheToken
 	// Cached API-key hits only store successful lookups, so err is expected
 	// to be nil. Keeping the Result shape here mirrors ttlcache and makes
 	// that contract explicit at the call site.
 	err            error
 	hit            bool
-	call           *ttlcache.Call[string]
+	call           *ttlcache.Call[cachedAPIKey]
 	owner          bool
 	generation     uint64
 	consistentRead bool
+}
+
+type apiKeyValidationCall struct {
+	done       chan struct{}
+	cacheToken apiKeyCacheToken
+	current    bool
+	err        error
 }
 
 // SlackBotTokenInstall is the workspace-scoped Slack OAuth material persisted
@@ -212,9 +238,9 @@ func (p *DDBProvider) nowOrDefault() time.Time {
 	return time.Now()
 }
 
-func (p *DDBProvider) apiKeyCache() *ttlcache.Cache[string] {
+func (p *DDBProvider) apiKeyCache() *ttlcache.Cache[cachedAPIKey] {
 	p.apiKeyCacheOnce.Do(func() {
-		p.apiKeyLookupCache = ttlcache.New[string](ttlcache.Options[string]{
+		p.apiKeyLookupCache = ttlcache.New[cachedAPIKey](ttlcache.Options[cachedAPIKey]{
 			SweepEvery: apiKeyCacheSweepEvery,
 			OnSweep: func(at time.Time) {
 				// OnSweep runs under the ttlcache lock; keep this hook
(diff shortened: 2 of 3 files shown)


justin-layerv:
Addressed Claude review round 1 on the latest head:\n\n- coalesced same-workspace/same-token cache validation reads so a hot workspace burst shares one strong projected GetItem instead of fanning out per request\n- detach stale validation calls on local set/delete/cache-token eviction, matching the existing fill invalidation shape\n- kept fail-closed validation behavior for revocation correctness, but made the DDB-read and fail-closed tradeoff explicit in the provider comment\n- added focused coverage for validation coalescing and new validation-state guard behavior\n\nLocal validation on this head: `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, and `GOLANGCI_LINT_CACHE=/tmp/qurl-integrations-golangci-cache-766 PATH=/tmp/qurl-integrations-bin:/Users/posey/.codex/tmp/arg0/codex-arg0z1PyRJ:/Users/posey/.local/bin:/opt/homebrew/opt/libpq/bin:/Users/posey/.jenv/shims:/Users/posey/.jenv/bin:/Users/posey/.sdkman/candidates/java/current/bin:/opt/homebrew/bin:/opt/homebrew/sbin:/usr/local/bin:/System/Cryptexes/App/usr/bin:/usr/bin:/bin:/usr/sbin:/sbin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/local/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/bin:/var/run/com.apple.security.cryptexd/codex.system/bootstrap/usr/appleinternal/bin:/pkg/env/global/bin://Applications/Topaz Gigapixel.app/Contents/Resources/bin:/Library/Apple/usr/bin:/Applications/Codex.app/Contents/Resources make check` pass.\n\nRe-running /simplify and waiting for exact-head CI + Claude review.

justin-layerv:
Addressed the second Claude review cycle on the latest head:

- Validation transport errors now fall back to the cached key so a DDB validation blip does not become a Slack outage.
- Definitive changed/deleted validation results still evict and refill strongly before returning a key.
- Context-shaped validation owner errors retry once with the existing shared-error retry shape before falling back.
- Added coverage for validation outage fallback and transient validation context-error retry.
- Re-ran: `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `GOLANGCI_LINT_CACHE=/tmp/qurl-integrations-golangci-cache-766c PATH=/tmp/qurl-integrations-bin:$PATH make check`.

justin-layerv:
Addressed the latest Claude review on the newest head:

- Confirmed and kept the PR description aligned with the fail-open-on-validation-transport-errors behavior.
- Added the requested strong-read rationale to the provider comment: sibling rotations/deletes take effect immediately when DDB is reachable rather than waiting for eventual-read replication lag.
- Re-ran `go test -count=1 ./shared/auth` and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check` on this head.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest non-blocking Claude nits on the newest head:

- Added comments on the token-mismatch and local-token-change re-loop paths explaining why they make progress.
- Switched cache validation `ProjectionExpression` to expression-attribute aliases and updated tests to assert the alias mapping.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check` on this head.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude minor items on the newest head:

- Added a bounded cache-validation recheck counter so token-mismatch/local-writer re-loops are provably finite.
- Re-check local cache identity before serving the cached key on fail-open validation transport errors, so a racing local `SetAPIKey` wins instead of returning captured stale plaintext.
- Added regression coverage for validation outage racing a local `SetAPIKey`.
- Kept the strong-read and validation-payload-reuse points as documented/accepted design trade-offs: strong reads preserve immediate sibling revocation when DDB is reachable, and rotation/delete remains uncommon enough that the simpler refill path is preferable to threading validation payloads through the cache-hit loop.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check` on this head.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Rebased onto current main and ported the API-key cache validation fix onto the shared ttlcache helper. Latest local validation on head a573f05: `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check` all pass. Re-running /simplify and waiting for exact-head CI + Claude review.

justin-layerv:
Addressed the latest exact-head Claude coverage feedback on head e2d83046:

- Added validation transport-error fail-open coverage.
- Added validation context-error retry coverage.
- Added successful-validation/local-SetAPIKey race coverage so a racing local writer wins instead of returning stale plaintext.
- Added bounded recheck convergence coverage for repeated local-token changes.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`. One first `make check` attempt hit golangci-lint parallel lock; rerun with a fresh cache passed.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude nits on head d8f15625:

- Added a same-token validation coalescing test that asserts owner/waiter share one strongly-consistent projected DDB validation read.
- Renamed the validation-error fallback locals to avoid shadowing the named returns in `apiKeyFromValidatedCache`.
- Updated the PR risk notes to explicitly document the strong-read RCU/latency tradeoff and the TTL-bounded fail-open behavior during validation transport outages.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude minor notes on head 160ae5ca:

- Added a comment that the validation recheck budget is deliberately shared across all re-loop reasons because each means the caller has not validated the currently local token.
- Added coverage for the validation-error/local-writer race (`fallbackErr == nil && !cachedAPIKeyStillLocal`) so a DDB validation outage racing local `SetAPIKey` revalidates and returns the new cached key.
- Left the direct `ctx.Err()` guard in `apiKeyAfterCacheValidationError`; it is intentionally what prevents a canceled caller from fail-opening to cached plaintext.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude optional feedback on head d8b7de94:

- Renamed the `apiKeyAfterCacheValidationError` cached-key parameter to avoid shadowing the `cachedAPIKey` struct type.
- Added direct coverage for the sibling `DeleteAPIKey`/missing-item validation path: warm cache, projected strong validation read returns no item, cache evicts, and the strong refill returns `ErrWorkspaceNotConfigured` without another decrypt.
- Moved token-mismatch/delete eviction ahead of the recheck-cap return so even a pathological non-convergence result does not leave the stale token cached until TTL.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude documentation considerations on head aa668552:

- Added a provider comment noting that the validation token intentionally tracks encrypted key material, so a future same-plaintext rewrap would invalidate warm cache entries and re-run KMS decrypts.
- Clarified why first-fill reads can remain eventually consistent while cache-hit validation uses strong reads: #766 is specifically about prompt cross-instance rotation/revocation visibility.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude nit on head 61eda825:

- `apiKeyAfterCacheValidationError` now returns the caller context error when validation is canceled/deadlined, instead of echoing the validation-layer error. This keeps the validation cancellation path consistent with the rest of `APIKey`.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude coverage suggestion on head b1fb2b89:

- Added a direct sibling-rotation convergence test: warm cache holds the old key, projected strong validation sees the new encrypted key material, the old token is evicted, the strong refill returns the new API key, and decrypt runs only for prime + rotated refill.
- Removed the unnecessary explicit `int` annotations from the cache retry-limit constants.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Addressed the latest Claude observability/readability suggestions on head 40ebc7d9:

- Added a `slog.WarnContext` on the rare `cache validation did not converge` branch with workspace ID, recheck count, and limit so churn-induced failures are distinguishable in logs.
- Added a short doc comment for `apiKeyFromValidatedCache` explaining the `done=false` re-enter-loop contract.
- Re-ran `go test -count=1 ./shared/auth`, `go test -race -count=1 ./shared/auth`, `go test -count=1 ./...`, and `tmp_cache=$(mktemp -d /tmp/qurl-golangci-XXXXXX) && GOLANGCI_LINT_CACHE="$tmp_cache" PATH=/tmp/qurl-integrations-bin:$PATH make check`.

Re-running /simplify and waiting for exact-head CI + Claude review again.

justin-layerv:
Explicit trade-off confirmation for final review: this PR intentionally uses a strongly-consistent projected DDB `GetItem` on every warm API-key cache hit. The goal for #766 is immediate cross-instance rotation/delete visibility whenever DDB is reachable, while preserving the main KMS `Decrypt` savings from the cache. That means the cache no longer avoids DDB reads on steady serial slash-command traffic; validation coalescing only collapses concurrent same-token bursts. The PR risk notes call out the added strong-read RCU/latency cost and the TTL-bounded fail-open behavior during validation transport outages. If production RCU/latency later needs a different point on the freshness/cost curve, the follow-up lever would be a short per-workspace validation throttle, knowingly reintroducing an N-second staleness window.

PR #783, opened 2026-06-13
fix(slack): lameduck health before shutdown

### LINKED ISSUE #154: deploy: ALB drain headroom (lameduck /health → 503) before srv.Shutdown on SIGTERM

justin-layerv:
## Summary
- Flip Slack /health to 503 during SIGTERM lameduck before closing the listener
- Keep the 13s lameduck phase inside the existing 25s shutdown budget, leaving drain headroom before Fargate SIGKILL
- Reserve lameduck for SIGTERM; SIGINT/local Ctrl-C now drains immediately
- Decouple handler async cancellation from the raw signal so requests accepted during lameduck do not inherit an already-canceled base context
- Add focused tests for health 503 behavior, shutdown budget headroom, immediate drain, and an in-flight request completing through the lameduck -> drain transition
- Pair with layervai/qurl-integrations-infra#1070 so ALB marks draining targets unhealthy within the bot's lameduck window

## Business relevance
This reduces user-visible deploy/replacement errors by giving ALB time to stop routing to a task before the Slack service closes its listener, while preserving bounded graceful shutdown inside the Fargate termination window.

## Validation
- `go test -count=1 ./apps/slack/cmd ./apps/slack/internal`
- `go test -count=1 ./...`
- `golangci-lint run --allow-parallel-runners --new-from-rev=$(git merge-base HEAD origin/main) --timeout=5m ./...`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.10.1 run --allow-parallel-runners --timeout=5m ./...`
- `go test -race -count=1 -coverprofile=coverage.out -covermode=atomic ./apps/slack/... ./shared/...`
- `go tool govulncheck ./...`
- `CGO_ENABLED=0 GOOS=linux GOARCH=arm64 go build -trimpath -ldflags="-w -s -X main.version=dev" -o release/slack/qurl-bot-slack ./apps/slack/cmd/`
- `git diff --check`

Closes #154

### DIFF
--- a/apps/slack/cmd/main.go
+++ b/apps/slack/cmd/main.go
@@ -18,6 +18,7 @@ import (
 	"os/signal"
 	"strconv"
 	"strings"
+	"sync"
 	"syscall"
 	"time"
 
@@ -84,6 +85,19 @@ const (
 	// headroom for the expected case while staying well inside
 	// Fargate's 30s SIGTERM→SIGKILL window.
 	shutdownTimeout = 25 * time.Second
+
+	// lameduckDuration is the ALB deregistration head start inside
+	// shutdownTimeout. On SIGTERM, /health flips to 503, we keep the
+	// listener open briefly so ALB can remove the target, then
+	// srv.Shutdown starts the actual HTTP drain with the remaining
+	// budget. Keep this greater than the ALB target group's unhealthy
+	// detection window; with qurl-webhook-runtime's 5s interval x
+	// 2-unhealthy-threshold target, 13s gives a small ALB propagation
+	// margin while leaving roughly 12s for in-flight requests and async
+	// workers before Fargate's 30s SIGTERM→SIGKILL window closes.
+	// TODO(upstream-contract): Keep this in lockstep with
+	// qurl-integrations-infra's qurl-webhook-runtime health check cadence.
+	lameduckDuration = 13 * time.Second
 	// maxHeaderBytes is well above Slack's realistic header size (sig +
 	// timestamp + standard headers fit comfortably in 2 KiB) but bounds
 	// the per-connection memory an attacker can force pre-handler.
@@ -145,11 +159,13 @@ func run() error {
 
 	// DDBProvider reads the workspace_state table populated by
 	// /oauth/qurl/callback. Missing WORKSPACE_STATE_TABLE fails
-	// startup distinctly from an empty table.
-	signalCtx, stop := signal.NotifyContext(context.Background(), syscall.SIGTERM, syscall.SIGINT)
-	defer stop()
+	// startup distinctly from an empty table. The signal watcher is
+	// armed before infra setup so SIGTERM/SIGINT during AWS config
+	// load cancels those calls promptly.
+	shutdownSignals := newShutdownSignalSource(syscall.SIGTERM, syscall.SIGINT)
+	defer shutdownSignals.stop()
 
-	ddbProvider, err := auth.NewDDBProvider(signalCtx,
+	ddbProvider, err := auth.NewDDBProvider(shutdownSignals.ctx,
 		auth.WithTableName(os.Getenv("WORKSPACE_STATE_TABLE")),
 		auth.WithKMSKeyARN(os.Getenv("WORKSPACE_STATE_KMS_KEY_ARN")),
 	)
@@ -162,7 +178,7 @@ func run() error {
 	maxConcurrentAsync := readMaxConcurrentAsync()
 	maxConcurrentFollowupAsync := readMaxConcurrentFollowupAsync()
 	maxConcurrentFollowupGateAsync := readMaxConcurrentFollowupGateAsync()
-	adminStore := buildAdminStore(signalCtx)
+	adminStore := buildAdminStore(shutdownSignals.ctx)
 	slackBotToken := strings.TrimSpace(os.Getenv("SLACK_BOT_TOKEN"))
 	if err := auth.ValidateSlackBotTokenShape(slackBotToken); err != nil {
 		return fmt.Errorf("invalid SLACK_BOT_TOKEN: %w", err)
@@ -247,7 +263,7 @@ func run() error {
 	var agentStore *slackdata.AgentStore
 	if !agentDisabled {
 		agentLLM = buildAgentLLM()
-		agentStore = buildAgentStore(signalCtx)
+		agentStore = buildAgentStore(shutdownSignals.ctx)
 	}
 	logAgentSurfaceState(agentSurfaceState{
 		llmWired:              agentLLM != nil,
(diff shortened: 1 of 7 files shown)


justin-layerv:
Addressed the Claude review items on the updated head:

- Paired the app change with layervai/qurl-integrations-infra#1070 so the target group health window is 5s x 2 = 10s, and increased app lameduck to 11s with the constraint documented next to the constant.
- SIGTERM still lameducks for ALB; SIGINT/local Ctrl-C now drains immediately.
- Added coverage for `duck=0` immediate drain, including no `SetHealthy(false)` and still cancel/shutdown/wait.
- Clarified that lameduck + `srv.Shutdown` share the same timeout and can intentionally squeeze async drain budget to zero under slow HTTP drain.

Local validation was re-run after these changes: full tests, pinned golangci-lint, race/coverage, govulncheck, linux/arm64 build, and `git diff --check`.

justin-layerv:
Followed up on the latest Claude questions:

- Increased SIGTERM lameduck from 11s to 13s so the 5s x 2 ALB unhealthy window has a small propagation margin instead of only ~1s.
- Kept the shutdown timeout at 25s, leaving roughly 12s of HTTP/async drain headroom inside Fargate's 30s SIGTERM-to-SIGKILL window.
- Added an explicit server comment that the slow OAuth callback can still be cut if it lands during task-stop; that is intentional because the deploy lameduck is protecting the customer-facing Slack ack path, and the OAuth path already can exceed the Fargate shutdown budget in the worst case.
- Re-ran the full local app validation; the direct `--new-from-rev=origin/main` lint form now includes unrelated base-drift files, so the PR body records the merge-base form that matches the actual PR diff and passes with 0 issues.

justin-layerv:
Addressed the latest non-blocking Claude notes too:

- Updated `Handler.baseCtx` documentation so it now says workers are canceled at HTTP drain start after lameduck, not directly on raw SIGTERM.
- Added a comment in `newShutdownSignalSource` documenting the intentional first-signal-only behavior: later signals do not shorten SIGTERM lameduck; Fargate SIGKILL is still the hard stop, and local SIGINT uses immediate drain.

Validation after this comment-only cleanup: `go test -count=1 ./apps/slack/cmd ./apps/slack/internal`, merge-base `golangci-lint`, `go vet`, and `git diff --check`.

justin-layerv:
Addressed the latest Claude pass on the final head:

- Fixed the `WaitTimeout(0)` edge case by tracking registered async workers separately from the waitgroup, so a fully consumed drain budget returns cleanly when no workers are pending instead of racing an immediate timer and logging a bogus timeout.
- Added focused tests for zero-budget/no-worker and zero-budget/worker-pending behavior.
- Reconfirmed the no-signal-escalation behavior is intentional and documented in `newShutdownSignalSource`: SIGTERM lameduck is not shortened by later signals; local SIGINT still maps to immediate drain, and Fargate SIGKILL is the hard stop.
- Reconfirmed listener-error/immediate-drain does not flip `/health` to 503 because there is no ALB head-start to preserve in that path; the SIGTERM task-stop path is the one that lameducks.

Local validation after this fix: `go test -count=1 ./...`, pinned golangci-lint, race/coverage for Slack/shared, `go vet`, `govulncheck`, merge-base PR-diff lint, linux/arm64 Slack build, and `git diff --check`.

justin-layerv:
Fixed the CI merge-ref failure on the final head:

- Rebased the branch onto current `origin/main` so GitHub tests the same base locally.
- Resolved the base-drift import issue by carrying the new `main.go` imports from `main` plus the shutdown path's `sync` usage; the prior PR merge ref failed with `undefined: sync` while the branch head still compiled.
- Re-ran validation after the rebase/import fix: `go test -count=1 ./...`, pinned golangci-lint, race/coverage for Slack/shared, `go vet`, `govulncheck`, merge-base PR-diff lint, linux/arm64 Slack build, and `git diff --check`.

justin-layerv:
Addressed Claude's latest convention/test notes on the final head:

- Added the requested `TODO(upstream-contract)` marker on `lameduckDuration` so the qurl-integrations-infra health-check cadence coupling is discoverable with `git grep TODO(upstream-contract)`.
- Added direct `shutdownSignalSource` unit coverage for first-signal-wins behavior, context cancellation, idempotent stop, and startup-context cancellation through `stop()`. The production signal.Notify path is unchanged; tests inject the signal channel to avoid sending real process signals.
- Re-ran validation after this change: `go test -count=1 ./...`, pinned golangci-lint, race/coverage for Slack/shared, `go vet`, `govulncheck`, merge-base PR-diff lint, linux/arm64 Slack build, and `git diff --check`.

justin-layerv:
Addressed the latest Claude nits on the final head:

- Moved the `TODO(upstream-contract)` marker to the end of the `lameduckDuration` comment block so it no longer splits the unhealthy-detection sentence.
- Changed `asyncDone` to call `wg.Done()` before decrementing `activeWorkers`, making the zero-budget drain counter conservative.

Validation after this cleanup: `go test -count=1 ./...`, pinned golangci-lint, race/coverage for Slack/shared, `go vet`, `govulncheck`, merge-base PR-diff lint, linux/arm64 Slack build, and `git diff --check`.

justin-layerv:
Addressed the latest Claude observation on the final head:

- If lameduck exhausts the shared shutdown budget before HTTP drain starts, we now still call `srv.Shutdown` to close the listener, but treat the expected `context.DeadlineExceeded` as a budget-exhaustion warning instead of logging a misleading graceful-shutdown error.
- Forced the async drain budget to zero on that exhausted-lameduck path and added a regression test for the behavior.

Validation after this cleanup: `go test -count=1 ./...`, pinned golangci-lint, race/coverage for Slack/shared, `go vet`, `govulncheck`, merge-base PR-diff lint, linux/arm64 Slack build, and `git diff --check`.

justin-layerv:
Confirmed the store-context lifetime assumption from the latest Claude review:

- `auth.NewDDBProvider(ctx)` uses the context only for `awsconfig.LoadDefaultConfig`; `DDBProvider` stores the AWS clients, table/encryptor/cache state, and clock, not the context. Per-call DDB/KMS work takes its own context.
- `slackdata.NewStore(ctx)` and `NewAgentStoreFromEnv(ctx)` follow the same pattern: constructor ctx only feeds AWS config load; returned `Store`/`AgentStore` structs do not retain it, and methods accept per-call contexts.
- So switching these constructors to `shutdownSignals.ctx` affects startup/config-load cancellation only; it does not yank store calls during lameduck/drain.

PR #802, opened 2026-06-14
test: bounded transient-status retry on connector-stack e2e fetches

justin-layerv:
## Summary

Stop the connector post-deploy **e2e smoke** from false-red'ing on transient
connector-stack errors. The smoke runs *while* the connector-v2 ECS rollout is
still in flight (the infra `terraform` apply has no `wait_for_steady_state` —
layervai/qurl-integrations-infra#1085), so a brief rolling/drain window can serve
a **503** (ALB has no healthy target) before the replacement task is healthy. The
upload + viewer fetches previously hard-failed on the first non-ok response,
turning a seconds-long blip into a red deploy gate. This is ask **#3** of that
issue.

## Scope

- [ ] `apps/slack/`
- [ ] `apps/teams/`
- [ ] `apps/discord/`
- [ ] `apps/cli/`
- [ ] `apps/zapier/`

> **`e2e/` test-infra only** — none of the app scopes apply. No `apps/`, `shared/`, `go.*`, or `.github/` files change, so the Go `shared/required` aggregates don't trigger.

## Changes

- Add `e2e/helpers/http.ts` — `fetchWithTransientRetry`, a **bounded, method-aware** retry on transient HTTP statuses for the connector-stack serving path.
- Route the connector `/upload` + fileviewer `/view` fetches through it:
  - `uploadMapLocation` + `fetchViewerPage` (`e2e/tests/google-maps.test.ts`)
  - `uploadFile` (`e2e/helpers/qurl-api.ts`) — connector `/upload`, used by the smoke
  - `uploadImage` + the pre-revoke `/view` read (`e2e/tests/file-revoke.test.ts`) — same serving paths (not in the smoke `testPathPatterns`, but identical paths; included for consistency, flagged by review + /simplify)

## Design — never masks a real outage (the issue's hard constraint)

- **Method-aware** so a non-idempotent `/upload` POST can never be retried into a duplicate resource:
  - **Any** method retries `{408, 425, 429, 503}` — statuses where the request provably did NOT reach/complete at the app (503 = the drain-gap; 408/425/429 = rejected/timed-out before processing).
  - **Idempotent** methods (GET/HEAD/…, e.g. the `/view` read) also retry `{502, 504}` — transient gateway failures where the backend *may* have already processed (safe to replay on a GET, unsafe on a POST).
  - `500` excluded everywhere (app error → fail fast); `403` excluded (WAF/IP-reputation block, blocks the runner run-wide → tracked in layervai/qurl-integrations-infra#1091).
- **Bounded** (3 attempts, 1s/2s linear backoff) → a *sustained* failure still surfaces; the final `Response` is returned for the caller's own `!res.ok` throw.
- Does **not** catch fetch *rejections* — the sustained `fileviewer.layerv.xyz:443` ConnectTimeout the ticket calls out is a connection failure that propagates immediately and fails fast.
- Each retry logs a `console.warn` (method, url, status, attempt) so a run that *recovered* after a blip is visible in CI logs.
- `mintLink` / `accessLink` / `revokeLink` / `getLinkStatus` hit qurl-service (a separate upstream, not the connector being deployed) — intentionally **out of scope**.

## Test Plan

- [x] `npx tsc --noEmit` clean on `e2e/` (the only local gate — `make check` is Go-only; the pre-existing `TS5107` moduleResolution deprecation is on `main`, unrelated).
- [ ] `make check` — N/A (Go targets; no Go files changed).
- [x] FormData body re-readability across retries verified (undici re-extracts `FormData`/`Blob` per `fetch()`).
- [ ] **Real validation is the next sandbox post-deploy e2e-smoke run** — the `e2e/` TS is not exercised by PR CI (it only runs post-deploy from the infra repo). Unit coverage for the helper + an `e2e/` PR-CI step are tracked in #804.

## Related Issues

Addresses layervai/qurl-integrations-infra#1085 (ask #3). Companion infra PR (topology fix + 403 investigation): layervai/qurl-integrations-infra#1093. Follow-up: #804 (e2e PR CI + unit coverage).

### DIFF
--- a/e2e/helpers/http.ts
+++ b/e2e/helpers/http.ts
@@ -16,7 +16,12 @@
  *       · ANY method retries {408, 425, 429, 503} — statuses where the request
  *         provably did NOT reach/complete at the app (503 = ALB has no healthy
  *         target, the drain-gap; 408/425/429 = rejected/timed-out before
- *         processing), so a retry can't duplicate work even on a POST.
+ *         processing), so a retry can't duplicate work even on a POST. NOTE the
+ *         503-on-POST guarantee is contingent on the connector signaling
+ *         rate-limits via HTTP 200 + `error` (not an app-level 503), so a 503
+ *         here is always the ALB no-healthy-target case, never a post-accept app
+ *         503. If the connector ever returns a real app-level 503 after partially
+ *         processing, move 503 to the idempotent-only set below.
  *       · IDEMPOTENT methods (GET/HEAD/…, e.g. the `/view` read) ALSO retry
  *         {502, 504} — transient gateway failures where the backend MAY have
  *         already processed the request before the response was lost. Retrying
--- a/e2e/helpers/http.ts
+++ b/e2e/helpers/http.ts
@@ -11,8 +11,18 @@
  * hard-failed the smoke on the very first response (a false red).
  *
  * What it deliberately does NOT do, so it can never MASK a real outage:
- *   - Retries only transient HTTP *statuses* (5xx + 408/425/429). Deterministic
- *     4xx (400/401/404/409) fail fast — those are real failures or test bugs.
+ *   - Retries only statuses where the request provably did NOT reach/complete at
+ *     the app: 502/503 (ALB couldn't reach a healthy target — the drain-gap we
+ *     target) plus 408/425/429 (rejected/timed-out before processing). This set
+ *     is safe to retry even on the non-idempotent `/upload` POST: none of them can
+ *     mean "the backend processed it but the response was lost," so a retry can't
+ *     duplicate a resource. Deterministic 4xx (400/401/404/409) fail fast — real
+ *     failures or test bugs.
+ *   - Excludes 500 and 504 on purpose: 500 is an app-level error (a real failure,
+ *     not a transient infra blip), and 504 (gateway timeout) can mean the backend
+ *     DID process the request before the response was lost — retrying a POST would
+ *     then duplicate the resource, and the orphan escapes the tests' afterAll
+ *     cleanup (which only tracks the final success). The drain-gap is 502/503.
  *   - Excludes 403: on this path a 403 is a WAF-layer block (e.g. AWS managed
  *     IP-reputation flagging the CI runner's egress IP), which blocks the runner
  *     run-wide — an intra-run retry (same IP) can't recover it and would only
@@ -25,18 +35,18 @@
  */
 
 /**
- * Transient HTTP statuses worth a bounded retry on the connector/fileviewer
- * serving path. See the module header for why 403 and the other 4xx are absent.
- * Module-private — the retry policy is an implementation detail of the helper.
+ * Statuses worth a bounded retry on the connector/fileviewer serving path —
+ * limited to ones where the request did NOT reach/complete at the app, so a retry
+ * is safe even on a non-idempotent POST. See the module header for why 500/504,
+ * 403, and the deterministic 4xx are absent. Module-private — the retry policy is
+ * an implementation detail of the helper.
  */
 const RETRYABLE_HTTP_STATUS: ReadonlySet<number> = new Set([
-  408, // Request Timeout
-  425, // Too Early
-  429, // Too Many Requests
-  500, // Internal Server Error
-  502, // Bad Gateway
-  503, // Service Unavailable — an ALB with no healthy target (the drain-gap)
-  504, // Gateway Timeout
+  408, // Request Timeout — request not fully received, so not processed
+  425, // Too Early — TLS early-data replay guard; safe to replay
+  429, // Too Many Requests — rejected before processing
+  502, // Bad Gateway — ALB couldn't reach a healthy target (drain-gap)
+  503, // Service Unavailable — ALB has no healthy target (the drain-gap)
 ]);
 
 /**
--- a/e2e/helpers/http.ts
+++ b/e2e/helpers/http.ts
@@ -1,0 +1,73 @@
+/**
+ * Shared HTTP helper for the E2E smoke: a BOUNDED retry on transient HTTP
+ * responses from the connector-stack serving path (connector `/upload` +
+ * fileviewer `/view`).
+ *
+ * Why this exists: the connector post-deploy smoke runs WHILE the connector-v2
+ * ECS rollout is still in flight — the infra `terraform` apply has no
+ * wait_for_steady_state, so the smoke races the rollout
+ * (qurl-integrations-infra#1085). A brief rolling/drain window can serve a 5xx
+ * from the ALB before the replacement task is healthy, which previously
+ * hard-failed the smoke on the very first response (a false red).
+ *
+ * What it deliberately does NOT do, so it can never MASK a real outage:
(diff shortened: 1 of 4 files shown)


justin-layerv:
Thanks — all four addressed in 60c2128.

**1. `file-revoke.test.ts` gap → fixed.** Routed both its connector-stack fetches through the helper (`uploadImage`'s `/upload` POST + the pre-revoke `/view` GET). One correction for the record: `file-revoke.test.ts` is **not** in the post-deploy smoke (`testPathPatterns='smoke|google-maps|discord-channels'` doesn't match it), so it can't false-red the deploy gate today — but it exercises the identical serving paths, so the consistency fix is right regardless (the /simplify reuse pass flagged it too). PR title updated to "…e2e fetches" (dropped "smoke") to match.

**2. Non-idempotent POST on 500/504 → fixed at the root.** Rather than document the duplicate-resource leak, I dropped **500 and 504** from the set — now `{408, 425, 429, 502, 503}`. Your classification is exactly the cut line: 502/503 (ALB has no healthy target — the drain-gap) never reach the backend, and 408/425/429 are rejected/timed-out before processing, so all five are retry-safe even on a non-idempotent POST. 504 (may have processed) and 500 (app-level error that should fail fast) are out. This eliminates the duplicate-resource / orphaned-`createdResourceIds` risk entirely instead of accepting it.

**3. Redundant `!res.ok` in the loop condition → kept (defensive).** As you noted it's harmless; I left it as belt-and-suspenders against a future mis-edit that adds a 2xx/3xx to the set. Happy to drop it if you'd prefer the tighter condition.

**4. No unit coverage for `fetchWithTransientRetry` → tracked.** Filed #804 to add PR CI for `e2e/` (`tsc --noEmit` + mocked-fetch unit tests for the helpers — status-set membership, budget exhaustion, body-cancel, rejection-propagation). Deferred rather than inline because the suite has no unit-test/mocked-fetch pattern today and no PR CI runs `e2e/` jest, so a unit test added here wouldn't execute automatically — #804 fixes that substrate first.

Confirmed-correct items (FormData re-readability, attempt accounting, body-cancel, latency) — appreciated the cross-checks.

justin-layerv:
Thanks — round-2 nits addressed in 09e336c.

1. **502-on-POST safety → fixed at the root (method-aware).** You're right that 502 has the same "may have processed" ambiguity as 504, so leaning on "provably didn't reach the app" to license retrying a POST was overclaiming. Rather than soften the wording, I made the set **method-aware** (your option (b), generalized):
   - **Any** method retries `{408, 425, 429, 503}` — provably not processed → safe even on the non-idempotent `/upload` POST.
   - **Idempotent** methods (GET/HEAD/…, e.g. the `/view` read) also retry `{502, 504}` — transient gateway failures that are safe to replay on a GET but risk a duplicate on a POST.
   So the no-duplicate invariant is now airtight for POST, *and* the idempotent viewer read regains 502/504 coverage that the previous flat `{…502,503}` set didn't even give it. 500 stays out everywhere.
2. **Description drift → fixed.** Rewrote the PR Design section to state the method-aware sets exactly (no more "5xx + …").
3. **Silent retries → fixed.** Each retry now `console.warn`s `method url -> status; retry n/N in Xms`, so a recovered-after-blip run is legible in CI logs.
4. **Combined budget** — noted, no change (orthogonal layers, well under 120s; the POST path now retries fewer statuses so it's tighter).
5. **Unit coverage** — tracked in #804 (updated it to include the new method-aware cases). Deferred only because no PR CI runs `e2e/` jest today and the smoke `testPathPatterns` would exclude a `http.test.ts`, so an inline test wouldn't execute — #804 adds that substrate (tsc + mocked-fetch unit tests) first.

justin-layerv:
Thanks — appreciate the thorough close-out. Round-3 notes handled:

1. **503-on-POST invariant is contingent → documented** (4f7bca9). Added a header note that the no-duplicate guarantee for retrying 503 on the `/upload` POST holds *because* the connector signals rate-limits via HTTP 200 + `error` (never an app-level 503), so a 503 here is always the ALB no-healthy-target case. If the connector ever returns a real app-level 503 after partially processing, the note says to move 503 to the idempotent-only set. Good catch — that's exactly the kind of assumption that should be explicit.
2. **Per-attempt wall-clock timeout → deferred to #804.** Agreed it's worth it (a backend that accepts then stalls has no client deadline today), but it's pre-existing and not a masking risk — jest's 120s testTimeout is the backstop, so a hung attempt still *fails*, just slowly. Folded an `AbortSignal.timeout(...)`-per-attempt item into #804 alongside the unit coverage.
3. Nested-budget multiplication / `!res.ok` redundancy / FormData re-extraction — noted, no change (all confirmed fine).

Unit coverage (#804) is the one real automation gap — agree on prioritizing it; the method-aware branching is exactly what regresses silently.

justin-layerv:
Round-4 notes handled (bceabcb):

2. **Viewer URL in CI logs → fixed.** Good catch — the fileviewer `/view/<mint-id>` path carries the capability mint-id. The retry `console.warn` now logs `new URL(input).origin` only (enough to say which host blipped — the drain-gap signal — without the id), guarded so a non-absolute input can't throw.
1. **FormData re-extraction invariant / unit test** — agreed it's the silent-failure mode most worth locking; tracked in #804 (which now also covers the method-aware branching + per-attempt timeout). Agree on landing it sooner given zero PR-CI exercise today.
3. **`maxAttempts`/`baseDelayMs` unused by callers** — keeping them per your suggestion (they let the #804 unit test drive fast retries without real backoff).
4. **Multiplicative budgets** — acknowledged; ~22s worst-case backoff, well under the 120s jest timeout, layers are orthogonal.

Thanks for the thorough review.

justin-layerv:
Thanks — addressing the round-5 points (all non-blocking, no code change needed):

1. **502-on-POST hard-fails by design — confirming it's the right call.** Honest answer on the empirical question: I don't have live ALB sampled data, but a drained AWS ALB returns **503** (no healthy target), which is the documented drain-gap status #1085 is about; a **502** would be a different mechanism (target reset mid-request). Two things make excluding 502 on POST the right tradeoff: (a) the companion infra fix (qurl-integrations-infra#1093, `min_healthy=100`) **removes the drain entirely**, so post-merge the smoke shouldn't see drain-induced 502/503 on any path — this retry is defense-in-depth; (b) on the non-idempotent `/upload` POST, **no-duplicate-resource beats de-flaking a rare 502** (the helper documents 502 as idempotent-only for exactly this reason). If a POST-502-during-rollout is ever actually observed, the escape hatch is the same as the 503 note: revisit the classification. I'll add that to #1085's thread so the tradeoff is on record.

2. **`429` ignores `Retry-After`** — moot on this path: the connector signals rate-limits via HTTP 200 + `error` (handled by the *app-level* 429 loop), so a *wire* 429 reaching `fetchWithTransientRetry` is a pre-app ALB/WAF edge. The bounded 1s/2s linear backoff is fine for that; honoring `Retry-After` would only matter if a real upstream started returning wire-429s with the header, which isn't the case today. Folded as a note into #804 alongside the per-attempt-timeout item.

3. **Composition timing** (~22s worst-case backoff < 120s) — acknowledged, no change.

4. **Unit coverage** — #804; agree the status-classification matrix is the cheap, high-value thing to lock down. Prioritized there.

PR #791, opened 2026-06-13
feat(slack): add explicit workspace key rotation

justin-layerv:
## Summary
- Add signed setup modes for `/qurl setup <email> --rotate` and `--repoint`, while normal `/qurl setup <email>` keeps idempotent key reuse.
- Store qURL `key_id`/prefix metadata with encrypted workspace keys and strongly read that metadata for explicit rotations.
- Revoke the old key before replacement mint, use idempotency for replacement retries, verify already-revoked owner state safely, and document/test the failure ordering.
- Pre-validate the replacement idempotency key before revoking, and give the owner-scoped revoked-key scan a dedicated multi-page timeout budget.
- Block plain setup from minting around an invalid metadata-bearing stored key; admins must retry `--rotate`/`--repoint` so a rotation persist-failure recovers via the old-key idempotency bucket instead of orphaning the prior replacement.
- Document the deleted-at-qURL stale-key dead end and operator recovery path.
- Expose separate operator-configurable replay-window mirrors for binding-backed setup retries and API-key replacement mint retries, so `retry_window_hours` maps to the right upstream idempotency contract.
- Remove the SHA-256 cache-token digest over encrypted API-key material; cache validation now compares copied encrypted-material tokens directly to satisfy CodeQL without changing plaintext exposure.

## Business relevance
This gives Slack workspace owners a safe self-service key rotation path after setup reruns became idempotent, preventing accidental API-key slot leaks while preserving the normal setup UX.

## Follow-up
- #790 tracks the non-owning cross-account repoint path once qURL exposes binding transfer support. This PR fails closed before minting when the authenticated qURL account cannot revoke the current workspace key.
- layervai/qurl-service#946 tracks replacing the bounded revoked-list scan with an owner-scoped direct key-status lookup. Source verification shows qurl-service lists API keys by `created_at` descending, not revoke time, so this follow-up is now labeled `bug` + `priority: high`; this PR remains safe because it fails closed and sends operators to key-status verification when the bounded scan cannot confirm ownership.
- layervai/qurl-service#954 tracks documenting the API-key mint idempotency TTL contract that Slack mirrors for rotation replacement retry windows.
- layervai/qurl-integrations-infra#1077 tracks wiring the actual CloudWatch alarm for `setup_rotation_replacement_persist_failure`; this PR emits the structured event and adds the app runbook query.

## Validation
- `go test ./apps/slack/internal/oauth ./apps/slack/internal ./shared/auth`
[...]

PR #803, opened 2026-06-14
feat(slack): revoke upstream qURL API key on /qurl uninstall (#792)

### LINKED ISSUE #792: Persist qURL API key IDs so /qurl uninstall can revoke upstream keys
## Problem

`/qurl uninstall` disconnects the Slack workspace by removing the locally stored qURL API key columns from `workspace_state`, but the row currently does not persist the qURL service key ID. Without that ID, existing workspace installs cannot revoke the upstream API key during uninstall.

## Acceptance criteria

- Persist the qURL service key ID when `/qurl setup` stores a workspace API key.
- Use that key ID to revoke the upstream qURL API key during `/qurl uninstall` before or alongside local credential removal.
- Keep legacy rows without a key ID on the current safe local-disconnect path, with user/operator-visible logging.
- Cover setup storage, uninstall revocation success, and legacy-no-key-ID fallback in tests.

Surfaced while making #268 merge-ready in #786.

### LINKED ISSUE #805: Confirm qurl-service self-revoke contract for /qurl uninstall (qurl:write key revoking its own key_id)
## Context

PR #803 (closes #792) wires `/qurl uninstall` to revoke the workspace's upstream qURL API key before removing local credentials. Because a slash-command uninstall has no live OAuth flow (unlike the rotation path in `oauth/callback`, which revokes with the Auth0 access token), uninstall **self-revokes**: the workspace's own `qurl:write` key authenticates `DELETE /v1/api-keys/{its-own-key_id}` via `client.Client.RevokeAPIKey`.

This assumes qurl-service permits a `qurl:write` key to revoke **itself**. We have evidence that a `qurl:write` key can revoke *other* keys (the tunnel bootstrap-cleanup path, `revokeBootstrapKeyAfterInstallFailure`, revokes child keys), but self-revoke is unverified against the qurl-service contract.

## Why it matters

The code degrades gracefully if the assumption is wrong: a `401`/`403` from the revoke maps to a **local-only disconnect** (the pre-#803 behavior) with the existing "not revoked outside Slack" caveat and an operator log line. So uninstall never breaks. **But** if self-revoke is unsupported, *every* uninstall silently lands on that branch and the upstream-revocation feature is a no-op — exactly the gap #792 set out to close.

Marker in code: `git grep "TODO(upstream-contract)" apps/slack/internal/handler.go` → `classifyUninstallRevokeError`.

## Acceptance criteria

- [ ] Confirm with the qurl-service team whether a `qurl:write` API key may `DELETE /v1/api-keys/{key_id}` for its **own** key id.
- [ ] If **supported**: remove the `TODO(upstream-contract)` marker in `classifyUninstallRevokeError` and note the confirmed contract in `apps/slack/docs/operating.md` ("Uninstall revocation visibility").
- [ ] If **not supported**: re-architect uninstall revocation to use an owner/operator-authenticated path (mirroring rotation's minter-token revoke) instead of self-revoke, or document that uninstall is local-only by design.
- [ ] Either way, update the operating.md note that currently says "confirm the self-revoke contract with the qURL service team."

Surfaced by the Claude review (cr) pass on #803 — both review runs flagged this as the one item that gates whether the feature works in production.

justin-layerv:
## What

Closes #792. Also closes #805 (the self-revoke contract follow-up, folded in — see "Confirmed contract" below).

`/qurl uninstall` previously removed only the locally stored workspace credentials, leaving the upstream qURL API key live. Now that setup persists the qURL `key_id` (via `SetAPIKeyWithMetadata`, #791), uninstall strongly reads that `key_id` and **attempts to revoke the upstream key before local credential removal**.

## How

- **Layering.** Revocation is orchestrated in the Slack handler, not in `shared/auth`. The auth package keeps its DDB+KMS-only dependency surface and never imports the qurl-service client — the same posture as the explicit-rotation path in `oauth/callback`. The stale `TODO(upstream-contract)` on `DeleteAPIKey` is replaced with a doc note; `DeleteAPIKey` stays pure local removal.
- **Capability interface.** A new `workspaceKeyRevoker { APIKeyID(...) }` is type-asserted on the configured provider (compile-time `var _ workspaceKeyRevoker = (*auth.DDBProvider)(nil)`). `*auth.DDBProvider` implements it; providers that can't (e.g. `auth.EnvProvider`) fall back to local-only disconnect.
- **Self-revoke mechanism.** The workspace key authenticates `DELETE /v1/api-keys/{key_id}` against its own id (a `qurl:write` self-revoke), reusing the existing `client.Client.RevokeAPIKey` already used by the tunnel bootstrap-cleanup path.

## Confirmed contract (#805)

The self-revoke assumption is now **confirmed against qurl-service**: an API key **cannot** manage API keys. `DELETE /v1/api-keys/{key_id}` returns `403` ("API keys cannot manage API keys. Use JWT authentication.") for **any** key id — including the caller's own — structurally and **before** any existence check. (Rotation can revoke because `oauth/callback` runs under the owner's freshly authenticated JWT; uninstall has only the workspace API key.)

Consequences, applied in this PR:

- **Upstream revocation from `/qurl uninstall` is local-only by design**, not a deployment quirk — a live workspace key always gets `403` and degrades to a local-only disconnect. `classifyUninstallRevokeError`'s `TODO(upstream-contract)` is removed and replaced with the confirmed contract.
- **The `RevokeAPIKey` call is intentionally kept** as the safe interim (per the #806 escape-hatch sequencing — flipping `401`/`403` to abort without #806 would leave admins unable to disconnect at all). It still does real work on the `404` path (key already gone upstream → `upstream_revoked=true`).
- **Log level `ERROR` → `WARN`** on the `401`/`403` degrade: `403` is now the expected every-uninstall outcome, so `ERROR` was alert noise. The structured `status` field distinguishes the still-live `403` from the already-dead `401` for operators.
- Pinned upstream by a qurl-service regression test: **layervai/qurl-service#959** (`TestRevokeApiKey_ForbiddenForAPIKeyAuth` asserts the `403` code+detail; `TestRevokeApiKey_APIKeyAuthForbiddenBeforeExistence` pins auth-before-existence).

- **Budget.** Stays synchronous within the existing 1.2s `adminSyncVerbBudget` (one strongly-read `APIKeyID`, the cached `APIKey` for the client, one HTTP DELETE, one `DeleteAPIKey` write) — well inside the 3s ack window.

### Failure matrix

| Revoke outcome | Behavior |
| --- | --- |
| 2xx (not reachable in prod under the confirmed contract — defensive), or 404 (already gone) | local removal + "key has been revoked" confirmation (404 self-heals a prior revoke-ok/delete-failed retry) |
| legacy row with no `key_id` | local-only disconnect + existing "not revoked outside Slack" caveat + operator `WARN` |
| 403 (self-revoke forbidden — the universal case for a live key) | degrade to local-only disconnect + caveat + operator `WARN` (`status=403`) |
| 401 (key already revoked/expired upstream, e.g. prior rotation) | degrade to local-only disconnect + caveat + operator `WARN` (`status=401`) |
| 5xx / transport / timeout / `key_id` read error | **abort before local removal** so the stored `key_id` survives for a retry — never orphan a live key |

## Tests

Red-first. `recordingAuthProvider` (no `APIKeyID`) keeps existing uninstall tests on the local-only path unchanged; a new `revokingAuthProvider` + a DELETE-recording customer server exercise:

- revoke success → disconnect + revoked confirmation (defensive: the 204 branch isn't reached under the confirmed self-revoke contract — kept for a future owner-auth path, #806)
- 404 → treated as revoked
- legacy no-`key_id` → local-only, no DELETE call
- 403 → local-only fallback (the confirmed universal prod outcome for a live key)
- 401 → local-only fallback (new — already-revoked-upstream sibling of the 403 case)
- 5xx → abort, `DeleteAPIKey` not called (key_id preserved)
- `key_id` read error → abort, `DeleteAPIKey` not called

Setup `key_id` persistence (AC#1) is already covered by `TestDDBProviderSetAPIKeyWithMetadata*` / `TestDDBProviderAPIKeyID`.

## Docs

- `apps/slack/README.md`: uninstall now revokes the upstream key (+ legacy fallback note).
- `apps/slack/docs/operating.md`: "Uninstall revocation visibility" section reframed as **local-only by design** (the confirmed contract), with the log events + a CloudWatch query and the per-`status` operator guidance.

## Verification

- `go build ./...`, `go test ./...` (slack + shared + cli) green
- `golangci-lint` (CI-pinned **v2.10.1**) clean repo-wide
- commits GPG-signed

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -52,7 +52,7 @@ workspace's Secure Access Agent supports.
 | `/qurl get <$id\|$alias> reason:"…"` | Mint the link and record a reason in the audit log. |
 | `/qurl list` | List the resources available to you in this channel. |
 | `/qurl aliases` | List this channel's aliases and the resource each one points to. |
-| `/qurl uninstall` | Owner/admin-gated: disconnect qURL from this workspace's Slack commands. This does not revoke the upstream qURL API key. |
+| `/qurl uninstall` | Owner/admin-gated: disconnect qURL from this workspace's Slack commands and revoke this workspace's qURL API key. Workspaces connected before key identity was stored fall back to a local-only disconnect (the key isn't revoked) — contact the operator to revoke it. |
 | `/qurl feedback` | Send a bug report or feature request to the qURL team. |
 | `/qurl help` | Show the user command help. |
 
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -53,7 +53,7 @@ workspace's Secure Access Agent supports.
 | `/qurl get <$id\|$alias> reason:"…"` | Mint the link and record a reason in the audit log. |
 | `/qurl list` | List the resources available to you in this channel. |
 | `/qurl aliases` | List this channel's aliases and the resource each one points to. |
-| `/qurl uninstall` | Owner/admin-gated: disconnect qURL from this workspace's Slack commands and revoke this workspace's qURL API key. Workspaces connected before key identity was stored fall back to a local-only disconnect (the key isn't revoked) — contact the operator to revoke it. |
+| `/qurl uninstall` | Owner/admin-gated: disconnects qURL from this workspace's Slack commands. This is a **local-only disconnect** — it does **not** revoke the workspace's qURL API key outside Slack (an API key can't revoke itself), so if you're disconnecting because the key may be exposed, revoke it through qURL key management or your operator. |
 | `/qurl feedback` | Send a bug report or feature request to the qURL team. |
 | `/qurl help` | Show the user command help. |
 
--- a/apps/slack/docs/operating.md
+++ b/apps/slack/docs/operating.md
@@ -359,6 +359,42 @@ instead of revoked, it will also be absent from the revoked list. Repeated
 retries will keep failing until an operator verifies the key status or uses a
 direct owner-scoped lookup when qURL exposes one.
 
+## Uninstall revocation visibility
+
+`/qurl uninstall` strongly reads the stored qURL `key_id` and revokes the
+upstream workspace key (a `qurl:write` self-revoke: the workspace key
+authenticates the `DELETE /v1/api-keys/{key_id}`) before removing local
+credentials. `event`-free log lines under the `/qurl uninstall:` prefix report
+the outcome; the terminal `/qurl uninstall: disconnected workspace Slack
+commands` line carries `upstream_revoked` so a single query separates real
+revokes from local-only disconnects:
+
+```text
+fields @timestamp, team_id, caller_user_id, key_id, upstream_revoked
+| filter @message like "/qurl uninstall:"
+| sort @timestamp desc
+| limit 50
+```
+
+Two outcomes need operator follow-up because the upstream key is **not** revoked
+but the workspace was disconnected locally:
+
+- `legacy workspace row has no qURL key id — local-only disconnect`: the
+  workspace connected before `key_id` persistence (issue #792), so Slack cannot
+  identify which key to revoke. Revoke it through qURL account/API-key
+  management if the disconnect was security-motivated.
+- `upstream refused workspace key self-revoke — local-only disconnect`
+  (`status=401`/`403`): the deployment's qURL service does not permit a
+  workspace key to revoke itself. Revoke through operator tooling and confirm
+  the self-revoke contract with the qURL service team.
+
+`upstream qURL key revoke failed — aborting to preserve key id` (and the
+`could not read stored qURL key id before revoke` variant) means nothing was
+disconnected: the stored `key_id` is intentionally preserved so the admin can
+retry. A transient failure self-heals on retry because an already-revoked key
+returns 404 and is then treated as revoked. Persistent failures past a few
+retries warrant the same upstream-key verification as the rotation path above.
+
 ## Endpoints
 
 | Endpoint | Purpose |
--- a/apps/slack/docs/operating.md
+++ b/apps/slack/docs/operating.md
@@ -492,12 +492,18 @@ or unexpected (5xx, transport, timeout), so the stored `key_id` is intentionally
 preserved for a retry that self-heals on the next uninstall. The abort is
 deliberate — it never severs local Slack access while leaving a possibly-live
 upstream key it can no longer identify (that would give false security on a
-compromised workspace, since the leaked key still works via the qURL API). If the
-upstream stays unreachable and Slack access must be cut urgently, revoke the key
-through qURL account/API-key tooling and then clear the `workspace_state` row's
-qURL columns directly — the manual equivalent of the local disconnect — rather
-than waiting on the retry path. A first-class admin/operator force/local-only
-escape hatch is tracked in #806.
+compromised workspace, since the leaked key still works via the qURL API).
+
+**Availability note:** this inverts the pre-upstream-revocation behavior, where
+`/qurl uninstall` always cut local Slack access. A persistent upstream outage (or
+a slow upstream that exhausts the sync budget) can now **block the disconnect**
+until it recovers — the trade-off for never orphaning a key the operator can no
+longer identify. If the upstream stays unreachable and Slack access must be cut
+urgently, revoke the key through qURL account/API-key tooling and then clear the
+`workspace_state` row's qURL columns directly — the manual equivalent of the local
+disconnect — rather than waiting on the retry path. A first-class admin/operator
+force/local-only escape hatch is tracked in #806 (sequence it soon, since this
+manual row edit is otherwise the only out during an upstream outage).
 
 ## Endpoints
 
--- a/apps/slack/docs/operating.md
+++ b/apps/slack/docs/operating.md
@@ -405,6 +405,15 @@ revoked (404), or — if the service checks auth before existence — lands on t
 failures past a few retries warrant the same upstream-key verification as the
 rotation path above.
 
+The abort is deliberate: it never severs local Slack access while leaving a
+possibly-live upstream key it can no longer identify (that would give false
+security on a compromised workspace, since the leaked key still works via the
+qURL API). If the upstream stays unreachable and Slack access must be cut
+urgently, revoke the key through qURL account/API-key tooling and then clear the
+`workspace_state` row's qURL columns directly — the manual equivalent of the
+local disconnect — rather than waiting on the retry path. A first-class
+admin/operator force/local-only escape hatch is tracked in #806.
+
 ## Endpoints
 
 | Endpoint | Purpose |
--- a/apps/slack/docs/operating.md
+++ b/apps/slack/docs/operating.md
@@ -361,13 +361,23 @@ direct owner-scoped lookup when qURL exposes one.
 
 ## Uninstall revocation visibility
 
-`/qurl uninstall` strongly reads the stored qURL `key_id` and revokes the
-upstream workspace key (a `qurl:write` self-revoke: the workspace key
-authenticates the `DELETE /v1/api-keys/{key_id}`) before removing local
-credentials. `event`-free log lines under the `/qurl uninstall:` prefix report
-the outcome; the terminal `/qurl uninstall: disconnected workspace Slack
-commands` line carries `upstream_revoked` so a single query separates real
-revokes from local-only disconnects:
+`/qurl uninstall` strongly reads the stored qURL `key_id` and attempts to revoke
+the upstream workspace key before removing local credentials. Because uninstall
+has no live OAuth flow, it can only authenticate that `DELETE /v1/api-keys/{key_id}`
+with the workspace's own API key — and qurl-service **categorically forbids an API
+key from managing API keys** (the DELETE returns `403` for any key id, including
+its own, before any existence check; confirmed in #805). So for a live workspace
+key the upstream revoke always degrades to a **local-only disconnect by design**:
+uninstall reliably severs Slack access but does not revoke the key outside Slack.
+Rotation, by contrast, *can* revoke, because it runs under the owner's freshly
+authenticated JWT (see "Workspace key rotation" above); an owner/operator-
+authenticated uninstall revoke is future work, tracked with the escape hatch in
+#806.
+
+`event`-free log lines under the `/qurl uninstall:` prefix report the outcome; the
+terminal `/qurl uninstall: disconnected workspace Slack commands` line carries
+`upstream_revoked` so a single query separates the rare real revoke from the
+expected local-only disconnect:
 
 ```text
 fields @timestamp, team_id, caller_user_id, key_id, upstream_revoked
@@ -376,43 +386,42 @@ fields @timestamp, team_id, caller_user_id, key_id, upstream_revoked
 | limit 50
 ```
 
-Two outcomes need operator follow-up because the upstream key is **not** revoked
-but the workspace was disconnected locally:
+Because upstream revoke is local-only by design, expect `upstream_revoked=false`
+on essentially every uninstall. The local-only log lines and their operator
+implications:
 
+- `upstream qURL key not revoked — local-only disconnect` (`status=403`): the
+  expected path for a live key — qurl-service forbids the self-revoke. The
+  upstream key is **still live**, so if the disconnect was security-motivated
+  (the key may be exposed), revoke it through qURL account/API-key management or
+  operator tooling.
+- `upstream qURL key not revoked — local-only disconnect` (`status=401`): the
+  workspace key was already revoked or expired upstream (e.g. a prior rotation),
+  so qurl-service rejected the credential before the revoke. The key is already
+  dead — no upstream action is needed.
 - `legacy workspace row has no qURL key id — local-only disconnect`: the
   workspace connected before `key_id` persistence (added in #791, which this
   uninstall revocation builds on), so Slack cannot identify which key to revoke.
   Revoke it through qURL account/API-key management if the disconnect was
   security-motivated.
-- `upstream refused workspace key self-revoke — local-only disconnect`
-  (`status=401`/`403`): the deployment's qURL service does not permit a
-  workspace key to revoke itself. Revoke through operator tooling and confirm
-  the self-revoke contract with the qURL service team.
 
-`upstream qURL key revoke failed — aborting to preserve key id` (and the
-`could not read stored qURL key id before revoke` / `could not build client to
-revoke upstream key` variants) means nothing was disconnected: the stored
-`key_id` is intentionally preserved so the admin can retry, and a transient
-failure self-heals on the next uninstall.
+`upstream qURL key already absent — treating as revoked` (`status=404`) is the one
+outcome that reports `upstream_revoked=true`: the key had already been deleted
+upstream (e.g. through operator tooling), so uninstall simply confirms it.
 
-One expected intermediate state: if a revoke succeeds but the local removal then
-fails, the workspace keeps pointing at a now-revoked key until the retry, so
-every `/qurl` command in that window returns 401. A short burst of
-workspace-wide 401s between a failed uninstall and its retry is expected, not a
-separate incident. On that retry the already-revoked key's DELETE is treated as
-revoked (404), or — if the service checks auth before existence — lands on the
-401/403 local-only path above; either way the disconnect completes. Persistent
-failures past a few retries warrant the same upstream-key verification as the
-rotation path above.
-
-The abort is deliberate: it never severs local Slack access while leaving a
-possibly-live upstream key it can no longer identify (that would give false
-security on a compromised workspace, since the leaked key still works via the
-qURL API). If the upstream stays unreachable and Slack access must be cut
-urgently, revoke the key through qURL account/API-key tooling and then clear the
-`workspace_state` row's qURL columns directly — the manual equivalent of the
-local disconnect — rather than waiting on the retry path. A first-class
-admin/operator force/local-only escape hatch is tracked in #806.
+`upstream qURL key revoke failed — aborting to preserve key id` (and the `could
+not read stored qURL key id before revoke` / `could not build client to revoke
+upstream key` variants) means nothing was disconnected: the failure was transient
+or unexpected (5xx, transport, timeout), so the stored `key_id` is intentionally
+preserved for a retry that self-heals on the next uninstall. The abort is
+deliberate — it never severs local Slack access while leaving a possibly-live
+upstream key it can no longer identify (that would give false security on a
+compromised workspace, since the leaked key still works via the qURL API). If the
+upstream stays unreachable and Slack access must be cut urgently, revoke the key
+through qURL account/API-key tooling and then clear the `workspace_state` row's
+qURL columns directly — the manual equivalent of the local disconnect — rather
+than waiting on the retry path. A first-class admin/operator force/local-only
+escape hatch is tracked in #806.
 
 ## Endpoints
 
--- a/apps/slack/docs/operating.md
+++ b/apps/slack/docs/operating.md
@@ -458,9 +458,12 @@ fields @timestamp, team_id, caller_user_id, key_id, upstream_revoked
 | limit 50
 ```
 
-Because upstream revoke is local-only by design, expect `upstream_revoked=false`
-on essentially every uninstall. The local-only log lines and their operator
-implications:
+Because upstream revoke is local-only by design, `upstream_revoked` is **always
+`false`** today: a self-revoke authenticates with the workspace's own key against
+its own `key_id`, so a live key 403s and a dead key 401s — both before any
+existence check — and the `revoked=true` (204/404) outcomes can't occur until the
+owner-authenticated revoke path (#806) lands. The local-only log lines and their
+operator implications:
 
 - `upstream qURL key not revoked — local-only disconnect` (`status=403`): the
   expected path for a live key — qurl-service forbids the self-revoke. The
@@ -481,9 +484,12 @@ implications:
   Revoke it through qURL account/API-key management if the disconnect was
   security-motivated.
 
-`upstream qURL key already absent — treating as revoked` (`status=404`) is the one
-outcome that reports `upstream_revoked=true`: the key had already been deleted
-upstream (e.g. through operator tooling), so uninstall simply confirms it.
+`upstream qURL key already absent — treating as revoked` (`status=404`) would be
+the one outcome reporting `upstream_revoked=true`, but it **can't fire for a
+self-revoke**: the credential is the target key, so a deleted key surfaces as a
+`401` (bad credential) at the auth layer, never a `404` at the existence check.
+This branch is defensive scaffolding for the #806 owner-authenticated revoke,
+where the credential and target differ and a real `404`/`204` becomes possible.
 
 `upstream qURL key revoke failed — aborting to preserve key id` (and the `could
 not read stored qURL key id before revoke` / `could not build client to revoke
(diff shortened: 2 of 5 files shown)


justin-layerv:
Thanks for the thorough reviews. Disposition of every finding across both runs:

**Fixed (pushed in 2532cbf2):**
- **`caller_user_id` missing on revoke log lines** (R1 #1, "action before merge"). Every `/qurl uninstall:` log line now carries `caller_user_id` — the key-id read error, the build-client error, and all three branches of `classifyUninstallRevokeError` (404 / 401-403 / abort), plus the generic `DeleteAPIKey failed` line. Threaded `userID` into `classifyUninstallRevokeError`. The operating.md runbook query now resolves the attributor on exactly the key-still-live cases.
- **Stale "2s ack envelope" comment** (R2 #2). Corrected to the real 1.2s verb budget within Slack's 3s ack window, and noted the revoke shares that ctx so a flapping upstream aborts (key_id preserved) rather than missing the ack.
- **`TODO(upstream-contract)` in-code marker** (R1 #2). Added in an earlier round (3b42f6a8) and now points at the durable follow-up below.

**Deferred (filed #805, labeled `pr-polish-followup`/`go`/`priority: high`):**
- **Self-revoke contract confirmation** (R1 #2, R2 #4 — the production-gating item). The 401/403→local-only degrade assumes qurl-service lets a `qurl:write` key `DELETE` its own `key_id`. We know `qurl:write` can revoke *child* keys (tunnel bootstrap cleanup), but self-revoke is unverified. #805 tracks confirming with the qurl-service team; if unsupported, every uninstall silently degrades to local-only and the path must move to an owner/operator-authenticated revoke. The TODO marker references #805.

**Considered and intentionally not changing (with reasoning):**
- **Budget headroom of the synchronous DELETE** (R2 #1). The happy path is ~100–400ms (two same-row DDB ops + one DELETE) — well under 1.2s. The only way to approach the budget is a flapping upstream where the prod client's bounded retries (`WithRetry(2)`, `waitForRetry` honors the ctx) burn time; that path is ctx-bounded and aborts safely with `key_id` preserved, and the next retry 404-self-heals. So no orphan and no missed ack. A dedicated sub-budget would add a magic number for a case the ctx already bounds — happy to add one if you'd prefer a hard guarantee that `DeleteAPIKey` always gets a fixed slice.
- **404 → "has been revoked" copy** (R1 #3, nit). The message states the key's *state* ("has been revoked"), which is accurate whether this call revoked it or it was already gone — it doesn't claim "we just revoked it." Leaving as-is per "harmless for uninstall semantics."
- **`upstream_revoked` log field not asserted** (R1, minor). The `revoked` boolean's effect is already asserted via the user-facing copy — the success test asserts "API key has been revoked" and the legacy/403 tests assert "does not revoke the qURL API key." Asserting the slog attribute too would duplicate that against brittle log-capture machinery the package's handler tests don't otherwise use.
- **Misleading reply on revoke-then-`NotConfigured` race** (R2 #3, not blocking). Narrow race; by the time the reply is sent the workspace genuinely *is* disconnected (the columns are gone, whether by this call or the concurrent one), and the revoke that happened was logged. Adding a `revoked`-aware branch to the `ErrWorkspaceNotConfigured` arm for a self-correcting race isn't worth the extra surface.

CI green; re-running the loop after this batch.

justin-layerv:
Thanks — third pass. Disposition:

**Fixed (8bcd75be):**
- **R3 #1 — "404 self-heals" over-claimed.** You're right: the revoke-ok/delete-fail retry authenticates the DELETE with the *already-revoked* workspace key, so qurl-service may return 401/403 (auth) before 404 (existence). Softened the `classifyUninstallRevokeError` comment to stop asserting the 404 mechanism — it now notes that either response completes the disconnect via the degrade path, the caveat then under-claims (safe direction), and the disconnect never depends on guessing 404-vs-401. Folded the auth-vs-existence ordering into #805 as an explicit acceptance item.

**Acknowledged, no change (informational):**
- **R3 #2 — brief broken-but-connected window after revoke-ok/delete-fail.** Agreed this is the accepted tradeoff of revoke-before-delete (the alternative ordering risks orphaning a live key, which the PR prioritizes against). Self-heals on the next uninstall. Calling it out in the thread for the record; no change, as you noted.

**Confirmed:**
- **R3 #3 — budget.** Thanks for independently verifying the revoke is ctx-bounded via `waitForRetry`'s `ctx.Done()` select — that matches the reasoning in my earlier reply, so I'm leaving the single 1.2s ctx as-is (no sub-budget).
- **Nit — budget comment drift.** The "2s envelope" wording is already gone in 2532cbf2 (now "1.2s … within Slack's 3s ack window"); no other site in the changed code carries the old number.

The one genuine production gate remains the self-revoke contract (#805) — degradation is graceful either way, but I've flagged it in code (`TODO(upstream-contract)`), the PR body, operating.md, and the tracking issue.

justin-layerv:
Fourth pass — thanks. Disposition:

**Fixed (b6dee5d1):**
- **Docs nit — wrong issue number.** Correct catch: `key_id` persistence landed in **#791**; #792 is this revocation work. operating.md now attributes the legacy fallback to #791.
- **Revoke-ok/delete-fail 401 window.** Added an operating.md note: a short burst of workspace-wide 401s between a failed uninstall and its retry is expected (the workspace points at a now-revoked key until the retry self-heals), not a separate incident.
- **Budget framing.** Tightened the `deleteWorkspaceAPIKey` comment — the qURL client is `WithRetry(2)`, so the revoke is best-effort within the ctx and a retry storm aborts (key_id preserved) rather than missing the ack. (Behavior already matched the "5xx → abort" matrix row, as you noted; this just stops the comment from understating the worst case.)
- **Uncovered abort branch.** Added `TestSlashCommandUninstallClientBuildErrorAbortsWithoutDelete` + an `apiKeyErr` field on `recordingAuthProvider`, exercising the `authenticatedClient` build failure (non-`ErrWorkspaceNotConfigured` from `APIKey()`) → abort before revoke and before local removal. All six prior matrix tests plus this one now pin every abort path.

**Acknowledged, no change (you flagged as fine to leave):**
- Revoke-before-delete recoverable-but-broken window and the `revoked`-then-`NotConfigured` race message — both are the accepted tradeoff of "never orphan a live key," self-heal on retry, and are now documented in operating.md.

**Confirmed intentional:** `key_id` is an identifier (not the secret key) and is deliberately logged + surfaced in the runbook query for operator traceability; the comment density documents non-obvious *why* (ordering invariant, self-revoke contract).

CI was fully green on the prior commit (incl. slack/docker); re-running the loop on b6dee5d1.

justin-layerv:
Fifth pass — thanks, much appreciated. Disposition:

**1. #805 self-revoke contract is load-bearing — confirmed/tracked.** Agreed the "API key has been revoked" confirmation can't be fully trusted until #805 lands (with, ideally, an integration test against a real qurl-service). It degrades safely (401/403 → local-only + the old caveat) and is flagged in code (`TODO(upstream-contract)`), the PR body, operating.md, and #805. Not advertising it as definitively revoking until #805 closes.

**2. Availability tradeoff on hard-abort — posture confirmed + escape hatch documented/tracked (cd2eb3d1):**
- The "preserve key_id, abort before local removal" posture is intentional. For a *compromised* workspace the urgent action is revoking the **live** key — severing only local Slack creds while the leaked key still works via the qURL API would give false security, and dropping local state would orphan a key we can no longer identify (the exact failure #792 set out to fix). So revoke-first is the security-correct default, not just the no-orphan default.
- You're right there was no documented override for a *persistently* stuck upstream. operating.md now states why the abort is deliberate and documents the operator manual escape hatch (revoke via qURL tooling, then clear the `workspace_state` row directly). A first-class admin/operator `--force`/local-only command is filed as **#806** (lower priority than #805).

**3. 404 → "revoked" wording — keeping (state-accurate).** The 404 branch means the key is absent upstream, so "has been revoked" describes the resulting state accurately for uninstall (the workspace's key is no longer live), even if a third party deleted it out-of-band. I softened the *mechanism* claim in the code comment in an earlier round; the user-facing copy stays. Noting it's been raised thrice and is consistently cosmetic.

**Minor (no action, confirmed):** the two provider reads are fine within budget; budget reasoning checks out against `WithRetry(2)` + the ctx bound.

CI was fully green on b6dee5d1; re-running the loop on cd2eb3d1.

justin-layerv:
Sixth pass — thank you. Disposition:

**Fixed (d2bbff68):**
- **#1 — `revoked==true` then `DeleteAPIKey→ErrWorkspaceNotConfigured` misleading reply.** Took the stronger option you offered: when the row's qURL columns are already gone (concurrent uninstall / partial row) but this call *did* revoke the upstream key, the reply now confirms the revoke instead of saying "isn't currently connected." Factored the revoked reply into one const shared by the clean path and the race path. (This had come up across three passes — fixing it rather than re-dispositioning.)
- **#3 — transient 401/403 false positive.** Good catch. Noted in the `classifyUninstallRevokeError` comment (and added to #805) that the degrade assumes 401/403 is *structural*; a transient 401/403 (just-rotated key, clock skew, gateway blip) would drop local state while the key is still live. The contract confirmation in #805 now includes routing transient 401/403 to the abort path if it's possible.
- **Coverage — revoke-success → DeleteAPIKey-failure.** Added two tests: row-already-gone → revoked confirmation (the new #1 behavior), and generic delete failure → retry guidance (the documented 401-burst window). Every uninstall transition now has a test.

**#2 — two provider reads / `APIKey` caching: confirmed.** `DDBProvider.APIKey` is cache-backed (5-min TTL `apiKeyLookupCache`); a warm hit avoids KMS and does only a strongly-consistent validation read, and the admin is actively in Slack so the key is typically warm. So the budget assumes `APIKeyID` (strong read) + `APIKey` (warm validation read) + one `WithRetry(2)` DELETE, all ctx-bounded under 1.2s — and a cold/slow upstream aborts rather than missing the ack.

**404 → "revoked" wording: holding.** State-accurate for uninstall (the key is absent upstream either way); softened the *mechanism* claim in the comment in an earlier round. Noting it's consistently been flagged as cosmetic/low-priority.

CI was fully green on cd2eb3d1; re-running the loop on d2bbff68.

justin-layerv:
Seventh pass — thank you, and glad it reads as merge-ready on correctness. Disposition of the two non-blocking items:

**Fixed (e3f109a6):**
- **#1 — cache/strong-read skew as a transient 401/403 source.** Good catch: `APIKeyID` strong-reads `key_id` but `authenticatedClient`→`APIKey` resolves the plaintext from the 5-min cache, so a stale cached key racing a freshly-rotated `key_id` is a fourth transient source that would degrade. Named it explicitly in the `classifyUninstallRevokeError` comment alongside the other three (it's also covered by the structural-vs-transient question in #805).
- **#2 — key-vanished-between-reads coverage.** Added `TestSlashCommandUninstallKeyVanishedBetweenReadsDisconnectsLocally` (`keyID` set, `apiKey:""` → `APIKey` returns `ErrWorkspaceNotConfigured`), asserting the local-only fall-through with no upstream DELETE. That was the last untested branch in `revokeWorkspaceUpstreamKey`.

**#3 — `revoked && DeleteUnsupported`: confirmed doubly unreachable, no change.** Agreed — `requireUninstallAvailableAndAuthorized` gates on `SupportsDeleteAPIKey()` before we get here, and only `*auth.DDBProvider` (which supports delete) implements `workspaceKeyRevoker`, so a revoke can't precede an "unsupported" delete. Left as-is rather than adding a dead `revoked` check.

The only open item remains the #805 self-revoke contract (degrades safely either way). CI green on d2bbff68; re-running the loop on e3f109a6.

justin-layerv:
Eighth pass — thank you. These are the two substantive asks; both are decisions rather than code changes, dispositioned below.

**#2 — should 401/403 preserve `key_id` (abort) instead of degrading to local-only?** Keeping the degrade **for now**, deliberately:
- Abort would be the safer direction *only once there's an escape hatch*. Today there isn't (the force/local-only command is #806, not yet built). If self-revoke turns out structurally unsupported, flipping 401/403 to abort would make **every** uninstall fail permanently with no way to disconnect — regressing a command that worked before this PR. Degrade lands on exactly that pre-PR behavior (local-only + the exposed-key caveat), so it is not a regression.
- Weighing the two failure modes under the unconfirmed contract: degrade's risk is a narrow, transient, caveat-mitigated window on the workspace's *own* key; abort's risk is total, permanent, unescapable breakage. The lower-expected-harm interim is the one that can never stop uninstall from working.
- So this reconciles with your ask rather than rejecting it: **abort becomes correct after #805 confirms self-revoke is supported, or after #806 ships the escape hatch** — not before. I've coupled that decision to #806 in #805. #3 (rotation strong-read/cache skew) is the same transient-401/403 family and rides on this decision.

**#1 — feature efficacy gated on the self-revoke contract.** Agreed it's load-bearing, and I'd flag it for DE/owner sign-off before relying on this in prod (tracked in #805). One clarification on the user-facing copy: "has been revoked" only renders on an actual 2xx/404 from qurl-service. If self-revoke were unsupported, every attempt returns 401/403 → the **caveat** reply, never "has been revoked" — so the confirmation language is already gated by the real response, not assumed. Whether to block *merge* on #805 vs. ship-degraded-and-confirm is a DE/owner call; I'm surfacing it, not deciding it.

**#4 / nits — no change.** The budget code-comment already notes the `WithRetry(2)` retry-storm-aborts case (the PR-body "one DELETE" line is the happy-path summary). The verbose `classifyUninstallRevokeError` comment is intentional given the security-sensitivity (prior passes agreed); the unguarded test counters are fine for the single synchronous uninstall call.

This is where the design converges: every remaining thread is the #805 self-revoke contract, which can't be resolved in-repo. CI is fully green on e3f109a6 and `/simplify` is clean. Treating the PR as merge-ready pending the #805 DE decision; not self-merging.

justin-layerv:
Thanks — addressed in `eddeb469` (folded into the #805 contract-confirmation work):

- **(1) key_id erased after the 403 degrade; log line is the only surviving pointer.** Good catch — this is the security-motivated path. Added an explicit note to operating.md's `status=403` bullet: the local disconnect removes `key_id` from the workspace row, so that `WARN` line's `key_id` is the only remaining pointer to the still-live key; operators must capture it promptly and ensure the log group's retention outlives the manual-revoke turnaround. I kept the behavior as-is (first-class `key_id` preservation on the 403-with-live-key path is the kind of change that belongs with the #806 escape hatch, which is where the abort-vs-degrade decision lives) and said so in the note.
- **(2) test gap: `APIKeyID` returning `ErrWorkspaceNotConfigured`.** Added `TestSlashCommandUninstallKeyIDReadNotConfiguredDisconnectsLocally`, pinning that first switch arm (no upstream DELETE, falls through to `DeleteAPIKey` local-only) distinctly from the `APIKey`/client-build fall-through (`KeyVanishedBetweenReads`) and the generic-error abort.
- **(3) shared-budget squeeze on the defensive 204 path.** Left as-is intentionally: under the confirmed contract the real path is 403 (fast, no retry), and the 204 branch is unreachable in prod — the budget interaction would only matter if an owner-auth revoke path (#806) makes 204 reachable, at which point it should be revisited with that work. The existing 404 self-heal already covers the documented retry.

CI is green on the merged-with-main branch (the earlier CONFLICTING state was a stale merge against #807, now resolved).

justin-layerv:
Thanks — this is the right thing to press on. Addressed in `a4476271`:

**(1) Is a transient 401 possible for a still-live key? Confirmed: no.** I traced qurl-service's auth middleware (`internal/api/middleware/apikey.go`):
- A live key (active, unexpired, present) **passes** auth and reaches the handler's 403 gate — it never 401s.
- 401 is returned **only** for a genuinely revoked / expired / unknown credential (deterministic).
- A transient infra problem (the `repo.GetByHash` datastore error path) surfaces as **500**, not 401 — so it lands on our 5xx **abort** arm, preserving `key_id`.
- The 60s auth cache can't manufacture a live-key 401 either: only *active* keys are cached, so the worst case is a just-revoked key still seeing 403 for up to the TTL — a dead key getting 403, which still degrades correctly.

That contract is pinned upstream by `TestAPIKeyAuth_ValidKey` (live → passes → 403), `TestAPIKeyAuth_RevokedKey` / `TestAPIKeyAuth_ExpiredKey` (dead → 401), and `TestAPIKeyAuth_RepoError_Returns500` (infra → 500, not 401). I made the assumption explicit in `classifyUninstallRevokeError` and cited those tests, so a future reader (or a qurl-service refactor) has the dependency in view. The "stale plaintext after rotation" case resolves the same way: rotation revokes the old key, so a stale stored key is a *revoked* key → 401 is correct (it's dead). I kept the degrade rather than aborting on 401, per the #805 decision-coupling (flipping to abort is sequenced behind #806).

**(2) Guaranteed-403 round trip — note in code.** Added a comment at the `RevokeAPIKey` call site in `revokeWorkspaceUpstreamKey` saying the always-403-then-degrade is by design, kept for the 404 path and the future owner-auth path (#806), so a reader doesn't mistake it for a bug.

**(3) 5xx abort blocks local disconnect.** Agreed this is the right default (don't sever local access while orphaning a key we can no longer identify), and the interim reply tells admins to retry; operating.md points at the manual escape hatch. The first-class fix is exactly #806 — noted its priority there.

Minors: kept the doc-comment volume (house style, load-bearing) and the `qURL`/caveat strings as-is.

justin-layerv:
Thanks — the README catch is a real one, and it's the security-sensitive surface. Fixed in `64f8fd51`.

**README accuracy (main point).** Reworded the `/qurl uninstall` row so it no longer claims the key is revoked or frames local-only as legacy-only: it now says uninstall is a local-only disconnect that does **not** revoke the upstream key (an API key can't revoke itself), with a pointer to revoke via qURL key management / operator if the disconnect is security-motivated — matching operating.md. I also grepped the rest of `apps/slack/README.md` and `docs/` for other "uninstall … revoke" overstatements; the only remaining hits are the accurate "future work (#806)" and legacy-row lines.

**Minors:**
- *Guaranteed-failing call on the hot path.* Deliberate (404 path + future #806), and now commented at the `RevokeAPIKey` call site. Agreed that if #806 slips, gating the call so a flapping/slow upstream can't eat the budget on a request that can't succeed is the right hardening — noted on #806.
- *Two sequential DDB reads.* Acknowledged; the second (`APIKey`) is ttlcache-backed and both sit inside the 1.2s budget with margin (the matrix tests pass within it).
- *Comment density.* I considered trimming `classifyUninstallRevokeError`, but kept it: it's the in-code executable record of the cross-repo contract, including the 401-degrade-safety rationale this review chain specifically asked me to make explicit, and it already cross-references operating.md for the operator-facing detail. I'd rather it drift loud (a failing assumption is visible at the code) than under-document the one classification whose safety depends on upstream behavior.
- *401 caveat over-warns.* Confirmed deliberate — over-warning about an already-dead key is the safe direction.

justin-layerv:
Addressed in `8b3d7ed0`.

**(1) Availability coupling — yes, intended.** Confirmed: the abort-before-`DeleteAPIKey` on an uncertain error is the deliberate conservative interim, and you've correctly identified that the confirmed contract makes it a pure availability-for-identifiability trade (the revoke never succeeds for a live key, so the only thing the abort buys is preserving `key_id` **in the row** — not just the log line — so an operator can still identify which key to manually revoke; a degrade would erase it). It does invert the prior always-cut-local-access guarantee, so I added an explicit **Availability note** to `operating.md`: a persistent upstream outage (or a budget-exhausting slow upstream) can now block the disconnect, with the manual `workspace_state` row edit as the interim out, and flagged that #806 should be sequenced soon precisely because that manual edit is otherwise the only escape during an outage. The abort-vs-degrade flip for these paths is the same decision sequenced behind #806 (per the #805 decision-coupling), so I kept the behavior and documented the trade rather than changing it here.

**(2) 404 "has been revoked" wording.** Good catch, and sharper under the contract: since the 204 path is unreachable in prod, the *only* prod path that shows this reply is the 404 (already-gone) one — so the old wording asserted an action that never happens in practice. Reworded to "…has been revoked (or was already revoked upstream)" with a comment explaining why the hedge is there.

**(3) Double workspace read.** Added the one-line note: `APIKeyID` returns the key_id and `authenticatedClient` needs the KMS-decrypted plaintext; rotation's combined `APIKeyIdentity` returns key_id + account (not the plaintext), so it can't be reused — the second read is ttlcache-backed and both stay inside the sync budget.

Thanks for the deep read on the failure semantics — the availability inversion was the right thing to make explicit.

justin-layerv:
Addressed in `52c54e52`.

**(1) Both `revoked=true` arms unreachable — yes, intended, and you're right it's 404 too.** Confirmed: shipping the (currently-doomed) revoke call now is the deliberate safe interim ahead of #806, rather than gating the whole attempt behind the owner-auth path — it keeps the wiring + tests in place and the 404/204 arms light up once #806 makes the credential and target distinct. And your sharper point lands: 404 is **as unreachable as 204** for a self-revoke, because the credential *is* the target key — a deleted key surfaces as a `401` at the auth layer, never a `404` at the existence check. So I stopped treating 404 as a live "already-revoked" outcome and flagged the genuinely-dead machinery as defensive-for-#806 in four places:
- `classifyUninstallRevokeError` (the 404 bullet + a "net for a self-revoke" summary: only 403/401→degrade and 5xx→abort occur; `upstream_revoked` is effectively always false),
- the `revokedReply` comment (it's not shown in prod today),
- `operating.md` — the intro now says `upstream_revoked` is **always** false today, and the 404 line explains why it can't fire, and
- the 404 test, marked defensive like the 204 one.

**(2) Abort copy.** Fixed — `:warning: Couldn't disconnect qURL right now…` instead of "Couldn't revoke…", since the same abort also covers the `key_id`-read and client-build (KMS) failures where no revoke was attempted. The "Nothing was disconnected — try again" guidance is unchanged.

**(3) Budget headroom.** Considered: the outbound DELETE (with its bounded `WithRetry`) plus the two reads and the `DeleteAPIKey` write all run under the one `adminSyncVerbBudget` ctx, so the failure mode of a slow/flapping upstream is a **clean ctx-deadline abort with `key_id` preserved**, not a missed ack — the budget can't be blown, only hit. The added synchronous hop does narrow the headroom vs the old DDB-only path, which is the other reason the gate-the-call hardening is worth doing if #806 slips (noted there).

justin-layerv:
Addressed in `d3eefccd`.

**(2) Comment-to-code ratio — done.** You're right that I'd let the same "403 universal / 401 already-dead / 404 unreachable / defensive-for-#806" reasoning drift into near-verbatim copies (a side effect of pinning the contract precisely over the last few revisions). Consolidated to `classifyUninstallRevokeError` as the single authoritative record and trimmed the `revokeWorkspaceUpstreamKey` doc, the `RevokeAPIKey` call-site comment, the `revokedReply` comment, and the 204/404 test docs to one-line pointers (net −15 lines). `operating.md` I left self-contained on purpose — it's the operator runbook, a different audience that won't read the handler — but it no longer needs to mirror the code's phrasing.

**(1) Net behavior / availability regression — confirmed, conscious choice.** Yes: landing the scaffolding now (the capability interface, the abort/404 machinery, the tests) rather than deferring the upstream call itself to #806 is deliberate — it keeps the wiring and the red-first matrix in place so #806 is a focused change, at the cost of the documented availability trade (an upstream 5xx can now block the disconnect) for zero current revocation benefit. That trade is captured in operating.md's **Availability note** and accepted as the interim; the cleaner end state (owner-auth revoke, or gating the doomed call) is #806. Flagging here so it's on record as accepted, not overlooked.

**(3) Minors.** The generic-`DeleteAPIKey`-failure reply is defensive (unreachable under today's contract, as you note). And confirmed: prod runs `WithRetry(2)` so a real 5xx is up to 3 DELETEs — still inside the 3s ack window and ctx-bounded by the 1.2s budget, as documented; the test client's `WithRetry(0)` is what makes the "exactly 1 DELETE" assertion exact.

Issue #790, opened 2026-06-13
Complete Slack cross-account repoint after qURL binding transfer support


justin-layerv:
## Context

Issue #609 needs Slack setup to support explicit rotation/repoint without leaking extra live workspace keys. The Slack-side rotation path now carries explicit signed setup intent, stores qURL key identity metadata, revokes the stored key before replacement mint, and fails closed when the authenticated qURL account cannot revoke the current key.

## Follow-up

Complete the non-owning cross-account repoint path once qURL exposes binding transfer/rotation support. Today Slack must fail closed before minting if the signed-in qURL account cannot revoke the current workspace key; otherwise it could leave the old account with a live key while storing a new account key.

## Acceptance

- Slack can repoint a workspace from one qURL account to another through an explicit owner-approved flow.
- qURL deactivates or transfers the old workspace binding/key before or atomically with storing the new account key.
- Failure ordering does not leave both the old and new account keys live.
- Normal /qurl setup <email> remains idempotent key reuse by default.

justin-layerv:
Decision + status update.

**Approach: ops-mediated transfer.** qurl-service has no tenant-facing cross-account binding transfer, and #910 deliberately preserves cross-tenant refusal, so an automatic same-flow takeover is off the table (it would risk leaving both the old and new account keys live). The safe shape is: Slack **detects** a cross-account move and routes the owner to a LayerV operator, who performs the transfer with guarded tooling.

This splits the work in two:
- **Slack-side detection + routing — #807 (advances this issue).** Persists the qURL account that minted each workspace key (`qurl_account_id`), promotes `--repoint` to a distinct mode, and unifies `--rotate`/`--repoint` so a cross-account move fails closed (no mint, no revoke), emits `setup_cross_account_repoint_requested`, and routes to the operator. Normal `/qurl setup <email>` stays idempotent reuse.
- **qurl-service operator reassign tooling — layervai/qurl-service#956.** The operator-guarded surface that atomically revokes the old account key and frees/reassigns the binding so the destination account can re-key. Until it ships the operator path is manual (documented in `apps/slack/docs/operating.md`).

Keeping this issue open until both land; #807 does not close it.

PR #807, opened 2026-06-14
feat(slack): detect cross-account repoint and route to operator transfer

justin-layerv:
## Summary
- Persist the qURL account (`id_token` `sub`, which qurl-service uses as the binding/api-key `owner_id`) that minted each workspace key as a new `qurl_account_id` attribute. This is **key provenance only** — the workspace ownership anchor stays the Slack user in `workspace_mappings` (#510). It is best-effort and tolerated empty on the sandbox/no-verifier path, and a blank value never erases prior provenance.
- Promote `--repoint` to a distinct signed `SetupModeRepoint` (it was a silent same-account alias for `--rotate`). Both explicit modes are now gated identically at the slash surface (AdminStore required, legacy-owner reclaim) with mode-aware copy.
- Unify `--rotate` and `--repoint` in the callback into one mode-aware `replaceWorkspaceAPIKey` that reads the stored key identity once (`APIKeyIdentity` → `key_id` + `qurl_account_id`) and branches:
  - **Same qURL account** → revoke-then-replace (the existing rotation).
  - **Different qURL account** → a cross-account move. Slack fails closed (no mint, no revoke), emits `setup_cross_account_repoint_requested` with both account ids and the `mode`, and routes the owner to an operator-assisted transfer. Detecting this from stored provenance also spares the doomed wrong-owner DELETE + revoked-list scan that a wrong-account `--rotate` used to attempt.
  - **Unknown provenance** (legacy row, no recorded account) → `--repoint` fails closed; `--rotate` proceeds and records the account so the owner can self-heal a legacy row.
- Docs: README + operating runbook for the cross-account event and the (still-manual) operator transfer.

## Why this shape
qurl-service has **no tenant-facing cross-account binding transfer** — `POST /v1/external-identity-bindings` refuses reassignment by design (cross-tenant refusal, layervai/qurl-service#910 preserves this). An automatic takeover would leave the old account's key live alongside a new one — the exact double-live-key leak #790 guards against. So the safe Slack-side behavior is to **detect cross-account and route to an operator**, which is the ops-mediated decision recorded for #790.

## Scope / follow-up
- **Advances #790** (does not close it): this is the Slack-side detection + routing half. The operator-guarded qurl-service reassign tooling that actually performs the transfer is tracked in **layervai/qurl-service#956**; the runbook points operators there (and is honest that today's manual path has no clean API).

## Validation
- `GOLANGCI_LINT_CACHE="$(mktemp -d)" PATH="$(go env GOPATH)/bin:$PATH" make check` — `golangci-lint` 0 issues, `go test -race ./...` all green.
- New tests: combined `APIKeyIdentity` read (single GetItem), `qurl_account_id` write/omit-on-empty/delete, `SetupModeRepoint` state round-trip + `Explicit()`, and callback coverage for same-account repoint (rotates), cross-account repoint **and** cross-account `--rotate` (route to operator, no revoke), legacy-row fail-closed, unconfigured-first-setup, and read-error 500.

### DIFF
--- a/apps/slack/README.md
+++ b/apps/slack/README.md
@@ -46,7 +46,8 @@ workspace's Secure Access Agent supports.
 | Command | What it does |
 |---------|--------------|
 | `/qurl setup <email>` | Connect qURL to your workspace. The first person to run it becomes the owner and is the only one who can re-run it. |
-| `/qurl setup <email> --rotate` | Owner-only: revoke the stored workspace key and replace it with a new one. `--repoint` is accepted as a same-account alias; cross-account transfer is not supported yet. |
+| `/qurl setup <email> --rotate` | Owner-only: revoke the stored workspace key and replace it with a new one, on the same qURL account. Signing in with a *different* qURL account is detected and routed to an operator-assisted transfer (same as `--repoint`), not silently rotated. |
+| `/qurl setup <email> --repoint` | Owner-only: move the workspace to a different qURL account. Signing in with the account that already holds the key behaves like `--rotate`; a genuine cross-account move is detected and routed to an operator-assisted transfer (qURL does not let one account take over another's workspace connection automatically). |
 | `/qurl get <$id\|$alias>` | Mint a one-time qURL link for a resource in this channel. |
 | `/qurl get <$id\|$alias> dm:true` | Mint the link and DM it to you instead of posting it in the channel. |
 | `/qurl get <$id\|$alias> reason:"…"` | Mint the link and record a reason in the audit log. |
@@ -144,10 +145,7 @@ re-pointed at a different qURL account. Ask the owner, or use the
 **How do I rotate the workspace key?** The owner runs
 `/qurl setup <email> --rotate`, then completes the Auth0 prompt. Slack revokes
 the old workspace key before storing the replacement, so a failed rotation does
-not leave an extra live key behind. `--repoint` currently behaves like
-`--rotate` and still requires the signed-in qURL account to own the current
-workspace key; cross-account transfer is tracked in
-[issue #790](https://github.com/layervai/qurl-integrations/issues/790).
+not leave an extra live key behind.
 If rotation fails after revoking the old key, retry with `--rotate` or
 `--repoint`; plain setup will not mint around the stored old key identity.
 If that old identity was deleted at qURL rather than revoked, Slack fails closed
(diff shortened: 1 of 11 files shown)


justin-layerv:
Addressed the Claude review:
- **Nit (fix-before-merge):** moved `setupModeFlag`/`setupModeAction` above `handleSetup`'s doc comment so the owner-gate rationale stays attached to `handleSetup` (it was being bound to the helper).
- **Optional:** added a one-line invariant note at the `storedAccountID != "" && storedAccountID != qurlAccountID` site documenting that `qurlAccountID` is non-empty for explicit modes (slash gating + `checkBindAllowed` ordering), so a future refactor can't silently make a same-account row look cross-account.

The three "no action required" notes were confirmations of intended behavior. `make check` still green (0 lint, `-race`).

justin-layerv:
Addressed the third review pass (all hardening/coverage items):
- **Empty signed-in account (belt-and-suspenders):** `replaceWorkspaceAPIKey` now explicitly fails closed (500) if it ever reaches a provenance-bearing row with an empty `qurlAccountID`, instead of relying solely on the upstream `checkBindAllowed` ordering — covered by `TestCallbackExplicitModeEmptyAccountAgainstProvenanceRowFailsClosed`.
- **Legacy `--rotate` self-heal coverage:** added `TestCallbackLegacyRotateRecordsQURLAccount` asserting a legacy-row `--rotate` records `qurl_account_id` (the "records the account for next time" contract).
- **`DeleteAPIKey` condition note:** added the one-line rationale for why `#qurl_account_id` is in REMOVE but not the `ConditionExpression`.

`make check` green (0 lint, `-race`).

justin-layerv:
Addressed the fourth review pass:
- **Observation 3 (optional) — structured event for legacy-repoint refusals:** added `setup_repoint_legacy_row_refused`, emitted when `--repoint` fails closed on a row with no recorded `qurl_account_id`, so operators can measure how often legacy rows block repoint. Asserted by the updated `TestCallbackRepointLegacyRowWithoutAccountFailsClosed`, with a one-line runbook note.
- **Observations 1 & 2:** confirmed as correct/by-design per the review — #1 (`DeleteAPIKey` condition invariant) is already documented in-code, and #2 (shared cross-account copy) is intentional: the page leads with the re-sign-in guidance for a mis-signed-in `--rotate`, and tailoring it per-mode would re-introduce the branching the unification removed. Left as-is.

`make check` green (0 lint, `-race`).

justin-layerv:
Addressed the fifth review pass (both optional doc clarifications):
- **README `--rotate` row:** now notes that signing in with a *different* qURL account on `--rotate` is detected and routed to the operator (like `--repoint`), not silently rotated.
- **Runbook stale-provenance edge case:** added a note to the cross-account section covering the one scenario where a legitimate owner can get stuck — if the stored `qurl_account_id` ever diverges from the key's actual qurl-service owner (e.g. an upstream binding migration that doesn't update the Slack row), the `--rotate` self-heal is blocked and the operator process is the resolution.

Doc-only; `make check` still green.

justin-layerv:
Addressed the sixth review pass:
- **Trademark consistency:** the error pages in `replaceWorkspaceAPIKey` now mark the first singular body mention as `qURL™` (headings stay plain, per CLAUDE.md), making the rewritten function internally consistent with the older setup/bind pages.
- **`setupModeAction` "key repoint" phrasing:** left as-is per the review — it is deliberately parallel to "key rotation" and matches the asserting tests; reshaping it would ripple through the success-URL/legacy-owner copy and tests for subjective benefit.

`make check` green (0 lint, `-race`).

justin-layerv:
Re the seventh review pass (trademark consistency):

My round-6 `qURL™` additions made the untouched pre-existing invalid-key page (callback.go:808, a #791 plain-`qURL` page) the odd one out. Rather than expand this feature PR into a file-wide brand sweep across unrelated pages (`confirmStoredKeyAlreadyRevoked`, `mintAndPersist`, `reuseStoredWorkspaceKey`), I've **reverted the ™ additions** so the PR is trademark-neutral: the repoint pages now use plain `qURL`, matching their adjacent #791 rotate-flow neighbors exactly. The PR introduces no new trademark inconsistency.

The file-wide `qURL™` uniformity (a pre-existing split: setup/mint pages mark it, the #791 rotate flow doesn't) is tracked as a focused cosmetic cleanup in #808 (`pr-polish-followup`, `priority: low`, `good first issue`).

`make check` green (0 lint, `-race`).

PR #809, opened 2026-06-14
test(slack): add Slack bot HTTP liveness smoke tests

justin-layerv:
## Summary

First of two changes adding a **prod smoke for the Slack bot** (the Slack app currently has no automated prod smoke, unlike Discord). This adds the Jest `e2e/` tests; the deploy/promote wiring lands separately and depends on this being on `main` first.

The suite drives the **live deployed Slack bot over HTTP** with synthetic, signed requests — the prod-gate analogue of the Discord smoke's bot-liveness checks. Scope is deliberately **liveness/signature/dispatch only** (read-only, side-effect-free):

- `GET /health` → 200 `{"status":"ok"}`
- signed `url_verification` → 200, echoes the challenge (proves the deployed `SLACK_SIGNING_SECRET` matches what we signed with + the endpoint is reachable end-to-end)
- **bad signature** → 401 (proves signature enforcement isn't bypassed)
- signed `/qurl help` → 200 help text (synchronous dispatch, no qurl-service call)

## Files

- `e2e/helpers/slack-sign.ts` (new) — ports the app's `v0` HMAC scheme (`apps/slack/internal/signature.go`).
- `e2e/tests/slack-liveness.smoke.test.ts` (new) — `describe.skip` when `SLACK_BOT_BASE_URL` is unset; signed cases additionally need `SLACK_SIGNING_SECRET`.
- `e2e/helpers/env.ts`, `e2e/.env.example` — optional Slack vars (test/staging only).

## Out of scope (by design)

Minting a qURL **through** Slack is not tested here: `response_url` is pinned to `hooks.slack.com` (`process.go`), so a synthetic command's async reply is discarded and can't be observed; and qURL functional coverage already exists globally via the shared suite. The hermetic mint→deliver→revoke path **with reply-text assertions** is already covered by `TestConfirm_GetApproveInDMMintsAndDeliversLinkInThread` + the `adminSlashInvoker` harness — so no new Go test was added.

Verifying a real user-visible link *through* Slack would need a test workspace + agent/browser harness — deferred.

## Verification

- `npx jest slack-liveness` with vars **unset** → suite skips (safe to run anywhere, incl. the Discord smoke whose `smoke|…` pattern also matches this file).
- With vars set against a local mock that reimplements the app's `v0` verification → **4/4 pass**, proving the signer produces signatures the app accepts and the assertions are correct.

## Sequencing

Must merge to `main` first — the deploy/promote smoke runs this suite from `main`.

### DIFF
--- a/e2e/.env.example
+++ b/e2e/.env.example
@@ -13,3 +13,14 @@ MINT_API_URL=https://api.example.com
 # Discord test guild/channel
 GUILD_ID=your_test_guild_id
 CHANNEL_ID=your_test_channel_id
+
+# --- Slack bot HTTP smoke (optional; slack-liveness.smoke.test.ts) ---
+# Liveness/signature/dispatch only — no minting, no workspace install needed.
+# Unset SLACK_BOT_BASE_URL to skip the suite; set it for health, plus
+# SLACK_SIGNING_SECRET for the signed cases. These are for running the suite
+# LOCALLY by hand against sandbox — keep production values out of your local
+# .env. (As a deploy gate, CI injects the real per-environment host/secret —
+# including prod — from the deployment config, not from this file.)
+# A trailing slash on SLACK_BOT_BASE_URL is trimmed.
+SLACK_BOT_BASE_URL=https://slackbot.example.com
+SLACK_SIGNING_SECRET=your_slack_signing_secret
--- a/e2e/helpers/env.ts
+++ b/e2e/helpers/env.ts
@@ -19,6 +19,14 @@ export interface E2EOptionalEnv {
   // unset. Per-environment values come from the same SSM source
   // as the bot's BASE_URL config.
   BOT_HTTP_URL?: string;
+
+  // --- Slack bot HTTP smoke (slack-liveness.smoke.test.ts) ---
+  // Public HTTPS base URL of the deployed Slack bot (e.g.
+  // https://slackbot.example.com); a trailing slash is trimmed by the suite.
+  SLACK_BOT_BASE_URL?: string;
+  // The deployed workspace signing secret, used to forge validly-signed
+  // probes. Its presence gates the signed cases (url_verification, /qurl help).
+  SLACK_SIGNING_SECRET?: string;
 }
 
 const REQUIRED: (keyof E2EEnv)[] = [
@@ -47,5 +55,7 @@ export function loadEnv(): E2EEnv {
 export function loadOptionalEnv(): E2EOptionalEnv {
   return {
     BOT_HTTP_URL: process.env.BOT_HTTP_URL,
+    SLACK_BOT_BASE_URL: process.env.SLACK_BOT_BASE_URL,
+    SLACK_SIGNING_SECRET: process.env.SLACK_SIGNING_SECRET,
   };
 }
--- a/e2e/helpers/slack-sign.ts
+++ b/e2e/helpers/slack-sign.ts
@@ -1,0 +1,65 @@
+/**
+ * Slack request signing for the Slack bot HTTP smoke tests.
+ *
+ * Mirrors the `v0` HMAC-SHA256 scheme the bot verifies in
+ * apps/slack/internal/signature.go (and the test helper `signSlackBody` in
+ * handler_test.go): the signature base string is `v0:{timestamp}:{body}`,
+ * keyed by the workspace signing secret, hex-encoded, prefixed `v0=`.
+ *
+ * Keeping this in lockstep with the app means a valid-signature smoke proves
+ * the *deployed* SLACK_SIGNING_SECRET matches what we signed with — the same
+ * config-coherence guarantee the Discord smoke gets from `me.id`.
+ */
+
+import { createHmac } from 'crypto';
+import { fetchWithTransientRetry } from './http';
+
+/** Slack's signature-version prefix (see signature.go `slackSignatureVersion`). */
+const SLACK_SIGNATURE_VERSION = 'v0';
+
+interface SlackSignatureHeaders {
+  'X-Slack-Signature': string;
+  'X-Slack-Request-Timestamp': string;
+}
+
+/**
+ * Computes the headers Slack would send to authenticate `body` at `timestamp`.
+ * `body` is signed verbatim and must equal the bytes POSTed.
+ */
+function signSlackRequest(body: string, signingSecret: string): SlackSignatureHeaders {
+  const ts = String(Math.floor(Date.now() / 1000));
+  const mac = createHmac('sha256', signingSecret);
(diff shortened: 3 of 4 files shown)


justin-layerv:
Thanks for the review — addressed the feedback, and a `/simplify` pass on top:

**Acted on**
- **Transient-retry (main suggestion):** all probes now route through `fetchWithTransientRetry`. Safe even on the signed POSTs since every probe is side-effect-free and the helper only retries statuses that provably didn't reach the app; verified end-to-end against a mock that returns `503→200` on `/health` (the drain-gap case) and recovers.
- **Trailing-slash duplication (nit):** removed — the suite now follows the sibling convention (`${baseUrl}${path}`, no-trailing-slash documented on `SLACK_BOT_BASE_URL`) instead of a bespoke join helper.
- **Content-Type note (nit):** added — documented that the app routes by path, so the header is cosmetic to the app.

**Declined, with rationale — per-request timeout (nit, "low priority")**
I'm not adding `AbortSignal.timeout` here. Two reasons: (1) every sibling smoke relies on jest's suite timeout (`testTimeout: 120_000`) to bound a hung probe — none use per-request timeouts, and diverging adds surface for marginal benefit; (2) a per-call `AbortSignal` combined with `fetchWithTransientRetry` (which reuses `init` across attempts) yields a single signal shared across retries — correct as a *total* budget, but subtle enough to mislead. Minting a per-attempt timeout *inside* the shared helper would be the right altitude, but that changes behavior for the connector smokes too and is out of scope for this PR. Net: relying on jest's bound keeps this consistent with the rest of the suite. Happy to revisit in a follow-up that adds per-attempt timeouts to `fetchWithTransientRetry` for all smokes if that's wanted.

Also simplified the helper surface (dropped an inert `fetchImpl` injection, an options bag with unused fields, and unused `SLACK_TEST_*` env vars) — net −16 lines, single `postSigned` export.

justin-layerv:
Round 4 follow-ups:

1. **Clock-skew false-red (fixed):** added a failure-mode note to the file header — signed cases use the runner's wall clock, so >5 min drift 401s the positive cases as stale (not a signing regression). Good triage-saver, thanks.
2. **"Redundant dotenv load":** keeping it — it's the universal convention here, not an outlier. All 11 other e2e test files call `dotenv.config({ path: … })` explicitly (`smoke.test.ts`, `qurl-oauth-setup.smoke.test.ts`, `file-revoke`, `google-maps`, …). The explicit absolute path also makes it cwd-robust vs. relying on `setupFiles`' cwd-relative load. Removing it from just this file would make it the inconsistent one.
3. **Unsigned-401 narrowness:** agreed, and that's by design — the *positive* signed `url_verification` echo is what proves the deployed secret is correct (it needs a valid HMAC for 200); the unsigned case only asserts missing-header rejection. No change.

Also fixed a skip-safety regression I'd introduced applying the round-3 trailing-slash suggestion: normalizing at `const base = baseUrl!.replace(...)` threw on `undefined` when unset (Jest still executes the `describe.skip` body), crashing the suite instead of skipping. Moved the trim to the module-level read with optional chaining (`optEnv.SLACK_BOT_BASE_URL?.replace(/\/+$/, '')`), so `undefined` stays undefined and the suite still skips cleanly when unconfigured. Verified: 5 skipped when unset, 5 passed against a mock (incl. trailing-slash base + a 503→200 drain-gap retry).

justin-layerv:
Round 5 follow-ups (all optional, addressed):

1. **Deploy-race on positive cases:** correct that the ~3s budget (`maxAttempts=3`) won't outlast a long drain — that's intended fail-fast gate behavior, and it matches the connector smokes (same shared helper, same rollout race). The durable fix is ensuring the *promote* step runs after steady-state; I'll make sure of that in the deploy/promote wiring change (the separate infra PR), rather than inflating the retry budget here.
2. **Body assertion on 401s (added):** both the unsigned and bad-signature cases now also assert `expect(await res.text()).not.toContain('should-not-echo')`, so a regression that 401s but still echoes would be caught.
3. **/health JSON guard (added):** assert `content-type` matches `/json/i` before `res.json()`, so a 200 non-JSON page fails with a clear assertion instead of an opaque parse error (matches the content-type assertion in `qurl-oauth-setup.smoke.test.ts`).

Verified: 5 skipped when unset, 5 passed against the mock.

justin-layerv:
Round 7 follow-ups:

1. **Deploy-gate false-green (most substantive) — tracked:** you're right that as a gate, a missing `SLACK_SIGNING_SECRET` would silently skip the signed cases and go green having validated only `/health` + unsigned-401. The test must stay skip-when-unset (safe to run in the general e2e suite), so the enforcement belongs in the caller/gate workflow. Filed [qurl-integrations-infra#1103](https://github.com/layervai/qurl-integrations-infra/issues/1103) (labeled `pr-polish-followup`/`area: ci`/`priority: medium`/`type: security`) to require the secret + treat a skipped signed case as a gate failure when the deploy/promote wiring lands.
2. **Unsigned-401 scope (fixed):** corrected the comment — it now states this proves "unsigned is rejected", NOT that the secret is configured (an empty secret also 401s via `errSlackSigningSecretEmpty`), and points to the positive signed `url_verification` case as the one that proves the secret is set. Same point captured in #1103.
3. **"prod smoke" vs "don't point at production" (fixed):** clarified the `.env.example` comment — those are *local* hand-run values (keep prod out of your local `.env`); CI injects the real per-environment host/secret from the deployment config, not from that file.
4. **Per-request timeout (declining, consistent with the earlier `/simplify` pass):** the timeout machinery was deliberately removed by `/simplify` as over-built and non-sibling-consistent, and a per-call `AbortSignal` over `fetchWithTransientRetry`'s reused `init` is a shared-signal-across-retries that two review passes found confusing. The right home for a fail-fast deadline is a per-attempt timeout *inside* `fetchWithTransientRetry` (benefiting all connector/slack smokes) — out of scope for this test-only PR. jest's 120s bounds a hung probe in the meantime.

Verified after changes: 5 skipped when unset, 5 passed against the mock.

justin-layerv:
Round 9 — all points are already addressed or by-design:

- **Skip-vs-fail in the deploy-gate context:** same point as round 7 #1; the test stays skip-when-unset (required to be safe in the general e2e suite), and the gate-side enforcement (require `SLACK_SIGNING_SECRET`, fail if signed cases skip) is tracked in [qurl-integrations-infra#1103](https://github.com/layervai/qurl-integrations-infra/issues/1103) for the deploy/promote wiring PR — which you noted is the natural place.
- **/health under a draining task** and **signer/app scheme drift:** both acknowledged as by-design/acceptable-for-a-smoke; no change.

A final `/simplify` pass came back clean across reuse/simplification/efficiency/altitude. Thanks for the thorough multi-round review.

PR #815, opened 2026-06-15
test: browser-knock Tier-2 for the fileviewer tunnel; drop maps tests

justin-layerv:
## Summary

**Why:** After qurl-integrations-infra#1111 decommissioned the legacy EC2 fileviewer host, the e2e suite's "direct-fetch the view URL" pattern times out (`ConnectTimeoutError: fileviewer.layerv.xyz`). Under render-at-mint + the NHP tunnel, a recipient reaches the view **only** via the minted `qurl.link` → SPA reads the `#at_` fragment in JS → knock → the view at `r_<id>.qurl.site/views/<mint-id>`. `node fetch`/curl can't do that; a real browser can. **Verified headless against sandbox:** `302 resolve…` (knock) → `200 …/views/<id>` (the view served).

**What:**
- `helpers/tunnelView.ts` — `viewViaQurlLink(qurlLink)`: headless Playwright drives the knock and resolves on the tunnel-view `…/views/<id>` response, returning its HTTP status (200 = the view served). Throws clearly when no view response arrives (the revoke/negative signal). The capability-bearing URL is deliberately **not** returned.
- `file-revoke.test.ts` — migrated the view check to `viewViaQurlLink` (the real recipient path); dropped the obsolete `/view/` assertion (`resource_url` no longer carries it post-#1113). Revoke/`getLinkStatus` assertions unchanged.
- `helpers/qurl-api.ts` — consolidated `uploadImage` into `uploadFile` (path **or** in-memory bytes); success keys on `resource_id` alone (the connector's documented marker) so the sibling viewer-ttl smoke's contract is untouched.
- `google-maps.test.ts` — **deleted.** Maps/url-type support is being dropped (unused; and url-type can't render on the tunnel — the dumb-proxy refuses inline HTML for XSS safety). The connector's now-dead maps render code, the discord bot maps command, and `location-variants.test.ts`'s maps fixtures are tracked for removal in **#816** (until then the connector maps-render path is live-but-untested — closed when the dead code is removed).
- `playwright` dependency added — the chromium browser is installed in CI by the companion **qurl-integrations-infra#1114**.

**Greens the post-deploy `e2e-smoke`** (the dead-host timeouts in google-maps + file-revoke). `smoke.test.ts` / `location-variants.test.ts` don't touch the dead host (unchanged).

## Dependency
Merge with **qurl-integrations-infra#1114** (installs Playwright chromium in `e2e-smoke.yml`). Without it the browser test errors "browser not found" (Playwright's own error names the `npx playwright install` fix). The install runs *after* `npm ci`, so the browser matches the lockfile-resolved Playwright version.


justin-layerv:
**Review-response summary** (all rounds). Addressed in code:
- Always-resolving tunnel-view waiter (no goto/waiter unhandled-rejection mis-attribution).
- `uploadFile` succeeds on `resource_id` alone — decouples the sibling viewer-ttl smoke; `qurl_link` typed optional with a throwing guard in file-revoke.
- Dropped the capability-bearing tunnel-view URL from the helper result.
- Documented the first-`/views/`-match-is-200 assumption (verified topology).
- Maps dead-code cleanup tracked in #816 (referenced from the body).

Consciously **declined**, with rationale:
- *Tighten `TUNNEL_VIEW_RE` after `.qurl.site` to `[:/]`*: that would break the deliberate host-suffix-agnostic match of `.qurl.site.layerv.xyz`/`.layerv.ai`. The helper only ever matches the response URL of a navigation to our **own** API-minted link, so a hostile `.qurl.site.evil.com` suffix can't be injected — test-only, trusted source.
- *Local-dev chromium README / postinstall hook*: Playwright's native launch error already names the exact fix (`npx playwright install`); a postinstall hook would force a ~150 MB chromium download on **every** CI `npm ci`, including jobs that never use it.
- *`playwright` under `devDependencies`*: the e2e package puts all deps (jest/ts-jest included) under `dependencies`; moving it would diverge from that convention.

`/simplify` returns clean (the two findings — FormData-in-retry-loop and cross-suite cleanup-helper extraction — are out-of-scope/not-safe and were skipped; details on request).

**Merge ordering:** land this on `main` BEFORE infra#1114 — see that PR's body (the `--no-install` guard there would otherwise auto-revert a healthy discord deploy).

PR #818, opened 2026-06-15
test: knock the connector-minted recipient view through the fileviewer tunnel (Tier-2)

justin-layerv:
## Why

The fileviewer reverse-tunnel Tier-2 smoke is the only automated check that a **recipient can actually open a file shared through the tunnel** — and it now gates every sandbox deploy (#815 + qurl-integrations-infra#1114). On its first live run it **failed**, for the wrong reason: it browser-knocked the wrong link and 404'd. Left unfixed, it would auto-revert *healthy* sandbox deploys while giving us **zero** real tunnel coverage. This fixes it to exercise the true recipient path, so the gate is both correct and meaningful.

## Root cause (verified live, not a regression)

In render-at-mint mode the connector's `/api/upload` `qurl_link` is the **"never-shared per-upload qURL"** — it targets `/resources/<md5>`, so its NHP knock 302s to `r_<uploaded-resource>.qurl.site/` (root) → **404**. The tunnel itself serves fine. Recipients only ever reach content through a **per-recipient view minted separately** via `POST /api/mint_link/<resource_id>`, whose knock resolves to `r_<tunnel>.qurl.site/views/<mint-id>` → **200**.

`file-revoke` (migrated in #815) knocked `/api/upload`'s `qurl_link` — the wrong one. (The 2026-06-15 manual check passed only because it happened to drive a separately-minted link.)

Network trace from the failed CI run + local repro:
```
upload link:  302 resolve.qurl.link → 404 r_<resource>.qurl.site/            ✗
minted link:  302 resolve.qurl.link → 200 r_<tunnel>.qurl.site/views/<id>    ✓
```

## What

- `helpers/qurl-api.ts` — add `mintConnectorView(uploadUrl, resourceId, apiKey, {expiresAt, oneTimeUse})` → `POST /api/mint_link/:resource_id`, returns the per-recipient `qurl_link`. Documented as distinct from `mintLink` (that hits the qurl-service `MINT_API_URL`; this hits the **connector**).
- `tests/file-revoke.test.ts` — first test now: upload → **mint a recipient view** → browser-knock the minted link → 200 → revoke → `getLinkStatus` 404. Removed the now-dead `upload.qurl_link` guard; revoke/404 assertions unchanged (they passed in the failed run). Timeout 60s→90s for the added mint + cold-CI chromium launch.
- **No infra change** — the deploy gate already selects `file-revoke`; it just tests the right link now.

## Verification

Verified end-to-end against sandbox with the **exact helper chain** this test runs: `uploadFile → mintConnectorView` (`expires_at` via `toISOString()`, millis form accepted by the connector) `→ viewViaQurlLink` → **status 200**. tsc clean.

**Not** CI-verified: local used a sandbox token; CI uses the SSM `QURL_API_KEY`. Upload already succeeded in the failed CI run with that key and `/api/mint_link` uses the same bearer, so mint scope is low-risk — but the **first post-merge sandbox deploy** is the true confirmation.

### DIFF
--- a/e2e/helpers/http.ts
+++ b/e2e/helpers/http.ts
@@ -17,11 +17,15 @@
  *         provably did NOT reach/complete at the app (503 = ALB has no healthy
  *         target, the drain-gap; 408/425/429 = rejected/timed-out before
  *         processing), so a retry can't duplicate work even on a POST. NOTE the
- *         503-on-POST guarantee is contingent on the connector signaling
- *         rate-limits via HTTP 200 + `error` (not an app-level 503), so a 503
- *         here is always the ALB no-healthy-target case, never a post-accept app
- *         503. If the connector ever returns a real app-level 503 after partially
- *         processing, move 503 to the idempotent-only set below.
+ *         503-on-POST guarantee is contingent on the connector never emitting an
+ *         app-level 503 *after* partially processing. Both POST endpoints honor
+ *         that: `/upload` signals rate-limits via HTTP 200 + `error` (never a
+ *         503), and `/api/mint_link` DOES emit one app-level 503 — its pre-bake
+ *         "render-at-mint not ready" guard — but that fires BEFORE any bake, so a
+ *         retry still can't duplicate a `views/<mint-id>` object (its post-bake
+ *         failures surface as 502, which POSTs don't retry). If any endpoint ever
+ *         returns a 503 AFTER partially processing, move 503 to the
+ *         idempotent-only set below.
  *       · IDEMPOTENT methods (GET/HEAD/…, e.g. the `/view` read) ALSO retry
  *         {502, 504} — transient gateway failures where the backend MAY have
  *         already processed the request before the response was lost. Retrying
--- a/e2e/helpers/http.ts
+++ b/e2e/helpers/http.ts
@@ -23,9 +23,10 @@
  *         503), and `/api/mint_link` DOES emit one app-level 503 — its pre-bake
  *         "render-at-mint not ready" guard — but that fires BEFORE any bake, so a
  *         retry still can't duplicate a `views/<mint-id>` object (its post-bake
- *         failures surface as 502, which POSTs don't retry). If any endpoint ever
- *         returns a 503 AFTER partially processing, move 503 to the
- *         idempotent-only set below.
+ *         failures surface as 502, which POSTs don't retry). TODO(upstream-contract):
+ *         this mirrors the qurl-s3-connector mint/upload contract (handler.go —
+ *         pre-bake 503 guard + 502 default) — if either endpoint ever returns a
+ *         503 AFTER partially processing, move 503 to the idempotent-only set below.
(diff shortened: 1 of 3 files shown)


justin-layerv:
Thanks — verified all four against the connector mint handler + `http.ts`, all resolved (no code-behavior change needed; added a doc block to `mintConnectorView` capturing the reasoning):

**1 & 2 (retry-safety + 429 backoff):** The mint endpoint honors `http.ts`'s exact 503-on-POST caveat. A 503 from `/api/mint_link` comes ONLY from the pre-bake `tunnelResourceID==""` "render-at-mint not ready" guard (handler.go:4086-4092, before any `mintOne`/bake) — so a retry can't double-bake a `views/<mint-id>`. Post-bake failures surface as **502** (`statusCode` default `StatusBadGateway`, handler.go:4125), which `fetchWithTransientRetry` does **not** retry for a POST (502/504 are idempotent-only). And a rate-limited mint is an actionable **HTTP 429** (not upload's 200+`error` shape), which the helper already backs off on — so no separate 200+error loop is needed here. So the POST retry is safe and rate-limit is covered.

**3 (`n: 1`):** documented now (mint exactly one link → read `links[0]`).

**4 (one-time-use + view-then-revoke):** confirmed fine — `viewViaQurlLink` does exactly one knock+navigation (one use), and the 404 assertion is resource-level (the sibling double-revoke test proves revoke 404s without any mint/view). No action.

justin-layerv:
Both addressed:

- **503 doc reconciliation:** updated `http.ts` to state that `/api/mint_link` DOES emit one app-level 503 (the pre-bake "render-at-mint not ready" guard) but it fires BEFORE any bake, so the 503-on-POST retry stays safe (post-bake failures surface as 502, which POSTs don't retry). The two docs now agree.
- **`success` gate:** dropped `!data.success` — gate on `link?.qurl_link` alone. `!res.ok` already catches error statuses and a 200 always carries `links[]`, so the link's presence is the authoritative signal (mirrors `uploadFile` keying on `resource_id`). Confirmed `success:true` is contractual on the 200 path (handler.go:4165) but we no longer depend on it.
- Added the `oneTimeUse` default note. Return-type `qurl_id`/`expires_at` left as-is (mirrors `MintResult`).

justin-layerv:
Both addressed: narrowed `mintConnectorView`'s return to `{ qurl_link: string }` (the only guard-checked + consumed field — no more typed-but-unenforced `qurl_id`/`expires_at`), and added a `TODO(upstream-contract)` marker on the `http.ts` 503 note pointing at the qurl-s3-connector mint/upload contract so the lockstep site is greppable. Timeout budget left at 90s (happy path well under; the view portion is the lever if degraded-CI flakes appear, as you noted).

justin-layerv:
Removed the redundant `expect(minted.qurl_link).toBeTruthy()` — `mintConnectorView` throws on a missing link + returns a typed-present `{ qurl_link }`, so it couldn't fail. The mid-deploy pre-bake-503 flakiness note is a real future tuning item (out of scope here, as you said) — noted for if the gate ever flaps. Thanks for the thorough passes.

PR #820, opened 2026-06-15
feat(slack): cascade-delete all workspace data on app uninstall / token revoke

Kevin-layerV:
## Why

This closes a **Slack Marketplace blocker**. Today, when a workspace uninstalls the qURL app from Slack, **nothing is deleted** — there is no handler for Slack's lifecycle events. And `/qurl uninstall` clears only the qURL API key columns: the **KMS-encrypted Slack bot token**, the `workspace_mappings` row (owner + admins), and all `channel_policies` rows persist forever. Marketplace review requires that uninstalling the app forgets the workspace's data.

## What

The bot is multi-tenant; per-workspace data lives in four DynamoDB tables (keyed by Slack `team_id`, or `enterprise_id` for org installs). This PR adds a full per-workspace purge and wires it to both the Slack lifecycle events and the `/qurl uninstall` command.

### 1. Deletion methods (each idempotent — an absent item is a no-op)

- **`shared/auth/ddb_provider.go`** — `DDBProvider.DeleteWorkspaceState(ctx, workspaceID)` deletes the **entire** `workspace_state` row: the encrypted Slack bot token + its data key, the encrypted qURL API key + its data key, and all install/setup metadata. The existing `DeleteAPIKey` (which clears *only* the qURL columns and deliberately preserves Slack install metadata for a key rotation) is left unchanged. `DeleteItem` is added to the package's `DynamoDBClient` interface.
- **`apps/slack/internal/slackdata/workspace.go`** — `Store.DeleteWorkspaceMapping(ctx, teamID)` deletes the `workspace_mappings` row (owner, admin set, agent toggle).
- **`apps/slack/internal/slackdata/policies.go`** — `Store.PurgeTeamChannelPolicies(ctx, teamID)` Queries every `channel_policies` row for the team (paged over `LastEvaluatedKey`) and `DeleteItem`s each. Sequential deletes (not `BatchWriteItem`) keep the client interface unchanged; per-team policy rows are bounded by the channels the workspace used the bot in.

### 2. Orchestrator + event handling (`apps/slack/internal/handler_lifecycle.go`, new)

- **`purgeWorkspace(ctx, log, workspaceID)`** — best-effort: it **attempts all three deletes** even if one fails, so a transient error on one table never strands the others. Each delete is independently idempotent, so a partial prior purge — or a re-delivered Slack event — converges on a retry. Failures are logged with the **workspace id only** (the deletes carry no token material, so nothing secret is logged). Upstream qURL-key revocation is deliberately *not* done here — it stays the caller's concern, keeping `purgeWorkspace` a pure local storage sweep and the `auth` package free of the qurl-service client dependency.
- **`handleLifecycleEvent`** — routes `app_uninstalled` and `tokens_revoked` event_callbacks. It is dispatched from `handleEvent` **before** `handleAgentEvent`, because `handleAgentEvent` returns early when conversation mode is unwired — the cascade must run regardless of agent wiring. It resolves the workspace id from `team_id`, falling back to `enterprise_id`, and **always acks `200 OK` promptly** (the purge runs asynchronously off the server base context via the existing wg-tracked `h.Go`), so Slack does not retry the delivery forever even if the purge is slow or fails.
- **`tokens_revoked` is treated as an uninstall.** The bot's only install credential is the workspace bot token (`shared/auth.SlackBotTokenInstall`); it holds no per-user tokens. Slack does not reliably distinguish a bot-token revoke from a lone user-token revoke in this payload, so any `tokens_revoked` for this app means the credential it depends on is being torn down. Re-install re-seeds the row from scratch.
- A `workspaceStateDeleter` capability interface (discovered by type assertion, mirroring the existing `workspaceKeyRevoker` pattern) keeps `DeleteWorkspaceState` off the base `auth.Provider` interface; a compile-time assertion ensures the production `*auth.DDBProvider` keeps satisfying it.

### 3. `/qurl uninstall` extension (`handler.go`)

After the existing best-effort upstream qURL-key revoke and `DeleteAPIKey`, the success path now calls `purgeWorkspace` so the command fully forgets the workspace (bot token + mappings + policies), not just the qURL key. It runs synchronously inside the same sync ack budget; because it is best-effort and the primary disconnect already succeeded, a sweep failure is logged and does not change the user-facing success copy (which is unchanged).

### Out of scope: `qurl_agent_state`

The conversation-transcript / audit table (`qurl_agent_state`) is **not** purged. It is a separate composite-key table owned by `AgentStore` (not the `Store` tables this PR touches), every item is already **DynamoDB-TTL'd**, and a team-wide sweep there would need its own paged Query + batch-delete on a different struct — non-trivial and unnecessary for this blocker. Left as a follow-up.

## Tests

- `shared/auth`: `TestDDBProviderDeleteWorkspaceState` — whole-row delete by `team_id`, unconditional (idempotent), absent-row no-op, empty-id rejected before any DDB call, transport error wrapped. (The package's test fake gained a `DeleteItem` method.)
- `slackdata`: `TestDeleteWorkspaceMapping` and `TestPurgeTeamChannelPolicies` — correct keys, idempotent absent-row/empty-team no-ops, 400 on empty id before DDB, multi-page Query sweep, and Query/Delete error surfacing.
- handler: `TestHandleLifecycleEvent_AppUninstalledPurgesWorkspace` / `_TokensRevokedPurgesWorkspace` feed a **signature-valid** event_callback through `ServeHTTP` and assert `200` + that all three deletes ran (workspace_state via the recording provider; mappings/policies verified gone by reading them back through the store). `_AckEvenWhenPurgeFails` proves the `200` survives a failing purge. `TestHandleEvent_NonLifecycleEventNotPurged` proves an ordinary `app_mention` does not trigger a purge. `TestSlashCommandUninstallPurgesWorkspace` proves `/qurl uninstall` now removes the bot token + mappings + policies on top of the qURL key.
- **Two pre-existing uninstall tests updated.** `TestSlashCommandUninstallAllowsWorkspaceAdminOrOwner` and the former `TestSlashCommandUninstallRepeatReportsNotConnected` assumed the `workspace_mappings` row *survived* a prior uninstall, so a repeat / second-admin uninstall still reached `DeleteAPIKey`. That assumption no longer holds now that uninstall forgets the row. They were restructured to verify the same intent (the admin/owner gate; the "isn't currently connected" copy) against fresh workspaces / a direct partial-state setup, rather than by uninstalling twice in a row.

## Local verification

- `go test ./apps/slack/... ./shared/...` — pass
- Pinned lint `golangci-lint@v2.10.1 run ./apps/slack/... ./shared/...` — `0 issues`
- `pre-commit run --files <changed>` — all hooks pass
- `go vet ./apps/slack/... ./shared/...` — clean

(Two failing tests in `apps/cli/cmd` — `TestMissingAPIKey`, `TestNewClient_MissingAPIKey` — are a pre-existing local-environment artifact, unrelated to this change: they fail identically on a clean `origin/main` checkout, and this PR touches nothing in `apps/cli`.)

## Pairs with a manifest PR

This handler PR pairs with a **qurl-integrations-infra** manifest change that subscribes the qURL Slack app to the `app_uninstalled` and `tokens_revoked` events. Those events will not fire in prod until that manifest PR merges and deploys, and the sandbox app needs the two events added too. **Recommend merging this handler PR first** so that a handler always exists before the events start arriving (an event with no handler is wasted; a handler with no events is simply dormant).

### DIFF
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1757,6 +1757,11 @@ type workspaceKeyRevoker interface {
 // uninstall to local-only.
 var _ workspaceKeyRevoker = (*auth.DDBProvider)(nil)
 
+// Ensure the production provider keeps satisfying the workspace-row teardown
+// capability so a refactor that drops DeleteWorkspaceState surfaces here rather
+// than silently leaving the encrypted bot token behind on uninstall.
+var _ workspaceStateDeleter = (*auth.DDBProvider)(nil)
+
 func (h *Handler) deleteWorkspaceAPIKey(w http.ResponseWriter, teamID, userID string) {
 	// Reuse the sync admin-verb budget (1.2s): after the owner/admin gate, the
 	// optional upstream revoke plus the DeleteAPIKey write stay inside Slack's 3s
@@ -1805,6 +1810,15 @@ func (h *Handler) deleteWorkspaceAPIKey(w http.ResponseWriter, teamID, userID st
 			return
 		}
 	}
+	// DeleteAPIKey cleared only the qURL key columns. Forget the rest of the
+	// workspace too — the encrypted Slack bot token + data key, the
+	// workspace_mappings row, and every channel_policies row — so `/qurl
+	// uninstall` leaves nothing behind (the same teardown a Slack app_uninstalled
+	// event triggers). Best-effort and idempotent: the primary disconnect already
+	// succeeded, so a sweep failure is logged inside purgeWorkspace and does not
+	// change the success reply. Runs inside the same sync budget as the slash ack.
+	purgeLog := slog.With("surface", "uninstall", "team_id", teamID, "caller_user_id", userID)
+	h.purgeWorkspace(ctx, purgeLog, teamID)
 	slog.Info("/qurl uninstall: disconnected workspace Slack commands", "team_id", teamID, "caller_user_id", userID, "upstream_revoked", revoked)
 	if revoked {
 		respondSlack(w, revokedReply)
@@ -2026,6 +2040,13 @@ func (h *Handler) handleEvent(w http.ResponseWriter, body []byte) {
 	case env.Type == "url_verification":
 		respondJSON(w, http.StatusOK, map[string]string{"challenge": env.Challenge})
 		return
+	case env.Type == "event_callback" && isLifecycleEvent(env.Event.Type):
+		// App uninstall / token revoke. Routed here BEFORE handleAgentEvent
+		// because the cascade must run regardless of conversation-mode wiring
+		// (handleAgentEvent returns early when the agent is disabled). It only
+		// schedules the async purge; the 200 below is the ack Slack needs to stop
+		// retrying.
+		h.handleLifecycleEvent(&env)
 	case env.Type == "event_callback":
 		// Conversation mode. handleAgentEvent only schedules async work (or
 		// no-ops when disabled/filtered); we always ack 200 below so Slack
--- a/apps/slack/internal/handler.go
+++ b/apps/slack/internal/handler.go
@@ -1816,9 +1816,18 @@ func (h *Handler) deleteWorkspaceAPIKey(w http.ResponseWriter, teamID, userID st
 	// uninstall` leaves nothing behind (the same teardown a Slack app_uninstalled
 	// event triggers). Best-effort and idempotent: the primary disconnect already
 	// succeeded, so a sweep failure is logged inside purgeWorkspace and does not
-	// change the success reply. Runs inside the same sync budget as the slash ack.
+	// change the success reply. Run it on a tracked async goroutine off h.baseCtx
+	// (NOT this request's ctx, which `defer cancel()`s on return) so the extra
+	// DeleteItem/Query round-trips — which can be several on a workspace used in
+	// many channels — stay off the slash ack's tight sync budget. h.Go is
+	// wg-tracked so a graceful shutdown drains an in-flight purge.
+	purgeTeamID := teamID
 	purgeLog := slog.With("surface", "uninstall", "team_id", teamID, "caller_user_id", userID)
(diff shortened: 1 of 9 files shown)


Kevin-layerV:
Claude follow-up status (updated July 8, 2026):

- `qurl_agent_state`: no longer TTL-only. The PR now explicitly purges the agent-state partition during lifecycle teardown and `/qurl uninstall`; TTL remains only a backup cleanup path.
- Grid partitioning: agent conversation/dedupe state now uses Slack authorization metadata so Grid workspace installs key by `team_id` and org installs key by `enterprise_id`, matching lifecycle purge scope.
- Toolchain: bumped Go / Slack builder image to 1.26.5 so `govulncheck` clears GO-2026-5856.
- Latest local verification is green; CI and fresh Claude review are pending on the new push.

Kevin-layerV:
@claude /simplify

Kevin-layerV:
@claude /simplify

Kevin-layerV:
@claude /simplify

Kevin-layerV:
Follow-up on the latest Claude gate:

- Confirmed the paired infra PR layervai/qurl-integrations-infra#1133 now pre-stages a separate `SlackWorkspaceMappingsScan` statement granting `dynamodb:Scan` only on the Slack `workspace_mappings` table.
- Infra CI is green on `c98d197`, including the `qurl-bot-slack/terraform` assertion that the Scan grant is scoped to `workspace_mappings`; the rollout ledger also requires applying that scoped IAM grant before lifecycle events are enabled in any rate-limit-enabled environment.
- Updated this PR body to call out #1133, the scoped IAM ordering, and the new #932 timestamp-helper follow-up.
- Updated #930 so the deferred performance work explicitly includes bounding or replacing the TTL-backed rate-limit counter Scan path before broad rate-limit-enabled Enterprise Grid rollout.
- Latest `/simplify` on this PR completed with no changes needed.

PR #823, opened 2026-06-16
feat(discord): add /qurl detect — recipient attribution + % match, guild-scoped

justin-layerv:
## Summary

User-facing watermark attribution (the Discord-bot half): a `/qurl detect` command — upload an image, get back the **original recipient's Discord handle + a % match**, scoped to the requester's guild. Pairs with the connector `/api/detect` endpoint (counterpart PR pending in `layervai/qurl-integrations-infra`).

## Scope

- [x] `apps/discord/`

## Changes

- `connector.js`: `mintLinks` sends `guild_id`; new `detectWatermark()` POSTs the image to the connector `/api/detect` (`Authorization` + `X-Guild-Id`) per the locked contract.
- `ddb-store.js`: `recordQURLSendBatch`/`recordQURLSend` persist `guild_id` (sparse; read back via the `qurl_id-index` GSI `projection=ALL` — no Terraform change).
- `commands.js`: thread `guild_id` through all four send paths (`executeSendPipeline` + `handleAddRecipients`); new `handleQurlDetect` — guild-only, per-`(guild,user)` cooldown, SSRF + content-type + size validation, **defense-in-depth same-guild filter that fails closed**, ephemeral reply, audit every call **without** the recipient id; the `detect` subcommand + dispatch.
- `constants.js`: `AUDIT_EVENTS.QURL_DETECT`.

**Cross-tenant isolation** is connector-ENFORCED (the `/api/detect` guild scope); the bot's same-guild filter is defense-in-depth.

## Test Plan

- [x] `npm test` (2742 tests) + `npm run lint` pass locally
- [x] New tests added (connector contract, store guild_id, `handleQurlDetect` incl. the cross-guild guard, end-to-end guild_id threading)
- [ ] Manual testing — the e2e smoke validates at deploy; `/qurl detect` integrates once the connector `/api/detect` deploys

## Related Issues

Part of layervai/qurl-integrations-infra#1101 (connector counterpart PR pending in that repo).


justin-layerv:
Thanks — sharp review, especially flagging the GSI projection as the load-bearing cross-repo fact.

**Merge-blocker (GSI projection) — CONFIRMED resolved.** The infra `qurl_id-index` GSI is `projection_type = "ALL"` (`qurl-integrations-infra` `modules/qurl-bot-ddb/main.tf:206`), so the new non-key `guild_id` IS projected onto the index `findSendsByQurlId` queries — the fail-closed same-guild filter reads a real value, not `undefined`. (You were right that the existing `qurl.expired` webhook reading non-key attrs off the same GSI implied `ALL`; now verified explicitly.) A real mint→detect e2e round-trip integrates once the connector `/api/detect` counterpart lands.

Addressing the rest now:
- **Cooldown on "not configured"** — clearing it (honest config error → immediate retry, matching the handler’s own rule).
- **SSRF-probe audit** — emitting a `result:rejected` `QURL_DETECT` audit on that path so the strongest abuse signal isn’t audit-invisible; broadening the `constants.js` scope comment.
- **Confidence display** — omitting `confidence` from the user-facing reply (keeping it in the audit/logs); fixes the mixed 0–100 / 0–1 scales and the "confidence 0" edge.
- **Test gaps** — adding image-too-large, CDN-`!ok`, apiKey-missing (asserting cooldown cleared), and required-option-missing.

Leaving the two minor notes (`Limit:2` under the one-row invariant; the 10000/20000 magic-number pair mirroring `sendCooldowns`) as-is. Counterpart connector PR (the `/api/detect` endpoint) is in flight in the infra repo.

justin-layerv:
Thanks for re-verifying the GSI projection + the helper existence end-to-end. Addressing all three:

- **Cooldown policy** — reconciling to the stated rule: CDN-fetch-`!ok` and connector-error clear the cooldown (honest, the expired-link path even says "re-upload and try again"), EXCEPT a connector **429** keeps it (back off); SSRF stays keep. Adding tests for each branch.
- **Ambiguity guard** — `handleQurlDetect` will now treat `sameGuild.length > 1` as ambiguous (no attribution, ephemeral "couldn’t determine a single recipient", `result:ambiguous` audit), matching the `qurl.expired` webhook’s stance rather than silently taking `[0]`.
- **`match_pct:0` nit** — switching the suffix guard to a truthy/`>0` check.

Leaving the comment-density note as-is per your call. Counterpart connector `/api/detect` PR is still in flight in the infra repo.

justin-layerv:
Thanks — addressing all four:
1. Folding `detectCooldowns` into the existing 60s sweep (real parity with `sendCooldowns`).
2. Softening the comment so the connector (`X-Guild-Id` + 429) is named the authority and the in-process cooldown is defense-in-depth (not shard-shared).
3. Wrapping the CDN fetch in its own try/catch so a throw (timeout/DNS) reports a download failure (+ clears the cooldown) rather than the misleading "detection unavailable".
4. Checking `bytes.length` after the fetch (closes the TOCTOU) + a `typeof`-safe pre-fetch size check.

Leaving the comment-density nit as already decided. Round-4 push incoming.

justin-layerv:
Thanks — addressing 1-3:
1. Adding `allowedMentions: { parse: [] }` to the matched reply so the deanonymized recipient is never pinged (explicit guarantee).
2. Adding `QURL_DETECT_COOLDOWN_MS` (default = `QURL_SEND_COOLDOWN_MS`, no behavior change) so the oracle throttle is decoupled from the send cadence and tunable independently.
3. Extracting a shared `evictOldest(map, max)` so the two cooldown buckets stay structurally in parity.

4 (legacy forward-only/fail-closed) and 5 (`match_pct>0` suppression, raw value in the audit) are correct as-designed — no change. Comment-density nit left as decided. Round-5 push incoming.

justin-layerv:
Good catch — the sweep/gate window mismatch is a real bug my round-4 decoupling introduced (the time-sweep kept the *send* window while the gate moved to the *detect* window), and it defeats the decoupling for detect>send. Fixing:
1. **Bug** — sweeping `detectCooldowns` against `QURL_DETECT_COOLDOWN_MS` + correcting the comment + a detect=90s/send=30s regression test (entry at ~31s must still be on cooldown).
2. **Hardening** — unifying the no-mark and cross-guild-filtered (`sameGuild===0`) replies to byte-identical copy so the bot-side defense-in-depth can not reintroduce the cross-guild signal if the connector scope ever regresses; the ambiguous-row message stays distinct (same-guild, no leak).
3. **Audit** — emitting `QURL_DETECT result:'unconfigured'` on the no-API-key branch so "audit on every invocation" holds.

Item 4 (`match_pct>0`): confirmed the connector rounds (`hammingToMatchPct` uses `math.Round`), so the assumption holds. Comment-volume left as decided. The connector counterpart PR is now open: **layervai/qurl-integrations-infra#1157**.

justin-layerv:
Thanks — addressed all five (`0750e5da`).

**1. Permission gate (decision).** Deliberate: `/qurl detect` is open to **any guild member**, NOT admin- or original-sender-gated — matching the un-gated sibling `/qurl send`. The only access boundary the feature *requires* is cross-guild (an image marked in guild A must never attribute in guild B), and that is enforced two ways (connector `X-Guild-Id` scope + the bot-side same-guild fail-closed filter). The user spec asked for exactly that boundary ("do not allow cross-tenant detections") and nothing narrower. I documented this as an explicit `ACCESS MODEL` note in the handler header so the next reader doesn't re-raise it; within-guild gating stays a reversible ~3-line product choice if it's ever wanted. Flagging it to the product owner in my summary so they can override.

**2. Audit logs rounded, comment claims raw (nit).** Fixed by the reviewer's first option — the matched audit now logs the **raw** `result.match_pct` (not the display-rounded `matchPct`), so the "keeps the raw value for forensics" comment is true and forensics gets the connector's exact value. Added a test: a fractional pct (92.7) displays rounded (`93% match`) but audits raw (`92.7`).

**3. "Audit on EVERY invocation" misses the failure paths (low).** Fixed both ways you suggested: (a) the connector **429** now audits as `result: 'rate_limited'` — it's an abuse signal that KEEPS the cooldown, so it mirrors the SSRF-probe audit and the throttled-spam pattern is now queryable; (b) reworded the header so it no longer over-claims "every invocation" — it now says it audits the attribution outcomes + abuse signals and that honest operational failures (CDN download, connector 5xx/network) log at warn/error, not audit. Tests: 429 audits `rate_limited`, 5xx does not.

**4. Pre-deploy sends silently un-attributable (info).** Confirmed expected UX — rows written before this change carry no `guild_id`, the same-guild filter drops them, and detect returns "no match" (fails closed, no leak). This is the forward-only behavior documented in the rollout-ledger entry; `guild_id` is backfillable later if attributing legacy images is ever wanted.

**5. Comment density (acknowledged).** Kept deliberately per the prior rounds — the *why* comments (cooldown policy, cross-guild rationale, the access model) are the load-bearing ones. Noted as non-blocking.

2763 tests pass; CI green.

justin-layerv:
Thanks — all three addressed (`b0c34897`).

**1. Unconditional registration vs pending connector (deciding deliberately).** Decided: **not** adding a bot-side flag. The feature is **already gated where it matters — on the connector side**: `/api/detect` returns `503 "attribution not available in this mode"` whenever the detector is unwired, and the whole connector watermark stack sits behind `watermark_service_enabled` / `WATERMARK_METASEAL_ENABLED` (sandbox-first, prod-off). So until the connector activates, the command is dark *by the existing activation gate*, and `handleQurlDetect` degrades gracefully to "unavailable right now" (pinned by `TestDetect … 5xx ⇒ graceful ephemeral`). A `DETECT_COMMAND_ENABLED` flag would be a **redundant second gate** on the same activation. The `MAP_COMMAND_ENABLED` analogy doesn't carry: `map` needs the flag because it has a **bot-side required secret** (`GOOGLE_MAPS_API_KEY`) with boot-validation; `detect` has no bot-side required config, and (as you noted) the e2e smoke-test name-set coupling means a flag wouldn't be a light change. Rollout sequences **connector-first** so the interim window is zero; the cross-repo coordination + graceful-degradation is documented in the #1157 rollout ledger. Net: deliberate no-gate, not an oversight.

**2. `AUDIT_EVENTS.QURL_DETECT` doc overstates coverage (low).** Fixed. Reworded the constants.js doc from "emitted once per invocation / records every query" to "audits the attribution **outcomes + abuse signals**," and explicitly listed what does NOT audit (CDN download, connector 5xx/network; and the pre-connector input rejects — cooldown, missing/non-image/oversize). Added the missing **`rate_limited`** outcome to the enumeration (the 429 branch emits it). This is the same drift I fixed in the commands.js handler header last round — thanks for catching the second copy.

**3. `evictOldest` thrash above the soft threshold (nit, pre-existing).** Agreed it's pre-existing (the verbatim `setCooldown` logic, now shared) and out of this change's altitude, so not touching it in-PR per your "no action needed." Filed as a tracked low-priority follow-up: layervai/qurl-integrations#824 (`priority: low`, `pr-polish-followup`) with the fix sketch (mirror the connector's `reclaimLocked` amortized eviction) + an acceptance test.

2763 tests pass; CI green.

justin-layerv:
Thanks — addressed (`49b477bb`); one item is now with the product manager.

**1. Access model — open-to-any-member oracle.** You're right to keep pressing this, and I agree it's a product call, not an eng default. I've escalated it to the **product manager for explicit sign-off** with the three options framed (any member / admins-mods only / original-sender only) and their trade-offs. The code stays open-to-any-member in the interim (the spec'd + tested default), and if product chooses admin- or sender-gating I'll land it before merge — a `requireAdmin` gate is ~10 lines + a flag-off test, and the in-code `ACCESS MODEL` comment already flags it as the reversible decision it is. **Holding merge on this sign-off.**

**2. Deploy ordering — command ships before its backend.** Sequenced **connector-first**, and there's a structural backstop beyond ordering: the connector's `/api/detect` returns `503 "attribution not available in this mode"` until the watermark stack is activated (`watermark_service_enabled` / `WATERMARK_METASEAL_ENABLED`, sandbox-first/prod-off), so the command is dark *by the existing activation gate* and `handleQurlDetect` degrades to "unavailable right now" (pinned by the 5xx-graceful test) in any interim window. The cross-repo coordination + this ordering are documented in the connector PR's rollout ledger (#1157).

**3. Comment density.** Acknowledged — deliberate, and as you note the prose is carrying the non-obvious security reasoning (the fail-closed filter, the byte-identical-copy invariant, the access model). Left per the prior rounds; not blocking.

**Minor — `findSendsByQurlId` `Limit: 2` correctness.** Added the one-line note you asked for at the query: the cap is safe for detect's same-guild filter ONLY because a `qurl_id` belongs to a single guild (one mint → one recipient → one guild), so every row it can return is same-guild and the cap can't truncate the caller's row; flagged to revisit if that one-guild-per-`qurl_id` invariant is ever relaxed.

**Minor — transient-failure cooldown clear / no 5xx back-pressure.** Intentional per the "honest errors clear" rule; the connector's per-guild limiter is the authoritative back-pressure on a 5xx storm (the in-process cooldown is best-effort and explicitly *not* the backstop). Noted in the catch-block rationale.

Also folded in a `/simplify` pass (clean): collapsed the single-use `SEND_/DETECT_COOLDOWNS_MAX` aliases to the shared `COOLDOWNS_MAX`. 2763 tests pass; CI green.

justin-layerv:
Thanks — this is a terminal pass for the bot side; addressing each:

**1. Within-guild open access.** Agreed it must be a conscious call — and it now is: escalated to the **product manager** for explicit sign-off (4th reviewer flag = exactly the signal to stop defaulting). The code stays open-to-any-member (the spec'd + tested default) pending their decision; a `requireAdmin` (or original-sender) gate is ~10 lines + a flag-off test if they choose to restrict. **#823 merge is held on that sign-off.**

**2. Realized-buffer guard runs after `arrayBuffer()`.** Acknowledged — it's not a true streaming ceiling; a lying Content-Length buys one full buffer before the post-fetch `bytes.length` check rejects. Low-risk for the reasons you note (the SSRF gate pins the source to the Discord CDN, and it matches the existing `/qurl send` path verbatim — consistency over a bespoke streaming cap here). Left as-is.

**3. Legacy rows (pre-#1101) lack `guild_id` → always "no match".** Confirmed acceptable: this is the intended forward-only, fail-closed behavior — an un-attributable image returns "no match" (never a wrong/cross-guild attribution). `guild_id` is backfillable later (`wm_id → qurl_id → bot mapping`) if attributing in-flight legacy sends ever matters; it's a future option, not a regression. Documented in the connector rollout ledger.

**4. Hard dependency on `qurl_id-index` `projection=ALL`.** Good catch — added a guard, but at the **change site** rather than the ledger (a re-projection is a permanent footgun; a delete-when-done rollout entry is the wrong medium). The GSI lives in the infra repo, so the guard went there: `modules/qurl-bot-ddb/main.tf` now carries a `DO NOT NARROW` note at the `qurl_id-index` definition explaining that `/qurl detect` reads `guild_id` straight off this projection and a re-projection to a subset would silently fail every attribution closed. Landed in the connector PR (layervai/qurl-integrations-infra#1157, `d12768b`).

**5. Comment density.** Acknowledged and deliberate — the security-rationale blocks (fail-closed filter, byte-identical-copy invariant, the access model) are the load-bearing ones. A descriptive-comment trim is reasonable as a later pass; not doing it now to avoid churning a converged, heavily-reviewed handler.

**Nits** — 👍 thanks. 2763 tests pass; CI green. Modulo the PM's access call, the bot side is merge-ready.

justin-layerv:
Thanks — all addressed (`b68d94db`).

**1. Visible-but-failing command window → GATED.** You (and the prior rounds) were right, and a topology fact settled it: `/api/detect` 503/404s until the watermark detect stack is ACTIVATED — a separate gated step AFTER the connector deploys (the metaseal cutover is its own future PR), so the window is plausibly weeks, not hours. So I gated `/qurl detect` behind `DETECT_COMMAND_ENABLED` (default **OFF**), mirroring `/qurl map`:
- registration gate (the builder IIFE) + a dispatch guard (stale-client → clean `QURL_DETECT_DISABLED_REPLY`, enforced even if the backend is live but the flag is off, so a stale detect can't reveal a recipient).
- the e2e smoke test is now flag-aware (detect conditional on `DETECT_COMMAND_ENABLED`) — kept LENIENT (default-off) so it's correct against a not-yet-activated deploy without a simultaneous infra CI change; the strict resolution + tfvar step is activation-time wiring, captured in the #1157 rollout ledger's new flip step.
- tests: flag-on suite sets the flag true; flag-off suite sets false + a dispatcher-disabled stale-client test. No `missingDetectCommandKeys` boot machinery — detect has no required bot-side secret (unlike map's `GOOGLE_MAPS_API_KEY`), so the gate stays light.

Default-off means #823 is **inert on merge** (no PR rot); the flip is an owned activation step (ledger).

**2. Standing is bot-only; connector enforces only guild-scope.** Exactly right, and now documented in the ACCESS MODEL: the standing gate is a bot-side UX/governance layer, NOT a server-enforced authz boundary. A holder of the guild's qURL API key (admin-tier) could call `/api/detect` directly and bypass it (guild-scope still holds — never cross-tenant). Closing that gap is a connector per-route auth factor, tracked in layervai/qurl-integrations-infra#1170 (deferred there as a connector-wide auth decision, behind the SG gate).

**3. Content-Type into the connector header.** Acknowledged — validated to `image/*`, Discord-supplied, and `fetch` rejects malformed header values; low-risk, noted in the data path.

**Nits:** the `evictOldest` consolidation is behavior-preserving (thanks for confirming no drift). On comment volume — fair; the byte-identical rationale is restated at each branch deliberately (each is a load-bearing security invariant a future editor must not break in isolation), but point taken; left as-is given the security sensitivity + house style.

2767 tests pass. Modulo the PM's confirmation on the two policy calls I surfaced (permission set + byte-identical UX), the bot side is merge-ready.

justin-layerv:
Thanks — both addressed (`7cbf0b7c`), neither blocking.

**1. Timing side-channel.** Good catch — the byte-identical reply closes the content channel but not timing. Documented as an explicit RESIDUAL in the ACCESS MODEL header: no-mark returns before the DDB lookup, no-standing after it, matched after a `users.fetch`, so latency weakly clusters fast/medium/slow. Accepted (not equalized with constant-time work) for the reasons you note — weak oracle, blunted by ephemeral replies + network jitter + the per-(guild,user) cooldown + the connector throttle, and the connector guild scope stays the authoritative boundary.

**2. Sub-0.5% match_pct.** Clarified: the `> 0` test is on the ROUNDED value, which is safe because the connector FLOORS a detected `match_pct` at 75 (Hamming) / 100 (exact) — so a value that rounds to 0 is genuinely malformed, never a real sub-1% match dropped. (You're right the original comment conflated raw vs rounded; the connector floor is why it holds.)

**Minor:** comment density — acknowledged, deliberate. `detectWatermark` empty-buffer guard — moot (the handler only calls it with a non-empty fetched buffer), agreed.

CI green, 2767 tests pass. With #1157 merged, this is the paired bot half — **merge-ready** (gated default-OFF, inert until the connector's `/api/detect` activation, which is itself blocked on layervai/qurl-integrations-infra#1170).

PR #826, opened 2026-06-17
fix(discord): url-migration cleanup in webhook-registrar (prevent orphan subs on base_url rename)

justin-layerv:
## Summary

When the bot's `base_url` is renamed (host change, same `/webhooks/qurl` path), the webhook registrar's canonical-URL matcher correctly finds zero subscriptions at the new URL, so it creates a fresh one. Without cleanup, the OLD subscription stays alive in qurl-service forever:

- The old URL's DNS typically still resolves (same ALB, different host header is acceptable).
- qurl-service keeps trying to deliver `qurl.accessed` / `qurl.expired` events to the orphan.
- Every delivery fails sig verification at the bot — the bot now reads the NEW sub's secret from SSM, the orphan's secret has not been rotated since its creation.
- Failure counter on the orphan climbs forever; clutter accumulates in the qurl-service subscription list.

This PR adds a cross-host orphan sweep that runs BEFORE the find/reuse/rotate/create branching (so the retry-on-next-boot guarantee holds even after a successful create on the previous boot). The sweep co-collects orphan candidates in `findExistingSubscriptions`'s existing cursor walk (single full-scan, not two), and filters to ones whose:

1. URL pathname matches the new bridge URL pathname (same route).
2. URL host DIFFERS from the new bridge URL host (cross-host = rename candidate).
3. Description matches the bot's stable description prefix exactly or with ` (` as the next character — i.e. the prefix is the full stable segment, not just any character-prefix. The prefix is a shared `DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX` constant consumed by both the bot path (`guild-webhook-link.js`) and the Lambda input-validator (which rejects a terraform-set `description` that doesn't begin with it).
4. Compound liveness gate: `last_delivery_success === false` AND `failure_count >= 30`. The floor tolerates not just transient blips but also short sustained outages (~10-30 minutes of consecutive failed deliveries) on an otherwise-healthy cross-host sibling; the strict `=== false` treats null/undefined as presumed-alive. Uses `Number.isFinite` (not `typeof === 'number'`) so a `NaN` `failure_count` cannot slip through.

Each match is DELETEd best-effort. A per-orphan failure (5xx, 4xx, gateway timeout) logs `URL-migration orphan delete failed` at error level and continues to the next orphan + to the create path. Blocking the create on a stale-orphan DELETE failure would leave the bot UN-registered AND still-orphaned — strictly worse than the orphan-only state we started in. Hoisting the cleanup above the branch makes the retry-on-next-boot guarantee actually true: every boot collects orphans and DELETEs them, regardless of whether the bot then takes existing/reuse/rotate/create.

## Why description-prefix matching matters

Under today's bot-API-key-shared model (see [`project_qurl_api_key_blast_radius`](https://github.com/layervai/qurl-integrations/blob/main/CLAUDE.md)), sibling services (e.g. `qurl-s3-connector`'s `resource.closed` subscription) live under the SAME `owner_id` as the bot's view-counter sub because they all mint via the bot's `QURL_API_KEY`. Pathname + host filtering alone would still match them. The description-prefix + boundary anchor is the load-bearing safety: connector subs have `description: "qurl-s3-connector resource.closed subscription"`, which does NOT start with `Discord bot view counter (` (or equal `Discord bot view counter`), so they are skipped.

Both callers of `ensureWebhookSubscription` (the central registrar Lambda + per-guild link from `guild-webhook-link.js`) interpolate the same `DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX` constant. The Lambda also validates that its terraform-set `description` input begins with that prefix at input-validation time, so a terraform drift fails the deploy LOUDLY instead of silently turning the sweep into a no-op for the central path.

## Active-active multi-region: hard kill-switch + #827 for the durable fix

The compound liveness gate cannot distinguish a true URL-migration orphan from a sibling region in a sustained outage — both produce `last_delivery_success=false` + climbing `failure_count`. Today's deployment is single-host, so this is hypothetical. Issue #827 tracks the durable fix (host allowlist / explicit old-host marker / region-tagged description) that must land BEFORE any active-active multi-region rollout under a shared `QURL_API_KEY`.

Defense-in-depth: a hard kill-switch (`QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP` env var) disables the sweep on BOTH call paths (Lambda wrapper + bot replicas via `guild-webhook-link.js`). Flipping the env var to any non-empty value before a multi-region rollout disables sweep classification + DELETE on every replica + the Lambda — a single config flip, no code change required. Disabled state logs `URL-migration orphan sweep DISABLED via env override`.

## Invocation-cadence note

The first-boot-after-rename case: the OLD sub's `last_delivery_success` may still be `true` for a brief window after the rename (the last pre-rename delivery succeeded). It flips to false only after the first post-rename delivery fails.

Two call paths with different cadences:
- **Central Lambda**: runs on `terraform apply` (deploy-cadence). The orphan accrues failures continuously between rename and next deploy — the motivating incident hit 1475 failures, well above the threshold of 30, so it would be swept by the next post-rename deploy.
- **Bot replicas via `guild-webhook-link.js`**: runs on EVERY guild-link event (OAuth callback, `/qurl setup` paste, backfill script) — i.e. many invocations per day. On the bot path the sweep retries far more frequently; the deploy-cadence concern doesn't apply.

If qurl-service ever stops attempting deliveries to a dead host (both `last_delivery_success` and `failure_count` would stall), this cadence stops being adequate — file an issue for a scheduled invocation then.

## Related operational context (this is distinct from #1148)

This PR is NOT a fix for #1148 (different bug). It addresses a separate symptom uncovered during the sandbox base-url rename `discord.layerv.xyz` -> `discord.connector.layerv.xyz`: an orphan subscription with description `Discord bot — view counter (sandbox, owner-scoped)` and 1475 failed deliveries was alive alongside the live sub at the new URL. That historical orphan has a different em-dash + wording in its description, so this code's `Discord bot view counter` prefix matcher would NOT have swept it (it has already been manually deleted). This PR prevents recurrence for orphans minted by the CURRENT description shape on future renames (region migrations, env-name flips, etc.).

## Test plan

- [x] Cross-host orphan with matching description + dead + many failures: DELETEd.
- [x] Cross-host sub with DIFFERENT description (sibling service): NOT deleted.
- [x] Cross-host sub with last_delivery_success=true (healthy sibling): NOT deleted.
- [x] Cross-host sub with last_delivery_success=null/undefined/missing (presumed-alive): NOT deleted.
- [x] Cross-host sub with last_delivery_success=false but failure_count under threshold (transient blip): NOT deleted.
- [x] Cross-host sub with failure_count missing/non-numeric (schema drift fail-closed): NOT deleted.
- [x] Cross-host sub with failure_count=NaN/Infinity/-Infinity (predicate-level fail-closed): NOT deleted.
- [x] Boundary anchor: `Discord bot view counterX (...)` over-match: NOT deleted.
- [x] Existing sub at the new URL alongside a stale orphan: orphan is still swept (retry-on-next-boot pinned).
- [x] DELETE 5xx on an orphan: logs error, continues to create; next invocation retries.
- [x] DELETE 404: treated as success (concurrent cleanup).
- [x] Same-host different-path subs: left alone (path filter).
- [x] Multi-orphan sweep (sequential renames): all DELETEd.
- [x] Per-orphan failure isolation: one DELETE 5xx does not shadow the next orphan.
- [x] Pagination cursor walked for the sweep list.
- [x] Empty description short-circuits to no-op (prevents `startsWith('')` matching everything).
- [x] Malformed URL on a sub: gracefully skipped (urlHost/urlPathname return null).
- [x] Near-miss observability log: emitted when candidates fail the liveness gate; NOT emitted at zero count.
- [x] Tri-state classifier pinned: `MATCH` / `NEAR_MISS_LIVENESS` / `NO_MATCH`.
- [x] Log shape pinned: `URL-migration orphan deleted` carries `old_url`, `webhook_id`, `description`, `failure_count`, `last_delivery_success`.
- [x] Lambda input rejects description shapes that don't start with `Discord bot view counter (` (5 drift cases).
- [x] Lambda accepts bare-prefix-no-paren (boundary parity with matcher).
- [x] Kill-switch covers BOTH call paths: Lambda env var disables sweep + warn line emitted; `linkGuildWebhookSubscription` honors the same env var.
- [x] All 2815 discord tests pass (101 in the registrar suite, multiple new Lambda + bot path tests).
- [x] `npm run lint` clean.

After merge:
- [ ] Next sandbox deploy: verify no new orphan accrues when the registrar Lambda re-runs.
- [ ] **Recurring**, every rename event (any env): confirm CloudWatch has `URL-migration orphan deleted` log line(s) for the previous host AND verify `failure_count` on the swept orphan(s) was past 30. This sanity-checks the deploy-cadence assumption + the `failure_count` increment semantics (if `failure_count` ever stops climbing for a dead host, the orphan would silently survive — file a follow-up issue then).
- [ ] **CloudWatch alarms (infra-side)**: land the five alarm definitions tracked in **qurl-integrations-infra#1205** so an unexpected sweep is loud, not silent. Alarms cover orphan-deleted spikes, delete-failure spikes, near-miss surges (schema drift signal), and kill-switch flips. Ideally lands within the same deploy window as this PR; the log lines exist and are runbook-grep-able from day one regardless.
- [ ] Before any active-active multi-region rollout under a shared `QURL_API_KEY`: set `QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP=1` on both the Lambda env and the bot tasks BEFORE deploying the second region. Track #827 for the durable replacement (region-tagged description / host allowlist). The destructive default-ON behavior IS bounded by per-env `QURL_API_KEY` owner-isolation — anything that consolidates keys across envs (cross-mount SSM, future shared-key decision) MUST flip the switch first.

### DIFF
--- a/apps/discord/lambda/webhook-registrar/index.js
+++ b/apps/discord/lambda/webhook-registrar/index.js
@@ -93,6 +93,7 @@ const {
   ensureWebhookSubscription,
   buildSsmPersistSecret,
   DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX,
+  isTruthyEnvFlag,
 } = require('../../src/qurl-webhook-registrar');
 
 // SSM client at module scope — Lambda reuses the same execution
@@ -245,9 +246,13 @@ exports.handler = async (event, context) => {
   // single-host deploy; flip to OFF before any active-active multi-
   // region rollout under a shared `QURL_API_KEY` (#827). Env-var
   // override (no code change required at flip time):
-  //   QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP=1
-  // Any non-empty value disables; empty / unset leaves it enabled.
-  const urlMigrationSweepEnabled = !process.env.QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP;
+  //   QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP=1   (or true/yes/on)
+  // Only those truthy literals disable; `=0`, `=false`, `=no`, `=off`,
+  // and empty / unset leave the sweep ENABLED. Without this
+  // normalization, an operator typing the intuitive `=0` would
+  // accidentally DISABLE the sweep (any non-empty string is truthy
+  // in JS) — the failure mode is benign (no deletions) but surprising.
+  const urlMigrationSweepEnabled = !isTruthyEnvFlag(process.env.QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP);
   if (!urlMigrationSweepEnabled) {
     console.warn(JSON.stringify({
       msg: 'webhook-registrar: URL-migration orphan sweep DISABLED via env override',
--- a/apps/discord/lambda/webhook-registrar/index.js
+++ b/apps/discord/lambda/webhook-registrar/index.js
@@ -92,6 +92,7 @@ const {
 const {
   ensureWebhookSubscription,
   buildSsmPersistSecret,
+  DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX,
 } = require('../../src/qurl-webhook-registrar');
 
 // SSM client at module scope — Lambda reuses the same execution
@@ -172,6 +173,21 @@ function validateAndNormalizeInput(event) {
       limit: DESCRIPTION_WARN_LEN,
     }));
   }
+  // Stable-prefix invariant: the URL-migration orphan sweep matches
+  // on the byte-identical pre-`(` segment of the description (see
+  // qurl-webhook-registrar.js::deriveDescriptionPrefix). If the
+  // terraform-set `description` ever drifts to NOT start with this
+  // prefix, every future rename leaves an uncatchable orphan — fail
+  // fast at the boundary so the next deploy surfaces the drift
+  // instead of silently disabling the sweep.
+  const expectedPrefix = `${DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX} (`;
+  if (event.description !== DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX
+      && !event.description.startsWith(expectedPrefix)) {
+    throw new Error(
+      `webhook-registrar: description must start with "${expectedPrefix}" (orphan-sweep matcher invariant); `
+      + `got: ${event.description.slice(0, 80)}`,
+    );
+  }
   const normalized = { ...event, bridgeUrl: event.bridgeUrl.replace(/\/$/, '') };
   return normalized;
 }
@@ -225,12 +241,26 @@ exports.handler = async (event, context) => {
   // never reaches the bot and the receiver 503s every webhook. Fail
   // the Lambda loudly → Terraform apply fails → operator notices
   // instead of half-registered state shipping silently.
+  // URL-migration orphan sweep hard guard. Default ON for today's
+  // single-host deploy; flip to OFF before any active-active multi-
+  // region rollout under a shared `QURL_API_KEY` (#827). Env-var
+  // override (no code change required at flip time):
+  //   QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP=1
+  // Any non-empty value disables; empty / unset leaves it enabled.
+  const urlMigrationSweepEnabled = !process.env.QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP;
+  if (!urlMigrationSweepEnabled) {
+    console.warn(JSON.stringify({
+      msg: 'webhook-registrar: URL-migration orphan sweep DISABLED via env override',
+    }));
+  }
+
   const result = await ensureWebhookSubscription({
     apiEndpoint: input.apiEndpoint,
     apiKey,
     bridgeUrl: input.bridgeUrl,
     description: input.description,
     initialSecret,
+    urlMigrationSweepEnabled,
     // persistSecret intentionally omitted — handled below with strict
     // throw-on-failure semantics.
   });
--- a/apps/discord/src/guild-webhook-link.js
+++ b/apps/discord/src/guild-webhook-link.js
@@ -123,6 +123,14 @@ async function linkGuildWebhookSubscription({ guildId, apiKey, descriptionContex
       // and the central registrar's coherently. Don't inline the
       // literal here; the constant is the single source of truth.
       description: `${DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX} (guild=${guildId}${descriptionContext ? `, ${descriptionContext}` : ''})`,
+      // Honor the same env-var kill-switch as the Lambda wrapper so
+      // a multi-region rollout (#827) can disable the sweep on every
+      // call path (Lambda + bot replicas) with a single config flip.
+      // The bot HTTP fleet is exactly the active-active topology
+      // that would cannibalize healthy peers, so this path must
+      // respect the override too — Lambda-only coverage would be
+      // false safety.
+      urlMigrationSweepEnabled: !process.env.QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP,
     });
   } catch (err) {
     logger.warn('Per-guild webhook subscription registration failed', {
--- a/apps/discord/src/guild-webhook-link.js
+++ b/apps/discord/src/guild-webhook-link.js
@@ -12,6 +12,7 @@ const {
   ensureWebhookSubscription,
   deleteSubscription,
   DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX,
+  isTruthyEnvFlag,
 } = require('./qurl-webhook-registrar');
 const subs = require('./webhook-subscriptions');
 
@@ -130,7 +131,7 @@ async function linkGuildWebhookSubscription({ guildId, apiKey, descriptionContex
       // that would cannibalize healthy peers, so this path must
       // respect the override too — Lambda-only coverage would be
       // false safety.
-      urlMigrationSweepEnabled: !process.env.QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP,
+      urlMigrationSweepEnabled: !isTruthyEnvFlag(process.env.QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP),
     });
   } catch (err) {
     logger.warn('Per-guild webhook subscription registration failed', {
--- a/apps/discord/src/guild-webhook-link.js
+++ b/apps/discord/src/guild-webhook-link.js
@@ -8,7 +8,11 @@ const config = require('./config');
 const db = require('./store');
 const logger = require('./logger');
 const { AUDIT_EVENTS } = require('./constants');
-const { ensureWebhookSubscription, deleteSubscription } = require('./qurl-webhook-registrar');
+const {
+  ensureWebhookSubscription,
+  deleteSubscription,
+  DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX,
+} = require('./qurl-webhook-registrar');
 const subs = require('./webhook-subscriptions');
 
 // Frozen discriminator for the `reason` field on link results.
@@ -112,7 +116,13 @@ async function linkGuildWebhookSubscription({ guildId, apiKey, descriptionContex
       apiEndpoint: config.QURL_ENDPOINT,
       apiKey,
       bridgeUrl: bridgeUrl(),
-      description: `Discord bot view counter (guild=${guildId}${descriptionContext ? `, ${descriptionContext}` : ''})`,
+      // Stable prefix is the shared DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX
+      // constant — same string the central registrar's terraform-set
+      // description must start with — so the orphan-sweep matcher in
+      // qurl-webhook-registrar.js classifies both this caller's subs
+      // and the central registrar's coherently. Don't inline the
+      // literal here; the constant is the single source of truth.
+      description: `${DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX} (guild=${guildId}${descriptionContext ? `, ${descriptionContext}` : ''})`,
     });
   } catch (err) {
     logger.warn('Per-guild webhook subscription registration failed', {
--- a/apps/discord/src/qurl-webhook-registrar.js
+++ b/apps/discord/src/qurl-webhook-registrar.js
@@ -200,6 +200,145 @@ async function deleteSubscription({ apiEndpoint, apiKey, webhookId }) {
   }
 }
 
+// Derive the stable description prefix from the bot's current description.
+// `description` shape today: `Discord bot view counter (region=..., env=...)`
+// (central registrar) or `Discord bot view counter (guild=..., ...)`
+// (per-guild link). Everything in parens varies with env/region/guild;
+// everything BEFORE the first ` (` is the stable identity of "subs this
+// bot creates". Extract that prefix so the URL-migration orphan sweep
+// only matches the bot's own historical subs, never a sibling service's.
+//
+// Falls back to the whole string if no ` (` present, so a caller passing
+// a description without parens still gets a sensible match (string-exact).
+function deriveDescriptionPrefix(description) {
+  if (typeof description !== 'string' || description.length === 0) return '';
+  const idx = description.indexOf(' (');
+  return idx === -1 ? description : description.slice(0, idx);
+}
+
+// Pull the host from a URL string. Returns null on parse failure so the
+// caller can skip the row defensively (same fallback shape as canonicalUrl).
+function urlHost(u) {
+  try { return new URL(u).host.toLowerCase(); } catch { return null; }
+}
+
+// Pull the pathname from a URL string with trailing-slash normalization
+// (mirrors canonicalUrl semantics so `/webhooks/qurl` and `/webhooks/qurl/`
+// compare equal). Returns null on parse failure.
+function urlPathname(u) {
+  try {
+    const url = new URL(u);
+    let p = url.pathname;
+    if (p.length > 1 && p.endsWith('/')) p = p.slice(0, -1);
+    return p;
+  } catch { return null; }
+}
+
+// URL-migration orphan sweep.
+//
+// Symptom (sandbox, 2026-06): the bot's `base_url` was renamed from
+// `discord.layerv.xyz` → `discord.connector.layerv.xyz`. The registrar's
+// canonical-URL matcher correctly found ZERO subs at the new URL, so it
+// CREATED a fresh subscription. The OLD subscription stayed alive in
+// qurl-service — its secret unchanged, qurl-service kept trying to
+// deliver to the old host (which still resolved to the same ALB), every
+// delivery failed sig-verification at the bot (the bot now reads the
+// NEW sub's secret from SSM). Net: 1475 failed deliveries to a permanent
+// orphan + cluttered subscription list.
+//
+// Cleanup criteria (conservative — false-positive deletion of someone
+// else's sub is far worse than a missed orphan):
+//   1. URL pathname matches the bot's bridge URL pathname (same route)
+//   2. URL host DIFFERS from the bot's current bridge URL host
+//      (cross-host = candidate for rename)
+//   3. Description STARTS WITH the bot's stable description prefix
+//      (e.g. `Discord bot view counter`) — this is the safety net that
+//      keeps connector subs (description: `qurl-s3-connector ...`) out
+//      of the deletion set even though they share owner_id with the bot
+//      under today's bot-API-key-shared model.
+//
+// Each match is DELETEd best-effort: a 5xx on one orphan logs an error
+// and continues to the next + to the normal create path. Next invocation
+// retries the orphan delete idempotently. We deliberately do NOT propagate
+// the failure: blocking the create on a stale-orphan delete failure would
+// leave the bot UN-registered (no new sub) AND still-orphaned (old sub
+// undeleted) — strictly worse than the orphan-only state we started in.
+async function findUrlMigrationOrphans({ apiEndpoint, apiKey, bridgeUrl, descriptionPrefix }) {
+  if (!descriptionPrefix) return [];
+  const targetHost = urlHost(bridgeUrl);
+  const targetPath = urlPathname(bridgeUrl);
+  if (!targetHost || !targetPath) return [];
+  const orphans = [];
+  let cursor = '';
+  for (let i = 0; i < 50; i++) {
+    const qs = cursor ? `?cursor=${encodeURIComponent(cursor)}&limit=100` : '?limit=100';
+    const path = `/v1/webhooks${qs}`;
+    const resp = await callQurlService({ method: 'GET', path, apiEndpoint, apiKey });
+    const subs = Array.isArray(resp?.data) ? resp.data : [];
+    for (const s of subs) {
+      const subHost = urlHost(s.url);
+      const subPath = urlPathname(s.url);
+      const subDesc = typeof s.description === 'string' ? s.description : '';
+      if (subHost === null || subPath === null) continue;
+      if (subPath !== targetPath) continue;
+      if (subHost === targetHost) continue;
+      if (!subDesc.startsWith(descriptionPrefix)) continue;
+      orphans.push(s);
+    }
+    const next = resp && resp.meta && resp.meta.next_cursor;
+    if (!next) return orphans;
+    cursor = next;
+  }
+  // Same fail-loud policy as findExistingSubscriptions: a stuck cursor
+  // must not silently fall through. The caller catches and logs so the
+  // create-fresh path can still proceed even if the orphan sweep failed.
+  throw new Error('findUrlMigrationOrphans: pagination cap hit (50 pages, ~5000 subs); possible stuck cursor');
+}
+
+async function cleanupUrlMigrationOrphans({ apiEndpoint, apiKey, bridgeUrl, descriptionPrefix }) {
+  let orphans;
+  try {
+    orphans = await findUrlMigrationOrphans({ apiEndpoint, apiKey, bridgeUrl, descriptionPrefix });
+  } catch (err) {
+    // Log + continue: orphan sweep is observability-grade cleanup, not
+    // load-bearing for registration. A pagination cap or transient list
+    // failure shouldn't block create-fresh.
+    logger.error('URL-migration orphan sweep list failed (continuing with create)', {
+      error: err.message, status: err.status, op: err.op,
+    });
+    return;
+  }
+  if (orphans.length === 0) return;
+  // Sequential DELETEs (vs parallel like the dedupe path) — orphan counts
+  // are typically 0-1 per migration, and serial gives the cleanest
+  // per-orphan failure isolation in the logs. Each failure is caught,
+  // logged, and skipped so the rest of the sweep + the create path
+  // still proceed.
+  for (const orphan of orphans) {
+    try {
+      await deleteSubscription({ apiEndpoint, apiKey, webhookId: orphan.webhook_id });
+      logger.info('URL-migration orphan deleted', {
+        old_url: orphan.url,
+        webhook_id: orphan.webhook_id,
+        description: orphan.description,
+        failure_count: orphan.failure_count,
+        last_delivery_success: orphan.last_delivery_success,
+      });
+    } catch (err) {
+      // Don't propagate: blocking create on a stale-orphan delete
+      // failure leaves the bot UN-registered AND orphaned — worse than
+      // the starting state. Next invocation retries the delete.
+      logger.error('URL-migration orphan delete failed (continuing — next invocation retries)', {
+        old_url: orphan.url,
+        webhook_id: orphan.webhook_id,
+        error: err.message,
+        status: err.status,
+        op: err.op,
+      });
+    }
+  }
+}
+
 // Pick a deterministic survivor across replicas so two replicas racing
 // on dedupe converge on the same one without coordination. Oldest-by-
 // created_at is the natural choice (first-to-exist wins); fall back to
@@ -469,6 +608,19 @@ async function ensureWebhookSubscription(opts) {
     }
     logger.info('qURL webhook subscription reconciled (existing found, bootstrap rotate)', { webhookId, url: bridgeUrl });
   } else {
+    // No existing sub at the requested URL. Before creating a fresh one,
+    // sweep for URL-migration orphans: prior subs this bot registered at
+    // a DIFFERENT host (e.g. `base_url` rename) on the same `/webhooks/qurl`
+    // path. Without this, a rename leaves the old sub alive in
+    // qurl-service forever, retrying deliveries that the bot can no
+    // longer verify. See findUrlMigrationOrphans for the safety criteria
+    // (description-prefix match keeps sibling-service subs out of scope).
+    await cleanupUrlMigrationOrphans({
+      apiEndpoint,
+      apiKey,
+      bridgeUrl,
+      descriptionPrefix: deriveDescriptionPrefix(description),
+    });
     const created = await createSubscription({ apiEndpoint, apiKey, bridgeUrl, description });
     webhookId = created.webhook_id;
     secret = created.secret;
@@ -533,5 +685,8 @@ module.exports = {
     pickSurvivor,
     redactSecret,
     QurlServiceError,
+    deriveDescriptionPrefix,
+    findUrlMigrationOrphans,
+    cleanupUrlMigrationOrphans,
   },
 };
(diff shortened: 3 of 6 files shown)


justin-layerv:
## cr feedback addressed

Pushed `040601ff` addressing every cr finding from both rounds.

**Round-2 cr-bot:**

1. **(High) Retry-on-next-boot now actually true.** Hoisted `cleanupUrlMigrationOrphans` BEFORE the find/reuse/rotate/create branching. Previously the sweep only ran in the `else` (create-fresh) branch, so a 5xx-swallowed DELETE on boot 1 would never get retried on boot 2 (existing branch returns early). New test pins boot 2 with both the new sub AND the still-alive orphan re-attempting the DELETE.

2. **(Medium) Byte-stable-prefix constraint documented.** Added a LOAD-BEARING CONSTRAINT block in `deriveDescriptionPrefix` covering: (a) the historical em-dash orphan that this code would NOT catch (and why that's fine for the recurrence-prevention goal), (b) the rule that any future edit to the pre-`(` portion of the description must be paired with a one-off cleanup pass.

3. **(Low) Boundary-anchored description match.** Now require `subDesc === descriptionPrefix || subDesc.startsWith(descriptionPrefix + ' (')` so the prefix matches the full stable segment, not an arbitrary character-prefix. Test added for `Discord bot view counterX (...)` over-match.

4. **(Nit) `findExistingSubscriptions` doc updated.** Always returns `{matches, orphans}`; the no-filter case just has an empty `orphans`.

**Round-1 cr-bot:**

5. **(Important) Liveness gate on `last_delivery_success === false`.** Cross-host healthy active-active siblings (same `QURL_API_KEY`, different `BASE_URL`) would have self-cannibalized: A sweeps B's healthy sub, B sweeps A's, ping-pong on every cold boot. Strict-false gate asymmetrically protects healthy + not-yet-tested subs (null/undefined liveness = presumed alive) while still catching the unambiguously-dead orphan (`failure_count=1475`, `last_delivery_success=false` in the motivating incident). Three new tests pin the behavior.

6. **(Minor) Double full-list walk** — already addressed in `2b75bab7` by co-collecting orphans during the existing cursor walk.

7. **(Nit) `urlPathname` vs `canonicalUrl` query-string handling.** Added a comment noting the divergence (canonicalUrl preserves query as route-distinguishing; urlPathname drops it) and that the bridge URL is unparametrized today so the divergence is moot.

CI: 85 registrar tests + 2787 discord tests pass; lint clean.

justin-layerv:
## cr round 2 addressed (`94a5c90a`)

1. **(Medium) Transient-failure hole in liveness gate closed.** The cr-bot's exact scenario (single transient delivery failure flips `last_delivery_success` to false on an otherwise-healthy sibling → peer reboot DELETEs the live sub) now requires `failure_count >= 3` as well. Sets `URL_MIGRATION_ORPHAN_MIN_FAILURES = 3` so one transient failure cannot trigger deletion. The motivating orphan had 1475, well above. Also fails closed when `failure_count` is missing / non-numeric so qurl-service schema drift cannot accidentally lower the bar.

2. **(Low) `TODO(upstream-contract)` marker added** on the schema field names (`last_delivery_success`, `failure_count`) for `git grep` discoverability.

3. **(Low) Near-miss observability log.** Re-classified the orphan filter as a tri-state (`MATCH` | `NEAR_MISS_LIVENESS` | `NO_MATCH`). When a sub matched host+path+description but failed the liveness gate, count it; after the scan, log `URL-migration orphan sweep — liveness-gated near-misses` with the count. Operators can now distinguish "qurl-service field missing (sweep silently disabled)" from "sibling still alive (sweep correctly held back)" without inspecting subs directly. No log emitted at near-miss-count=0 to avoid steady-state noise.

4. **(Low) First-boot-after-rename caveat documented.** Old sub's `last_delivery_success` may still be `true` (last delivery succeeded just before the rename); it flips to false after the first post-rename delivery attempt fails sig-verification. So the immediate first post-rename boot may not sweep — caught on a subsequent boot. Acceptable because the hoisted sweep retries every boot.

5. **(Nit) `deriveDescriptionPrefix` parsing fragility documented.** Splits on the FIRST ` (`, so if the stable segment ever embeds its own parenthetical, the derived prefix silently shrinks.

Test gaps the cr-bot flagged also added:
- Transient failure (failure_count under threshold) → not deleted
- Missing / non-numeric failure_count → fail-closed
- Near-miss log emitted with count + url
- No near-miss log when zero candidates
- Malformed URL gracefully skipped
- `buildUrlMigrationOrphanFilter` returns null on unparseable `bridgeUrl`
- Tri-state classifier pinned: `MATCH` / `NEAR_MISS_LIVENESS` / `NO_MATCH`

97 registrar tests + 2799 discord tests pass; lint clean.

justin-layerv:
## cr round 3 addressed (`1bbe8b74`)

All three findings are doc-only — no behavior change.

1. **(Important) Softened the liveness-gate claim.** The cr-bot is right: the compound gate cannot distinguish "dead URL-migration orphan" from "sibling region in a sustained outage" — both produce `last_delivery_success === false` + climbing `failure_count`. Updated the in-code comment to call out that the gate only protects siblings whose LAST delivery succeeded (or hasn't happened yet), and that the durable fix for hypothetical active-active under a shared `QURL_API_KEY` is a non-signal-based discriminator (host allowlist, explicit old-host marker), NOT a higher `failure_count` floor. Test name updated to drop the over-strong "protects active-active siblings" framing.

2. **(Low) Failure-class symmetry documented.** Added a comment block to `cleanupUrlMigrationOrphans` noting that we intentionally do NOT alarm-tier-differentiate transient 5xx vs persistent 4xx. If the same orphan-delete error repeats, the repeating log line IS the signal (CloudWatch alarm pattern on `URL-migration orphan delete failed` is the runbook hook).

3. **(Low) Near-miss steady-state-noise edge case noted.** In the hypothetical active-active scenario a healthy cross-host sibling is a perpetual near-miss, so the log would emit every boot. Today's single-host deploy makes this a non-issue; flagged in the comment so a future multi-region rollout doesn't quietly re-introduce log noise (downgrade to debug or rate-limit then).

97 registrar tests + 2799 discord tests still pass; lint clean.

justin-layerv:
## cr round 4 addressed (`39e64193`)

1. **(Medium) Active-active cannibalization now has a tracking home.** Filed #827 with acceptance criteria, three durable-fix options (host allowlist, explicit old-host marker, region tag in description), and a clear repro for the multi-region rollout that would trip it. Added a `TODO(active-active-cannibalization, #827)` inline marker so `git grep` finds it before any multi-region rollout.

2. **(Low) Near-miss log field renamed.** `orphan_deletes` → `orphan_delete_attempts` per the cr's exact suggestion. A 5xx'd DELETE still counts toward `orphans.length`, so the old name would have misled an operator reading the log alongside fewer "URL-migration orphan deleted" confirmation lines. The per-orphan INFO/ERROR remain the authoritative record. Test updated.

3. **(Nit) Same-host query-string variant non-coverage confirmed intentional.** Added a paragraph to `urlPathname`'s comment noting that `<bridgeUrl>?x=1` is neither matched nor swept (canonicalUrl's query-preservation makes it a strict-eq miss; the same-host check rejects it from the orphan filter). Today's bridge URL is unparametrized so it's moot; flagged with the trigger to revisit if a query param ever enters the bridge URL.

97 registrar + 2799 discord tests still pass; lint clean.

justin-layerv:
## cr round 5 addressed (`a820e286`)

Two non-blocking polish items + acknowledging the informational note.

1. **(Informational, no fix needed)** Confirmed: the motivating em-dash orphan (`Discord bot — view counter (sandbox, owner-scoped)`) is not swept by this code because its prefix doesn't match the current `Discord bot view counter`. PR body already notes the orphan was manually deleted; this PR is recurrence-prevention only. The `deriveDescriptionPrefix` doc block already calls out the byte-stable-prefix constraint.

2. **(Optional) Hoisted `ORPHAN_FILTER_RESULTS`.** Moved above `findExistingSubscriptions` so the consumer reads below the definition (previously a forward reference; safe at runtime because the async function only executes after module-load TDZ, but cleaner ordering).

3. **(Nit) Fixed the "fails closed (skip)" wording.** A missing/non-numeric `failure_count` returns `NEAR_MISS_LIVENESS`, not a silent `NO_MATCH`. The row IS held back from deletion (fail-closed) AND counted toward `nearMissCount`, so the schema-drift case surfaces in the observability log instead of silently disappearing.

2799 tests still pass; lint clean.

justin-layerv:
## cr round 6 addressed (`604df71f`)

All three findings:

1. **(Medium → confirmed safe + documented) Per-guild host invariant.** Confirmed via `guild-webhook-link.js`: both per-guild link and central registrar use `\`\${config.BASE_URL}/webhooks/qurl\`` as their `bridgeUrl`, so per-guild subs always register at the SAME host as the central registrar. The host-equality check in criterion 2 keeps them out of the orphan filter; the cr's concern only materializes if per-guild registration ever moves to a different host. Added a `Per-guild safety` paragraph to the criterion-3 doc block pinning this invariant + the trigger that would dissolve it.

2. **(Low) Trim duplicated narrative.** Condensed ~50 lines that restated PR body / #827 content. Active-active discussion, first-boot caveat, failure-class symmetry, and steady-state-noise notes are now load-bearing-only; detailed write-ups live in PR body + #827, the inline comments point there.

3. **(Nit) Removed `urlHost`'s redundant `.toLowerCase()`.** The WHATWG URL parser already lowercases `URL.host`. Added a one-line comment noting why no fold is needed.

2799 tests + lint clean.

justin-layerv:
## cr round 7 addressed (`f65a5633`)

1. **(Medium) Per-guild interaction note corrected.** The previous "host-equality protects per-guild subs" framing was misleading. Actually: qurl-service owner-scoping + this file's dedupe path means per-guild and central calls converge on ONE sub per owner (same URL → first call CREATEs, every subsequent call REUSES whichever description landed first). On a rename, that one old-host sub is exactly the sweep's target — regardless of whether the first-to-register was the central Lambda or a guild link. Replaced the protective-invariant framing with the correct convergence note.

2. **(Minor)** Informational — orphan-not-cleaned acknowledged; already documented in `deriveDescriptionPrefix` doc-block.

3. **(Nit)** Continuing-trim noted — this round mostly preserved trimming from round 6; the remaining inline prose is load-bearing now.

2799 tests + lint clean.

justin-layerv:
## cr round 8 addressed (`8f424d11`)

1. **(Question) Invocation cadence ↔ first-boot-after-rename gap.** Confirmed in the PR body: the registrar Lambda runs on `terraform apply` (deploy-cadence-only). The motivating incident hit `failure_count=1475` between rename and next deploy — qurl-service continuously attempts deliveries, so by the time the next deploy fires the Lambda, the orphan is well past the floor of 3. Added an `## Invocation-cadence note` section to the PR body covering this, plus an after-merge checklist item to sanity-check the swept orphan's `failure_count` once a real rename happens. If qurl-service ever stops attempting deliveries to dead hosts, we'd file a scheduled-invocation issue.

2. **(Minor) Self-contradictory hoist comment trimmed.** Replaced with a clean "defined above its only consumer for reading order" + the truthiness-footgun rationale for the uniform-string choice.

3. **(Minor) `NO_MATCH: false` → `NO_MATCH: 'no-match'`.** Every result value is now a string. Strict `===` against the constants is the only valid check; a future `if (verdict)` can't accidentally treat NO_MATCH as falsy. Test assertions use `.toBe()` (strict equality), so they survive the change without modification.

4. **(Nit) Comment density.** Acknowledged. Rounds 6-7 trimmed a meaningful amount; the remaining inline prose is load-bearing (rationale for the conservative filter stack, the safety invariants, the limitations the gate doesn't cover).

2799 tests + lint clean.

justin-layerv:
## cr round 9 addressed (`e888cd89`)

1. **(Low → real bug fixed) `failure_count: NaN` fail-open closed.** The cr-bot spotted that `typeof s.failure_count !== 'number' || s.failure_count < 3` would let `NaN` slip through (typeof 'number' AND `NaN < 3` is false) → classify as MATCH → DELETE. Same problem for `Infinity`. Switched to `Number.isFinite(s.failure_count)` which rejects NaN, +/-Infinity, and non-numbers uniformly, matching the stated fail-closed invariant.

   In practice `JSON.parse` can't produce these values (no JSON literal), so qurl-service responses can't deliver them — but the predicate-level guarantee should be wire-independent (a future non-JSON SDK or in-process injection could otherwise hit it).

   Tests:
   - E2E: `failure_count: -1` → not deleted (the only wire-reachable shape that can slip past `< 3`).
   - Unit on `buildUrlMigrationOrphanFilter` directly: NaN/Infinity/-Infinity → `NEAR_MISS_LIVENESS`, not MATCH. Pins the predicate-level invariant wire-independently.

2. **(Info, already addressed)** Invocation cadence + first-boot caveat — PR body added an "Invocation-cadence note" section and an after-merge checkbox to sanity-check the swept orphan's `failure_count`.

3. **(Nit)** Comment density acknowledged; rounds 6-7 trimmed substantially. The remaining inline prose is the load-bearing safety reasoning the cr-bot has called "valuable" in every round.

4. **(Nit)** Sequential vs parallel DELETE asymmetry confirmed intentional (0-1 orphans expected; sequential gives clean per-orphan log isolation).

101 registrar tests + 2803 discord tests pass; lint clean.

justin-layerv:
## cr round 10 addressed (`35410ae2`)

Both the high-priority items:

1. **(Highest priority) Description-prefix drift between callers closed.** The cr-bot caught a real divergence risk: the per-guild link hardcoded `Discord bot view counter` in one place, while the central Lambda's description came from terraform (out-of-repo) in another. A one-character drift on either would silently disable the sweep.

   Three fixes for symmetry:
   - Extracted `DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX` as a single source of truth, exported from `qurl-webhook-registrar.js`.
   - `guild-webhook-link.js` now interpolates the constant instead of the literal.
   - `lambda/webhook-registrar/index.js` validates `event.description` starts with `<prefix> (` (or exactly equals the prefix) at input-validation time. A terraform drift now fails the deploy LOUDLY at the Lambda boundary with a clear error message instead of silently turning the sweep into a no-op.

   Tests: 5 Lambda validation cases reject drift shapes (capital B, dash variant, sibling description, no-boundary suffix, etc.). 1 boundary case accepts exact-prefix-no-paren.

2. **(Hard guard for active-active cannibalization)** Added `urlMigrationSweepEnabled` opt on `ensureWebhookSubscription` (defaults to `true`). Lambda wrapper reads `QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP` env var so a multi-region rollout can flip it WITHOUT a code change — exactly the kind of defense-in-depth the destructive-op warrants. Disabled state logs a clear `URL-migration orphan sweep DISABLED via env override` warn so operators see the gate is off.

   Tests: with opt=false, a row that would otherwise be a confirmed orphan is NOT deleted; default sweep still works; env-var path tested end-to-end through the Lambda handler.

3. **(Nit)** Continuing comment-density nudge acknowledged; the remaining inline prose is load-bearing safety reasoning. Will not trim further.

2812 tests + lint clean.

justin-layerv:
## cr round 11 addressed (`f5ab6611`)

**(High) Kill-switch now covers both call paths.** The cr-bot spotted that the previous round's `QURL_WEBHOOK_REGISTRAR_DISABLE_URL_MIGRATION_SWEEP` env var was only honored by the Lambda — bot replicas (`linkGuildWebhookSubscription`) defaulted to `true` regardless. Since the bot HTTP fleet IS the active-active multi-region target, Lambda-only coverage was false safety: flipping the env var would let bot replicas keep cannibalizing healthy peers.

Fix: `linkGuildWebhookSubscription` now reads the same env var and forwards `urlMigrationSweepEnabled` to `ensureWebhookSubscription`. A single config flip disables the sweep on EVERY call path — Lambda + every bot replica — with no code change required.

Tests pin the contract:
- env var unset → `urlMigrationSweepEnabled=true` (default)
- env var set → `urlMigrationSweepEnabled=false` (kill-switch honored on bot path)
- Description interpolation uses the shared `DISCORD_BOT_VIEW_COUNTER_DESCRIPTION_PREFIX` constant (drift safety test catches a future literal-vs-constant divergence)
- Test-mock exposes the real prefix constant so the description template interpolates correctly

PR body's Invocation-cadence section updated to reflect both paths (Lambda deploy-cadence vs bot per-guild-link cadence). Added an after-merge checklist item for the multi-region rollout procedure (flip env var on Lambda + bot tasks BEFORE deploying the second region).

**(Secondary / minor)** Operational note about the bot-path's more-frequent cadence already added to the PR body in the same edit.

2815 tests + lint clean.

justin-layerv:
## cr round 12 addressed (`9c2609a5`)

1. **(Highest) Cross-environment cannibalization pinned.** The cr-bot correctly noted that the description prefix is identical across envs (`Discord bot view counter`), so the ONLY structural separator between this sweep and another env's live sub is owner_id scope. Today's deployment uses per-env SSM `QURL_API_KEY` parameters (distinct keys → distinct owner_ids), so the invariant holds by configuration. Added a LOAD-BEARING ENVIRONMENT-ISOLATION ASSUMPTION block to the orphan-filter doc pinning this explicitly: any change that lets two envs share an owner_id (key migration, cross-mount, future shared-key decision) MUST flip the kill-switch BEFORE landing, or the sweep must gain an explicit env-discriminator first (#827 covers the durable fix).

   The cr-bot also asked "do any two environments share an `owner_id`?" — answer is no (per-env keys); the assumption block now documents that.

2. **(Low) 404 vs deleted log distinction.** `deleteSubscription` now returns a `DELETE_OUTCOMES` enum (`'deleted'` vs `'already-absent'`); `cleanupUrlMigrationOrphans` emits `URL-migration orphan already absent (concurrent cleanup)` on 404 instead of false-attributing the DELETE to itself. The dedupe path doesn't care about the distinction (callers ignore the return value, signature still no-throw on 404). Test updated to pin: on 404, the `URL-migration orphan deleted` line is NOT emitted.

3. **(Informational)** Cadence × failure-count-floor interaction acknowledged. The first post-rename deploy may classify as NEAR_MISS if `failure_count < 3` and only sweep on a subsequent deploy. Already documented; behavior is correct.

2815 tests + lint clean.

justin-layerv:
## cr round 13 addressed (`6905546a`)

1. **(Cross-repo coordination) Infra side verified.** Checked `/Users/posey/code/layerv/qurl-integrations-infra/qurl-bot-discord/terraform/webhook_registrar_lambda.tf` — current `webhook_registrar_description = "Discord bot view counter (region=${var.aws_region}, env=${var.environment})"`. Starts with `Discord bot view counter (`, conforms to the new Lambda input-validator. **No infra-side change needed**; this PR is safe to merge standalone.

2. **(Robustness nit) Kill-switch env-var normalized.** The cr-bot's exact concern: `!process.env.VAR` treats any non-empty string as truthy → `=0`, `=false`, etc. all DISABLED the sweep, opposite of the intuitive reading. Extracted a shared `isTruthyEnvFlag(value)` helper in `qurl-webhook-registrar.js` that:
   - Allowlists `1` / `true` / `yes` / `on` (case-insensitive, whitespace-tolerant) as truthy.
   - Treats everything else (`0`, `false`, `no`, `off`, empty string, unset, non-string) as falsy.

   Both call paths (Lambda + bot's `linkGuildWebhookSubscription`) now use the same helper for consistent semantics across flip points.

   Tests added: 8 unit tests on `isTruthyEnvFlag` direct; 12 parameterized bot-path tests (6 truthy → sweep disabled, 6 non-truthy → sweep enabled); 1 Lambda end-to-end test confirming `=0` does NOT disable the sweep (footgun guard).

3. **(Comment density)** Acknowledged. The remaining inline prose is invariant-pinning / load-bearing; further trimming would lose safety reasoning the cr-bot has called valuable.

2845 tests + lint clean.

justin-layerv:
## cr round 14 addressed (`17e82fa2`)

1. **(Substantive) `failure_count` threshold tracked.** Added a `TODO(upstream-contract)` marker on `URL_MIGRATION_ORPHAN_MIN_FAILURES = 3` documenting the threshold's dependence on `failure_count` incrementing ONCE per failed delivery (not per retry attempt). If qurl-service ever publishes counter-semantics docs, `git grep` finds this site to pin against them.

2. **(Substantive) CloudWatch alarms filed as qurl-integrations-infra#1205.** Five alarms:
   - `URL-migration orphan deleted` spike (unexpected sweep)
   - `URL-migration orphan delete failed` spike (qurl-service degradation / IAM)
   - `liveness-gated near-misses` surge (schema drift / contract regression)
   - `URL-migration orphan sweep DISABLED via env override` first-occurrence (deliberate flip visibility)
   - (no alarm on `already absent` — concurrent cleanup is normal)

   The PR body's after-merge checklist now includes the infra alarm checkbox. The log lines exist and are runbook-grep-able from day one; this PR is safe to merge before the alarms land.

3. **(Cross-host deletion default-ON)** The cr-bot raised whether this should default OFF. Trade-offs:
   - Default OFF means every existing single-host deployment misses the recurrence-prevention benefit unless opted in (silent regression to the current bug).
   - Default ON with kill-switch + per-env owner-isolation invariant + alarms is the same defense-in-depth pattern as the dedupe path (also default-on, also performs DELETEs).
   - The blast-radius gates (path + cross-host + description-prefix + compound liveness + failure_count floor) are conservative enough that the false-positive risk on a single-host deploy is approximately zero.

   Keeping default-ON; net safety wins are higher for the recurrence-prevention goal. The kill-switch + alarms (qurl-integrations-infra#1205) give the multi-region rollout path a clean handoff.

4. **(Style)** Comment density acknowledged final time; remaining inline prose is invariant-pinning. Further trimming would lose load-bearing safety reasoning.

2845 tests + lint clean.

justin-layerv:
## cr round 15 addressed (`bbfebcee`)

1. **(Minor — fixed)** Stale comment that said `failure_count` is "checked as `typeof === 'number'`" replaced. The actual code uses `Number.isFinite`, and the very next comment explicitly says NOT to use `typeof`. The first comment would have mis-led a future reader. Replaced with phrasing that describes the routing (both checks → NEAR_MISS_LIVENESS) without pinning the wrong check shape.

2. **(Substantive — restated, no code change)** Destructive-default trade-off reasoning made stronger in the PR body. The cr-bot accepts the current call (single-host today, kill-switch + #827 + alarms are the right mitigations). Strengthened the after-merge checklist's third item to spell out: per-env `QURL_API_KEY` owner-isolation is the load-bearing invariant; any key consolidation MUST flip the switch first.

3. **(Cadence)** After-merge `failure_count >= 3` sanity check upgraded from one-time to **recurring per rename event**. If qurl-service ever stops attempting deliveries to a dead host, both signals would stall — recurring check catches that drift before it becomes a recurrence-prevention regression.

2845 tests + lint clean.

justin-layerv:
## cr round 16 addressed (`7df1fcc1`)

Two substantive defense-in-depth tightenings:

1. **(Risk-framing) `URL_MIGRATION_ORPHAN_MIN_FAILURES` raised 3 → 30.** The cr-bot is right that a small floor bought almost no protection against the active-active cannibalization scenario — a sustained-but-recoverable peer outage hits `failure_count >= 3` within minutes. Raising to 30 means a peer needs ~10-30 minutes of consecutive failures before becoming sweep-eligible (the kind of sustained outage where operator intervention is already in progress). Cost is asymmetric: a true URL-migration orphan's `failure_count` climbs unbounded, so the higher floor just delays cleanup by a bounded amount; a false-positive deletion is unrecoverable.

   The motivating orphan had `failure_count=1475` — comfortably above the new floor. All existing E2E tests use `failure_count` values well above 30 (1475 in the `deadOrphan()` fixture, 99 in the log-shape test, 100 in path-filter, etc.), so no test breakage.

2. **(Minor) `urlHost` switched from `URL.host` → `URL.hostname` (port-excluded).** A port-flip rename (`:8080` → `:8443`, or implicit-443 vs explicit-`:443`) at the same hostname is NOT a URL-migration (new port still resolves to the same backend, no orphan accrues). With the previous `URL.host`, that hypothetical would have classified same-hostname-different-port as cross-host and falsely swept. The bot's bridge URL is port-less today, so this is purely defense-in-depth.

   Test added: a sub at `discord.connector.layerv.xyz:8443/webhooks/qurl` with otherwise-orphan-shaped fields is NOT swept against an unparametrized bridge URL.

3. **(Acknowledged)** Multi-region near-miss log noise + comment density — both already documented + tracked; no further changes.

4. **(Cadence-stall risk)** Already documented in the threshold's `TODO(upstream-contract)` (added round 14): if qurl-service stops attempting deliveries to dead hosts, both signals stall and the sweep silently never fires. Recurring after-merge check catches this drift.

2846 tests + lint clean.

justin-layerv:
## cr round 17 addressed (`0b994030`)

**(Non-blocking but operationally important)** All `3` → `30` reference drift reconciled:

- `tests/qurl-webhook-registrar.test.js:1478` — comment `// below MIN_FAILURES (3)` → `(30)`.
- `tests/qurl-webhook-registrar.test.js:1783` — `< 3` reference made generic (`< N`) since the NaN/Infinity slip-through point holds for any finite floor.
- **PR description**: Summary (criterion 4), Invocation-cadence section (motivating incident comparison), and after-merge checklist (recurring verify step) all now say `30` instead of `3`.

The runtime behavior was already correct (constant + predicate use `URL_MIGRATION_ORPHAN_MIN_FAILURES`), and the tests pinned to that constant via `_internals` continued to pass — this was purely a comment + operator-facing-docs sync.

2846 tests + lint clean.

justin-layerv:
## cr round 18 addressed (`ce3e123d`)

**(Low/medium) Persistent 4xx log-spam edge case fixed.** The cr-bot's exact scenario: bot path runs many times/day across N replicas; a permanently-undeletable orphan (e.g. 403 if the API key loses delete scope on that resource) would emit an error line on every guild-link forever, drowning the CloudWatch alarm signal.

Fix: `cleanupUrlMigrationOrphans` now distinguishes failure classes:
- **5xx + network errors**: log at ERROR on EVERY invocation. Repeating IS the alarm signal.
- **4xx (non-404)**: log at WARN ONCE per (`webhook_id`, `status`) per process lifetime, then suppress identical subsequent logs. The DELETE itself is still attempted on every invocation (idempotent; the next boot's API key MAY have different scope) — only the log noise is suppressed.

Implementation: module-level `persistentClientErrorSuppressionCache` Set keyed by `${webhook_id}:${status}`. Test seam `_resetUrlMigrationOrphanDeleteSuppressionCache` exported via `_internals` so tests pin behavior across multiple invocations.

Two new tests:
- 403 on the same orphan across two invocations: WARN logged once, ERROR never logged, DELETE attempted both times.
- 503 on the same orphan across two invocations: ERROR logged BOTH times, WARN never logged (transient is the signal; don't suppress).

Acknowledged (no change):
- **(Style)** Comment density — continuing nudge, no further trim.
- **(Sequential vs parallel)** Justified by orphan count 0-1.
- **(Upstream-contract)** `last_delivery_success`/`failure_count` field-name dependencies are pinned with `TODO(upstream-contract)`; current qurl-service response shape verified to match (confirmed by the motivating incident's `failure_count: 1475, last_delivery_success: false` fixture data).

2848 tests + lint clean.

justin-layerv:
## cr round 19 addressed (`e6e7491c`)

Three findings:

1. **(Doc nit) Stale `past 3` reference in the env-isolation comment** bumped to `past 30` so the load-bearing safety narrative matches the constant. (qurl-webhook-registrar.js:416)

2. **(Edge case) 429 routing fixed.** The previous `isClientError` lumped 429 with persistent 4xx, so a sustained rate-limit would log once at WARN then silently suppress. Renamed to `isPersistentClientError` and explicitly exclude 429 (alongside the existing 404 exclusion). 429 now routes to the transient/ERROR branch like 5xx — log on every invocation so the alarm signal stays visible.

3. **(Symmetry / log spam) Near-miss log now deduped per-process** by `(bridgeUrl, count)` using the same Set-based strategy as the persistent-4xx suppression. Active-active hypothetical (#827): a healthy cross-host sibling is a perpetual near-miss, and on the bot path the line would fire on every guild-link event many times/day. The `(bridgeUrl, count)` key means an INCREASE or DECREASE in near-misses re-fires the log (something new became sweep-eligible), so the cohort change is still loud.

Three new tests:
- 429 across two invocations → ERROR logged BOTH times (transient).
- Healthy cross-host sibling across two invocations → near-miss log fires once, suppressed second time (steady state).
- Healthy sibling count `1 → 1 → 2` across three invocations → near-miss log fires twice (first + third), suppressed on second.

Acknowledged (no change):
- **(Cache cardinality)** `persistentClientErrorSuppressionCache` + `nearMissLogSuppressionCache` are unbounded module-level Sets. Distinct `(webhook_id, status)` / `(bridgeUrl, count)` cardinality is tiny in practice (0–1 orphans, single-host bridge URL).

2851 tests + lint clean.

justin-layerv:
## cr round 20 — pausing on these two items

Both findings here are substantive but they reverse direction on machinery the cr-bot itself asked for across rounds 2-18 (kill-switch in round 10, near-miss observability in round 8, 4xx suppression in round 18, failure_count compound gate in rounds 7-10). They are also non-blocking by the cr-bot's own assessment ("Nothing here is a correctness blocker"). I'm not going to ship code for either this turn — both are architectural posture decisions, not correctness fixes, and the appropriate venue is operator sign-off, not another iteration of the same self-contradicting loop.

### Item 1 (default ON vs OFF)
The cr-bot is right that an unrecoverable destructive op defaulting to ON is the riskier direction in isolation. The counter-argument: inverting to opt-in means every existing single-host environment that doesn't set the flag silently gets ZERO recurrence-prevention — the bug this PR exists for. The discriminator is whether the infra PR that lands alongside this one wires the enabling flag for the deployed envs (sandbox + prod). If yes, opt-in is strictly better; if no, opt-in neuters the feature.

### Item 2 (scope reduction)
The observability + kill-switch machinery exists because earlier cr-bot rounds asked for it. Moving it to ship "with #827" doesn't make it simpler — it makes it a follow-up PR. The current scope is justified by the safety properties the cr-bot itself called for.

### What I'd like a human call on

1. **Default posture**: ON (current) with kill-switch, OR OFF requiring per-env opt-in? Determines whether the recurrence-prevention is automatic or operator-gated.
2. **Scope**: ship as-is, OR strip the active-active-related observability/suppression and re-add when #827 lands?

Either direction is defensible. The PR has been mergeable on correctness grounds for several rounds. Open to feedback from a maintainer / DE.

2851 tests + lint clean; no open blockers from CI or correctness.

PR #828, opened 2026-06-17
feat(discord): reach watermark-detect over the qURL reverse-tunnel (#1101, PR-3)

justin-layerv:
## What

`/api/detect` moved off the public connector to a tunnel-only service (#1170 / #1187), so the bot's `detectWatermark` — which POSTed to `${CONNECTOR_URL}/api/detect` — now hits a host that no longer answers detect. This redirects it through the **qURL reverse-tunnel** via `@layervai/qurl` (CJS-compatible since 0.2.0, pinned `~0.2.0` — patch-only until the SDK reaches 1.0, so a `0.x` minor can't change `resolve`/knock semantics unreviewed).

## The mechanism (why this fixes the bot's dynamic-IP blocker)

`QurlClient.resolve({access_token: DETECT_ACCESS_TOKEN})` **issues an NHP knock for the bot's current egress IP** and returns `target_url`; the image POST then goes out from that same IP within the knock window. **Resolve-per-call** — `target_url` is never cached (a stale one was granted to a prior IP/window). This is exactly what lets the bot's dynamic, no-NAT IP reach a tunnel that only admits knocked IPs — the original blocker that drove the tunnel redesign.

## Auth

- **Resolve Bearer:** the global `config.QURL_API_KEY` — **must carry `qurl:resolve` scope** (a key-provisioning step; enforced server-side by qURL, not in this code).
- **Detect POST Bearer:** the per-call `apiKey` (presence-only on the detect side; the tunnel + NHP-knock is the real gate).
- **`at_` token:** the new `config.DETECT_ACCESS_TOKEN` (multi-use; the resolve credential).
- **Behavioral note (rollout):** detect now **hard-requires `QURL_API_KEY`** even when a per-call `apiKey` is passed — a tightening from the prior `apiKey || QURL_API_KEY` guard, because the resolve Bearer is *always* the global key (a per-call key can't substitute for it). In practice `QURL_API_KEY` is set in every deployment, so no live caller is affected; flagged for completeness and pinned by a dedicated test.

## SSRF guard

`assertPublicHttpsTarget` on the resolved `target_url`: `https:` only, reject embedded userinfo (the `https://good@127.0.0.1/` bypass), reject private/loopback/link-local (reuses `qurl.js`'s `isPrivateHost`, now exported). Syntactic-only — the target comes from the trusted qURL `resolve`, so this is defense-in-depth. A further host-pin to the expected qURL tunnel domain is a pre-activation gate (see below, #832).

## Observability

The two failure modes of this oracle path are logged distinctly (`logger.warn`, message-only — never the token or raw target_url): a failed knock/transport vs. a target rejected by the SSRF guard, so an activation-time failure is diagnosable rather than an undistinguished throw at `handleQurlDetect`.

## Unchanged

`detectWatermark` keeps its headers, 60s timeout, and `{detected,qurl_id,match_pct,confidence}` normalization. `handleQurlDetect`'s access model (SENDER/STAFF standing gate), cooldowns, same-guild filter, and Gate-0 recipient reveal are all untouched.

## Gated — and the activation prerequisites

`DETECT_COMMAND_ENABLED` stays default-OFF, so this is inert on merge. Activation (slots into the detect rollout sequence — **after** the sandbox detect soak) requires:
0. **Confirm the live `@layervai/qurl` 0.2.0 `resolve()` shape** — `{ access_token } → { target_url }` top-level (verified here against the SDK's `ResolveOutput` type; the `~0.2.0` patch-pin plus the `assertPublicHttpsTarget(undefined)` → "unparseable target URL" fail-safe guard against drift, but confirm against the live endpoint once before flipping the gate).
1. **Host-pin the resolved target (#832, defense-in-depth)** — confirm the observed sandbox detect tunnel hostname during the soak, then land #832 to assert `target_url` is under the expected qURL domain (likely `DESTINATION_DOMAIN = 'qurl.link'`) before flipping the gate. Deferred from this PR because the correct pin value is an activation-time observable, and a guessed domain risks an over-strict guard silently breaking detect.
2. **Seed `DETECT_ACCESS_TOKEN`** (the multi-use `at_` token) in the bot's SSM.
3. **Grant `qurl:resolve` scope** on the bot's `QURL_API_KEY`.
4. **Flip `DETECT_COMMAND_ENABLED=true`.**

(qurl-integrations has no prod-rollout-ledger convention; these prereqs are documented here + in the code comments, and slot into the infra-repo detect-activation runbook.)

## Tests

The detect suite is retargeted to the two-leg `resolve→POST` flow + new cases: SSRF private-target / non-https / public-host-with-userinfo (each independent of `isPrivateHost`), unset `DETECT_ACCESS_TOKEN` and unset `QURL_API_KEY` (clear configured-errors, no knock/POST), `guildId`-before-resolve ordering, `resolve()` returning no `target_url` and `resolve()` rejecting (both → no POST, each with its failure-branch breadcrumb asserted), and direct `isPrivateHost` cases pinning the fc/fd-ULA and fe80::/10 + fec0::/10 vs. public-DNS classification. Full suite **2776 pass**, lint clean, `@layervai/qurl` adds **zero transitive deps**.

## Follow-ups

- **#830** — consolidate the bot onto one qURL client (migrate `qurlFetch` → SDK), per the decision to standardize on `@layervai/qurl`.
- **#832** — pre-activation host-pin (above).
- **#833** — detect test polish (assert `getQurlClient` construction args; hoist the detect describe to a sibling).

### DIFF
--- a/apps/discord/package-lock.json
+++ b/apps/discord/package-lock.json
@@ -15,7 +15,7 @@
         "@aws-sdk/lib-dynamodb": "^3.1036.0",
         "@discordjs/rest": "~2.6.0",
         "@discordjs/ws": "~1.2.3",
-        "@layervai/qurl": "^0.2.0",
+        "@layervai/qurl": "~0.2.0",
         "discord-api-types": "^0.38.33",
         "discord.js": "14.25.1",
         "express": "~5.2.1",
--- a/apps/discord/package-lock.json
+++ b/apps/discord/package-lock.json
@@ -15,6 +15,7 @@
         "@aws-sdk/lib-dynamodb": "^3.1036.0",
         "@discordjs/rest": "~2.6.0",
         "@discordjs/ws": "~1.2.3",
+        "@layervai/qurl": "^0.2.0",
         "discord-api-types": "^0.38.33",
         "discord.js": "14.25.1",
         "express": "~5.2.1",
@@ -2066,6 +2067,15 @@
       "dependencies": {
         "@jridgewell/resolve-uri": "^3.1.0",
         "@jridgewell/sourcemap-codec": "^1.4.14"
+      }
+    },
+    "node_modules/@layervai/qurl": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/@layervai/qurl/-/qurl-0.2.0.tgz",
+      "integrity": "sha512-+ldlWnnXPBQKAkGO/5v00FdK6bOvK2kMW7Cii1/BO4RG5Vu7SiBinvt37YcgmNJtkQY+bw3kw127q5o2Hij0CA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20"
       }
     },
     "node_modules/@napi-rs/wasm-runtime": {
--- a/apps/discord/package.json
+++ b/apps/discord/package.json
@@ -26,6 +26,7 @@
     "@aws-sdk/lib-dynamodb": "^3.1036.0",
     "@discordjs/rest": "~2.6.0",
     "@discordjs/ws": "~1.2.3",
+    "@layervai/qurl": "^0.2.0",
     "discord-api-types": "^0.38.33",
     "discord.js": "14.25.1",
     "express": "~5.2.1",
--- a/apps/discord/package.json
+++ b/apps/discord/package.json
@@ -26,7 +26,7 @@
     "@aws-sdk/lib-dynamodb": "^3.1036.0",
     "@discordjs/rest": "~2.6.0",
     "@discordjs/ws": "~1.2.3",
-    "@layervai/qurl": "^0.2.0",
+    "@layervai/qurl": "~0.2.0",
     "discord-api-types": "^0.38.33",
     "discord.js": "14.25.1",
     "express": "~5.2.1",
--- a/apps/discord/src/config.js
+++ b/apps/discord/src/config.js
@@ -431,6 +431,17 @@ module.exports = {
   QURL_ENDPOINT: process.env.QURL_ENDPOINT
     || (process.env.NODE_ENV === 'production' ? 'https://api.layerv.ai' : 'http://localhost:8080'),
 
+  // Multi-use qURL access token (`at_...`) the bot resolves to reach the
+  // watermark-detect endpoint over the qURL reverse-tunnel (PR-3, #1101).
+  // connector.js's resolveDetectTarget() calls QurlClient.resolve({
+  // access_token: DETECT_ACCESS_TOKEN }) — which issues an NHP knock for the
+  // bot's current IP — immediately before each /api/detect POST. Secret-
+  // shaped (read verbatim from env like QURL_API_KEY / QURL_WEBHOOK_SECRET);
+  // no default. When unset, /qurl detect surfaces a clear configured-error
+  // (resolveDetectTarget throws) rather than silently failing. SSM-seeded at
+  // detect activation, the same gated step that flips DETECT_COMMAND_ENABLED.
+  DETECT_ACCESS_TOKEN: process.env.DETECT_ACCESS_TOKEN,
+
   // qurl-s3-connector
   CONNECTOR_URL: process.env.CONNECTOR_URL
     || (process.env.NODE_ENV === 'production' ? 'https://get.qurl.link:9808' : 'http://localhost:9808'),
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -504,10 +504,27 @@ async function resolveDetectTarget() {
   if (!config.DETECT_ACCESS_TOKEN) {
     throw new Error('DETECT_ACCESS_TOKEN is not configured (required to resolve the detect tunnel target)');
   }
-  const { target_url: targetUrl } = await getQurlClient().resolve({
-    access_token: config.DETECT_ACCESS_TOKEN,
-  });
-  return assertPublicHttpsTarget(targetUrl);
+  // Breadcrumb the two distinct failure modes of this oracle path so an
+  // activation-time failure is diagnosable — a failed knock/transport vs. a
+  // rejected target — rather than an undistinguished throw at the handler. Log
+  // ONLY the error message: never the access_token, and never the raw
+  // target_url (assertPublicHttpsTarget's messages are static and URL-free, and
+  // a malformed target could carry userinfo).
+  let targetUrl;
+  try {
+    ({ target_url: targetUrl } = await getQurlClient().resolve({
+      access_token: config.DETECT_ACCESS_TOKEN,
+    }));
+  } catch (err) {
+    logger.warn('Detect tunnel resolve failed (knock/transport)', { error: err.message });
+    throw err;
+  }
+  try {
+    return assertPublicHttpsTarget(targetUrl);
+  } catch (err) {
+    logger.warn('Detect tunnel target rejected by SSRF guard', { error: err.message });
+    throw err;
+  }
 }
 
 /**
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -417,8 +417,8 @@ async function mintLinks(resourceId, { expiresAt, n, apiKey, selfDestructSeconds
 // tests can inject a mocked @layervai/qurl before the first call.
 //
 // CACHE THE CLIENT, NEVER THE RESOLVED target_url: resolve() issues a fresh
-// NHP knock per call (see resolveDetectTarget) — a stale target_url would
-// have been opened for a previous IP/knock-window and must not be reused.
+// NHP knock per call (see resolveDetectTarget) — a stale target_url's network
+// access was granted to a previous IP/knock-window and must not be reused.
 //
 // Bearer note: the SDK's `apiKey` is the qURL API Bearer for the /v1/resolve
 // call and MUST carry the `qurl:resolve` scope (enforced server-side by qURL,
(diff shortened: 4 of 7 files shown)


justin-layerv:
Thanks — all three addressed in 9f3fd102, plus the unrelated red check.

**1. resolve() timeout (medium) — answered + made explicit.** The SDK *does* enforce its own request timeout: `ClientOptions.timeout` defaults to **30000 ms, per attempt** (qurl-typescript `types.ts:882` / `DEFAULT_TIMEOUT` `client.ts:92`), so resolve was already bounded at 30s/attempt — not an unbounded hang. I've now set it **explicitly** on the client (`new QurlClient({ …, timeout: 30000 })`) so the bound lives in our code and survives any SDK-default drift; a stalled resolve aborts the same way the POST's `AbortSignal.timeout(60000)` does. resolve() is a fast knock+lookup, so 30s sits well under the POST's 60s.

**2. apiKey-set / QURL_API_KEY-unset inconsistency (minor) — fixed.** `detectWatermark` now guards `if (!config.QURL_API_KEY)` directly. resolve() always authenticates with the global key (see `getQurlClient`); a per-call `apiKey` overrides only the POST Bearer, so it can't substitute for the resolve Bearer — the old `!apiKey && !config.QURL_API_KEY` would have let resolve go out with an undefined Bearer. The *other* functions keep `!apiKey && !config.QURL_API_KEY` (correct: those POST directly and never resolve). New test pins it: QURL_API_KEY unset + apiKey set ⇒ clean \"not configured\" error, **no resolve, no POST**.

**3. userinfo SSRF branch untested (minor) — added.** New case resolves `https://attacker@detect-tunnel.qurl.link/api/detect` (public host + userinfo): `isPrivateHost` wouldn't fire and the scheme is https, so **only** the userinfo guard can reject it — POST never happens. Pins that branch independently of the private-host / scheme cases.

**Confirming items:** the `baseUrl`→`/v1/resolve` prepend is a one-time live integration check at activation (it's in the rollout runbook's verify list); the syntactic-only guard (defense-in-depth on a trusted resolve) and the benign self-IP knock on reject are as you read them.

**Unrelated red check:** `Validate PR Title` was failing on `Unknown scope \"detect\"` — retitled scope **detect→discord** (the valid scope for `apps/discord` changes; \"detect\" isn't in the semantic-PR allowlist).

CI locally: full discord suite **2772/2772**, lint clean. The two new tests bring the detect contract to 54 cases.

justin-layerv:
Round-2 addressed in ee0c7b68, plus a `/simplify` pass.

**\"firewall\" terminology (must-fix) — done.** Reworded both doc comments to \"grants network access\" per CLAUDE.md, and the one \"opened for a previous IP\" line in the client-cache note for good measure. `grep -i firewall` on the diff is now clean.

**No retry on the resolve leg — confirmed + pinned.** The SDK *does* retry internally: `maxRetries` defaults to **3** on 429/5xx + transport errors (`qurl-typescript` `types.ts:879`, `DEFAULT_MAX_RETRIES` `client.ts:91`), i.e. the same 3-attempt resilience `qurlFetch` gives the other qURL calls. I've pinned it **explicitly** (`maxRetries: 3` next to `timeout: 30000`) so a single transient blip on resolve doesn't fail the whole detect interaction, and the resilience is visible in-code rather than relying on an SDK default. Worst-case is `timeout*(maxRetries+1)+backoff`, still well inside Discord's 15-min deferred-interaction window.

**SSRF syntactic-only asymmetry (no change required) — added a cross-reference anyway.** The guard comment now points at `qurl.js`'s `assertNotPrivateAfterResolve` and states the syntactic-vs-DNS difference is intentional (trusted `resolve()` source + tight knock window), so a future reader won't assume parity.

**Behavior-change callout (minor) — acknowledged.** The `!config.QURL_API_KEY` guard tightening is deliberate (resolve always uses the global key) and is pinned by a dedicated test.

---

**`/simplify` finding worth surfacing:** the pass flagged that the `@layervai/qurl` SDK I use for `resolve()` duplicates the bot's existing `qurlFetch` (same `/v1` base, Bearer, 30s timeout, 3-retry backoff), and that `qurl.js`'s header comment said the bot *avoids* the SDK. I verified from the SDK source that `resolve()` is a plain `POST /v1/resolve` (the NHP knock is server-side), so `qurlFetch` would be wire-identical. I raised switch-vs-keep with the maintainer; **decision: keep the SDK and standardize the bot on it going forward** (this PR is the first adopter). I've updated the now-stale `qurl.js` comment to reflect the transition and filed #830 to consolidate `qurlFetch` → SDK repo-wide (preserving the auth-audit + body-redaction behaviors). The remaining `/simplify` items (two duplicated-comment trims) are applied in ee0c7b68.

CI: full `apps/discord` suite **2772/2772**, lint clean.

justin-layerv:
Round-3 addressed in 2cc8817e.

**1. `isPrivateHost` fc/fd false-positive (latent correctness) — fixed.** Good catch — the bare `startsWith('fc')||startsWith('fd')` would have misclassified a public DNS host like `fd-detect.qurl.link` as private, and since this PR extends the guard to the qURL-provisioned tunnel target, that would have broken `/qurl detect` silently. Tightened to require the IPv6 shape: `(h.startsWith('fc')||h.startsWith('fd')) && h.includes(':')` — bracket-stripped ULA literals always carry a colon, DNS names don't (`fe80:` already had its). No SSRF regression: real IPv6 ULA literals still match (the existing `[fd00::1]`/`[fe80::1]` cases stay green), and DNS-name→private-IP was never a *syntactic* guard's job anyway (that's `assertNotPrivateAfterResolve` on the link path). Pinned with **direct `isPrivateHost` unit tests** (leveraging the new export): ULA/link-local literals → private; `fd-detect.qurl.link`/`fcdn.example.com` → not.

**2. Test gaps — both added.**
- `resolve()` returns no `target_url` (`{resource_id}` only) → `assertPublicHttpsTarget(undefined)` → `new URL(undefined)` throws → \"unparseable target URL\", **no POST**.
- `resolve()` rejects (knock/transport failure after the SDK's retries) → propagates to the handler, **no POST** attempted.

**3. Resolve Bearer snapshotted at first construction (informational) — acknowledged, no change.** Correct as you note: `config.QURL_API_KEY` is env-frozen at boot, so the first-construction snapshot can't drift. (The per-call `connectorAuthHeaders` asymmetry exists because the POST Bearer legitimately varies per call via the `apiKey` override; the resolve Bearer never does.)

CI: full `apps/discord` suite **2776/2776** (+4), lint clean.

justin-layerv:
Round-4 addressed in 7dfc7dc7 (+ PR-body updates). Thanks for confirming the `isPrivateHost` fix.

**3. `@layervai/qurl@^0.2.0` → `~0.2.0` — done.** Agree on the supply-chain reasoning for a control-plane (resolve/knock) client; pinned to patch-only so a `0.x` minor (semver-permitted to break) can't auto-resolve a changed `resolve()` contract unreviewed. Bonus: it matches the repo's existing `~`-pinning for `@discordjs/*`, `express`, `helmet`, `jose`, `node-cron`. Resolved version unchanged (0.2.0); lockfile spec synced (one-line diff).

**1. Hard-requires `QURL_API_KEY` (behavioral change) — called out in the rollout notes.** Added a **Behavioral note** under *Auth* in the PR body: detect now requires `QURL_API_KEY` even with a per-call `apiKey` (the resolve Bearer is always the global key). In practice it's set in every deployment so no live caller is affected, and it's pinned by a dedicated test — but it's now documented for the activation runbook.

**2. Per-detect tail-latency — acknowledged, accepted.** The extra resolve round-trip + knock before the 60s POST is the cost of the reach requirement; worst case stays within Discord's 15-min deferred window (documented). No change.

CI: full `apps/discord` suite **2776/2776**, lint clean. PR body's stale test count (2770) corrected to 2776 and the new test cases enumerated.

justin-layerv:
Round-5 addressed in 7c962e0f.

**`fe80::/10` + `fec0::/10` gap — fixed.** You're right that `fe80:` only caught `fe80::`, missing the rest of link-local (fe80–febf) and deprecated site-local fec0::/10. Pre-existing, but since this PR widens `isPrivateHost`'s reach (the tunnel target *and* the user-input link path both consume it), I tightened it: `/^fe[89a-f][0-9a-f]:/` covers the full fe80–feff first-hextet span, colon-gated like the fc/fd fix so a public DNS name (`feb-cdn.example.com`) can't be misclassified. Direct unit tests cover `febf::`/`fec0::`/`feff::` + the fe-DNS non-match. The existing `[fe80::1]` link-path case stays green.

**SDK `resolve()` contract — verified against source.** `ResolveOutput` (`qurl-typescript types.ts:280-281`) declares `target_url: string` **top-level + required**, and `resolve()` returns `request<ResolveOutput>(...)` → `envelope.data` — so the mock matches the real 0.2.0 SDK, not just an assumption. Added a **pre-activation item 0** to the PR body to confirm against the live endpoint once before flipping the gate; the `~0.2.0` pin + the tested `assertPublicHttpsTarget(undefined)` → "unparseable target URL" fail-safe are the drift hedges if the shape ever changes.

**Observability — added.** `resolveDetectTarget` now logs the two failure modes distinctly: `logger.warn('Detect tunnel resolve failed (knock/transport)', …)` vs. `logger.warn('Detect tunnel target rejected by SSRF guard', …)` — **message-only**, never the access_token or the raw target_url (the guard's messages are static + URL-free; a malformed target could carry userinfo). Both breadcrumbs are asserted in tests. Given the deanonymization-oracle nature, those are exactly the branches worth a trail, as you noted.

**Minor (snapshot / engines)** — informational, no change; the `QURL_API_KEY` guard runs before `getQurlClient()` constructs, so it can't snapshot an undefined key.

CI: full `apps/discord` suite **2776/2776**, lint clean.

justin-layerv:
Thanks — dispositioning round 6. No code push this round: each item is defense-in-depth / pre-existing / cosmetic, handled via issues + a pre-activation gate rather than reopening the review loop on a default-OFF PR.

**1. Host-pin the resolved target (defense-in-depth) → deferred to #832 + added as a pre-activation gate.** Agree it's worth doing on a Bearer-carrying leg. Deferring rather than guessing: the *correct pin value is the detect tunnel's actual hostname*, which is an activation-time observable (the sandbox soak surfaces it). `DESTINATION_DOMAIN = 'qurl.link'` (`constants.js`) is the likely value — the connector tunnel is `get.qurl.link` — but the detect tunnel is a distinct slug/service and could differ per env, and pinning to a *guessed* domain risks exactly the over-strict-guard-silently-breaks-detect failure we just fixed for `isPrivateHost` fc/fd. So #832 sequences it: confirm the observed host during the soak, then enforce before flipping `DETECT_COMMAND_ENABLED`. It's now PR-body pre-activation **item 1**, next to the resolve()-shape check.

**2. Bare expanded loopback (`0:0:0:0:0:0:0:1`) → acknowledged, not reachable today.** Both production call sites parse through `new URL(...)`, which normalizes IPv6 to the compressed form (`::1`, exact-matched) before `isPrivateHost` sees it — so the expanded form can't reach the function in prod. Correct to flag now that it's exported; a future *direct* caller is the only way to hit it. Folding the "harden for direct callers" thought into #832's neighborhood if we touch the guard there; not changing it here to avoid reopening the loop for an unreachable path.

**3. `getQurlClient` construction-args assertion → #833.** Fair — it's currently only *indirectly* pinned (the unset-`QURL_API_KEY` test proves resolve depends on the global key). #833 adds the direct `expect(QurlClient).toHaveBeenCalledWith(...)`.

**4. Detect describe nested under the MD5 describe → #833.** Pre-existing placement nit; #833 hoists it to a sibling with its own `beforeEach`.

That dispositions every round-6 item (fixed-or-deferred-with-issue). Given the PR has been "ready to merge / nothing blocking" for four consecutive rounds with items decaying to pre-existing/defense-in-depth, I'm treating this as converged: **DE-accept ✓, CI green ✓, cr considered-and-dispositioned ✓, /simplify clean ✓.** Follow-ups #830/#832/#833 are linked in the PR body.

PR #835, opened 2026-06-17
fix(discord): flip recipient DM to "closed" when a one-time qURL is consumed

justin-layerv:
## Summary

When a recipient opens a **one-time qURL** and consumes it, the link is dead for them — they can no longer reach the content. But until now their delivered DM kept showing the present-tense **"🕐 Closes `<t:expiresAt:R>`"** embed (e.g. *"Closes in ~25m"*) right up until the 30-minute link expiry fired the `qurl.expired` webhook (or the sender ran `/qurl revoke`). Seeing *"Closes in 25m"* on a link that is **already dead** is confusing and quietly undermines the one-time-use trust signal — the whole point of a one-time link is that opening it closes the door. This change makes **consumption flip the DM immediately**.

When a `qurl.accessed` webhook arrives with `data.consumed === true`, the bot now flips the recipient's DM to a past-tense closed state:

> 🔒 You opened this one-time qURL.
> It has been used and is no longer active.

## Scope

- [ ] `apps/slack/`
- [ ] `apps/teams/`
- [x] `apps/discord/`
- [ ] `apps/cli/`
- [ ] `apps/zapier/`

## Changes

- **`dm-payloads.js`** — new `buildConsumedDMPayload()`. **Correctness note:** the copy is deliberately *marker-free* (static past/perfect tense, no `<t:N:R>`). `buildExpiredDMPayload` renders a relative-time marker against `expires_at`, but at consumption time `expires_at` is still in the **future**, so reusing it would render *"expired in 25 minutes"* — the same future-tense bug. Static copy is the fix.
- **`routes/qurl-webhook.js`** —
  - Extracted a shared `flipRecipientDMToClosed` helper (GSI lookup → editability gate → sibling-marker cross-check → revoke guard → idempotency-marker claim → `editDM` → marker rollback on transient failure). It **returns a verdict and never writes to `res`**; each caller owns its HTTP-status mapping. `handleQurlExpired` now delegates to it with unchanged external behavior.
  - On `qurl.accessed` with `consumed === true`, runs the flip **fire-and-forget** off the already-returned 200 (mirrors the existing view-counter publish), **not** `handleQurlExpired`'s 503 coupling: the view already recorded, so a flip failure must not make qurl-service retry the whole accessed event.
- **`store/ddb-store.js`** + **`store/contract.js`** — new `consumed_edited_at` idempotency marker (`markConsumedDMEdited` / `clearConsumedDMEdited`), **separate** from `expired_edited_at`. A one-time qURL gets consumed **and** later hits its TTL — `qurl.expired` fires regardless of prior state — so each path needs its own dedup and must not clobber the other's copy. The expired path now skips when `consumed_edited_at` is set, and the consumed path skips when `expired_edited_at` is set.

### Idempotency + the revoke / expired races
- **Redelivery:** the consumed flip is gated on `consumed === true` (not `dbResult === 'recorded'`), so a redelivered `qurl.accessed` re-enters and recovers a transiently-missed flip; the `consumed_edited_at` marker short-circuits the redundant edit.
- **Revoke race:** reuses the existing `isSendRevoked` guard — a sender-revoke ("🚪 …closed the door") is never overwritten.
- **Consume-then-expire:** the later `qurl.expired` event finds `consumed_edited_at` set and skips, so the more-accurate "you opened it" copy survives.
- **Transient flip failure:** rolls the `consumed_edited_at` marker back so a redelivery (or, failing that, the `qurl.expired` backstop) re-attempts. A permanent failure (recipient blocked the bot / deleted the DM) keeps the marker.

## Test Plan

- [x] `npm run lint` (eslint, `--max-warnings 0`) green
- [x] `npm test` (`jest --coverage`) green — **2798 tests across 82 suites**, coverage thresholds satisfied (`qurl-webhook.js` at 96%)
- [x] New tests added: `tests/qurl-webhook-consumed.test.js` (happy path, no-flip on `consumed: false` / stringified `"true"`, every skip branch incl. the sibling-marker cross-check, edit-failure handling, and the **no-future-marker payload assertion that pins the fix**) + a cross-marker skip test in `tests/qurl-webhook-expired.test.js` + store-level tests for the new markers in `tests/ddb-store.test.js`
- [x] For app-impacting changes: branch is current with `main`

## Related Issues

Closes #

### DIFF
--- a/apps/discord/src/dm-payloads.js
+++ b/apps/discord/src/dm-payloads.js
@@ -63,12 +63,9 @@ function buildExpiredDMPayload({ expiresAtSeconds }) {
 // `<t:now:R>` "opened just now" anchor is possible but adds nothing —
 // the recipient just clicked, they know when.)
 //
-// UX choices mirror buildExpiredDMPayload: the embed wholly replaces
-// the original delivery embed (PATCH only supports whole-array embed
-// replacement, and the "tap to step through" framing is misleading once
-// the one-time content is gone), color stays on QURL_BRAND for symmetry
-// with the expired/revoke paths, and `components: []` clears the now-
-// dead Step Through button (see module CONTRACT above).
+// UX choices (whole-embed replacement, QURL_BRAND color, `components: []`
+// to clear the now-dead Step Through button) mirror buildExpiredDMPayload
+// — see that function's comment for the why.
 function buildConsumedDMPayload() {
   const embed = new EmbedBuilder()
     .setColor(COLORS.QURL_BRAND)
--- a/apps/discord/src/dm-payloads.js
+++ b/apps/discord/src/dm-payloads.js
@@ -68,10 +68,9 @@ function buildExpiredDMPayload({ expiresAtSeconds }) {
 // FUTURE. Rendering it through Discord's relative-time marker (as the
 // expired payload does) would read "expired in 25 minutes" — a
 // future-tense "expired", exactly the confusing state this whole change
-// kills. Static copy is the fix: it says "you opened it, it's done"
-// without any time anchor that could re-render future-tense. (A
-// `<t:now:R>` "opened just now" anchor is possible but adds nothing —
-// the recipient just clicked, they know when.)
+// kills. Static copy is the fix: no time anchor that could re-render
+// future-tense. (This is the single canonical home for the marker-free
+// rationale; the route + test comments point here.)
 //
 // UX choices (whole-embed replacement, QURL_BRAND color, `components: []`
 // to clear the now-dead Step Through button) mirror buildExpiredDMPayload
--- a/apps/discord/src/dm-payloads.js
+++ b/apps/discord/src/dm-payloads.js
@@ -5,9 +5,9 @@
 // require graph (Discord REST, view-update-registry, flow-state,
 // etc.) at module-init time during tests.
 //
-// Today only buildExpiredDMPayload lives here. buildRevokedDMPayload
-// stays in commands.js next to its sole caller (the /qurl revoke
-// editTargets loop) until a second consumer needs it.
+// Today buildExpiredDMPayload + buildConsumedDMPayload live here.
+// buildRevokedDMPayload stays in commands.js next to its sole caller
+// (the /qurl revoke editTargets loop) until a second consumer needs it.
 //
 // CONTRACT: every builder MUST pass `components: []` explicitly.
 // Discord's PATCH /messages does NOT clear fields that aren't
@@ -48,4 +48,32 @@ function buildExpiredDMPayload({ expiresAtSeconds }) {
   return { embeds: [embed], components: [] };
 }
 
-module.exports = { buildExpiredDMPayload };
+// Companion to buildExpiredDMPayload for the qurl.accessed webhook path
+// when `data.consumed === true`: the recipient opened a one-time qURL
+// and the content has been served, so for THEM the door has already
+// closed — the link is dead even though its 30m expiry hasn't elapsed.
+//
+// COPY IS DELIBERATELY MARKER-FREE (past/perfect tense, no <t:N:R>):
+// at consumption time the link's `expires_at` is still ~minutes in the
+// FUTURE. Rendering it through Discord's relative-time marker (as the
+// expired payload does) would read "expired in 25 minutes" — a
+// future-tense "expired", exactly the confusing state this whole change
+// kills. Static copy is the fix: it says "you opened it, it's done"
+// without any time anchor that could re-render future-tense. (A
+// `<t:now:R>` "opened just now" anchor is possible but adds nothing —
+// the recipient just clicked, they know when.)
+//
+// UX choices mirror buildExpiredDMPayload: the embed wholly replaces
+// the original delivery embed (PATCH only supports whole-array embed
+// replacement, and the "tap to step through" framing is misleading once
+// the one-time content is gone), color stays on QURL_BRAND for symmetry
+// with the expired/revoke paths, and `components: []` clears the now-
+// dead Step Through button (see module CONTRACT above).
+function buildConsumedDMPayload() {
+  const embed = new EmbedBuilder()
+    .setColor(COLORS.QURL_BRAND)
+    .setDescription('🔓 You opened this one-time qURL.\nIt has been used and is no longer active.');
+  return { embeds: [embed], components: [] };
+}
+
+module.exports = { buildExpiredDMPayload, buildConsumedDMPayload };
--- a/apps/discord/src/dm-payloads.js
+++ b/apps/discord/src/dm-payloads.js
@@ -53,6 +53,14 @@ function buildExpiredDMPayload({ expiresAtSeconds }) {
 // and the content has been served, so for THEM the door has already
 // closed — the link is dead even though its 30m expiry hasn't elapsed.
 //
+// The "one-time qURL" wording is a contract assumption: qurl-service
+// sets `consumed: true` ONLY on the one-time-use branch (resolve_service
+// commitConsume → publishAccessEvent); a multi-use link hitting its
+// final allowed access increments the use counter and emits
+// `consumed: false`, never true. So consumed===true strictly implies a
+// one-time link here. If that emit-side invariant ever changes, this
+// copy must change with it.
+//
 // COPY IS DELIBERATELY MARKER-FREE (past/perfect tense, no <t:N:R>):
 // at consumption time the link's `expires_at` is still ~minutes in the
 // FUTURE. Rendering it through Discord's relative-time marker (as the
--- a/apps/discord/src/dm-payloads.js
+++ b/apps/discord/src/dm-payloads.js
@@ -60,6 +60,8 @@ function buildExpiredDMPayload({ expiresAtSeconds }) {
 // `consumed: false`, never true. So consumed===true strictly implies a
 // one-time link here. If that emit-side invariant ever changes, this
 // copy must change with it.
+// TODO(upstream-contract): consumed===true implies the one-time-use
+// branch in qurl-service; keep this copy in lockstep if that changes.
 //
 // COPY IS DELIBERATELY MARKER-FREE (past/perfect tense, no <t:N:R>):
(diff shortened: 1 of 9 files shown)


justin-layerv:
Thanks for the review — addressed in 09ed9b2b (comment + test hardening only, no behavior change). Per-item:

**Q1 — consumed/expired simultaneity double-edit.** Agreed, documented. Added a TOCTOU note in the sibling cross-check comment mirroring the existing revoke-race treatment: if both events pass the cross-check before either marker is written, both editDM and it's last-writer-wins, end state still "closed" (just possibly the less-specific copy). I also named the single-shared-marker alternative there (a unified `dm_closed_at` claimed via conditional write would serialize the two paths and close the race outright) and why it's out of scope for this PR — it's a schema migration of the pre-existing `expired_edited_at` attribute for a race whose worst case is the less-specific copy. No locking.

**Q2 — does `consumed: true` ever fire for a multi-use link's final access?** Verified in qurl-service: **no.** `commitConsume` (internal/service/resolve_service.go) sets `consumed = true` **only** inside the `if rc.token.OneTimeUse` branch on a successful `ConsumeTokenByHash`; the multi-use branch calls `IncrementUseCountByHash` and returns `false` unconditionally. There's no "final allowed access flips consumed" concept. So `consumed: true` strictly implies `OneTimeUse`, and the "You opened this one-time qURL" copy is provably correct. No change needed.

**Q3 — fire-and-forget loss on process exit.** Confirmed intended durability bound: a flip lost to a SIGTERM/freeze before the microtask drains is recovered by the `qurl.expired` backstop (which fires regardless of prior consume state and skips only when `consumed_edited_at` is set). The function comment already states this.

**Minor — `payload-build-failed` status removed.** Confirmed safe: grepped qurl-integrations and the -infra dashboards/alerts — nothing references the string. qurl-service reads the HTTP code, not the body `status`. The consolidation (validating `buildPayload` before the marker claim, so no claim → no rollback) is a net simplification.

**Minor — transient-verdict coverage seam.** Closed. The consumed path's transient verdicts (`lookup-error`, `mark-error`, `edit-failed-rollback-failed`) have no HTTP mapping, so they're now asserted via the terminal verdict debug log. Also added a `{status, transient}` verdict assertion to every skip + edit-failure case. +3 tests (2801 total), suite + lint green.

Also ran /simplify: trimmed three comment blocks that re-derived the "why" already canonical next to `buildExpiredDMPayload`.

justin-layerv:
Round 2 addressed in 7a59baa8 (comment + test only):

- **#1 (confirm consumed:true only for one-time links):** verified on the emit side and anchored the contract assumption **in-code** in `buildConsumedDMPayload` (not just this thread): `commitConsume` sets `consumed = true` only inside `if rc.token.OneTimeUse`; the multi-use branch increments the counter and returns `false`. `publishAccessEvent` is the sole `qurl.accessed` emitter and passes that boolean straight through. So `consumed===true` strictly implies a one-time link.
- **#2 (comment drift, marker-free "why" stated in 3 places):** consolidated to ONE canonical source (`buildConsumedDMPayload`); the route and test-header comments are now pointers, not re-derivations.
- **Noted seam (split-emission flow):** added an end-to-end test — a `consumed:false` event then a `consumed:true` event on the same qURL flips ONLY on the burn, exercising `recordQurlView`'s documented false→true follow-on shape through the route.

Suite + lint green (2802 tests).

justin-layerv:
Round 3 addressed in 580a2e1b:

- **#1 (tag the contract assumption greppable):** added `// TODO(upstream-contract): consumed===true implies the one-time-use branch in qurl-service; keep this copy in lockstep if that changes.` in `buildConsumedDMPayload`, wiring it into the repo's existing `git grep TODO(upstream-contract)` lockstep convention (CLAUDE.md).
- **#2 (redelivery through the route):** added a test that drives `recordQurlView` → `'dedup'` (replayed event_id) and asserts the flip STILL fires (`markConsumedDMEdited` + `editDM` run, verdict `edited`). This pins the headline decision to gate on `consumed`, not `dbResult === 'recorded'` — the `already-edited` test covered the marker level; this covers the route level.
- **#3 (no concurrency bound on the fire-and-forget chain):** conscious choice, no change — consistent with the existing `viewUpdatePublisher.publish` pattern, and one-time-link opens aren't a burst workload. Flagging acknowledged.

Suite + lint green (2803 tests).

justin-layerv:
Round 4 addressed in e76f5c16 (comment accuracy only):

- **#1 (process-restart degrades the copy):** good catch — tightened the fire-and-forget comment. A bounce after the 200 but before the deferred flip runs leaves `consumed_edited_at` unclaimed, so the `qurl.expired` backstop flips the DM to the less-specific "expired" copy, not "you opened it". Self-heals to a correct "closed" state regardless; only copy specificity degrades, bounded by the 30m TTL. Now documented precisely (it previously implied a clean recovery).
- **#2 (outer `.catch` unreachable in tests):** keeping as-is — it's correct defense-in-depth against exactly the sync-throw refactor of the helper it guards; no value in a contrived test.
- **#3 (comment density):** judgment call acknowledged. Three blocks were already trimmed to pointers in earlier rounds; the rest is load-bearing failure-mode reasoning in this file's established style, so I've left it.

Brand spelling / strict-equality / marker-claim-ordering confirmations all noted — thanks.

justin-layerv:
Round 5 addressed in 8e1461f6 (copy + comments only):

- **#4 (🔓 reads as "still accessible"):** good UX catch — switched to 🔒. The closed lock signals "done" and avoids the open-padlock ambiguity, consistent with the sibling copies (`⌛`/`🚪`). The whole point of this change is to stop confusing the recipient, so an emoji that could read "still open" undercut it.
- **#1 (GSI-projects-ALL is load-bearing):** strengthened the cross-check comment to name the failure mode — if `qurl_id-index` ever narrows from `ALL`, the cross-check silently always-misses and clobbers the sibling copy. It's a pre-existing dependency (the expired path already reads `created_at`/`dm_status` off the same projection); flagged greppably at the read site.
- **#2 (access_count gate precedes the flip):** added a comment noting it's intentional — a malformed/zero `access_count` short-circuits at `invalid-payload`; by contract a consumed event always carries `access_count >= 1`, and the `qurl.expired` backstop covers the theoretical miss.
- **#3 (process-bounce):** confirmed — the webhook receiver is a long-lived Express server (not a freeze-after-response runtime), and the window is already documented in the flip comment.
- **#5 (comment density):** locally consistent with this file's style — left as-is.

Suite + lint green (2803 tests).

justin-layerv:
Round 6 addressed in dbe02552:

- **#1 (unguarded `buildPayload(row)` could reject into Express on the expired path):** fixed at the seam — wrapped `buildPayload` in try/catch inside `flipRecipientDMToClosed`, so a throw becomes a permanent `payload-build-error` skip instead of escaping. This covers BOTH callers (the consumed path was already insulated by the background `.catch`; this closes the expired path, which awaits the delegation outside the route's `try`). Chose the try/catch over the "one-line note" alternative you offered since it's the stronger fix. Following up with a test for the branch (see next push).
- **#2 (no graceful-shutdown drain):** matches the existing `viewUpdatePublisher.publish` fire-and-forget pattern; recovery leans on redelivery + the 30m-TTL-bounded `qurl.expired` backstop, already documented. No change.
- **#3 (comment density):** the new TOCTOU/GSI-projection notes are at the upper end but load-bearing; left as-is per your "fine to keep."

justin-layerv:
Follow-up to round-6 #1 in 83b7ce31: added a test for the new `buildPayload`-throw guard after all. I'd called it hard to test without testing hypothetical code, but it's cleanly fencible — `qurl-webhook-consumed-payload-throw.test.js` mocks `dm-payloads` so `buildConsumedDMPayload` throws, drives a consumed event through the route, and asserts the flip skips with the `payload-build-error` verdict, claims no marker, never edits, and surfaces the throw as the guard's error log (NOT the outer flip-threw `.catch`) — proving the throw is contained inside `flipRecipientDMToClosed` and never reaches Express. Covers the previously-untested branch; qurl-webhook.js coverage 96.23%. 83 suites / 2804 tests green.

justin-layerv:
Round 7 nit addressed in 446d66e0: trimmed the canonical marker-free block in `buildConsumedDMPayload` — dropped the `<t:now:R>` digression and labeled it explicitly as the single canonical home (the route + test comments have been pointers to it since rounds 1–2, which is the one-home + pointers structure you're recommending). Same information, less to keep in sync. Comment-only; suite + lint green.

justin-layerv:
Round 8 addressed in c09866d1:

- **#4 (test-helper duplication):** extracted `flipVerdict`/`drainTicks`/`flushFlip` into `tests/helpers/consumed-flip.js` (matching the repo's existing `tests/helpers/` convention), parameterized by the caller's mocked logger. Both consumed suites now bind thin wrappers; call sites unchanged. 83 suites / 2804 tests green.
- **#1 (comment density):** noted for a possible future design-note pass; the inline failure-mode prose is load-bearing and the marker-free rationale is already one-canonical-home + pointers.
- **#2 (status-string surface shrank):** correct — `payload-build-failed` is gone and the new `payload-build-error` added; both are response-body `status` strings only (qurl-service reads the HTTP code), so external behavior is unchanged as the PR body claims. Nothing greps either string.
- **#3 (shutdown gap):** matches the existing `viewUpdatePublisher.publish` fire-and-forget pattern; self-heals via the 30m-TTL-bounded backstop, documented. No change.

Thanks for the careful pass across the consume/expire/revoke races.

justin-layerv:
Round 9: thanks. The one concrete item (#3, 🔓 in the PR description vs 🔒 in the code) was a PR-body inconsistency — fixed the description to 🔒 to match the code. No code change. #1 (Lambda-runtime latent constraint), #2 (`TODO(upstream-contract)` invariant), and #4 (comment density) are all acknowledged/documented with no change needed per your read.

This converges the review loop: substantive code feedback ended at the round-6 `buildPayload`-throw guard (added + tested); rounds 7–9 were progressively smaller nits (comment canonicalization, test-helper DRY, PR-body emoji), each addressed. CI is fully green on the latest commit (c09866d1), `/simplify` is clean, and there are no remaining code-actionable items. Calling it merge-ready.

PR #838, opened 2026-06-17
fix(discord): host-pin the detect tunnel target to the qURL tunnel domain (#832)

### LINKED ISSUE #832: Pin the detect tunnel target host to the expected qURL domain (defense-in-depth, before activation)
## Context

From the [#828](https://github.com/layervai/qurl-integrations/pull/828) review. `detectWatermark` POSTs the raw image bytes **and** `Authorization: Bearer <apiKey>` to whatever host `resolve()` returns. `assertPublicHttpsTarget` blocks only *private/loopback/link-local* hosts and rejects non-https/userinfo — it does **not** constrain the target to the known qURL tunnel domain.

## Threat (defense-in-depth)

If the qURL control plane were ever compromised/spoofed and `resolve()` returned a malicious-but-**public** https host, the bot would ship a deanonymization-oracle payload (a recipient image) + its API-key Bearer to an arbitrary public endpoint. The control plane is authenticated (QURL_API_KEY over TLS), so this is a strong precondition — hence defense-in-depth, not a live hole — but the leg carries a Bearer + sensitive payload, so pinning the host is cheap insurance.

## Why deferred from #828 (not fixed inline)

The **correct pin value is the detect tunnel's actual hostname**, which is an activation-time observable (the sandbox detect soak surfaces it). `DESTINATION_DOMAIN = 'qurl.link'` (`apps/discord/src/constants.js`) is the likely value — the connector tunnel already lives at `get.qurl.link` — but the detect tunnel is a distinct slug/service and the domain could differ per env. Pinning to a *guessed* domain risks the same failure class as the `isPrivateHost` fc/fd false-positive fixed in #828: an over-strict host guard silently breaks the feature. So this is sequenced as a **pre-activation** task (added to #828's PR-body pre-activation gates), to be confirmed against the observed tunnel host before enforcing.

## Acceptance criteria

- In `assertPublicHttpsTarget` (or `resolveDetectTarget`), assert the resolved `target_url` hostname `=== <expected>` **or** `endsWith('.' + <expected>)`, where `<expected>` is the confirmed detect tunnel domain (default `DESTINATION_DOMAIN`; overridable via env if a non-`qurl.link` tunnel domain is ever used).
- Reject a mismatch with a clear error + the existing SSRF-rejection breadcrumb (message-only, no URL).
- Tests: matching host/subdomain allowed; a public non-qURL host rejected with no POST.
- **Confirm against the observed sandbox detect tunnel host first**, then enforce — before flipping `DETECT_COMMAND_ENABLED`.

justin-layerv:
## What
`assertPublicHttpsTarget` now requires the `resolve()`-returned `target_url` host to be under **`qurl.site`** (the qURL reverse-tunnel domain). Closes #832.

## Why
The detect leg is a deanonymization oracle that POSTs the image bytes **and** the bot's API-key Bearer to whatever host `resolve()` returns. The prior guard blocked private/non-https/userinfo hosts but **not** an arbitrary *public* host — so a compromised or spoofed qURL control plane returning `https://evil.example.com/...` would have been handed the payload. Pinning the host to the tunnel domain closes that gap (defense-in-depth — `resolve()` is still trusted; this is belt-and-suspenders on a Bearer-carrying leg).

## Domain correction (qurl.link → qurl.site)
The earlier #828/#832 references to `qurl.link` were **wrong** for the tunnel. Verified against the infra: the qURL reverse-tunnel resource host is **`r_<id>.qurl.site`** (`qurl-s3-connector/terraform/fileviewer-tunnel-v2.tf`; the `view_domain` validator pins `^https://r_[a-z0-9_-]{11}\.qurl\.site$`, mirroring qurl-service `resourceIDPattern`). `qurl.link` is the short-link/ALB domain. This PR corrects the test fixtures accordingly.

## Pin strength
The enforced check is the **`.qurl.site` suffix** (`host === 'qurl.site' || endsWith('.qurl.site')`) — robust, won't false-reject if the detect tunnel host differs slightly from the per-resource form. Once the **sandbox soak confirms the observed host**, it can tighten to `/^r_[a-z0-9_-]{11}\.qurl\.site$/`.

## Tests
- public non-qURL host (`evil.example.com`) → rejected, no POST, logged via the SSRF breadcrumb;
- `evilqurl.site` suffix-boundary → rejected (no dot separator);
- happy-path + userinfo cases now use a real `r_<id>.qurl.site` target.
Full `apps/discord` suite **2778** pass, lint clean.

## Ordering / companion
Companion to the merged #828 (bot resolve-then-POST) and the merged #1209 (infra activation wiring). Listed as Phase-0 gate in the production-release runbook (qurl-integrations-infra#1219). Pin value to be confirmed at the sandbox soak.

### DIFF
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -457,8 +457,9 @@ function getQurlClient() {
 
 // SSRF guard for the resolve()-returned tunnel target. Must be a PUBLIC
 // `https:` URL; reject any non-https scheme, embedded userinfo (the
-// `https://good@127.0.0.1/` hostname-confusion bypass), and any
-// private/loopback/link-local host (reusing qurl.js's syntactic isPrivateHost).
+// `https://good@127.0.0.1/` hostname-confusion bypass), any
+// private/loopback/link-local host (reusing qurl.js's syntactic isPrivateHost),
+// and any host NOT under the qURL reverse-tunnel domain (qurl.site).
 // Deliberately NOT port-locked (the tunnel target may sit on a non-standard
 // port) and NOT DNS-resolved — a syntactic check ONLY, unlike the link-minting
 // path's assertNotPrivateAfterResolve in qurl.js, which adds a DNS-level
@@ -481,6 +482,17 @@ function assertPublicHttpsTarget(targetUrl) {
   }
   if (isPrivateHost(parsed.hostname)) {
     throw new Error('Detect tunnel target points to a private/internal address');
+  }
+  // Host-pin (defense-in-depth on this Bearer-carrying oracle leg): the resolved
+  // target MUST be under the qURL reverse-tunnel domain, qurl.site. The tunnel
+  // resource host is `r_<id>.qurl.site` (qurl-service resourceIDPattern); we pin
+  // the `.qurl.site` suffix so a compromised or spoofed resolve() that returned a
+  // public NON-qURL host can't be POSTed the image bytes + our API-key Bearer.
+  // (NOT qurl.link — that's the short-link/ALB domain, not the tunnel.) Once the
+  // sandbox soak confirms the observed host, tighten to /^r_[a-z0-9_-]{11}\.qurl\.site$/.
+  const host = parsed.hostname.toLowerCase();
+  if (host !== 'qurl.site' && !host.endsWith('.qurl.site')) {
+    throw new Error('Detect tunnel target host is not under the expected qURL tunnel domain (qurl.site)');
   }
   return targetUrl;
 }
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -484,14 +484,15 @@ function assertPublicHttpsTarget(targetUrl) {
     throw new Error('Detect tunnel target points to a private/internal address');
   }
   // Host-pin (defense-in-depth on this Bearer-carrying oracle leg): the resolved
-  // target MUST be under the qURL reverse-tunnel domain, qurl.site. The tunnel
-  // resource host is `r_<id>.qurl.site` (qurl-service resourceIDPattern); we pin
-  // the `.qurl.site` suffix so a compromised or spoofed resolve() that returned a
-  // public NON-qURL host can't be POSTed the image bytes + our API-key Bearer.
-  // (NOT qurl.link — that's the short-link/ALB domain, not the tunnel.) Once the
+  // target MUST be a SUBDOMAIN of the qURL reverse-tunnel domain. The tunnel
+  // resource host is `r_<id>.qurl.site` (qurl-service resourceIDPattern) — always
+  // a subdomain, so we require the `.qurl.site` suffix; the bare apex is never a
+  // detect target. A compromised or spoofed resolve() returning a public NON-qURL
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks — LGTM noted, both nits addressed in 45244657.

1. **Apex branch untested** — added `host-pin allows the apex \`qurl.site\` host (POST proceeds)`, exercising the `host === 'qurl.site'` arm (the `*.qurl.site` arm stays covered by the happy-path `TUNNEL_TARGET`). 59/59.
2. **Misleading test title** — reworded to `host-pin rejects the look-alike suffix \`evilqurl.site\` (no dot separator)`.
3. **Future-tighten reminder** — already tracked: the production-release runbook ([qurl-integrations-infra#1219](https://github.com/layervai/qurl-integrations-infra/pull/1219)) Phase 1 step 6 makes "confirm the observed `target_url` host = `r_<id>.qurl.site`" a soak gate, and the #832 Phase-0 line says to tighten the pin to `^r_[a-z0-9_-]{11}\.qurl\.site$` once confirmed. So the broad pin can't quietly become permanent.

Re the homograph/apex/DNS-control points — agreed on all; the pin holds for the realistic threat (a spoofed resolve() returning a non-qURL public host).

justin-layerv:
Thanks — round-2 addressed.

- **Apex arm → dropped.** Tightened to subdomain-only (`!host.endsWith('.qurl.site')`); the real host is always `r_<id>.qurl.site`, so the bare apex was dead weight — and dropping it also makes the round-1 apex-coverage note moot (no apex branch to test). Removed the apex test; `*.qurl.site` stays covered by `TUNNEL_TARGET`. 58/58.
- **Fail-closed blast radius — tracked, agreed.** The runbook ([qurl-integrations-infra#1219](https://github.com/layervai/qurl-integrations-infra/pull/1219)) Phase 1 step 6 makes "confirm the observed `target_url` host = `r_<id>.qurl.site`" a **soak gate**, and Phase 0 says to tighten the pin only after that. So the pin can't relied-on in prod before the host is confirmed.
- **Literal vs constant (nit):** leaving `qurl.site` as a literal per CLAUDE.md (you noted it's not a rebrand target); after the apex drop it's a single guard literal, so a constant would be over-centralizing.
- **Trailing-dot FQDN / homograph:** agreed — all fail closed (rejected), not bypasses.

justin-layerv:
Thanks — LGTM noted. Dispositioning the three optional nits without another churn round (all are "leave it"/defensive):

1. **`toLowerCase()` redundant** — correct, `URL.hostname` is already lowercased/punycode for `https:`. Kept purely as defensive intent-documentation; harmless belt-on-belt.
2. **Trailing-dot FQDN** — agreed, it rejects (safe direction). Noted for the planned `r_<id>` regex tighten so a legitimately-rooted observed host isn't false-rejected then.
3. **Scope of the guarantee** — explicit acknowledgement: the suffix pin does **not** defend against an attacker who controls a `*.qurl.site` subdomain; that residual is what the post-soak tighten to `^r_[a-z0-9_-]{11}\.qurl\.site$` shrinks (gated in the runbook, qurl-integrations-infra#1219). `resolve()` remains the trust boundary; this is belt-and-suspenders, as the PR body states.

No code change — these are all conscious/safe as-is. Merge-ready.

PR #839, opened 2026-06-17
fix(discord): dense early-poll so the view counter ticks within seconds, not ~60s

justin-layerv:
## Summary

The sender's `/qurl send` confirmation shows a live "👀 N viewed / M pending" counter — the only feedback a sender gets that the thing they shared was actually opened. Operators reported it stuck at "0 viewed" after a recipient opened the link; confirmed live on 2026-06-17, where it only toggled "after a full minute." For a counter the product surfaces as live, a ~60s lag reads as broken.

This makes the counter tick within a few seconds of a view.

## Root cause

The counter is fed by two mechanisms:
1. A **sub-second SQS view-update push** (fast path), and
2. A **polling fallback** in `monitorLinkStatus` (the correctness primitive).

The push is **best-effort and silent-drops** on the N-1 of N HTTP replicas that don't host the sender's in-memory monitor — this is load-bearing by design (`view-update-registry.js`'s SILENT DROP doc). With 3 replicas, the push renders sub-second only ~1/3 of the time; the rest fall through to the poll. That poll ran on a **flat interval** of `max(15s, min(60s, expiryMs/10))` — which is **60s for any expiry ≥ ~2.5min** (i.e. the common 30m/1h/7d sends). So a view that landed seconds after the send wasn't reflected until the next 60s tick.

Diagnosis was verified end-to-end against the sandbox: the `qurl.accessed` webhook recorded the view, the SQS push published + was consumed (queue metrics: 1 sent / 1 received, DLQ empty), and the monitor's replica stayed alive — the only gap was the poll cadence. The deployed image's view-counter code was confirmed byte-identical to `main`, ruling out a deploy-specific bug.

## Fix

Add a **dense early-poll phase**: poll every 5s for the first 90s of the monitor's life (recipients overwhelmingly open within a minute or two), then fall back to the existing steady interval to keep an idle monitor's DDB cost bounded.

Implementation: replace the two-phase `setTimeout`→`setInterval` scheduler + the `pollCount`-based throttle decay with a **single self-rescheduling `setTimeout`** whose delay is chosen by elapsed time (`nextPollDelay`). This also makes ticks **strictly non-overlapping** (each reschedules only after its own `await` resolves), which lets the old concurrent-tick race — and the defensive comment block guarding it — be deleted outright.

Notes:
- **Polling cadence change only — no extra Discord traffic.** `runTick` edits the message ONLY on a real pending→opened transition (`changed`), and each link flips once, so total edits over the monitor's life are bounded by recipient count (≤50), not poll frequency. At ≥5s spacing this is far under Discord's editReply rate limit — no 429 exposure.
- The push stays best-effort; **no SNS fan-out** (disproportionate for a documented non-load-bearing path; the lead scoped this to latency).
- **Known limitation (unchanged, out of scope):** the in-memory monitor still does not survive a replica restart — a mid-send rolling deploy freezes that send's counter until the link expires.

## Scope

- [x] `apps/discord/`

## Changes

- `apps/discord/src/commands.js` — dense early-poll ramp via a single self-rescheduling timer in `monitorLinkStatus`; retire the `pollCount` decay + two-phase scheduler; `addRecipients` re-enters the dense phase by refreshing `earlyPhaseUntil` (not `startTime`, preserving the 14-min life cap).
- `apps/discord/tests/send-pipeline-back-half.test.js` — regression test (a view recorded seconds after a 30m-expiry send is reflected within seconds, not ~60s; fails on the old flat-interval code), a cadence-shape test (dense early → decay to steady), and a `beforeEach` `getQurlViews` mock-queue reset so the suite isn't coupled to exact tick counts.

## Test Plan

- [x] `npm run lint` clean
- [x] `npm test` green (full discord suite: 81 suites / 2777 tests)
- [x] New regression test verified red on current code, green after the fix
- [ ] Post-deploy: one sandbox tiny-file `/qurl send` → view to confirm the counter ticks within seconds (verification, not diagnosis)

## Related Issues

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1248,7 +1248,7 @@ async function persistDispatchResult(sendId, recipientDiscordId, result) {
 
 // --- Link status monitor ---
 // Track live monitors so a burst of `/qurl send` + `/qurl map` commands
-// can't stack more than MAX_CONCURRENT_MONITORS setIntervals. When we
+// can't stack more than MAX_CONCURRENT_MONITORS poll loops. When we
 // cross the cap, the oldest monitor is stopped to make room (the user
 // can still `/qurl revoke`; they just stop seeing live status updates in
 // the original message).
@@ -1259,14 +1259,14 @@ const activeMonitors = new Set();
 // not via bot-token channel PATCH. Both /qurl send and /qurl map
 // deferReply ephemeral, so the confirmation is unreachable once the
 // token expires. There's no cross-token fallback — we just cap the
-// monitor below the 15-min cliff so we don't waste setIntervals
+// monitor below the 15-min cliff so we don't waste poll loops
 // against a dead token.
 
 function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, expiresIn, baseMsg, buttonRowArg, delivered) {
   // Rebind params as closure-mutable so stop() can null them out for GC.
   // Long-running monitors (up to MAX_MONITOR_DURATION_MS ×
   // MAX_CONCURRENT_MONITORS=50) otherwise pin interaction/recipients/
-  // buttonRow in the setInterval closure.
+  // buttonRow in the poll-loop closure.
   let interaction = interactionArg;
   let qurlLinks = qurlLinksArg;
   let recipients = recipientsArg;
@@ -1332,7 +1332,7 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
     safeEdit,
     onAllDone: () => {
       allDone = true;
-      clearInterval(timer);
+      clearTimeout(timer);
     },
     logger,
   });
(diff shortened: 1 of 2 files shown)


justin-layerv:
Thanks for the careful review — addressed all four in the latest push (rebased onto current main).

**1. Timer orphan when `addRecipients` re-arm races a mid-await tick.** Real race, fixed. Added a `loopGen` generation guard: `arm()` bumps it and the scheduled tick captures it; a tick resuming from `await runTick()` only reschedules if its capture still matches `loopGen`. So a stale tick whose chain was superseded by a re-arm becomes a no-op, keeping the "single timer handle" invariant exact rather than commented-around.

On severity, to be precise: I traced it and the impact is narrower than "pins the closure refs until the cap." `stop()` nulls `interaction`/`linkStatus`/`buttonRow` directly, and the orphaned chain's next `runTick` short-circuits at `isTerminated()`/`!interaction` and does not re-arm — so it self-heals within one tick at `stop()` (or the 14-min cap). The true residual without the guard was **bounded double-polling** (two chains → ~2× BatchGet rate) until stop/cap, with no leak past stop. The guard is cheap insurance that removes even that.

Regression test added (`re-arm racing a mid-await tick keeps a single poll chain (loopGen guard)`): it enables the push, suspends the first tick via a blocking `getQurlViews`, drives `registry.dispatch → onAllDone` + `/qurl add` re-arm while suspended, then measures the post-race poll **rate** (one chain ≈ 3 ticks / 15s vs. ~6 for two chains). Verified red without the guard (6 ticks), green with it (≤4). (A post-stop call-count assertion can't detect this — `isTerminated()` short-circuits the orphan at stop — so the discriminator is poll rate, as you'd expect.)

**2. `/qurl add` latency asymmetry on a still-pending monitor.** Confirmed intended, and expanded the comment to say why: when the loop is still running we deliberately do NOT `clearTimeout`+re-arm — the live tick owns `timer`, and restarting would either fire a redundant tick or (mid-await) need the same `loopGen` guard to avoid the orphan. The running loop picks up the refreshed `earlyPhaseUntil` at its next reschedule, so a `/qurl add` past the 90s window waits at most one in-flight steady interval (≤60s) — an accepted, bounded asymmetry vs. paying double-timer complexity for the rarer pending-add case.

**3. Stale `armLoop` comment reference.** Fixed → `arm()`/`tick()`.

**4. Idle DDB cost (flat steady vs. old every-4th decay).** Conscious sign-off — the cost comment now calls out the few× idle-BatchGet increase explicitly and why it's acceptable (bounded by `MAX_CONCURRENT_MONITORS`, individually cheap, ≤14-min monitor life).

Full discord suite green (81 suites / 2778 tests), lint clean.

justin-layerv:
Cycle 2 nit addressed: swept the four stale `setInterval` references in the monitor header/cap comments (commands.js ~1251, ~1262, ~1269, ~1460) → "poll loop(s)". The remaining `setInterval` mentions are intentional — the historical "self-destruct-era setInterval concurrency race is gone" note, the "never a setInterval" contrast in the scheduler comment, and the unrelated module-level `sweepCooldowns` interval. The steady-phase / idle-BatchGet and `/qurl add` pending-asymmetry items are conscious sign-offs as you noted (no change requested). Full discord suite green, lint clean.

PR #844, opened 2026-06-17
refactor(discord): consolidate the bot onto the @layervai/qurl SDK

### LINKED ISSUE #830: Consolidate the bot to one qURL client: migrate qurlFetch → @layervai/qurl SDK
## Context

[PR #828](https://github.com/layervai/qurl-integrations/pull/828) (detect-over-tunnel) adopted the **`@layervai/qurl` SDK** for the detect `resolve()` call, per a product decision to standardize the Discord bot on the SDK going forward (PR-3 is the first adopter).

The bot also has a hand-rolled fetch client **`qurlFetch`** (`apps/discord/src/qurl.js`). A `/simplify` pass on #828 flagged that the two now coexist — same `/v1` base, `Bearer` auth, 30s timeout, 3-attempt backoff — and that `qurl.js`'s header comment (*"Avoids ESM/CJS compatibility issues with the @layervai/qurl SDK"*) is now stale, since the bot uses the SDK. Source check confirmed the SDK's `resolve()` is wire-identical to a `qurlFetch` POST (the NHP "knock" is server-side — qURL reads the request's TCP source IP).

The decision on #828 was **keep the SDK**. This issue tracks the strategic cleanup it implies: migrate the remaining `qurlFetch` callers to the SDK so the bot has **one** qURL client.

## Scope

- Replace `qurlFetch` callers (`createOneTimeLink`, send/revoke paths, etc. — enumerate from `apps/discord/src/`) with SDK equivalents (`createQurl`/`resolve`/…).
- **Preserve the bot-specific behaviors `qurlFetch` owns** (these are the real risk — do not lose them):
  - the auth-failure audit emit (`AUDIT_EVENTS.DEPENDENCY_AUTH_FAILURE` on 401/403), with the **EMIT-ONCE invariant** (401/403 stay out of the retry set so the metric doesn't multiply per attempt);
  - **error-body redaction** in logs (only status + body length, never the raw body — it may echo tokens);
  - 30s/attempt timeout + 3-attempt backoff (the SDK's `timeout`/`maxRetries` defaults already match; verify on the migrated paths).
  - Wrap the SDK or add a thin adapter if any of these isn't covered natively.
- Remove/replace the now-stale anti-SDK header comment in `qurl.js`.
- The SSRF guards (`isPrivateHost`, `assertNotPrivateAfterResolve`, `assertPublicHttpsTarget`) are client-independent — keep/relocate as-is.

## Acceptance criteria

- Exactly one qURL client remains in the bot (the SDK); `qurlFetch` removed.
- The auth-audit emit, emit-once invariant, and body-redaction behaviors are preserved and still pinned by `tests/qurl-coverage.test.js`.
- Full `apps/discord` suite green.

## Out of scope

- #828 itself stays on the SDK; no behavior change to the detect path.

Kevin-layerV:
## What & why

Closes #830. The Discord bot had **two** qURL clients: the hand-rolled `qurlFetch` in `apps/discord/src/qurl.js` and the `@layervai/qurl` SDK adopted for the detect path in #828. This consolidates onto **one** — the SDK — and deletes `qurlFetch`.

`createOneTimeLink` / `deleteLink` / `getResourceStatus` now call the SDK's `create` / `delete` / `get`. A thin `callQurl()` adapter layers on the bot-specific behaviors the SDK doesn't own, and the client-independent SSRF guards stay put.

## Behaviors preserved

- **`DEPENDENCY_AUTH_FAILURE` audit emit on 401/403**, fired before the throw so a caller's catch can't suppress the alarm.
- **Emit-once invariant** — the SDK keeps 401/403 out of its retryable set (`{429, 502, 503, 504}`), so the audit fires once per request, not per attempt.
- **Error-body redaction, end-to-end.** `callQurl` logs only `status` + error `code`, and — on any real HTTP-status failure — re-throws a **status-only** `Error` (`qURL API <M> <P> failed (<N>)`, the prior wire-contract shape). The SDK's `detail`-bearing message is never propagated, so the revoke caller's `logger.error(..., { error: reason.message })` stays redacted. A test pins that a token planted in the error body reaches neither the log nor the thrown error. (status-0 network/timeout errors carry no server body and propagate unchanged.)
- **Error-message contract** — failures keep the same `qURL API <M> <P> failed (<N>)` message shape, so callers that log it are unaffected.
- **SSRF guards** (`isPrivateHost`, `assertNotPrivateAfterResolve`) and the resource-id charset guard — unchanged, client-independent.
- **Wire contract** — 30s/attempt timeout, **3 total attempts** (`maxRetries: 2`, matching `qurlFetch`'s documented "initial + 2 retries"), and the `qurl-discord-bot/1.0` User-Agent are pinned explicitly so an SDK-default drift can't silently change this path. (Each now has a test.)

## Intentional changes from adopting the SDK

- `getResourceStatus` returns the SDK's `QURL` shape — access tokens under **`access_tokens`** (the SDK renames the API's wire-format `qurls`). No `src/` caller reads this field today.
- **Retry set narrowed to the SDK's policy.** `qurlFetch` retried `{408, 429, 500, 502, 503, 504}` on every method; the SDK retries `{429, 502, 503, 504}` for idempotent GET/DELETE and **only 429** for POST/PATCH. Net effect: POST is no longer retried on 5xx (removes `qurlFetch`'s latent duplicate-create risk), and GET/DELETE no longer retry 408/500 (a 500 is usually non-transient). All three behaviors are pinned by tests.
- `delete` now requires an `r_`-prefixed resource ID (the qurl-service contract). The revoke path stores exactly that id, so it's satisfied; malformed ids fail client-side instead of round-tripping to a server 4xx.
- A GET that returns `204` now throws (a status read expects a body) instead of returning `null` — an unrealistic case with no `src/` caller.
- **Retry observability (accepted trade-off).** `qurlFetch` logged a `warn` on each transient retry; retries are now internal to the SDK, which exposes no per-retry hook (only a general `debug` callback), so that per-attempt breadcrumb is gone. The terminal-failure breadcrumb is retained (`logger.debug('qURL API error', { status, code })`). Net: less routine `warn`-noise on transient blips that succeed, at the cost of per-attempt retry visibility. The shared-client-factory follow-up could wire the SDK's `debug` callback if per-attempt telemetry is wanted back.

## Test plan

- Full `apps/discord` suite green: **2812 passed**, eslint clean, `qurl.js` coverage 82.4/76.2/100/84.4 (above the 78/68/78/78 floors).
- `qurl-coverage.test.js` keeps the audit / emit-once / no-emit-on-non-auth tests and adds: end-to-end redaction (log **and** thrown error), a User-Agent assertion, POST-not-retried-on-503, and DELETE-retries-on-503. Fetch doubles were upgraded to SDK-parseable shape (the SDK reads `.json()` for RFC-7807 envelopes and `.headers.get()` for Retry-After).
- `/simplify` reviewed (reuse / simplification / efficiency / altitude); review feedback on the redaction seam folded in.

## Out of scope (follow-ups, not in this PR)

This PR is scoped to `qurl.js` per the issue (the detect path stays as-is). Two adjacent items surfaced and are intentionally **not** touched here:

1. **Latent bug in `connector.js` (from #828):** it imports `{ QurlClient }`, but the SDK only exports `QURLClient` — so `new QurlClient(...)` would throw `is not a constructor` the first time the detect path constructs its client (the test mock masks it). Fixed in the fast-follow **#845**, which also de-rots the now-stale `qurlFetch` comments in `connector.js` and adds a no-mock SDK-export contract test.
2. A shared `callQurl` + client-options factory between `qurl.js` and `connector.js` would give the resolve leg the same audit-emit + redaction and remove the duplicated client construction / `validateResourceId` regex.

### DIFF
--- a/apps/discord/src/qurl.js
+++ b/apps/discord/src/qurl.js
@@ -1,4 +1,9 @@
-const { QURLClient } = require('@layervai/qurl');
+const {
+  QURLClient,
+  ERROR_CODE_NETWORK,
+  ERROR_CODE_TIMEOUT,
+  ERROR_CODE_CLIENT_VALIDATION,
+} = require('@layervai/qurl');
 const config = require('./config');
 const logger = require('./logger');
 const { AUDIT_EVENTS } = require('./constants');
@@ -28,6 +33,16 @@ const MAX_RETRIES = 2;
 // User-Agent the qURL service sees for the bot's calls. Preserved verbatim
 // across the SDK migration (a literal wire identifier — see CLAUDE.md).
 const USER_AGENT = 'qurl-discord-bot/1.0';
+
+// status-0 SDK error codes whose message the SDK synthesizes itself (no server
+// body) — the only status-0 errors callQurl surfaces verbatim. See its
+// REDACTION note: anything else at status 0 is re-wrapped, so the no-body-leak
+// invariant holds structurally rather than by trusting SDK internals.
+const SAFE_STATUS0_CODES = new Set([
+  ERROR_CODE_NETWORK,
+  ERROR_CODE_TIMEOUT,
+  ERROR_CODE_CLIENT_VALIDATION,
+]);
 
 // Construct a per-call SDK client. Per-call (not cached) because each call
 // carries its own apiKey (the bot is multi-tenant) and because these are rare
@@ -65,11 +80,13 @@ function makeClient(apiKey) {
  *     detail`, where `detail` is parsed from the server body (which can echo
  *     request headers or tokens). So for any positive status we log only status
  *     + code and re-throw a status-only Error (callers such as the revoke path
- *     log the thrown `.message`, so the body must not reach it). status-0 errors
- *     propagate unchanged: the SDK uses status 0 only for network / timeout /
- *     client-validation / unexpected-shape errors, whose messages it synthesizes
- *     itself (never from a server body), so they're safe to surface and more
- *     useful than a generic string. Pinned by tests/qurl-coverage.test.js.
+ *     log the thrown `.message`, so the body must not reach it). At status 0,
+ *     only SDK-synthesized, body-free errors (network / timeout / client-
+ *     validation — see SAFE_STATUS0_CODES) propagate verbatim; any other
+ *     status-0 error (e.g. an unexpected-response shape error that could embed a
+ *     body snippet) is re-wrapped to a code-only message, so the invariant holds
+ *     structurally rather than by trusting SDK internals. Pinned by
+ *     tests/qurl-coverage.test.js.
  */
 async function callQurl(method, path, fn) {
   try {
@@ -90,11 +107,17 @@ async function callQurl(method, path, fn) {
     }
     // A real HTTP status means the SDK error wraps a server response body — throw
     // a status-only error so that body can't leak through a caller that logs
-    // `err.message`. status 0 (no server body) propagates unchanged.
+    // `err.message`.
     if (status > 0) {
       throw new Error(`qURL API ${method} ${path} failed (${status})`);
     }
-    throw err;
+    // status 0: surface only SDK-synthesized, body-free errors verbatim;
+    // re-wrap anything else to a code-only message (defense-in-depth — the SDK
+    // doesn't embed bodies in status-0 messages today, but we don't rely on it).
+    if (err && SAFE_STATUS0_CODES.has(err.code)) {
+      throw err;
+    }
+    throw new Error(`qURL API ${method} ${path} failed (${err?.code || 'unknown error'})`);
   }
 }
 
--- a/apps/discord/src/qurl.js
+++ b/apps/discord/src/qurl.js
@@ -60,14 +60,16 @@ function makeClient(apiKey) {
  *     {429, 502, 503, 504}), so this fires once per request, not once per
  *     attempt. If that ever changes, the audit count would multiply on a single
  *     auth failure. Pinned by tests/qurl-coverage.test.js.
- *   - REDACTION: never let a qURL error body escape this module. The SDK's
- *     QURLError.message is `Title (status): detail`, and `detail` can echo
- *     request headers or tokens — so on a real HTTP-status failure we log only
- *     status + code and re-throw a status-only Error (callers such as the revoke
- *     path log the thrown `.message`, so the body must not reach it). status-0
- *     errors (network / timeout / client-validation) carry no server body, so
- *     they propagate unchanged — their message is transport/SDK text and is
- *     useful. Pinned by tests/qurl-coverage.test.js.
+ *   - REDACTION: never let a qURL error body escape this module. On an
+ *     HTTP-status failure the SDK's `QURLError.message` is `Title (status):
+ *     detail`, where `detail` is parsed from the server body (which can echo
+ *     request headers or tokens). So for any positive status we log only status
+ *     + code and re-throw a status-only Error (callers such as the revoke path
+ *     log the thrown `.message`, so the body must not reach it). status-0 errors
+ *     propagate unchanged: the SDK uses status 0 only for network / timeout /
+ *     client-validation / unexpected-shape errors, whose messages it synthesizes
+ *     itself (never from a server body), so they're safe to surface and more
+ *     useful than a generic string. Pinned by tests/qurl-coverage.test.js.
  */
 async function callQurl(method, path, fn) {
   try {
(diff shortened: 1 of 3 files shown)


Kevin-layerV:
Thanks for the thorough passes. Round-2 items landed in `edf97c7c` (GET 408/500 narrowing documented + pinned by a test, status-0 docblock tightened, 204 test decoupled from SDK wording, `err?.status`). On the round-3 notes:

- **status-0 "residual leak":** verified non-reachable under the pinned SDK. The non-JSON-body-on-2xx path throws with `status = response.status` (>0 → sanitized to a status-only error), and its `detail` is a fixed `"Expected JSON response body on HTTP <n>…"` string, never the body. The only status-0 `unexpected_response` (a 204 on GET) is the SDK-synthesized `"Unexpected 204 No Content…"` string — also body-free. So every status-0 message is SDK-authored, never a server body; the docblock now says exactly that.
- **`err.status` / `userAgent` assumptions:** validated by the suite, which drives the *real* pinned SDK (these specs don't module-mock `@layervai/qurl`) — the redaction and UA assertions fail loudly if either name drifts.
- **`connector.js` `{ QurlClient }` bug + the now-dangling `qurlFetch` comments:** both go in the fast-follow `fix(discord)` PR (correct import + its test mock + comment cleanup), kept out of this refactor so the crash-fix lands as its own changelog entry.
- **per-attempt retry log / `~0.2.0` patch range:** accepted — retries are the SDK's concern now (the final-failure breadcrumb remains), and the new retry-policy tests are the guard against a patch-level set change.

Kevin-layerV:
Thanks — all four addressed:

- **Medium (status-0 redaction rests on an SDK assumption):** fixed structurally in `51175402`. `callQurl` now propagates a status-0 SDK error verbatim **only** when its `code` ∈ `{network_error, timeout, client_validation}` (`SAFE_STATUS0_CODES`); any other status-0 error is re-wrapped to a code-only message. So the no-body-leak invariant no longer depends on the SDK never embedding a body snippet in a status-0 message. The 204 path (status 0, `unexpected_response`) now pins the **re-wrap** branch and asserts the SDK's own text doesn't escape; the network-error test pins the **verbatim** branch.
- **Low (retry observability):** to be precise — the *final-failure* breadcrumb was `logger.debug` in the pre-SDK client too (the `!resp.ok` path logged at debug before throwing), so its level is unchanged. What's gone is the *intermediate* per-retry `warn`, now that retries are SDK-internal. I judged restoring it not worth coupling `qurl.js` to the SDK's debug-message format (the SDK only exposes a general `debug` callback, not a retry-specific hook); the failure signal itself is intact. Easy to revisit in the shared-client-factory follow-up if the team wants per-attempt telemetry.
- **Low/info (no `src/` consumer for `createOneTimeLink`/`getResourceStatus`):** agreed and already called out — the `access_tokens` rename and 204-throws are mock-validated only; a future real caller would be the first prod exercise. `deleteLink` (the one prod consumer) is fully covered.
- **Nit (persistent-503 test cost):** checked — full `apps/discord` suite runs in ~11s; the real backoff sleeps across the 503/429/DELETE-retry tests add ~2–3s total, which isn't moving the needle, so I left them on real timers rather than add fake-timer scaffolding.

Kevin-layerV:
Addressed the observability point in `fc80252b`: the status-0 re-wrap now fires **only** for a coded SDK error outside the safe set (the body-snippet risk). Body-free SDK errors and non-SDK throws (a programming `TypeError`) now propagate verbatim, so their message + stack survive — without re-introducing a body-leak surface via `err.cause`.

The connector.js `QurlClient` → `QURLClient` fast-follow is already up as #845 (independent of this PR, CI green, includes a no-mock contract test so it can't regress silently again).

Kevin-layerV:
All addressed:

1. **Retry observability (medium):** the SDK exposes no per-retry/`onRetry` hook (only a general `debug` callback), so I've called this out as an **accepted trade-off in the PR body** rather than wire the general callback (which would only restore visibility at `debug` level, not the old `warn`). Net: the terminal-failure breadcrumb is retained; we lose per-attempt `warn` on transient blips that succeed — arguably less routine `warn`-noise. The shared-client-factory follow-up can wire the `debug` callback if the team wants per-attempt telemetry back.
2. **Stale `qurlFetch` comments:** fixed in the fast-follow **#845**, which reworks `connector.js:433/443` to be client-agnostic (alongside the `QURLClient` import fix). Left out of this PR so the detect path stays untouched here.
3. **`get()` vs the `r_` guard:** deliberate. `get()` doesn't require an `r_` prefix (only `delete()` does), so `getResourceStatus` keeps just the charset guard; a non-`r_` id would surface as a `client_validation` (status 0, in `SAFE_STATUS0_CODES`) and propagate — fine, and there are no `src/` callers regardless.
4. **status-0 no-`code` residual:** a status-0 error with no `.code` is a **non-SDK throw** (e.g. a programming `TypeError`) — the SDK always sets a `.code` (`createError` defaults to `unknown`). Those carry no server body, so propagating them verbatim is both safe and what preserves the stack you flagged in the last round. The re-wrap deliberately targets only *coded* SDK errors.
5. **Retry-policy tests pin SDK internals:** intended — they're drift-detectors for the SDK's retry classification, exactly so a patch-level change to the retry set surfaces in CI. Will go red on an SDK retry-policy change by design.

PR #846, opened 2026-06-17
fix(discord): block add recipients during revoke

### LINKED ISSUE #199: discord: Revoke and Add-Recipients are not mutually exclusive

Kevin-layerV:
## Summary
- Cross-block `/qurl send` management actions so Revoke and Add Recipients cannot run concurrently for the same send.
- Reject Add Recipients while revoke is in flight/terminal, and reject Revoke while an Add Recipients flow holds the per-send lock.
- Add a regression test that fires Revoke and then races Add Recipients on the same collector, proving no post-revoke mint happens and the revoke summary stays scoped to the original recipient.

Fixes #199.

## Validation
- `npx jest --runInBand tests/send-pipeline-back-half.test.js --coverage=false`
- `npm run lint` from `apps/discord`
- `npm test -- --runInBand` from `apps/discord` (2,810 tests passed)
- `git diff --check`

## Notes
- `go test ./...` is polluted by this workstation's persisted qURL CLI config: `apps/cli/cmd` missing-API-key tests see a local API key and fail. Non-CLI packages pass; this PR only changes Discord JS code/tests.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -561,6 +561,13 @@ function isAllowedFileType(contentType) {
 // collector instances (e.g. flow_state RESUME on a bot restart loading
 // unfinished sends).
 const addRecipientsLocks = new Set();
+// Same-process per-send Revoke lock. Collector-local `revokeInFlight`
+// handles duplicate clicks inside one management collector; this Set lets
+// another collector in the same process see a Revoke already mutating the
+// send. Cross-process safety still relies on revoked_at/#862.
+const revokingSendLocks = new Set();
+const ADD_RECIPIENTS_IN_PROGRESS_MSG = 'Already processing an "Add Recipients" action.';
+const ALREADY_REVOKING_SEND_MSG = 'Already revoking links for this send.';
 
 const sendCooldowns = new Map();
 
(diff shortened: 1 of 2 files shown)


Kevin-layerV:
Follow-up from the latest `/simplify`: the pre-existing recovery gap where `revoked_at` is set but qURL DELETEs fail is tracked in #876. This PR keeps Add Recipients terminal in that state to close #199; #876 covers re-driving/recovering the failed DELETEs without reopening Add.

Issue #179, opened 2026-05-04
Discord OAuth Stage-2: signed install state + admin-change alarm (cross-repo)


vikramlayerv:
Split out from #178 as the cross-repo subset that PR #177 couldn't land in-place. Both items mitigate the Stage-2 confused-deputy attack class (attacker pre-runs Discord install in their browser, forwards the chained Auth0-redirect URL to a victim, victim's qURL account binds to attacker's guild). PR #177 ships the user-visible mitigation (success-page binding readout, JWKS-verified email); these are the technical mitigations layered on top.

## A.1 Signed state on the layerv.ai install link
Coordinate with the layerv.ai marketing/onboarding page so the "Add to Discord" link carries a short-lived signed state that `/oauth/discord/callback` validates (rejects unverified states, narrows the attack window to the link-mint TTL).

- layerv.ai mints state via shared-secret HMAC (or a tiny endpoint that returns a signed token)
- Stage-2 callback adds: `if (state && !verifyMarketingState(state)) reject`
- Backwards-compat: accept missing state for the rollout window, then enforce after marketing flips

**Owner:** marketing-page repo + qurl-integrations
**Cost:** ~1 day

## A.2 Observability metric/alarm
Until A.1 lands (or alongside it), add a CloudWatch metric on Stage-2 callbacks where the resolved `discord_user_id` doesn't match the prior owner of the `guild_id` (queried from `guild_configs.configured_by`). Anomaly = re-bind from a different admin = potential confused-deputy attempt.

- Metric: `qurl_setup_admin_changed{env}`
- Alarm: > 1/min sustained → page on-call
- Logs the (old_admin_id, new_admin_id, guild_id) tuple for triage

**Owner:** qurl-integrations + qurl-integrations-infra (Terraform)
**Cost:** ~half day

## Context
- Originally tracked in #178; closed when PR #177 landed the in-process subset (B + C.1–C.8).
- Stage-2 CSRF posture comment (`LOAD-BEARING — do not remove without replacing`) lives in `apps/discord/src/routes/discord-install.js`.

PR #850, opened 2026-06-17
fix(discord): verify signed install state

Kevin-layerV:
## Summary
- verify signed layerv.ai Discord install `state` values when present, while accepting missing state during rollout
- add `DISCORD_INSTALL_STATE_REQUIRED` so missing state can be rejected after signed links are fully rolled out
- document and trim `DISCORD_INSTALL_STATE_SECRET` for the shared marketing/bot HMAC contract
- emit `qurl_setup_admin_changed` audit events when Discord setup rebinds an existing guild to a different admin

## Why
Issue #179 tracks defense-in-depth for the Stage-2 confused-deputy flow. This adds the bot-side verification and the log signal consumed by the companion qurl-integrations-infra alarm PR.

## Rollout
Set `DISCORD_INSTALL_STATE_SECRET` in the Discord bot environment before marketing starts sending install links with signed `state`. Missing `state` stays accepted while `DISCORD_INSTALL_STATE_REQUIRED=false`; once every layerv.ai install link sends signed state, flip `DISCORD_INSTALL_STATE_REQUIRED=true` to reject callbacks that omit state.

## Validation
- `npm run lint` in `apps/discord`
- `npx jest --coverage=false --runTestsByPath tests/marketing-install-state.test.js tests/discord-install.test.js tests/config-int-env.test.js tests/boot-requirements.test.js tests/ddb-store.test.js --runInBand` in `apps/discord`
- `npm test -- --runInBand` in `apps/discord`

Refs #179.

### DIFF
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -253,10 +253,11 @@ OAUTH_STATE_SECRET=
 # but rejects a provided state unless this secret verifies it. Use the
 # same 64-character hex secret string on the marketing page and the bot;
 # both sides use the hex string itself as the raw HMAC key (do not
-# hex-decode before signing). Rollout order is strict: set this on the
-# bot before the marketing page starts sending state-bearing links, then
-# flip DISCORD_INSTALL_STATE_REQUIRED. State-bearing links fail closed
-# with 400 until this secret is deployed.
+# hex-decode before signing). The state signature segment is lowercase
+# SHA-256 hex, matching Node's digest('hex') output. Rollout order is
+# strict: set this on the bot before the marketing page starts sending
+# state-bearing links, then flip DISCORD_INSTALL_STATE_REQUIRED.
+# State-bearing links fail closed with 400 until this secret is deployed.
 DISCORD_INSTALL_STATE_SECRET=
 # After every layerv.ai install link sends signed state, set this to the
 # literal string "true" to reject callbacks that omit state. Keep false
--- a/apps/discord/.env.example
+++ b/apps/discord/.env.example
@@ -248,6 +248,21 @@ GITHUB_WEBHOOK_SECRET=your_webhook_secret
 #   openssl rand -hex 32
 OAUTH_STATE_SECRET=
 
+# Shared HMAC key for signed Discord install links minted by layerv.ai.
+# Optional during rollout: /oauth/discord/callback accepts missing state,
+# but rejects a provided state unless this secret verifies it. Use the
+# same 64-character hex secret string on the marketing page and the bot;
+# both sides use the hex string itself as the raw HMAC key (do not
+# hex-decode before signing). Rollout order is strict: set this on the
+# bot before the marketing page starts sending state-bearing links, then
+# flip DISCORD_INSTALL_STATE_REQUIRED. State-bearing links fail closed
+# with 400 until this secret is deployed.
+DISCORD_INSTALL_STATE_SECRET=
+# After every layerv.ai install link sends signed state, set this to the
+# literal string "true" to reject callbacks that omit state. Keep false
+# until the bot secret is deployed and marketing has fully flipped links.
+DISCORD_INSTALL_STATE_REQUIRED=false
+
 # Allowed GitHub organizations for contributor tracking (comma-separated)
 ALLOWED_GITHUB_ORGS=your_github_org
 
--- a/apps/discord/README.md
+++ b/apps/discord/README.md
@@ -93,7 +93,7 @@ setup) means required to use that feature.
 | `QURL_ENDPOINT` | No | qURL API base URL (defaults to production; localhost in dev) |
 | `CONNECTOR_URL` | No | qURL connector URL for file upload + serving |
 | `BASE_URL` | OAuth setup | Public `https://` origin of the bot; required to complete the OAuth `/qurl setup` flow (defaults to `http://localhost:3000`). |
-| `DISCORD_INSTALL_STATE_SECRET` | Discord install state | Shared 64-character hex HMAC secret for signed layerv.ai "Add to Discord" links; the bot and marketing both use the hex string as the raw HMAC key. Set on the bot before marketing flips the link to include `state`. |
+| `DISCORD_INSTALL_STATE_SECRET` | Discord install state | Shared 64-character hex HMAC secret for signed layerv.ai "Add to Discord" links; the bot and marketing both use the hex string as the raw HMAC key, and the state signature segment is lowercase SHA-256 hex. Set on the bot before marketing flips the link to include `state`. |
 | `DISCORD_INSTALL_STATE_REQUIRED` | Discord install state | Set to `true` only after the bot secret is deployed and all layerv.ai install links send signed `state`; then callbacks missing `state` are rejected. |
 | `KEY_ENCRYPTION_KEY` | Production | 32 random bytes, base64 — encrypts stored keys at rest |
 | `METRICS_TOKEN` | Production | Bearer token guarding the `/metrics` endpoint |
--- a/apps/discord/README.md
+++ b/apps/discord/README.md
@@ -93,12 +93,20 @@ setup) means required to use that feature.
 | `QURL_ENDPOINT` | No | qURL API base URL (defaults to production; localhost in dev) |
 | `CONNECTOR_URL` | No | qURL connector URL for file upload + serving |
 | `BASE_URL` | OAuth setup | Public `https://` origin of the bot; required to complete the OAuth `/qurl setup` flow (defaults to `http://localhost:3000`). |
+| `DISCORD_INSTALL_STATE_SECRET` | Discord install state | Shared 64-character hex HMAC secret for signed layerv.ai "Add to Discord" links; the bot and marketing both use the hex string as the raw HMAC key. Set on the bot before marketing flips the link to include `state`. |
+| `DISCORD_INSTALL_STATE_REQUIRED` | Discord install state | Set to `true` only after the bot secret is deployed and all layerv.ai install links send signed `state`; then callbacks missing `state` are rejected. |
 | `KEY_ENCRYPTION_KEY` | Production | 32 random bytes, base64 — encrypts stored keys at rest |
 | `METRICS_TOKEN` | Production | Bearer token guarding the `/metrics` endpoint |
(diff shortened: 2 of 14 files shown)


Kevin-layerV:
Follow-up for the non-blocking replay/guild-binding hardening note is tracked in #867. Keeping this PR scoped to #179's stated TTL-narrowing + rollout-enforcement objective.

Kevin-layerV:
Follow-up on the latest Claude notes:

- Marketing-side minter wire contract is now tracked in layervai/website#561: raw 64-char hex string as HMAC key, lowercase SHA-256 hex signature segment, per-render short-TTL state.
- Replay / nonce-cache hardening remains tracked in #867.
- The bot-side docs/comments were updated in this PR for both the lowercase signature contract and the `UPDATED_OLD` audit path handling of old encrypted `qurl_api_key`.

Kevin-layerV:
## Review disposition (178abf58 → 1a29b5c0)

Branch is merged with current `main` (8d0dcfad, clean merge; PR diff stays the same three Discord files).

| # | Finding | Disposition |
|---|---------|-------------|
| 1 | Delete-then-setup rebind is silent end-to-end | Follow-up, tracked in #1455. One correction to the premise: `_removeGuildApiKeyRaw` has no production caller today (its header requires a webhook-teardown orchestrator before any `/qurl unlink` ships), so the one-command bypass does not exist yet. The gap becomes live the moment a delete command lands; #1455 asks for a `config_deleted` audit event to ride in with it, plus the legacy-row caveat and a rollout-ledger note. |
| 2 | Constant doc narrows the event to OAuth | Fixed. Comment now says the event is emitted by `setGuildApiKey` for any successful guild setup, OAuth callback or `/qurl setup` paste. Did not thread a `via` field: that widens the store signature and both call sites for a dimension on-call can recover from the adjacent `descriptionContext` log line. |
| 3 | `res.Attributes` unguarded on `res` | Fixed with `res?.Attributes?.configured_by`. The "audit must not fail the write" half is already true: `logger.audit` coerces non-object meta and wraps serialization in try/catch, so nothing in the audit block after the guard can throw. |
| 4 | Negative test assertions weaker than the behavior | Fixed. The first-setup/same-admin test now asserts on the emitted event names (`mock.calls.map(([event]) => event)`), which also catches a meta-less emit. Added `does not audit when the write rejects`. |
| 5 | `fix` vs `feat` commit type | Reply, keeping `fix`. The infra filter/alarm (#1224) merged first and has been a silent no-op; this restores the behavior it already expects, which is a patch-track correction rather than a new capability. The changelog entry lands under Bug Fixes with the event name, so correlating the alarm to a deployed version still works. |

**Validation at 1a29b5c0:** `npm run lint` clean; `npx jest` 93 suites / 3571 tests passed (+1 test).

Kevin-layerV:
Final convergence round at `51b5b15c`.

| # | Finding | Disposition |
|---|---|---|
| 1 | Rebind on a row with no `configured_by` is silent | **Fixed.** The guard now keys on the prior `qurl_api_key` (present in the same `UPDATED_OLD` response) and reports `old_admin_id: null`. Test added. `logger.audit` passes `null` through `JSON.stringify` unchanged, and the infra filter keys on `event` + `agent`, so the payload shape is stable. |
| 2 | Delete-then-setup evades the audit | **PR body.** Stated plainly under "Known gap"; the companion delete audit is #1455. No code change here. |
| 3 | Every `setGuildApiKey` caller passes the acting human's ID | **Verified.** Two real callers: `commands.js:4010` passes `interaction.user.id` (`/qurl setup` modal) and `routes/qurl-oauth.js:441` passes `discordUserId` from the verified setup-session payload. `scripts/provision-guild-subscriptions.js` and the other hits only mention it in comments; no backfill, provisioning, or rollback path writes the key. Stated in the PR body. |
| 4 | Type strictness on the comparison | **Fixed.** `String()` on both sides, one line. |
| note | Same-admin fixture returns `guild_id` under `UPDATED_OLD` | **Fixed.** The fixture now returns only touched non-key attributes. |
| note | `TODO(upstream-contract)` on only one event | **No change.** As noted, not worth changing in this PR. |
| note | `fix` vs `feat` commit type | **No change.** Kept `fix`: it closes an unaudited security path (#179) rather than adding user-facing capability, so a patch bump is intended. |
| note | `logger.audit` safety, cardinality guidance | No change requested, none made. |

Counts: 3 fixed, 1 PR-body change, 1 verified, 3 replied, 0 open. The CI checkbox stays unchecked for CI to decide. `npm run lint` clean; full Jest 93 suites / 3572 tests green at `51b5b15c`.

Kevin-layerV:
Review convergence: the reviewer's latest pass reports no blocking correctness defect at the current head. The remaining items (follow-ups, activation-time checks, nits) are tracked in https://github.com/layervai/qurl-integrations/issues/1456 so they are not lost at merge; this PR is ready for human review.

Kevin-layerV:
Review disposition for 7c0cc5e, applied in the next push:
- #1: `prior_had_key` added.
- #2: `UNKNOWN` is no longer accepted as an input door.
- #3: the wrapper test has its own describe, an omitted-door case and a settled chain.

Not changed:
- #4: an options object for two callers. The enum plus the warning is the chosen guard.
- The suggested `qurl-oauth.test.js` "audit throws, no revoke" integration case: that suite mocks the store module, so `setGuildApiKey` never reaches the audit there. The invariant is pinned at the store level, which is the only place that can throw after a landed write.

Local `git grep -n 'setGuildApiKey(' apps/discord/src apps/discord/scripts` confirms exactly two callers (`commands.js`, `routes/qurl-oauth.js`) plus the definition.

Kevin-layerV:
Review disposition for 209a9dd (the reviewer's verdict: ready to land), applied in the next push:
- #3: the echoed door is truncated to 64 characters.
- #5: presence semantics (`'qurl_api_key' in prior`) now match the `attribute_exists` rollout query.

Not changed:
- #1: an options object or AST call-site guard. `git grep -n 'setGuildApiKey(' apps/discord` shows exactly two callers plus the definition, and each caller test pins its fourth argument.
- #2: the warning stays in the store so `constants.js` keeps no logger dependency. The wrapper's two callers go through the store first.
- #4: unreachable. An undefined `configuredBy` fails DocumentClient marshalling of `:b` before the audit runs.
- #6: the name is left as is.

Kevin-layerV:
Review disposition for 709d71a, applied in 8284ea7:
- #1: the warning text is now "any admin-change audit for this write records via=unknown".
- #2: the null-to-null damaged-row case is listed as the third blind spot.
- #3: the post-write logic is extracted into `auditSetupAdminChange`.

#4 (inner catch) is kept; an earlier review requested it so the store doesn't rely on the logger's no-throw contract.

Kevin-layerV:
Review disposition for 8284ea7, applied in 7ec98fc:
- #1: the rollout gate now includes "same-admin re-key in sandbox emits nothing".
- #2: the null-admin test is renamed to say what it pins (unreachable against real DynamoDB).
- #3: the non-auditing-door extension point is named at the constant.
- #4: a single `bestEffortLog` helper.
- #5: a shared `DOOR_WARN` test constant.
- Coverage: an injection-shaped door collapses to `unknown` in both the wrapper and `normalizeSetupVia` tests.

Not changed: #6 (warning in the wrapper) and `mockReset`, per earlier dispositions. The logger mock factory is a bare `jest.fn()` set, so nothing is lost.

Kevin-layerV:
Review disposition for 7ec98fc, applied in ad13cdc:
- #2: unrecognized doors are logged through `describeSetupVia` (only short slug-shaped values are echoed; anything else becomes `[unrecognized]` with its type). A test pins that a key-shaped misplaced argument never reaches logs.
- #3: admin IDs are compared as strings, with a test for a numeric ID.
- #4: the wrapper warns on an unrecognized door too, and its comment says not to await it in production.
- #5: the audit helper uses an early return.
- #6: the rollout gate now says benign handoffs page too and how to triage them.

Deferred: #1, a non-paging key-rotation audit for same-admin re-keys. It is a new detection feature on the same code path as the delete audit, so I added it to #1455. The separate `SETUP_VIA_UNKNOWN` export is not changed; the sentinel is documented and tested.

Kevin-layerV:
Review disposition for ad13cdc, applied in b27cc35: #1 (echo allowlist is letter-led, max 16 chars; a hex secret is not echoed), #3 (`bestEffortLog` comment), #4 (alarm scope excludes same-admin key replacement, pointer to #1455), #5 (exactly-one warning and no warning for known doors). #2: the second warning is intentional and now commented; it covers future link-only callers.

Kevin-layerV:
Review disposition for b27cc35 (verdict: sound, no defects): #5 fixed in the next push (test spacing, hoisted require). #1 and #3 were settled in earlier rounds (the second warning is intentional and commented; positional args stay with two callers). #2: the 16-character cap is the deliberate trade against echoing secrets. #4 is tracked in #1455.

Kevin-layerV:
Review disposition for afe6c75, applied in the next push: #1 (the whole `auditSetupAdminChange` call is best-effort, with a `configuredBy: Object.create(null)` test) and the wrapper-warning-throws test. #2–#4 are non-blocking and folded into #1455's follow-up scope as suggested. The `AUDIT_SECRET_KEYS` drift guard (`every AUDIT_SECRET_KEYS entry is also redacted by info()`) passes with the new member, because `api_key` substring-redacts `qurl_api_key` on the info path.

Kevin-layerV:
Review disposition for 3403bdc (verdict: solid), applied in the next push: #1 (the whole audit helper body is guarded, so failures log `Failed to emit…` instead of vanishing; the test asserts the error line) and #2 (`prior_updated_at` semantics corrected in the code comment and the rollout triage text).

Not changed: #3, a DynamoDB Local integration test. The Discord Jest suite has no DDB Local lane, and adding one for a single assertion is out of scope. The sandbox rebind and same-admin gates settle it against real DynamoDB before promotion. #4–#6 were settled in earlier dispositions.

Kevin-layerV:
Review disposition for b4479e6 (verdict: no correctness bugs), applied in 1085cbf: #1 (the door echo keeps case, max 16 chars), #2 (the helper receives only non-secret fields plus a key-presence flag), #3 (redundant outer guard dropped; the helper never throws), #4 (an empty `configured_by` reports `null`). #5/#6 were settled earlier.

PR #855, opened 2026-06-17
feat(discord): cross-replica sub-second view counter via portable interaction-token edit

justin-layerv:
## Summary

The sender's `/qurl send` confirmation is the only sender-visible signal that recipients opened what was shared. This PR upgrades that counter from poll-paced updates to sub-second webhook-driven updates for all sends, including high fan-out sends, without adding SNS/SQS fan-out or new public surface.

## Architecture

- Any HTTP replica can edit the sender confirmation through the persisted Discord interaction webhook token: `PATCH /webhooks/{app}/{token}/messages/@original`.
- The interaction token is envelope-encrypted at rest, scrubbed from errors/logs, stripped from generic send-config reads, and treated as absent after `confirm_expires_at`.
- First distinct views are detected from `recordQurlView(...)` using DynamoDB `ALL_OLD`, so only the first access for each qurl id increments the sender aggregate.
- The sender aggregate is stored as synthetic sharded rows in `qurl_views`, scaled by `expectedCount` up to 64 shards. This avoids a hot `qurl_send_configs` row at 20k-recipient fan-out.
- The fast path uses the same pure `renderViewCounter` body as the monitor and edits content-only so Add/Revoke buttons are preserved.
- `last_rendered_count` advances only after a confirmed Discord edit. Failed edits leave the floor untouched and stamp only the debounce clock, so the poll backstop can self-heal instead of getting stuck.

## Edit Coalescing

`QURL_VIEW_COUNTER_COALESCE_MS` defaults to `900`; higher overrides reject back to the default. The path now has two debounce layers:

- A local/render-state clock skips same-replica bursts before the expensive count read.
- A DDB-conditional `tryClaimRenderAttempt` claim runs before the shard-sum read/PATCH, so only one replica per send/window performs the count read and Discord edit.

Views inside the window still increment their shard and schedule a trailing flush. The trailing flush renders the settled count through the webhook path inside the sub-second window rather than waiting for the 15s poll.

## Cost Surface

Coalescing makes Discord edits and shard-sum reads flat per send/window, but the webhook model still has per-first-view work before coalescing: `findSendsByQurlId(qurl_id)` GSI Query, render-state read/cache participation, and the sharded aggregate Update. For a 20k-recipient burst that is intentionally O(views) for lookup/accounting and O(windows) for expensive reads/edits. #888 tracks live metrics and cost-owner sign-off for the residual GSI/read path before raising the fan-out envelope.

## Security

`interaction_token` is a live ~15-minute Discord bearer credential. This PR encrypts it before writing to `qurl_send_configs`, scrubs raw and percent-encoded token forms from network errors, adds it to exact-match audit redaction, strips it from `getSendConfig`, and keeps a single production caller of decrypted render state. Physical DDB TTL for `confirm_expires_at` is tracked in qurl-integrations-infra#1227; the bot is correct without it because use is guarded by the expiry check.

## Reviewer Follow-Ups

Tracked follow-ups:

- layervai/qurl-integrations-infra#1227: enable DDB TTL on `confirm_expires_at` for physical token reaping.
- layervai/qurl-integrations#875: rip out dead SQS view-update wiring and rename `confirm_terminal` to `confirm_fast_path_off`.
- layervai/qurl-integrations#887: migrate/document the monitor poll source-of-truth versus sharded aggregate relationship.
- layervai/qurl-integrations#888: validate max-fanout DDB cost and decide whether a qurl_id-to-send cache/denormalization is worth adding.
- layervai/qurl-integrations-infra#1234: unblock sandbox deploys after a canceled run left a stale Terraform state lock and the cleanup action failed to load.
- layervai/qurl-integrations#889: restore Claude Code `/simplify` issue-comment workflow after repeat SDK failures before review output.

Addressed in this PR:

- Cross-replica sender counter fast path with content-only Discord edit.
- Distributed edit/read coalescing via `tryClaimRenderAttempt`.
- Hot send-row write avoidance via fanout-scaled sharded aggregate rows.
- Commit-after-edit monotonic floor shared by fast path and monitor poll.
- Failure-path coalescing without advancing the displayed floor.
- Token-at-rest encryption, token log/error redaction, and render-state expiry self-defense.
- Store contract coverage for the new render-attempt claim.

## Test Plan

- [x] `npm run lint`
- [x] `npx jest --runTestsByPath tests/store-contract.test.js tests/ddb-store.test.js tests/qurl-webhook-counter.test.js` (3 suites / 189 tests)
- [x] `npm test` (85 suites / 2921 tests)
- [x] PR checks green on `9c1a873e353a35bca831c8baa505396694b6f753`
- [x] Claude review clean on final head: no blockers found
- [ ] `/simplify`: attempted twice, but both issue-comment runs failed inside the Claude Code action before review output. Tracked in layervai/qurl-integrations#889.
- [ ] LIVE final-head sandbox deploy/e2e smoke: attempted in layervai/qurl-integrations-infra run 27731975389, but blocked before smoke by stale Terraform lock from a canceled prior deploy. Tracked in layervai/qurl-integrations-infra#1234.

## Live Verification Status

Automated/unit coverage proves the invariants and mock-verified Discord edit path, but the one thing tests cannot prove is live cross-replica `PATCH .../messages/@original` plus button preservation. I attempted final-head sandbox verification for `9c1a873e353a35bca831c8baa505396694b6f753`; the deploy failed before e2e smoke because infra could not acquire the Terraform state lock left by canceled run 27731886565.

This is not an app-code failure, but it does leave live Discord verification blocked until qurl-integrations-infra#1234 is resolved.

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2442,9 +2442,12 @@ async function executeSendPipeline(interaction, {
         // baseMsg, WITHOUT the "👀 …" counter line (confirmMsg is that
         // base; monitor.getFullMsg() would double-stamp the counter).
         confirmBaseMsg: confirmMsg,
-        // The send's full qurl_id set, so the fast-path counts views off
-        // its single GetItem (no recipient-row Query). Filter falsy —
-        // legacy/non-guild links may omit qurlId.
+        // Optional inline qurl_id fallback cache. saveSendConfirmState
+        // caps large sends to [] before writing so qurl_send_configs
+        // never approaches DDB's item-size limit; the normal path renders
+        // from the sharded aggregate, and rare fallback reads recipient
+        // rows via getSendItems. Filter falsy — legacy/non-guild links may
+        // omit qurlId.
         confirmQurlIds: qurlLinks.map(l => l.qurlId).filter(Boolean),
         viewedCount: 0,
         // Epoch seconds. Aligned to the real Discord interaction-token TTL
@@ -2800,7 +2803,8 @@ async function executeSendPipeline(interaction, {
               } else {
                 // Re-arm the fast-path with the post-add totals so a view
                 // landing after /qurl add renders against the new base +
-                // count + qurl_id set. PARTIAL update — omitting
+                // count + optional inline qurl_id fallback cache. PARTIAL
+                // update — omitting
                 // interactionToken/appId leaves them untouched
                 // (saveSendConfirmState skips undefined keys) so it can't
                 // null the live token. Best-effort + logged-swallowed.
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1489,24 +1489,12 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // recipient burns the link before any interval would fire.
   const FIRST_POLL_DELAY_MS = 3000;
   // Dense early-poll phase. The webhook fast-path
-  // (routes/qurl-webhook.js) owns sub-second latency: it edits the
-  // sender's confirmation from any replica the instant a qurl.accessed
-  // view records, keeps distinct-view counts in sharded qurl_views
-  // counters, and schedules its own trailing flush for coalesced bursts.
-  // This poll is now a
-  // BACKSTOP / self-heal floor: it covers a legacy/pre-feature send (no
-  // persisted token), a fast-path edit that transiently failed (it
-  // deliberately does NOT advance last_rendered_count on a failed edit,
-  // so the next poll re-renders and self-heals), and the expand/collapse
-  // toggle the content-only fast-path doesn't restore.
-  //
-  // Because the fast-path carries the latency, #839's aggressive 5s dense
-  // ramp is relaxed back to a MODEST one: the first tick still fires at
-  // ~3s (FIRST_POLL_DELAY_MS), then EARLY_POLL_INTERVAL_MS (~15s) through
-  // the early window, then the flat steadyPollInterval (≤60s). A view the
-  // fast-path missed is still re-rendered within ~15s — fine for a
-  // backstop — at a fraction of the idle BatchGet volume the 5s ramp
-  // incurred.
+  // (routes/qurl-webhook.js) owns normal sub-second latency, but the
+  // cross-replica PATCH primitive is live-only. Keep #839's 5s dense
+  // backstop until that primitive is verified so a bad fast-path
+  // assumption cannot regress send counters below the pre-PR fallback.
+  // The poll also self-heals failed edits and restores expand/collapse
+  // content-only flicker.
   //
   // Cost: this replaces the old pollCount decay (which throttled an idle
   // monitor toward every-4th-tick), so the steady phase is a flat
@@ -1515,12 +1503,12 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // (`changed`) and each link flips once, so total edits over the
   // monitor's life are bounded by the count of distinct recipient views
   // (≤ QURL_SEND_MAX_RECIPIENTS, default 20000 — NOT 50), spread across
-  // the link's whole lifetime. At ≥15s spacing this is far under
-  // Discord's editReply rate limit — no 429 exposure. (The webhook
+  // the link's whole lifetime. At ≥5s spacing in the early window this is
+  // still far under Discord's editReply rate limit. (The webhook
   // fast-path, which fires per-view rather than per-tick, is the path
   // that needs the burst coalescing above; this poll is naturally rate-
-  // limited by its own ≥15s tick spacing.)
-  const EARLY_POLL_INTERVAL_MS = 15000;
+  // limited by its own poll tick spacing.)
+  const EARLY_POLL_INTERVAL_MS = 5000;
   const EARLY_POLL_WINDOW_MS = 90000;
   const startTime = Date.now();
   // Anchor for the dense early phase. Distinct from startTime (which
@@ -2570,7 +2558,7 @@ async function executeSendPipeline(interaction, {
         // deliberately NOT persisted: the webhook fast-path always renders
         // the persisted COLLAPSED base (content-only edit), so right after
         // an expand a concurrent fast-path counter edit can briefly show
-        // the collapsed list — the next poll tick (≤15s) restores the
+        // the collapsed list — the next early poll tick (≤5s) restores the
         // expanded view off the updated in-memory base. Accepted minor
         // toggle-flicker (option c); the poll is the self-heal floor.
         monitor.updateBaseMsg(confirmMsg);
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1484,8 +1484,9 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // Dense early-poll phase. The webhook fast-path
   // (routes/qurl-webhook.js) owns sub-second latency: it edits the
   // sender's confirmation from any replica the instant a qurl.accessed
-  // view records, keeps viewed_count as an O(1) aggregate, and schedules
-  // its own trailing flush for coalesced bursts. This poll is now a
+  // view records, keeps distinct-view counts in sharded qurl_views
+  // counters, and schedules its own trailing flush for coalesced bursts.
+  // This poll is now a
   // BACKSTOP / self-heal floor: it covers a legacy/pre-feature send (no
   // persisted token), a fast-path edit that transiently failed (it
   // deliberately does NOT advance last_rendered_count on a failed edit,
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -1481,27 +1481,16 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
   // Fast first tick catches sub-second self-destruct sends where the
   // recipient burns the link before any interval would fire.
   const FIRST_POLL_DELAY_MS = 3000;
-  // Dense early-poll phase. The sub-second SQS view-update push is
-  // best-effort and silent-drops on the N-1/N replicas that don't host
-  // this monitor (view-update-registry.js's SILENT DROP doc), so on
-  // most sends the polling tick below is what actually renders the
-  // counter. The webhook fast-path (routes/qurl-webhook.js) now OWNS
-  // sub-second latency — it edits the sender's confirmation from any
-  // replica the instant a qurl.accessed view records, no poll involved.
-  // So this poll is now a BACKSTOP / self-heal floor: it covers a
-  // legacy/pre-feature send (no persisted token), a fast-path edit that
-  // transiently failed (it deliberately does NOT advance last_rendered_count
-  // on a failed edit, so the next poll re-renders and self-heals), the
-  // expand/collapse toggle the content-only fast-path doesn't restore, and
-  // — since the fast-path is now leading-edge-coalesced
-  // (QURL_VIEW_COUNTER_COALESCE_MS, see routes/qurl-webhook.js step 4b) —
-  // the TRAILING-EDGE FLUSH of a burst's final-window tail: the last few
-  // views of a high-fan-out send land inside one coalesce window and are
-  // skipped by the fast-path, so this poll is what renders the settled
-  // final count. That makes the poll REQUIRED for convergence on a
-  // coalesced send, not merely a nice-to-have — do NOT remove it or widen
-  // its steady interval past the point where a coalesced tail would sit
-  // un-flushed for an unacceptably long time.
+  // Dense early-poll phase. The webhook fast-path
+  // (routes/qurl-webhook.js) owns sub-second latency: it edits the
+  // sender's confirmation from any replica the instant a qurl.accessed
+  // view records, keeps viewed_count as an O(1) aggregate, and schedules
+  // its own trailing flush for coalesced bursts. This poll is now a
+  // BACKSTOP / self-heal floor: it covers a legacy/pre-feature send (no
+  // persisted token), a fast-path edit that transiently failed (it
+  // deliberately does NOT advance last_rendered_count on a failed edit,
+  // so the next poll re-renders and self-heals), and the expand/collapse
+  // toggle the content-only fast-path doesn't restore.
   //
   // Because the fast-path carries the latency, #839's aggressive 5s dense
   // ramp is relaxed back to a MODEST one: the first tick still fires at
@@ -1637,12 +1626,9 @@ function monitorLinkStatus(sendId, interactionArg, qurlLinksArg, recipientsArg, 
         // or skip the allDone bookkeeping above. (tryAdvanceRenderedCount
         // also stamps last_rendered_at, arming the coalesce clock — so a
         // genuine NEW view landing within QURL_VIEW_COUNTER_COALESCE_MS
-        // (~1.5s) AFTER this tick IS coalesced by the fast-path's step 4b
-        // and deferred to the next poll tick. Benign: the poll just
-        // rendered the current count, and the ≤15s backstop catches the
-        // follower — the window is far shorter than the tick spacing, so
-        // this only shifts a sub-1.5s-after-tick view from fast-path to
-        // poll, never strands it.) NOT advanced on the degraded
+        // after this tick is coalesced by the fast-path's step 4b and
+        // rendered by the webhook trailing flush, not stranded for the
+        // next poll.) NOT advanced on the degraded
         // early-return or the terminal-freeze render — neither displays a
         // live counter, so advancing would strand it.
         if (ok) {
@@ -2443,6 +2429,7 @@ async function executeSendPipeline(interaction, {
         // its single GetItem (no recipient-row Query). Filter falsy —
         // legacy/non-guild links may omit qurlId.
         confirmQurlIds: qurlLinks.map(l => l.qurlId).filter(Boolean),
+        viewedCount: 0,
         // Epoch seconds. Aligned to the real Discord interaction-token TTL
         // (~15 min), NOT a minute past it: getSendRenderState self-defends
         // by treating a past value as absent, so setting this to ~15 min
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2443,12 +2443,19 @@ async function executeSendPipeline(interaction, {
         // its single GetItem (no recipient-row Query). Filter falsy —
         // legacy/non-guild links may omit qurlId.
         confirmQurlIds: qurlLinks.map(l => l.qurlId).filter(Boolean),
-        // Epoch seconds; bounds the sensitive token's life just past the
-        // 14-min monitor cap (16 min). getSendRenderState self-defends on
-        // this (treats a past value as absent) so the bot stops trusting
-        // a dead token even before the qurl-bot-ddb terraform enables a
-        // DDB TTL on confirm_expires_at to physically reap the row.
-        confirmExpiresAt: Math.floor(Date.now() / 1000) + 16 * 60,
+        // Epoch seconds. Aligned to the real Discord interaction-token TTL
+        // (~15 min), NOT a minute past it: getSendRenderState self-defends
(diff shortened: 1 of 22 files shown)


justin-layerv:
Addressed cr round 9 (HEAD 1012461f):

- **#1 dead-token retry window — FIXED.** `confirm_expires_at` was `+16min`; aligned it to `~15min` (the real Discord interaction-token TTL), kept above the 14-min monitor cap. `getSendRenderState`'s self-defense now treats the token as absent right when Discord kills it, closing the ~15–16min window where a view fired a 401ing PATCH. The fast-path still stays live for the token's whole valid window.
- **#2 CAS orders stored count, not rendered frame — acknowledged, no change.** Correct: the monotonic CAS guarantees `last_rendered_count` never regresses; concurrent PATCH *arrivals* from different replicas can still land out of order and briefly show a lower frame until the poll heals (≤15s). This is the documented "residual concurrent-leading-edge tail" (route COALESCING header + PR body) — the guarantee is on the persisted count, not the displayed frame, as you note.
- **#3 `getQurlViews` latency scales with N — acknowledged, no change.** Already documented (getQurlViews doc + step-5 comment + PR body): a non-coalesced edit on a max-fan-out send pays ⌈N/100⌉ BatchGets, off the fire-and-forget path so it can't delay the webhook ack; "sub-second" holds for small/medium sends, degrades with N on a large non-coalesced edit. Counter-sharding scoped out under the realistic cap.
- **Style (verbose comments) — partially trimmed.** Removed the success/failure-split rationale duplicated between `tryAdvanceRenderedCount`'s doc and the route's step-8 (store doc now keeps the write-mechanism detail; route owns the split's "why").
- **Rollout prerequisites — both tracked.** DDB `TimeToLive` on `confirm_expires_at` is filed as qurl-integrations-infra#1227 (gated on this PR, lead owns); live cross-replica verification is the post-deploy step documented in the PR body's "Un-draft / merge path" (safe failure mode → no regression on fallback).

Also folded in a /simplify nit (rename the token-strip throwaway `_omit` → `_drop` to match the `getGuildConfig` precedent).

justin-layerv:
Addressed cr round 10:

- **#1 terminal-fence call sites untested — reply (no harness).** Real catch, and I verified the cost: all three `markConfirmTerminal` call sites live *inside collector event handlers* (`collector.on('collect')` for revoke + add-degrade at commands.js:2620/2732, `collector.on('end')` for window-close at :2803), not inside the directly-exported `revokeAllLinks`/`handleAddRecipients` the send-pipeline suite drives — those are separate functions the collector calls alongside `markConfirmTerminal`. Asserting the call sites would require standing up a Discord button-collector harness (simulate `createMessageComponentCollector` + `on('collect')`/`on('end')` events), which is disproportionate to the risk: the *mechanism* is pinned by the store-layer unit test, the *read* side (`terminal:true` → skip) by the counter suite, and each call site carries an explicit "freeze so a late fast-path edit can't resurrect the counter" comment. The failure mode if a refactor dropped one is cosmetic + self-healing (a late view re-renders the counter once; the poll is already stopped). Flagging as the honest residual rather than building a collector harness for a converged PR — happy to add it if you'd rather.
- **#2 editInteractionReply swallows throws — reply, intentional.** The broad catch is deliberate: an interaction token routinely 401/404s at the ~15-min TTL, so resilience-over-the-expected-failure is correct, and the payload is a static `{content}` (throws before the request are network/Discord-API by construction, not programming errors). Escalating an unexpected class to `error` risks paging on a transient 5xx, which is the common case here — net worse. The poll backstop covers any miss regardless.
- **#3 GSI read amplification — reply, documented.** Already called out (getQurlViews doc + step-5 + PR body): 2 reads/recorded-view above the coalesce gate, off the awaited 200, can only degrade to poll latency. The `qurl_id-index` capacity/auto-scaling headroom for steady-state per-view query rate is an infra-side confirmation for the DDB-cost owner — noting for visibility; no bot change.
- **#4 ENABLE_VIEW_UPDATE_PUSH no-op + rip-out — FILED.** qurl-integrations#875 (rip out the dead publisher/consumer/registry trio + the no-op flag). Referenced in the PR body.
- **#5 comment density — reply.** Trimmed twice already (the COALESCING header restatement, and this round the success/failure-split duplication between `tryAdvanceRenderedCount` and the route). The remainder is load-bearing concurrency rationale (commit-after-edit, the clock-split, the terminal overload) where the *why* is genuinely non-obvious; further trimming risks cutting the invariants. Holding here.
- **#6 confirm_terminal rename — FILED.** qurl-integrations#875 (same issue), with the migration note (sandbox rows already carry the old name).

Net: zero warranted code changes this round (all reply/defer-class); the round-9 dead-token fix was the last code delta.

justin-layerv:
Addressed the latest Claude review consistency/rate-headroom item on the current head (8d06e36b): `QURL_VIEW_COUNTER_COALESCE_MS` now defaults to 900ms with max 900ms, and the PR body now matches the code. This keeps every coalesced sender-counter update sub-second while giving Discord the largest available edit-rate buffer.

Local verification after the tweak:
- `npm run lint`
- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/qurl-webhook.test.js tests/config-int-env.test.js`
- `npm test` (84 suites / 2864 tests)

justin-layerv:
Addressed Claude's latest performance/comment-density feedback on current head `cdfd9ab9`:

- Corrected the coalescing comments and PR body to say the debounce bounds Discord edits, not the per-view `viewed_count` write.
- Added an explicit `aggregate-update-error` fast-path verdict when `incrementSendViewedCount` throttles/fails, leaving the poll backstop to count from qurl-view rows.
- Added regression coverage for that aggregate-throttle fallback.
- Trimmed the long `markConfirmTerminal` and interaction-token scrub comments while keeping the load-bearing invariants.

Verification after this patch:
- `npm run lint`
- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/qurl-webhook.test.js tests/config-int-env.test.js` (5 suites / 399 tests)
- `npm test` (84 suites / 2865 tests)

justin-layerv:
Addressed Claude's latest coalescing semantics nit on current head `95077527`:

- Reworded the config comment and PR body so `QURL_VIEW_COUNTER_COALESCE_MS` is described as reject-to-default above 900ms, not clamped.
- Added direct config coverage for the view-counter coalesce knob: unset -> 900, in-range override -> accepted, `1500` -> warning + default 900.

Verification after this patch:
- `npm run lint`
- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/qurl-webhook.test.js tests/config-int-env.test.js` (5 suites / 402 tests)
- `npm test` (84 suites / 2868 tests)

justin-layerv:
@claude /simplify

Please run the simplify pass on PR #855 at current head `95077527`. Look for required and optional simplification opportunities, including comment density, unnecessary abstractions, redundant code paths, or doc/code mismatch. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed Claude's latest review on current head `ae8c127c`:

- Removed the dead defensive `recordQurlView` string-return fallback; the route now treats `returnMeta: true` as the contract.
- Migrated webhook tests/mocks to the `{ result, firstView }` metadata shape.
- Added coverage for the real scheduled trailing flush timer.
- Added coverage for the `repairFloor` re-render when a lower edit loses the post-edit CAS.
- Trimmed the coalescing and old SQS supersession comment blocks.

Verification after this patch:
- `npm run lint`
- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/qurl-webhook.test.js tests/qurl-webhook-consumed.test.js tests/qurl-webhook-consumed-payload-throw.test.js tests/qurl-webhook-flow.test.js tests/webhook-subscriptions-receiver.integration.test.js tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/config-int-env.test.js` (9 suites / 431 tests)
- `npm test` (84 suites / 2870 tests)

justin-layerv:
@claude /simplify

Please rerun the simplify pass on PR #855 at current head `ae8c127c`. The prior automatic review nits were addressed on this head: dead `recordQurlView` fallback removed, real trailing flush + repair-floor tests added, and comment density trimmed. Look for any remaining required or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the latest Claude/DE feedback on current head `991e6b6b`:\n\n- Replaced the hot `qurl_send_configs.viewed_count` aggregate with 64 sharded counter rows in `qurl_views`, keyed by stable qurl id hash, so high-fanout sends do not funnel every first-view write through one DDB item.\n- Fast-path renders now strongly read a fixed 64-key shard sum; coalesced events skip even that read and let the trailing flush render the settled count.\n- Encrypted `interaction_token` at rest using the existing app-layer envelope crypto; `getSendRenderState` decrypts only for the narrow edit path.\n- Removed the dead `recordQurlView(..., returnMeta)` polymorphism; it now always returns `{ result, firstView }` and rejects non-positive access counts.\n- Updated PR body and tests for the sharded-counter design.\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/ddb-store.test.js` (2 suites / 152 tests)\n- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/qurl-webhook.test.js tests/qurl-webhook-consumed.test.js tests/qurl-webhook-consumed-payload-throw.test.js tests/qurl-webhook-flow.test.js tests/webhook-subscriptions-receiver.integration.test.js tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/config-int-env.test.js` (9 suites / 432 tests)\n- `npm test` (84 suites / 2871 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `991e6b6b`. Latest patch addresses the prior medium hot-item concern by moving distinct first-view counts to sharded `qurl_views` counter rows with strong 64-key shard reads, encrypts `interaction_token` at rest, and removes the dead `recordQurlView(..., returnMeta)` polymorphism. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the latest automatic-review feedback on current head `107de181`:\n\n- Scaled sender view-counter shards by `expectedCount`: small sends use 1 shard/read, larger sends scale by fan-out up to 64 shards at max size. This keeps high-fanout hot-write protection without charging every one-recipient send for 64 strong reads.\n- Kept shard growth monotonic for `/qurl add`: since `expectedCount` only increases, reading more shards includes earlier lower-shard counts.\n- Documented the single-retry `UnprocessedKeys` behavior: partial sums are corrected by the next non-coalesced render and the poll backstop.\n- Replaced the dense aggregate fallback condition with a named `haveAggregateSignal`.\n- Updated PR body with scaled shard costs and the rolling-deploy legacy-floor note.\n\nVerification after this patch:\n- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/ddb-store.test.js` (2 suites / 153 tests)\n- `npm run lint`\n- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js tests/qurl-webhook.test.js tests/qurl-webhook-consumed.test.js tests/qurl-webhook-consumed-payload-throw.test.js tests/qurl-webhook-flow.test.js tests/webhook-subscriptions-receiver.integration.test.js tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/config-int-env.test.js` (9 suites / 433 tests)\n- `npm test` (84 suites / 2872 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `107de181`. Latest patch addresses the small-send read amplification review item by scaling shard count from 1 to 64 based on expected fan-out, documents the single-retry UnprocessedKeys behavior, names the aggregate fallback condition, and updates the PR body/rolling-deploy note. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the final simplify optional nit on current head `30023ddb`: added the missing rationale for `SEND_VIEW_COUNTER_TARGET_WRITES_PER_SHARD = 500` (roughly half the DDB single-partition WCU/s ceiling, leaving burst headroom before scaling to the next shard step).\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/ddb-store.test.js` (1 suite / 136 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `30023ddb`. Latest patch addresses the final optional nit by documenting why the shard write target is 500. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the latest non-blocking automatic-review note on current head `0feec632`: added the call-site invariant tying `recordQurlView`'s `firstView` contract to exactly one shard increment per distinct `qurl_id`.\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/qurl-webhook-counter.test.js` (1 suite / 17 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `0feec632`. Latest patch only adds the first-view/shard-increment invariant comment requested by automatic review. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the latest automatic-review optional items on current head `70dae928`: added a source-ownership regression test so `getSendRenderState` remains limited to the one production caller that handles the decrypted interaction token, and trimmed the repeated poll-backstop / supersession prose called out as cosmetic.\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/ddb-store.test.js tests/qurl-webhook-counter.test.js` (2 suites / 154 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `70dae928`. Latest patch addresses the automatic-review optional items: a source-ownership guard for the decrypted-token render-state getter, plus comment-density trims around the poll backstop / supersession prose. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `b3699359`. Latest patch addresses the extra low-severity automatic-review items: one-shard invalid expectedCount fallback, explicit best-effort shard aggregate comment, and `auth:false` for the interaction webhook PATCH. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the additional low-severity automatic-review items on current head `b3699359`:\n- invalid/legacy `expectedCount` now falls back to the one-shard floor, preserving the shard superset invariant if a row is later repaired\n- the sharded aggregate miss path now documents that the aggregate is best-effort and the source `qurl_views` + poll/floor path is the correctness anchor\n- `editInteractionReply` now passes `auth: false`, so the interaction webhook PATCH does not attach bot Authorization\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/ddb-store.test.js tests/discord-rest.test.js tests/qurl-webhook-counter.test.js` (3 suites / 178 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `2ab326f0`. Latest patch only clarifies the latest automatic-review notes: the two-write source/shard split, the bounded one-shot repair PATCH, and seed-only legacy `viewed_count`. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
Addressed the latest automatic-review notes on current head `2ab326f0`:\n- documented the intentional two-write split between source view recording and best-effort shard aggregate\n- documented the one-shot repair PATCH tradeoff after a monotonic-floor race\n- clarified that legacy `viewed_count` is seed-only for the new path and remains only as the rolling-deploy floor\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/ddb-store.test.js tests/qurl-webhook-counter.test.js` (2 suites / 155 tests)

justin-layerv:
Addressed the automatic-review poll→fast-path backwards-flicker finding on current head `0bc51451`:\n- added token-free `getSendRenderedCount(sendId)` with a strong read of `last_rendered_count` only\n- monitor poll renders now clamp to `Math.max(localPollViewed, persistedFloor)` before editing, so an eventual `getQurlViews` read cannot display below the fast-path floor\n- local monitor `viewed` is not mutated by the clamp, so later source-row catch-up does not double count pending flips\n\nVerification after this patch:\n- `npm run lint`\n- `npx jest --runTestsByPath tests/ddb-store.test.js tests/send-pipeline-back-half.test.js tests/qurl-webhook-flow.test.js tests/store-contract.test.js` (4 suites / 341 tests)

justin-layerv:
@claude /simplify\n\nPlease rerun simplify on PR #855 at current head `0bc51451`. Latest patch fixes the automatic-review poll→fast-path backwards-flicker finding by adding a token-free displayed-floor read and clamping monitor poll renders to that persisted floor. Look for any remaining required, nit, or optional simplification items. If nothing remains, please say the simplify pass is clean.

justin-layerv:
@claude /simplify

Please rerun simplify on PR #855 at current head `8574de00`. Since the prior review, I added the expected_count monotonic DDB condition plus tests, pinned the monitor floor-read fallback, and added the shard-expansion regression requested in review. Please look for any remaining required, nit, or optional simplification items; if clean, say so explicitly.

justin-layerv:
@claude /simplify

Please rerun simplify/review on PR #855 at current head `9a21d808`. Since the last pass, I addressed the remaining minor items: legacy qurl_view rows missing `access_count` no longer double-count as first views, the `last_rendered_at` comment now reflects conditional-write semantics, `saveSendConfirmState` documents the fail-closed expected_count condition, and the shard-sum read-failure fallback is directly tested. Please report any remaining required/nit/optional items, or explicitly say clean.

justin-layerv:
@claude /simplify

Please rerun simplify/review on PR #855 at current head `ae9cde80`. Since the last pass, I addressed the non-first-view read amplification by skipping repeat-access events before shard-sum reads unless force=true, and updated tests so trailing flush uses the real force=true timer path. Please report any remaining required/nit/optional items, or explicitly say clean.

justin-layerv:
@claude /simplify

Please rerun simplify/review on PR #855 at current head `0711c9a0`. I moved the repeat-access skip before any send lookup/render-state DDB reads, simplified the coalesce branch now that non-first non-force events are unreachable there, and added a comment explaining the legacy/no-floor fallback reachability. Please report any remaining required/nit/optional items, or explicitly say clean.

justin-layerv:
@claude /simplify

Retrying after the prior two Claude Code action runs failed from runtime errors. Current PR head is `0711c9a0` and all required PR checks are green. Please run the simplify pass on this exact head and explicitly say clean or list any remaining required/nit/optional items.

justin-layerv:
@claude /simplify

Single-trigger retry because the previous Claude Code run was cancelled before it returned findings. Please run the simplify pass against current PR head `0711c9a0`; all required PR checks are green, and the remaining read-load observation is now documented/confirmed as PAY_PER_REQUEST + graceful poll fallback. Please explicitly say clean or list any remaining required/nit/optional items.

justin-layerv:
@claude /simplify

Final pass on head `9c1a873e353a35bca831c8baa505396694b6f753`. Claude review feedback has been addressed or tracked:
- fixed distributed edit/read coalescing with `tryClaimRenderAttempt` before shard-sum read/PATCH
- added `tryClaimRenderAttempt` to `STORE_METHODS` and verified store-contract tests
- documented pre-arm source-of-truth reconciliation
- documented per-first-view DDB cost and created #888
- final-head sandbox deploy was attempted and is blocked by infra stale tfstate lock, tracked in layervai/qurl-integrations-infra#1234

Please return clean if there are no remaining required/nit/optional simplifications, or list every remaining item explicitly.

justin-layerv:
@claude /simplify

Retrying final simplify gate after run 27732290651 failed inside the Claude Code action before returning review output. Same head: `9c1a873e353a35bca831c8baa505396694b6f753`.

Please return clean if there are no remaining required/nit/optional simplifications, or list every remaining item explicitly.

Issue #834, opened 2026-06-17
Salvage from #831: boot-time fail-fast when DETECT_COMMAND_ENABLED=true but token/key missing or PLACEHOLDER


justin-layerv:
## Context

The merged [#828](https://github.com/layervai/qurl-integrations/pull/828) validates the detect credentials at **call time** (`detectWatermark` throws if `QURL_API_KEY` / `DETECT_ACCESS_TOKEN` is unset). The superseded parallel PR [#831](https://github.com/layervai/qurl-integrations/pull/831) (branch `justin/fix-1170-detect-tunnel-resolve`) had a stronger **boot-time** guard worth salvaging: a `boot-requirements.js` that fails fast at startup when `DETECT_COMMAND_ENABLED=true` but `QURL_API_KEY` or `DETECT_ACCESS_TOKEN` is missing, whitespace-only, or still the `PLACEHOLDER` sentinel.

Boot-time fail-fast is better for activation safety: a misconfigured flip (gate on, token still `PLACEHOLDER`) is caught at deploy/startup with a clear message, instead of every `/qurl detect` invocation failing at runtime after the bot has gone live.

## Scope

- Port #831's `apps/discord/src/boot-requirements.js` pattern (+ `boot-requirements.test.js`) onto the merged base: when `config.DETECT_COMMAND_ENABLED`, require `QURL_API_KEY` and `DETECT_ACCESS_TOKEN` to be present, non-whitespace, and not the placeholder sentinel — fail startup loudly otherwise.
- Reconcile with #828's existing call-time guards (keep them; the boot check is the early tripwire, not a replacement).
- Reuse the existing placeholder-sentinel convention if the bot already has one.

## Acceptance criteria

- Bot refuses to boot with `DETECT_COMMAND_ENABLED=true` + a missing/placeholder token or key, with a clear error naming the offending var.
- Boots normally when the gate is off (default) or when both creds are real.
- Tests cover: gate-off (no requirement), gate-on + each missing/placeholder credential.

Sequence before flipping `DETECT_COMMAND_ENABLED=true` (relevant to the #1209 activation wiring).

PR #863, opened 2026-06-17
refactor(discord): self-mint an ephemeral qURL for detect reach

justin-layerv:
## Summary

`/qurl detect` reaches the connector's `/api/detect` deanonymization oracle over the **qurl-detect reverse tunnel** (never the public ALB). This PR repoints the bot's tunnel reach from a **pre-seeded multi-use `DETECT_ACCESS_TOKEN`** to **self-minting an ephemeral qURL per call** — the bot already owns a `QURL_API_KEY`, so it mints + resolves its own short-lived link via the `@layervai/qurl` SDK:

`listResources({ slug, status:'active' })` → `createQurlForResource(rid, { target_path:'/api/detect' })` → `resolve({ access_token })` → POST raw image bytes to the resolved `target_url`.

The `at_…` access token comes from the minted `qurl_link` fragment; only the stable, non-secret `resource_id` is cached.

### What changed
- `resolveDetectTarget()` (`connector.js`): self-mint-then-resolve flow via the SDK's `QURLClient`. `_detectResourceId` is cached across calls (stable id); the ephemeral mint + the `resolve` NHP-knock run **per call** (fresh token, fresh IP knock). The cache **self-heals** — a mint failure clears it so a deleted/recreated resource re-resolves on the next detect instead of 404-looping until restart.
- `DETECT_ACCESS_TOKEN` dropped from `config`; `DETECT_TUNNEL_SLUG` added (plain non-secret env). The infra side (the slug tfvar + env wiring) is layervai/qurl-integrations-infra#1232; the `DETECT_ACCESS_TOKEN` SSM-param removal is layervai/qurl-integrations-infra#1233.
- `@layervai/qurl` bumped `~0.2.0` → `~0.3.0` — `target_path` support landed in layervai/qurl-typescript#145 (`0.3.0`). CI is green on 0.2.0 only because the tests mock the SDK; the deployed runtime needs 0.3.0.

### Security posture (preserved)
- The `assertPublicHttpsTarget` host-pin is byte-identical: rejects non-https, embedded userinfo, private hosts, and any host not ending in `.qurl.site`.
- Logs only `err.message` on every failure leg — never the access token or the raw `target_url`. Every failure path asserts **no POST**.
- The SDK Bearer (`QURL_API_KEY`, which must carry the `qurl:resolve` scope) is distinct from the per-call `apiKey` threaded into the detect POST's Bearer.

### Note on the SDK (correcting the earlier description)
This **keeps** the `@layervai/qurl` SDK for the detect reach (the standardized client) — an earlier draft of this PR dropped it for the bare `qurlFetch`, which was reverted. The bot still has both clients (`qurlFetch` in `qurl.js` + the SDK here); **consolidating to one client (#830) is out of scope and remains open — this PR does not resolve it.**

## Type of Change

- [x] New feature
- [x] Refactoring

## Related Issues

Relates to #830 (SDK consolidation — **NOT** resolved here), #834 (boot fail-fast follow-up). Infra: layervai/qurl-integrations-infra#1232 (slug) + #1233 (token removal). SDK: layervai/qurl-typescript#145 (`target_path`).

## Testing

- [x] Unit tests added/updated
- [x] All existing tests pass

Full bot suite: **2814 passed** (83 suites). The detect tests mock the SDK; the live SDK `0.3.0` surface (`QURLClient`, `listResources`, `createQurlForResource(id, { target_path })`, `resolve`) was verified against the installed package's `.d.ts`.

### DIFF
--- a/apps/discord/package-lock.json
+++ b/apps/discord/package-lock.json
@@ -15,7 +15,6 @@
         "@aws-sdk/lib-dynamodb": "^3.1036.0",
         "@discordjs/rest": "~2.6.0",
         "@discordjs/ws": "~1.2.3",
-        "@layervai/qurl": "~0.2.0",
         "discord-api-types": "^0.38.33",
         "discord.js": "14.25.1",
         "express": "~5.2.1",
@@ -2067,15 +2066,6 @@
       "dependencies": {
         "@jridgewell/resolve-uri": "^3.1.0",
         "@jridgewell/sourcemap-codec": "^1.4.14"
-      }
-    },
-    "node_modules/@layervai/qurl": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/@layervai/qurl/-/qurl-0.2.0.tgz",
-      "integrity": "sha512-+ldlWnnXPBQKAkGO/5v00FdK6bOvK2kMW7Cii1/BO4RG5Vu7SiBinvt37YcgmNJtkQY+bw3kw127q5o2Hij0CA==",
-      "license": "MIT",
-      "engines": {
-        "node": ">=20"
       }
     },
     "node_modules/@napi-rs/wasm-runtime": {
--- a/apps/discord/package-lock.json
+++ b/apps/discord/package-lock.json
@@ -15,7 +15,7 @@
         "@aws-sdk/lib-dynamodb": "^3.1036.0",
         "@discordjs/rest": "~2.6.0",
         "@discordjs/ws": "~1.2.3",
-        "@layervai/qurl": "~0.2.0",
+        "@layervai/qurl": "~0.3.0",
         "discord-api-types": "^0.38.33",
         "discord.js": "14.25.1",
         "express": "~5.2.1",
@@ -2070,9 +2070,9 @@
       }
     },
     "node_modules/@layervai/qurl": {
-      "version": "0.2.0",
-      "resolved": "https://registry.npmjs.org/@layervai/qurl/-/qurl-0.2.0.tgz",
-      "integrity": "sha512-+ldlWnnXPBQKAkGO/5v00FdK6bOvK2kMW7Cii1/BO4RG5Vu7SiBinvt37YcgmNJtkQY+bw3kw127q5o2Hij0CA==",
+      "version": "0.3.0",
+      "resolved": "https://registry.npmjs.org/@layervai/qurl/-/qurl-0.3.0.tgz",
+      "integrity": "sha512-pjkBnaSnzQyq25Z1BWYPPEUQzThihA465rDoPBHYTANdXcgI3dd5vqRkCO+ouyxx+BKFBshQI3HWPfr7Az5kFQ==",
       "license": "MIT",
       "engines": {
         "node": ">=20"
--- a/apps/discord/package-lock.json
+++ b/apps/discord/package-lock.json
@@ -15,6 +15,7 @@
         "@aws-sdk/lib-dynamodb": "^3.1036.0",
         "@discordjs/rest": "~2.6.0",
         "@discordjs/ws": "~1.2.3",
+        "@layervai/qurl": "~0.2.0",
         "discord-api-types": "^0.38.33",
         "discord.js": "14.25.1",
         "express": "~5.2.1",
@@ -2066,6 +2067,15 @@
       "dependencies": {
         "@jridgewell/resolve-uri": "^3.1.0",
         "@jridgewell/sourcemap-codec": "^1.4.14"
+      }
+    },
+    "node_modules/@layervai/qurl": {
+      "version": "0.2.0",
+      "resolved": "https://registry.npmjs.org/@layervai/qurl/-/qurl-0.2.0.tgz",
+      "integrity": "sha512-+ldlWnnXPBQKAkGO/5v00FdK6bOvK2kMW7Cii1/BO4RG5Vu7SiBinvt37YcgmNJtkQY+bw3kw127q5o2Hij0CA==",
+      "license": "MIT",
+      "engines": {
+        "node": ">=20"
       }
     },
     "node_modules/@napi-rs/wasm-runtime": {
--- a/apps/discord/package.json
+++ b/apps/discord/package.json
@@ -26,7 +26,7 @@
     "@aws-sdk/lib-dynamodb": "^3.1036.0",
     "@discordjs/rest": "~2.6.0",
     "@discordjs/ws": "~1.2.3",
-    "@layervai/qurl": "~0.2.0",
+    "@layervai/qurl": "~0.3.0",
     "discord-api-types": "^0.38.33",
     "discord.js": "14.25.1",
     "express": "~5.2.1",
--- a/apps/discord/package.json
+++ b/apps/discord/package.json
@@ -26,6 +26,7 @@
     "@aws-sdk/lib-dynamodb": "^3.1036.0",
     "@discordjs/rest": "~2.6.0",
     "@discordjs/ws": "~1.2.3",
+    "@layervai/qurl": "~0.2.0",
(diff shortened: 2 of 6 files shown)


justin-layerv:
Thanks — all addressed in `ec04d968` (+ the body rewrite):

1. **PR body ↔ code mismatch (blocking)** — rewrote the description to match what the diff actually does (keeps the `@layervai/qurl` SDK, bumps to `~0.3.0`) and **removed `resolves #830`**. #830 (SDK consolidation) is now explicitly called out as still-open / not resolved here — the bot keeps both `qurlFetch` and the SDK by design for now.
2. **Stale `getQurlClient` comment (blocking)** — corrected: `target_path` needs `@layervai/qurl >= 0.3.0` (landed in qurl-typescript#145), package.json pins `~0.3.0`. Dropped the wrong `~0.2.1`/"follow-up" wording.
3. **SDK surface vs. empirical validation** — verified the installed `0.3.0` `.d.ts` exposes exactly what the code calls: `QURLClient` (client.d.ts:3), `resolve(ResolveInput|string)` (:190), `listResources(ResourceListInput?)` (:196), `createQurlForResource(id, CreateQurlForResourceInput?)` (:213) with `target_path?` (types.d.ts:382). The mocks match the real contract.
4. **`_detectResourceId` self-heal** — the mint catch now clears `_detectResourceId`, so a deleted/recreated resource re-resolves the slug on the next detect instead of 404-looping until restart. Added a test driving mint success→fail→success asserting `listResources` re-runs (clearing on any mint error is intentional — worst case is one extra cheap `listResources`).

Minor: token extraction (everything after the first `#`, `startsWith('at_')`) and `resources[0]` are kept as documented assumptions of the current wire format + unique slugs — agreed they're fine as-is; flagging in case the wire format ever changes.

justin-layerv:
Thanks — addressed in `b65162c5`:

1. **Stale title/body** — the body was already rewritten to the SDK-based design (you reviewed a pre-edit snapshot); I've now also **retitled** to drop "drop the SDK" (the SDK is kept — only `DETECT_ACCESS_TOKEN` is dropped, and `resolves #830` is gone, so #830 stays open).
2. **listResources breadcrumb** — wrapped the slug-lookup leg with `logger.warn('Detect tunnel slug lookup failed', { error })` (message only), matching the mint/resolve legs, + a test asserting the breadcrumb and that no mint/POST happens on a transport failure.
3. **Ephemeral-mint cleanup** — acknowledged: detect is low-frequency and the ~5-min TTL reaps the mints; no explicit delete by design. Flagging it as an accepted implicit dependency on server-side expiry.

justin-layerv:
Thanks — addressed in `a559b55e`:

1. **Permissive fragment parsing** — hardened: extracts only the token via `fragment.split(/[&?#]/)[0]` (strips any trailing `&`/`?` params) before the `at_` check, so a future `qurl_link` with extra fragment data can't thread garbage into `resolve()`. + a test asserting `#at_tok123&utm=x` resolves with just `at_tok123`.
2. **`resources[0]` / slug uniqueness** — added a comment asserting the qurl-service `slug` filter is **exact-match**, so `[0]` is the unique active resource, not a prefix/substring co-match.
3. **Self-heal clears on any mint error** — confirmed intentional (already documented); worst case is one extra cheap `listResources`. No change.

Re sequencing: #1232 (slug wiring) is merge-ready and lands with this; #1233 (param removal) is a held **draft** (its `prevent_destroy` plan-gate is by-design until the sandbox `state rm`). `/qurl detect` stays gated **off** (`detect_command_enabled="false"`) until after the deploy + soak.

justin-layerv:
Thanks — addressed the two actionable items in `2c408879`:

**2. Token in resolve `err.message`** — verified **and** hardened. The SDK builds error messages from the RFC-7807 *response* envelope (`${title} (${status}): ${detail}`, `errors.js:62`), not the request body, so there's no leak in @layervai/qurl 0.3.0 today. But since "never log the token" must not depend on code this bot doesn't own, the resolve breadcrumb now scrubs any `at_<token>` from `err.message` before logging (+ a test). Self-enforced across SDK versions.

**4. Exact-slug-match contract** — added a `TODO(upstream-contract)` marker (the repo's lockstep convention per CLAUDE.md) so a future qurl-service fuzzy-slug change surfaces in `git grep TODO(upstream-contract)`.

**1. SDK live-surface vs mocked CI** — agreed; that's the sandbox `/qurl detect` soak, which is the explicit next step (the slug PR #1232 + this land first, then I soak before flipping `detect_command_enabled`).

**3, 5. Ephemeral-mint accumulation / cache-clear-on-any-error** — both documented + accepted as intentional (low-frequency detect makes accumulation negligible; the over-eager clear costs at most one extra `listResources`).

justin-layerv:
Good catch — fixed in `f7300024`. You're right that the token originates in the mint **response** (`qurl_link` fragment), so the mint breadcrumb is the more likely vector. Factored the scrub into a `redactAccessToken` helper and applied it to **both** the mint and resolve `catch` blocks (the slug-lookup leg is token-free); + a mint-leg redaction test mirroring the resolve one. Updated the JSDoc SECURITY note to match.

2, 3 (cold-cache best-effort concurrency; self-heal is mint-scoped) — agreed, both intentional/harmless as noted.

justin-layerv:
Good catch — fixed in `ecd20266` with your preferred option (explicit expiry). `createQurlForResource` now passes **`expires_in: '5m'`**, so the "~5-min ephemeral" premise is true by construction rather than resting on an unverified server default — bounding both the credential lifetime **and** mint accumulation (the bot relies on expiry, never deletes). Updated the docstrings + the call assertion. Marked `TODO(upstream-contract)` to confirm qurl-service honors `expires_in` on a resource mint during the sandbox soak (which is item 2 — CI mocks the SDK, so the live surface incl. this is exactly what the soak exercises).

Minor items (cache-clear on a shape error / cold-cache best-effort concurrency) — agreed, both intentional + benign as noted.

justin-layerv:
Thanks — addressed in `487213ab`:

**1. (Medium) `expires_in` cleanup is unverified** — agreed, this is the real gate. I'm making "confirm qurl-service honors `expires_in: '5m'` on the detect resource mint (the qURL self-reaps after expiry — no accumulation)" a **hard checklist item in the sandbox soak** before `detect_command_enabled` flips on. I'll verify it directly (mint via `/qurl detect` → confirm the qURL is gone post-expiry) and, per your suggestion, note a live-qURL-count check on the detect resource. Detect stays gated **off** until that passes.
**3. (Nit) slug breadcrumb null-guard** — applied `redactAccessToken` uniformly to all three legs (a no-op on the token-free slug message, but it null-guards + keeps the pattern consistent).
**Cosmetic comment wrap** — reflowed.
**2. (Low) cache-clear on a shape error** — agreed, intentional + harmless (at most one extra `listResources`); left as-is.

justin-layerv:
Thanks — addressed in `3f0c6105`:

**1. (Important) Understated scope requirement** — fixed. The `getQurlClient` Bearer note + the `detectWatermark` docstring now state `QURL_API_KEY` needs **`qurl:read` + `qurl:write` + `qurl:resolve`** (list / mint / resolve), with the explicit warning that a resolve-only key passes mocked CI then 403s on the first list/mint. The sandbox key is already verified working (mint 201 + resolve 200 earlier this cycle), and "confirm the key carries all three scopes" is now a hard soak gate before the flip.

**2 / 4 (soak gates)** — both now hard checklist items before `detect_command_enabled` flips: confirm `expires_in` is honored (the qURL self-reaps), and watch for a per-resource active-qURL cap (`"token limit per QURL reached"`) under a detect burst.

**3** — yes, `QurlClient`→`QURLClient` was a latent-bug fix (the mock masked it); now consistent with `qurl.js`.

**5, 6 (minor)** — cache-clear-on-shape-error + cold-cache concurrency: both intentional + benign, left as-is.

PR #897, opened 2026-06-18
fix(discord): retry meta-seal capped mints singly

justin-layerv:
## Summary
- tag connector meta-seal batch-cap errors as `mint_batch_cap_exceeded`
- keep the normal batch mint path first, then retry multi-recipient Discord sends as sequential `n=1` mints only for that exact cap error
- log already-minted qURL IDs if the single-link fallback fails mid-loop, without logging qURL links/tokens
- add regression coverage for cap=1 fallback, generic cap non-fallback, fallback guard failures, mid-loop failure logging, and typed error tagging

## Business relevance
This keeps Discord multi-recipient `/qurl send` working while invisible watermarking/meta-seal is enabled with `WATERMARK_BATCH_CAP=1`. Without this fallback, multi-recipient sends with 2-10 recipients fail before any links are minted, which blocks safely enabling the Discord detect rollout.

## Related
- https://github.com/layervai/qurl-integrations-infra/issues/1088

## Validation
- `npm ci`
- `npm run lint`
- `npx jest --runTestsByPath tests/connector-coverage.test.js --runInBand --coverage=false`
- `npx jest --runTestsByPath tests/send-pipeline-helpers.test.js --runInBand --coverage=false`
- `npm test -- --ci --silent --runInBand`
- `git diff --check`

### DIFF
--- a/apps/discord/src/connector.js
+++ b/apps/discord/src/connector.js
@@ -63,6 +63,11 @@ const QUOTA_EXCEEDED_PATTERNS = [
   /token limit per QURL reached/i,
   /per[\s_-]?resource (token|link|mint) (limit|cap)/i,
 ];
+// TODO(upstream-contract): replace this English-string match once the
+// connector exposes a typed meta-seal batch-cap code.
+const META_SEAL_BATCH_CAP_PATTERNS = [
+  /n must not exceed \d+ when invisible watermarking is enabled/i,
+];
 
 async function throwConnectorError(label, response) {
   let bodyText = '';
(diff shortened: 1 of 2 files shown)


justin-layerv:
Convergence note after the final Claude pass:

- Confirmed the meta-seal `n > WATERMARK_BATCH_CAP` rejection is atomic before minting: the connector returns from the cap gate before the mint/link path runs.
- The fallback remains bounded by the existing Discord batcher: `mintLinksInBatches` still sends at most `TOKENS_PER_RESOURCE` recipients per resource before re-uploading, so the single-mint fallback cannot exceed that per-resource budget.
- Mid-loop fallback failures now log already-minted `qurl_id`s for reconciliation without logging `qurl_link` or access tokens.
- `/simplify` was posted on the latest head; no separate bot reply surfaced, so I also did the manual simplify pass on the final diff. No further simplification changes found.

justin-layerv:
Closing this client-side fallback in favor of the connector-side architectural fix: layervai/qurl-integrations-infra#1243 accepts the public n<=10 contract and fans render-at-mint work out concurrently under WATERMARK_CONCURRENCY/service capacity instead of making Discord retry singles.

PR #900, opened 2026-06-19
fix(discord): surface partial mint links for reconciliation

### LINKED ISSUE #899: Handle connector mint_link partial-success links on non-2xx
## Problem

The connector's `POST /api/mint_link/:resource_id` partial-success contract can return a non-2xx response with a non-empty `links` array. The Discord connector client currently calls `throwConnectorError()` on `!response.ok`, so those successfully minted sibling qURLs are discarded by the bot.

The infra fix in layervai/qurl-integrations-infra#1243 reduces the old `WATERMARK_BATCH_CAP=1` failure by accepting Discord-sized batches and processing render-at-mint jobs concurrently. That makes the existing partial-success contract more important: if a concurrent batch has a partial failure, successful sibling links may exist even though the HTTP status is non-2xx.

## Scope

Update the Discord mint client / send pipeline so non-2xx `mint_link` responses with valid `links[]` are handled intentionally rather than being silently discarded. The exact UX can be fail-with-reconciliation, partial delivery, or explicit compensation if a qURL revoke/delete seam exists, but the live qURL IDs must be observable and not lost.

## Acceptance criteria

- `apps/discord/src/connector.js` parses `links[]` from non-2xx `mint_link` responses before throwing or returning.
- The send/add-recipients paths either persist/deliver valid partial links intentionally or log/record enough qurl_ids for reconciliation without exposing qurl_link access tokens.
- Tests cover a non-2xx `mint_link` body containing successful sibling links.
- The behavior is compatible with the connector partial-success contract documented in layervai/qurl-integrations-infra#1243.

justin-layerv:
## Business relevance

Discord multi-recipient sends can receive partial-success `mint_link` bodies after the SDK-backed connector mints links but returns a non-2xx response. Without surfacing those qURL IDs, operators cannot reconcile live sibling links and users can lose visibility into minted access. This preserves only qURL IDs/counts for recovery while avoiding access-token or full-link leakage.

Closes #899.

## Summary

- Parse non-2xx connector `mint_link` bodies before throwing.
- Attach partial qURL ID/count metadata to the thrown connector error and higher-level send/add-recipient logs.
- Log only qURL IDs/counts for partial mint reconciliation; never log `qurl_link` or access tokens.
- Add regression coverage for partial non-2xx mint bodies and downstream add-recipient failure logging.

## Validation

- `npx jest --runTestsByPath tests/connector-coverage.test.js --runInBand --coverage=false`
- `npx jest --runTestsByPath tests/send-pipeline-helpers.test.js --runInBand --coverage=false`
- `npx jest --runTestsByPath tests/send-pipeline-back-half.test.js --runInBand --coverage=false`
- `npm run lint`
- `git diff --check`
- `npm test -- --ci --silent --runInBand`

### DIFF
--- a/apps/discord/src/commands.js
+++ b/apps/discord/src/commands.js
@@ -2127,6 +2127,10 @@ async function executeSendPipeline(interaction, {
       error: error.message,
       apiCode: error.apiCode,
       status: error.status,
+      ...(error.partialLinkCount ? {
+        partial_link_count: error.partialLinkCount,
+        partial_qurl_ids: error.partialQurlIds,
+      } : {}),
       sendId,
     });
     clearCooldown(interaction.user.id); // allow retry on failure
@@ -3034,7 +3038,15 @@ async function handleAddRecipients(sendId, usersCollection, originalInteraction,
           ? 'Original attachment URL has expired. Please create a new send.'
           : 'Failed to prepare links. Please try again, or create a new send if the issue persists.';
         logger.error('addRecipients file re-upload failed', {
-          sendId, error: err.message, apiCode: err.apiCode, status: err.status, isExpired,
+          sendId,
+          error: err.message,
+          apiCode: err.apiCode,
+          status: err.status,
+          ...(err.partialLinkCount ? {
+            partial_link_count: err.partialLinkCount,
+            partial_qurl_ids: err.partialQurlIds,
+          } : {}),
+          isExpired,
         });
         // Always emit — every failure here (CDN re-download, connector
         // re-upload, or mint) is a "couldn't create links" event. A rare,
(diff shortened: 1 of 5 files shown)


justin-layerv:
Final Claude pass considered:

- Keeping connector `warn` + command-layer `error`: they are complementary correlation views (`resource_id` at connector layer, `sendId` at command layer).
- Keeping full `partial_qurl_ids`: per-mint batches are bounded and the full ID list is the reconciliation payload.
- Not deduping `qurl_id`s locally: duplicate IDs would be a connector-contract violation worth surfacing rather than silently hiding.
- Keeping `mintedLinksWithId` + `qurlIdsFromLinks` split: the guard/projection separation is clearer than collapsing them.

No remaining blocking or actionable review items from the final Claude pass.

